PR #72, opened 2017-05-30
Adding tmpfs to addon config

bestlibre:
I'm not able to test this PR at the moment. If you want to refactor the docker code, do not wait for me, I will redo the modification after. Or you could include them in your refactoring.


hassio/dock/addon.py
```diff
@@ -100,7 +109,8 @@ def _run(self):
                 ports=self.addons_data.get_ports(self.addon),
                 devices=self.addons_data.get_devices(self.addon),
                 environment=self.environment,
-                volumes=self.volumes
+                volumes=self.volumes,
+                tmpfs=self.tmpfs
```
pvizeli: you can call `self.addons_data.get_tempfs` without property.

hassio/addons/validate.py
```diff
@@ -52,6 +52,7 @@ def check_network(data):
     vol.Optional(ATTR_PORTS): dict,
     vol.Optional(ATTR_HOST_NETWORK, default=False): vol.Boolean(),
     vol.Optional(ATTR_DEVICES): [vol.Match(r"^(.*):(.*):([rwm]{1,3})$")],
+    vol.Optional(ATTR_TMPFS): vol.Coerce(str)},
```
pvizeli: Please make a `vol.Match` to validate the Input for that.

hassio/dock/addon.py
bestlibre: Should I modify addons_data.get_tempfs to return a dict then ? I was thinking that the get_tmpfs should return a string since the parameter in config will be a string

hassio/addons/validate.py
bestlibre: There is a lot of possibility since it is mount option. I don't know how to cover them all. I can write one for my use case if needed (size and uid).

hassio/dock/addon.py
pvizeli: It return None if no `tmpfs` is set and that disabled it.

pvizeli:
Oh I was wrong, I will revert it. Sorry

hassio/dock/addon.py
bestlibre: Done. I have modified the vol.Match to be directly compatible with the tmpfs api option

bestlibre: My push was rejected, I've tried it moments after your commit. I'm ok with the code as it is now, but if you prefer I can ditch the property and have a list in the options instead of a string.

pvizeli:
I like you simplify that we have only a /tmpfs :+1: good work

hassio/dock/addon.py
pvizeli: No I like you code. I missunderstand it on first read :smiley:

PR #75, opened 2017-06-02
Allow map special device to homeassistant docker

pvizeli:


luxus:
does special device means something like a razberry? if yes how can I turn it on? @pvizeli

PR #136, opened 2017-08-07
Allow privileged capability SYS_RAWIO

### LINKED ISSUE #134: Allow privilege SYS_RAWIO
I'm trying to use write an addon for some GPIO connected devices on my Raspberry Pi using the RPi.GPIO Python plugin. However, it needs access to /dev/mem and it's not sufficient to specify it with `"devices":["/dev/mem:/dev/mem:rwm"]`, because it leads to the following error:

`RuntimeError: No access to /dev/mem.  Try running as root!`

I think we need to set SYS_RAWIO on the container to allow this, but only NET_ADMIN and SYS_ADMIN is allowed in the validation script, see line 40:
https://github.com/home-assistant/hassio/blob/dev/hassio/addons/validate.py#L40

Can we add SYS_RAWIO to the allowed list as well? I haven't been able to test if that solves everything, otherwise I would've sent a PR instead :)

radhus:
In order to allow writes to /dev/mem, which is needed e.g. to use the
GPIO ports on Raspberry Pi, the SYS_RAWIO capability needs to be
granted.

Fixes #134.

**NOTE:** I haven't been able to test this, so please verify it as you please before merging :)

### DIFF
--- a/hassio/addons/validate.py
+++ b/hassio/addons/validate.py
@@ -38,6 +38,7 @@ STARTUP_ALL = [
 PRIVILEGED_ALL = [
     "NET_ADMIN",
     "SYS_ADMIN",
+    "SYS_RAWIO"
 ]
 
 


homeassistant:
Hi @radhus, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;hassio#136).

Thanks!

pvizeli:
Looks good. Please add it also to https://home-assistant.io/developers/hassio/addon_config/

radhus:
Thanks, opened home-assistant/home-assistant.github.io#3151 :)

PR #191, opened 2017-09-16
Update build system to origin docker

pvizeli:
We need wait until we release a new image with new docker version


hassio/const.py
```diff
@@ -117,6 +118,8 @@
 ATTR_DISK = 'disk'
 ATTR_SERIAL = 'serial'
 ATTR_SECURITY = 'security'
+ATTR_BASE_IMAGE = 'base_image'
```
frenck: Maybe rename it to "from" or "build_from" to keep naming with Docker term consistent?

pvizeli: that would be also not correct. We don't know if they would be build from them. We give only a possibility to work with our or custom base images but that's a option. With `build_from` we suggest that he realy will build from this. I think for the build env it would be a option to have a build-from while that would be realy build from this if we use this option as user

pvizeli: Ok

frenck:
I have several concerns, which I will combine into a single post/review reaction soon.

Issue #201, opened 2017-09-27
Running addons with --privileged


pschmitt:
Hi.

I'm trying to create an addon for [pilight](https://www.pilight.org/): https://github.com/pschmitt/hassio-addons/tree/master/pilight
Unfortunately for this to work I'd need to run the container with `--privileged`. AFAIK this is not possible right now. This and `--pid=host` may be useful for other addons as well (dind, zabbix-agent)

gollo:
Is the device mapping in config.json (eg ```"devices": [ "/dev/vchiq:/dev/vchiq:rwm" ]```) not sufficient?

pschmitt:
No. I tried both with `--device /dev/mem:/dev/mem:rw` and `--cad_add SYS_RAWIO`. The service starts okay, I can connect to it and all but nothing happens when I connect ask the daemon to issue commands (they are logged as successful - but that's all). Something else must be missing since starting the container with `--privileged` makes it work as expected.

If you happen to have a RF transmitter you can try it for yourself:

```bash 
# Do not work
docker run -it --rm --name pilight --device /dev/mem --cap-add SYS_RAWIO pschmitt/hassio-addon-armhf-pilight
docker exec -it pilight pilight-send -p pollin -u 1 -s 1 -t
# Works
docker run -it --rm --name pilight --privileged pschmitt/hassio-addon-armhf-pilight
docker exec -it pilight pilight-send -p pollin -u 1 -s 1 -t

gollo:
Sure. I guess the next question is why this is an addon and not just a component/platform in main HA (which runs with ```--privileged```). I suppose I'm just establishing a definitive requirement to then base the discussion on.  I'm assuming the default position is to avoid having to do this if at all possible (so a 3rd party addon doesn't kill the the entire appliance), but others know more/better than me...

pschmitt:
I guess pilight *could* be implemented as a component, but that is a ton of work: just take a look at the [source code](https://github.com/pilight/pilight). 
There already is a client component in hass. But nah - why not just use it?

Anyway this `--privileged` requirement will certainly come up for other addons. Like for example `zabbix-agent`. Ideally the agent would need to run with `--privileged` *and* `--pid=host` which is not available to addons as well.

pvizeli:
For zabbix and official docs (https://hub.docker.com/r/zabbix/zabbix-agent/)
[...]

PR #202, opened 2017-09-29
Add support for kernel gpio interface

pvizeli:
Fix part of #201

### DIFF
--- a/API.md
+++ b/API.md
@@ -342,6 +342,7 @@ Get all available addons.
             "url": "null|url",
             "logo": "bool",
             "audio": "bool",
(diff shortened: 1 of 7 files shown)


API.md
```diff
@@ -342,6 +342,7 @@ Get all available addons.
             "url": "null|url",
             "logo": "bool",
             "audio": "bool",
+            "audio": "bool",
```
pschmitt: Shouldn't this here be `"gpio": "bool"`?

pvizeli: Yes I fix it. But that is only inside description of API

PR #266, opened 2017-12-04
Disable AppArmor

pvizeli:


frenck:
Just wondering, this is being added to the caps? Shouldn't this be added to the security opts?

From Docker-py docs:

> security_opt (list) – A list of string values to customize labels for MLS systems, such as SELinux.

https://docker-py.readthedocs.io/en/stable/api.html#module-docker.api.container

pvizeli:
Yes that is. We call privileged (hassio) as security options from docker

PR #298, opened 2018-01-07
Update panel for 0.80

pvizeli:


frenck:
The change is in there... now the Travis builds are failing though...

PR #299, opened 2018-01-07
Disable ipv6 / Clear search domain names

pvizeli:
It is default false, but if so we are sure for future.


frenck:
I doubt this would work. 
the `homeassistant` container runs on the host network, so that still applies.
(Some goes for add-ons running on the host network).

This goes south when the host network has IPv6 available.

PR #360, opened 2018-02-14
Add timezone to snapshot timestamp

c727:
```
old: 2018-02-14T15:13:46.391829
new: 2018-02-14T15:13:46.391829+00:00
```
...required to display the correct time on frontend (without dirty hacks)


pvizeli:
Can you export this as function into `utils/dt.py`?

PR #435, opened 2018-04-08
Bugfix, remove unsupported characters

pvizeli:
CC @frenck


frenck:
Unsupported?
They are supported, right?

https://microbadger.com/images/hassioaddons/shinobi-aarch64

pvizeli:
Yes, but not with docker API :(

frenck:
In that case; go ahead and merge it in!
It is not a big deal, since this is for local builds only.

PR #552, opened 2018-07-04
Fix problem with Repositories

pvizeli:
Implement:
Fix: #551 

Fixup:
Fix: https://github.com/home-assistant/hassio-addons/issues/303
Fix: #553

### DIFF
--- a/hassio/addons/git.py
+++ b/hassio/addons/git.py
@@ -91,10 +91,15 @@ class GitRepo(CoreSysAttributes):
             return False
 
         async with self.lock:
+            git_args = {
+                'depth': 1,
+                'update-shallow': True,
+            }
+
             try:
                 _LOGGER.info("Pull addon %s repository", self.url)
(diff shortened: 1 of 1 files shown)


frenck:
Nice one, that would save some disk space in some cases as well.

pvizeli:
@frenck please make a final review

hassio/addons/git.py
```diff
@@ -120,13 +135,7 @@ def __init__(self, coresys, url):
...
-        """Remove a custom addon."""
-        if self.path.is_dir():
-            _LOGGER.info("Remove custom addon repository %s", self.url)
-
-            def log_err(funct, path, _):
-                """Log error."""
-                _LOGGER.warning("Can't remove %s", path)
-
-            shutil.rmtree(str(self.path), onerror=log_err)
+    def _remove(self):
```
frenck: This should be `def remove(self):` right?

PR #575, opened 2018-07-19
Leverage access and refresh tokens if available

balloob:
Authenticate with Home Assistant using access tokens.

**Not tested yet. Assumes Python 3.7.**

### DIFF
--- a/hassio/homeassistant.py
+++ b/hassio/homeassistant.py
@@ -402,6 +402,8 @@ class HomeAssistant(JsonConfig, CoreSysAttributes):
                 yield resp
                 return
 
+        raise HomeAssistantAPIError()
+
     async def check_api_state(self):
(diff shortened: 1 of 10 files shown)


hassio/homeassistant.py
```diff
@@ -317,51 +332,90 @@ def in_progress(self):
...
+            if resp.status != 401 or not self.refresh_token:
+                yield resp
+                return
+
+            # Access token expired
+            self.access_token = None
+            async with self.make_request(
+                    method, path, json, content_type, data, timeout,
+            ) as resp:
+                yield resp
```
pvizeli: I think we should loop that and use the token refresh on start of that loop. If home assistant have a problem, we crash on ARM after some reqursive calls while the stack size is very small.

PR #584, opened 2018-07-21
Bugfix passwrod

pvizeli:


pvizeli:
@balloob hmm, my test env was corrupt, otherwise I had see it .)

PR #631, opened 2018-08-07
Add SYS_RESOURCE to list of valid privileges

ndarilek:
I'm trying to create a PulseAudio addon. In order to enable real-time scheduling, the addon minimally needs the SYS_RESOURCE capability. SYS_NICE is only sufficient for high-priority. Unfortunately, I'm experiencing lots of underruns, and strongly suspect the lack of real-time scheduling in the addon as the cause.


pvizeli:
Thanks

PR #651, opened 2018-08-15
Improve docker build cache for supervisor

pvizeli:


Dockerfile
```diff
@@ -1,15 +1,17 @@
 ARG BUILD_FROM
 FROM $BUILD_FROM
 
-# Setup base
+# Install base
+apk add --no-cache \
```
frenck: Missing `RUN`?

PR #659, opened 2018-08-19
Detect running record migration

pvizeli:


hassio/homeassistant.py
```diff
@@ -452,21 +456,35 @@ def check_port():
...
+            # Running DB Migration
+            if migration_file.exists():
+                if not migration_progress:
+                    migration_progress = True
+                    _LOGGER.info("HomeAssistant record migration in progress")
+                continue
+            elif migration_progress:
+                migration_progress = False
+                start_time = time.monotonic()
+                _LOGGER.info("HomeAssistant record migration done")
```
balloob: We should only log this if migration_process was not True.

pvizeli: that will be on line 473

PR #713, opened 2018-09-24
Add log to blacklist / reduce free calls

pvizeli:


frenck:
It would be cool if there is 1 API call for all kinds of info.
For example `/versions` or `/info` (just examples) with the following:
- Home Assistant version
- OS type / version
- Hassio Supervisor version
- Beta/dev/stable channel information
- The architecture of the system

In general, displaying this kind of information in the add-on logs, greatly helps debugging issues.

Issue #656, opened 2018-08-18
Option to get access to VideoCore


frenck:
Currently, the only way to gain access to the VideoCore of an Raspberry PI, is to add the follwing devices to your add-on configuration:

- `/dev/vchiq`
- `/dev/vcsm`

The problem with this is that "other" devices don't have these. This will cause the add-on to fail to start on those devices.

Possible solutions:
- Add in another switch to enable the VideoCore capabilities and only actually use them on a Pi (or other supported platforms).
- A more general solution: Check and filter the devices list of an add-on before trying to start it.

frenck:
Related to #184

Issue #184, opened 2017-09-11
rpi_camera 'Image not available'


robmarkcole:
I'm seeing 'Image not available' under 0.53 on Pi-zero-W and pi-3 with `platform: rpi_camera`

![image](https://user-images.githubusercontent.com/11855322/30260262-388516a2-96be-11e7-9c47-093ef40d8cf3.png)

pvizeli:
Maybe it use the RT/XT like zwave module and GPIO. Try that: https://home-assistant.io/hassio/zwave/

pvizeli:
Try: `curl -d '{"devices": ["ttyAMA0"]}' http://hassio/homeassistant/options` to map the GPIO interface into homeassistant container

pvizeli:
I think, if we solve this with that command, I add a autodetect for `ttyAMA0` device and remove this command

pvizeli:
You need reboot the host or use the new function of hassio 0.61 to stop/start the homeassistant container with ssh

pvizeli:
Try also first to update config.txt:
![image](https://user-images.githubusercontent.com/15338540/30282550-e8b3d444-9715-11e7-8ac8-378c6f8b95a7.png)

robmarkcole:
Hi, where is config.txt? I have samba share setup as below.
Cheers

![image](https://user-images.githubusercontent.com/11855322/30288497-db21f2e2-9720-11e7-8360-f77705c8a053.png)

pvizeli:
Yeah you need do that on SD card. But @gollo have try it and mean that will not work

gollo:
I'm looking into it when I have the time...

gsbrown:
Same issue, hoping for a fix but have a workaround with a dropcam for now

pvizeli:
The problem look like with a blob firmware that need to be load and docker/alpine. But It is hard to discovery that without hardware :(

gsbrown:
Anyone figure this out yet?

jasiekzar:
I have the same problem - I have enabled the camera in the config file on the sd card but still "image not available"...

thesplitter:
I have tried editing config.txt to enable the camera and gpumem to 128 (as required) but the result is the same

gollo:
The root problem is that the module does not run in the Alpine docker
container.  Ideally, if I get time, I will attempt to get it to  run in
native Alpine on the Pi, then move to attempting to get it to run in the
container.  Attempting to get it running in the container has not worked
[...]

PR #716, opened 2018-09-24
WIP: Adds devices existence checks before using

frenck:
Check devices if they exist before actually using them.

This allows an add-on developer to specify devices that might not be available on all platforms.
For example: We could add the Raspberry camera device to the SSH add-on, without breaking it on other platforms.

Fixes: #656
Ref: #184

### DIFF
--- a/hassio/docker/addon.py
+++ b/hassio/docker/addon.py
@@ -1,10 +1,10 @@
 """Init file for HassIO addon docker object."""
 import logging
 import os
+import re
 from pathlib import Path
 
 import docker
-import re
 import requests
 
 from .interface import DockerInterface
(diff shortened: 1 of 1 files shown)


pvizeli:
That don't work because, the supervisor run without any privileged.

You need go the way with hardware.py and /sys/ or /prog/

frenck:
🤦‍♂️ 

Will revisit it. Meanwhile, I'll add the 🚧 WIP tag.

pvizeli:
Small input, maybe we can use the `machine` tag. Like we do on audio to resolve the name of input. I think we need also add a functionality that you can add a config option like `machine: [rpi]` and this can only be installed on a rpi platform.

frenck:
I'm closing this PR. It has to be rewritten anyway and I'm short on time as well.

PR #743, opened 2018-10-06
Adds support for privilege DAC_READ_SEARCH

frenck:
This PR adds support for DAC_READ_SEARCH privilege.

It allows a bypass on file read permission checks and directory read and execute permission checks.

This is for example required to be able to mount Samba shares inside Docker containers.


frenck:
I've added `DAC_READ_SEARCH` to lower the rating on use since it does bypass permissions.

Since this is mainly used by mounts, and mounts require SYS_ADMIN, I've given it the same rating value.

PR #747, opened 2018-10-09
Add missing tinker board

mbo18:
Since I can't install a new add-on since hassio release 134 because of error 'addon not supported on armhf'. I think that support for tinker board is missing


homeassistant:
Hi @mbo18, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;hassio#747).

Thanks!

PR #760, opened 2018-10-14
Fix proxy handling with failing connection

pvizeli:


hassio/api/proxy.py
```diff
@@ -117,7 +113,7 @@ def _check_access(self, request):
 
         try:
             client = await self.sys_websession_ssl.ws_connect(
-                url, heartbeat=60, verify_ssl=False)
+                url, heartbeat=30, verify_ssl=False)
```
balloob: We still need this?

pvizeli: I think it make no pains and it could help to detect faster wrong connections

PR #814, opened 2018-11-16
Provide options for legacy add-ons

cschroeter:
Solves https://github.com/home-assistant/hassio/issues/805

### DIFF
--- a/hassio/docker/addon.py
+++ b/hassio/docker/addon.py
@@ -81,18 +81,11 @@ class DockerAddon(DockerInterface):
         """Return environment for Docker add-on."""
         addon_env = self.addon.environment or {}
 
-        # Need audio settings
-        if self.addon.with_audio:
-            addon_env.update({
-                'ALSA_OUTPUT': self.addon.audio_output,
-                'ALSA_INPUT': self.addon.audio_input,
-            })
-
         # Provide options for legacy add-ons
         if self.addon.legacy:
             for key, value in self.addon.options.items():
                 if isinstance(value, (int, str)):
-                    addon_env.update({key: value})
+                    addon_env[key] = value
                 else:
                     _LOGGER.warning(
                         "Can not set nested option %s as Docker env", key)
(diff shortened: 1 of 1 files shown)


hassio/docker/addon.py
```diff
@@ -88,6 +88,12 @@ def environment(self):
                 'ALSA_INPUT': self.addon.audio_input,
             })
 
+        # Provide options for legacy add-ons
+        if self.addon.legacy:
+            addon_env.update({
+                **self.addon.options
+            })
+
```
pvizeli: ENV support only one depth. We should make sure, that there is no nested struct inside. `**{value: key in value, key for options.items() if isinstance(key, str, int)}`

cschroeter:
Mh if you flag your add-on with legacy (non-hassio) the options you pass may not be nested. I would rather check this in the validation part of the add-on than in the Docker part. 
What you say?

pvizeli:
Sure, that was my second opinion.

pvizeli:
https://github.com/home-assistant/hassio/blob/dev/hassio/addons/validate.py#L111

You can create it like `vol.Schema(vol.All({...}, _validate_legacy, ...)` and the function `validate_legacy` trow a `vol.Invalid` if legacy is set but the options is not valid.

cschroeter:
Jus because I set the flag _legacy_ my add-ons with nesteds options won't start?  Doesn't legacy add-ons also recieve the volumen with those options.json? It feels like a misuse of this flag.

I also don't feel comfortable passing those options as Docker env with out further notice. Maybe we could introduce something like a new flag - "pass options as docker env". What you say? If you don't agree I would vote to just pass the first level of options instead of deny a user that flags his add-on _legacy_  to use nested options

hassio/docker/addon.py
```diff
@@ -88,6 +88,15 @@ def environment(self):
                 'ALSA_INPUT': self.addon.audio_input,
             })
 
+        # Provide options for legacy add-ons
+        if self.addon.legacy:
+            for key, value in self.addon.options.items():
+                if isinstance(value, (int, str)):
+                    addon_env.update({key: value})
```
pvizeli: `addon_env[key] = value`

pvizeli:
Thanks

PR #832, opened 2018-11-21
Add generic platforms

deece:
Thinking a bit more about @pvizeli 's suggestion about 'orangepi-h5', I came to the conclusion that having some clearly labelled generic platforms would be helpful.

I spent quite some trying trying to figure out how the specify which machine to use on my (then unsupported) Orange Pi Prime, before realising that the qemu machine was fairly generic.

This patch adds some generic platforms that can either be used outright, or as a template for new boards.

### DIFF
--- a/hassio/addons/validate.py
+++ b/hassio/addons/validate.py
@@ -60,6 +60,7 @@ ARCH_ALL = [
 ]
 
 MACHINE_ALL = [
+    'generic-aarch64', 'generic-amd64', 'generic-armhf', 'generic-i386',
     'intel-nuc',
     'odroid-c2', 'odroid-xu',
     'orangepi-prime',


pvizeli:
The generic variants are the qemuXXX versions. I don't know why all call that qemu but that is very common.

The naming for the machine depends on bootloader and device-tree to boot the Linux kernel. So any image has his machine type. I've planned to make HassOS also for some OrangePis, but currently, without any developer they won't help me on that, it's not possible.

deece:
I would expect the qemu* platforms would be used for testing/CI, which has some overlap, but may not necessarily be the same as what is required for a template.

An alternative would be to improve the documentation (which I'm happy to do) to direct users to the qemu* platforms, however, I do think this is a less valuable approach, as the goals for a generic vs testing platform may diverge.

What assistance do you require? I've got plenty of Orange Pis on hand.  ATM, I'm running the docker images under Armbian.

frenck:
@deece the qemu version are not testing...

deece:
Is there any reason they shouldn't be named generic then?

pvizeli:
> Is there any reason they shouldn't be named generic
-> it's common to call this platforms qemu* in the world of embedded systems and low-level integration. Don't ask me why.

How good are your skills in system integration, build systems, kernel building and bootloader debugging (uart)? I can point you how you can port HassOS to orangepi-prime.

deece:
I should point out that my goal of having a generic config is not for HassOS, but for those who want to run the docker images on top of other Linux distros.

pvizeli:
@deece please open an Issue on HassOS for orangepi-prime if you want help to porting HassOS and create an image for it. I will point to the points that needed to make this available. Should be a 1-2 days of work.

PR #838, opened 2018-11-27
fix log message if there is no update available

sholzmayer:


pvizeli:
https://travis-ci.org/home-assistant/hassio/builds/460187909?utm_source=github_status&utm_medium=notification

I think we should make that as warning

hassio/core.py
```diff
@@ -66,7 +66,7 @@ def __init__(self, coresys):
             if await self.sys_supervisor.update():
                 return
         else:
-            _LOGGER.info("Ignore Hass.io auto updates on dev channel")
+            _LOGGER.info("Your are either on DEV-channel or there is currently no Hass.io update available")
```
pvizeli: maybe `elif self.sys_dev:`?

PR #841, opened 2018-11-29
Improve development instructions

OttoWinter:
This took me a while to figure out... Creating a PR so that others don't need to repeat this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ The development of the supervisor is a bit tricky. Not difficult but tricky.
 
 - You can use the builder to build your supervisor: https://github.com/home-assistant/hassio-build/tree/master/builder
 - Go into a HassOS device or VM and pull your supervisor.
-- Set the developer modus on updater.json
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -20,7 +20,7 @@ The development of the supervisor is a bit tricky. Not difficult but tricky.
 
 - You can use the builder to build your supervisor: https://github.com/home-assistant/hassio-build/tree/master/builder
 - Go into a HassOS device or VM and pull your supervisor.
-- Set the developer modus on updater.json
+- Turn on the developer modus in the `updater.json` file (usually `/usr/share/hassio/updater.json`) by setting `channel` to `dev`
```
pvizeli: There is an API call for it. The file can be differents places. Maybe we can maintain the API call and a small comment to `updater.json` inside hassio data folder?

pvizeli:
`hassio supervisor options --o channel=dev` (could be changed with new 2.0)

README.md
OttoWinter: Oh, you mean POST `/supervisor/options`? Yeah that would also be possible. Though I personally find editing a file quicker than having to type out a full curl command (+ finding the ip address of the VM/machine)

PR #892, opened 2019-01-15
Support armv7 and allow support of multible arch types per CPU

pvizeli:


hassio/__main__.py
```diff
@@ -9,21 +9,25 @@
 _LOGGER = logging.getLogger(__name__)
 
 
-def attempt_use_uvloop():
+def initialize_event_loop():
     """Attempt to use uvloop."""
     try:
         import uvloop
         asyncio.set_event_loop_policy(uvloop.EventLoopPolicy())
```
balloob: `uvloop.install()` has been introduced in [uvloop 0.12](https://github.com/MagicStack/uvloop/releases/tag/v0.12.0)

PR #944, opened 2019-03-09
Include serial device node links in container device mapping to allow…

cadwal:
… for persistent names in the HA serial config

This is intended as a fix for issue #911 and I might have managed to compile test it but no more.

I claim no credit for it either, kpine came up with the solution in the issue discussion.

PR #956, opened 2019-03-18
Revert dev link

pvizeli:
revert #944 

That will not work inside a container. Maybe if we map all rules into the container and do the same as on host, it could work.

Anyway, we move over to privileged modus for supervisor. Maybe, that fix also some issue


cadwal:
So this fails because the code is actually being run in the supervisor container, not the on base OS?

pvizeli:
yes

PR #1073, opened 2019-05-04
Fix validation of image and last_version

carstenschroeder:


hassio/api/homeassistant.py
```diff
@@ -41,8 +41,8 @@
 SCHEMA_OPTIONS = vol.Schema(
     {
         vol.Optional(ATTR_BOOT): vol.Boolean(),
-        vol.Inclusive(ATTR_IMAGE, "custom_hass"): vol.Maybe(vol.Coerce(str)),
-        vol.Inclusive(ATTR_LAST_VERSION, "custom_hass"): vol.Any(None, DOCKER_IMAGE),
+        vol.Inclusive(ATTR_IMAGE, "custom_hass"): vol.Any(None, DOCKER_IMAGE),
```
pvizeli: Use also `Maybe`, because the other one is the old way

PR #1131, opened 2019-06-19
Append devlinks to serial dev_list

lenisko:
Would solve https://github.com/home-assistant/hassio/issues/1111 (at least from `hassio` side).


pvizeli:
Please add a screenshot that this will work. I mean it's a bit complexer because you need also spawn a udev service in background to handle the udev simlinks.

lenisko:
You are correct, I didn't noticed we don't have working instance of `udevd` inside hassio docker.

We could solve that by installing `eudev` package inside hassio docker image. 
After that we have to run `udevd --daemon && udevadm control --reload-rules && udevadm trigger`

And devices should be passed correctly.

2019-06-21_18-33-58-firefox

What do you think @pvizeli ?

lenisko:
I have pushed few more changes.

* related to `Dockerfile`, so `udev` service and initialization.
* to include only symlinks starting with `/dev/serial/by-id`. `/dev/serial/by-path` was erroring on `str` verification. Propably due to `:` in path?

pvizeli:
We should do it like this: https://github.com/home-assistant/hassio-homeassistant/tree/master/base and add a entry point

pvizeli:
I think also we should provide as new struct and not inside serial devices. I mean that I can remember me, that you can run into issues if you map device links with `--device` command, but my last experiments with that is longer away.. That mean, the add-on need self run a udev inside container (like we had a long time before).

lenisko:
@pvizeli That's a good question.
I guess we need to test it with dummy add-on, as `deCONZ` is using `debian:stretch-slim`.
I will find some time tomorrow, or maybe next week to check it.

By entry point you mean, we should create `sh` script to initialize `udev` and run `hassio` module instad using `RUN` in `Dockerfile`? (I'm not an expert in dockers).

pvizeli:
Yeah you can copy it from my link and add an entry point script without touch the CMD (don't forget file chmod a+x) . I can also help to test after you add the entry point. In example the core addon ssh would be a good test case since they include serial devices

lenisko:
@pvizeli I have added those changes
- added back `eudev-libs` to packages, same as in your `Dockerfile`
- `entry.sh` has correct permission in repository

I didn't had time to test it yet

PR #1177, opened 2019-07-27
enable armv7 and armhf for qemuarm-64 machine

agners:
The aarc64 Qemu machine supports the 32-bit ARM state as well. This
is also useful since this machine is used on machines without an
official machine architecture in hass.io.

### DIFF
--- a/hassio/arch.json
+++ b/hassio/arch.json
@@ -49,7 +49,9 @@
         "armhf"
     ],
     "qemuarm-64": [
-        "aarch64"
+        "aarch64",
+        "armv7",
+        "armhf"
     ],
     "intel-nuc": [
         "amd64",


agners:
I am aware that some aarch64 CPU do not support aarch32. I found those are mostly CPU's more server focused.

I would suggest to go the other way around: Add a new generic aarch64 machine which only supports aarch64. Qemu aarch64 supports aarch32: I able to run aarch32 binaries when using  `qemu-system-aarch64`... Also, we have `i386` in `qemux86-64`...

How about creating `aarch64-only` machine? We could also consider just enabling it. Most things will just work anyway since `aarch64` is the default architecture then. Of course, it is not very nice if you install a armv7 container and then it does not work. However, if a user on a generic aarch64 picks the wrong machine, he ends up with the same situation...

pvizeli:
The naming was coming from first base images 2.5 years ago. They were from Resin.io, and later we crated own base images. Today, I would love do not stick with those names. Now we have so many devices out, that we need forward on that schema. Our qemu variants try to have the most significant match for arch support. The arm32 are also only armv6 because of that work even on armv7.

If I look to our stats, there would be more as 10k devices, and they were not run correctly after these changes. So there is no options to merge this.

But, we can add an option to config, that allows to overwrote the arch detections. That's fine for me. Another possible solution would be an `qemuarm32-64`.

agners:
> The naming was coming from first base images 2.5 years ago. They were from Resin.io, and later we created our base images. Today, I would love do not stick with those names. Now we have so many devices out, that we need forward on that schema. Our Qemu variants try to have the most significant match for arch support. The arm32 are also only supporting armv6 because of that work even on armv7.

Ok yeah I understand we have a user base which uses that machine, so we need to be careful what we change here.

> 
> If I look to our stats, there would be more as 10k devices, and they were not run correctly after these changes. So there are no options to merge this.

I don't quite get why they would not run correctly anymore. As far as I understand changing the supported arch would at first change nothing for them. It would however enable installing containers which were not supported before (e.g. deCONZ), which probably will work for 9k of those devices since they actually support aarch32... So one would argue we actually correct something for 9k of users.

> But, we can add an option to config, that allows to overwrote the arch detections. That's fine for me. Another possible solution would be an `qemuarm32-64`.

I kinda really dislike the idea creating another qemu machine, I think its confusing.

I rather prefer an option to config. Or the best solution which would be auto detection, but I am not sure how that can be done.

pvizeli:
> the best solution which would be auto detection, but I am not sure how that can be done.

I agree. For armv6/7 it's easy but this solution work not for aarch64 / aarch32. The option is maybe not beautiful but also not to bad. The CPU autodetection of docker or python/pip also work not perfect and end up in a lot of issues. That is the reason why we handle that a bit stricter.

agners:
FYI, I am still working on this. I have a potential solution using the `personality` syscall, but I am not sure if this is the right approach.

> The CPU autodetection of docker or python/pip works not perfect and end up in a lot of issues.

What auto detection was that exactly?

Meanwhile, I also learned quite something about aarch64 support, and wanted to report here since it is probably also interesting for the hass.io/HA community.

The main reason I am interested in running AArch32 kernels on my ARMv8/AArch64 machine is to run deCONZ. Hass.io officially supports only armhf (aka ARMv6 with hard-float) container, which is based off Raspian Debian stretch. After applying this patch, I was able to install and start the container from Hass.io, however the container immediately crashed (with empty `docker logs`! However, `dmesg` had some suspcious messages about a process exiting). I am 100% sure that the CPU supports AArch32 since `docker run -it --rm arm32v7/debian:buster-slim` works just fine! Also the relevant kernel config `CONFIG_COMPAT` which enables "Kernel support for 32-bit EL0" was enabled.

Debugging into it I started to realize that the processes exit due to an illegal instruction. The instructions were memory barrier calls such as `mcr p15, #0, r0, c7, c10, #5`. It turns out that those are deprecated already in ARMv7, and seemingly not supported in ARMv8/AArch32. However, it seems the kernel has mechanism to work around those missing instructions through the config symbol `CONFIG_ARMV8_DEPRECATED`. With that enabled, memory barriers etc. can get emulated. See also:
https://lists.debian.org/debian-kernel/2017/04/msg00345.html

I guess there might be even more reasons why a (older/obscure) ARM binary does not run on ARMv8/AArch32 anymore...

tldr; It is not only about CPU support, it also depends on the binaries/container running and on the hosts kernel configuration...

pvizeli:
For deConz, there are some aarch64 binary around, maybe I can add support for it

agners:
@pvizeli yes, and I had it running. Quickly pushed my branches how I made it work, but probably needs a bit cleanup:
https://github.com/home-assistant/hassio-base/pull/27
https://github.com/home-assistant/hassio-addons/pull/649

However, the aarch64 version is not really useful right now as it has a 5 device limit:
https://github.com/dresden-elektronik/deconz-rest-plugin/issues/1588

I contacted support, and they confirmed the problem. It should be resolved in next version.

frenck:
aarch64 is limited for deCONZ and is still beta/alpha level.
Adding that to our collection will make it unusable/unstable on devices that support fallback to 32 bits variants. I would advise against it at this point.

pvizeli:
I agree, we can't add deconz 64bit official if that is working so bad. Rpi3/4 and Rock64 work pretty well with aarch32. But I think deConz work on a better aarch64 (pretty new) support.

agners:
@frenck stability is actually quite good, I had running it for a week in real-life environment :-) However, it only supports 5 devices right now, which is definitely a problem. So I agree, we should not push it as long as it is not on pair with 32-bit builds since it would screw those users.

However, I expect it to stay alpha for a while, is there a way we can make it opt-in? Maybe just create a new "deCONZ (alpha)" add-on?

agners:
Btw, I do run deCONZ armhf variant with `CONFIG_ARMV8_DEPRECATED` enabled, and it works. I get a bunch of dmesg errors, but I guess those are expected. A proper AArch64 would definitly be nicer:

```
[34188.581048] cp15barrier_handler: 81032 callbacks suppressed
[34188.581058] "QThread" (7327) uses deprecated CP15 Barrier instruction at 0xf68f83e8
[34188.581070] "QThread" (7327) uses deprecated CP15 Barrier instruction at 0xf688b460
[34188.581077] "QThread" (7327) uses deprecated CP15 Barrier instruction at 0xf6888c84
[34188.581084] "QThread" (7327) uses deprecated CP15 Barrier instruction at 0xf688b37c
[34188.581092] "QThread" (7327) uses deprecated CP15 Barrier instruction at 0xf6ca4368
[34188.581098] "QThread" (7327) uses deprecated CP15 Barrier instruction at 0xf6ca4380
[34188.581105] "QThread" (7327) uses deprecated CP15 Barrier instruction at 0xf6ca439c
[34188.581111] "QThread" (7327) uses deprecated CP15 Barrier instruction at 0xf6ca43b8
[34188.581118] "QThread" (7327) uses deprecated CP15 Barrier instruction at 0xf688745c
[34188.581125] "QThread" (7327) uses deprecated CP15 Barrier instruction at 0xf68874b8
```

pvizeli:
I close this PR, and we can discuss on the addons PR.

pezlin:
> I close this PR, and we can discuss on the addons PR.

I know that this is closed but I also have a constant flow of _[34188.581125] "QThread" (7327) uses deprecated CP15 Barrier instruction at 0xf68874b8_ in my HA Supervisor - Debian installation. Do you know any solution to that? I haven't found anything else than this post.

frenck:
@pezlin If you have an issue, please create one in the appropriate place. Thanks 👍

PR #1210, opened 2019-08-14
WIP: Only connect to D-Bus interface for rauc on HassOS

### LINKED ISSUE #560: Introspection of de.pengutronix.rauc fails on generic linux
# Problem/Motivation

Introspection of de.pengutronix.rauc fails on generic linux, which is to be expected since RAUC is only available on HassOS. 

Log:

```
18-07-06 16:52:50 INFO (MainThread) [hassio.utils.gdbus] Introspect de.pengutronix.rauc on /
18-07-06 16:52:50 ERROR (MainThread) [hassio.utils.gdbus] DBus return error: b'Error: GDBus.Error:org.freedesktop.DBus.Error.ServiceUnknown: The name de.pengutronix.rauc was not provided by any .service files\n'
18-07-06 16:52:50 WARNING (MainThread) [hassio.dbus.rauc] Can't connect to rauc
```

# Expected behavoir

I did not expect the error to show up in the logs. I can imagine it might confuse less technical users.

# Suggested solution

Only introspect for RAUC, in case the underlying system is HassOS.

frenck:
This stops Hass.io from connecting to the D-Bus interface for rauc, which causes a warning in the logs for non-HassOS users. While cosmetic, it throws off a lot of new users.

This PR checks if the underlying system is actually HassOS, before continuing connecting.
If the host is HassOS, the behavior is unchanged and it will still throw the warning when the connection fails.

Implementation decision reasoning: I've passed in coresys, which might be overkill, however, it does match the general pattern in the codebase. Another option is passing in the D-Bus hostname instance directly (since it actually only depends on that).

Fixes #560

### DIFF
--- a/hassio/dbus/__init__.py
+++ b/hassio/dbus/__init__.py
@@ -15,7 +15,7 @@ class DBusManager(CoreSysAttributes):
 
         self._systemd = Systemd()
         self._hostname = Hostname()
-        self._rauc = Rauc()
+        self._rauc = Rauc(coresys)
 
     @property
     def systemd(self):
--- a/hassio/dbus/rauc.py
+++ b/hassio/dbus/rauc.py
@@ -1,10 +1,13 @@
 """D-Bus interface for rauc."""
 import logging
 
-from .interface import DBusInterface
-from .utils import dbus_connected
+from cpe import CPE
+
+from ..coresys import CoreSys, CoreSysAttributes
 from ..exceptions import DBusError
 from ..utils.gdbus import DBus
+from .interface import DBusInterface
+from .utils import dbus_connected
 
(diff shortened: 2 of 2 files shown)


pvizeli:
We should filter `GDBus.Error:org.freedesktop.DBus.Error.ServiceUnknown` on gdbus and trow an ie. ServiceUnknown exception. On rauc interface, we can catch this and show information or nothing

frenck:
I do not agree on that, if there is an issue with the service on HassOS (whatever the reason), the ServiceUnkown should trigger the current behavior and should not be filtered.

pvizeli:
That is addressed if we print a message later after we catch the ServiceUnkown on rauc

pvizeli:
Addressed before

PR #1235, opened 2019-08-17
Replace Google DNS by Quad9, prefer CloudFlare

frenck:
Currently, the Hass.io DNS falls back to Google DNS & CloudFlare's one dot one dot one dot one.

Just hours after releasing Hassio DNS, people started commenting on the privacy concerns with Google DNS. 

This PR is addressing this privacy concern, by changing the defaults to the 2 biggest privacy-aware DNS providers out there: CloudFlare & Quad9. Both have an excellent track record. 

Cloudflare is considered the fastest DNS server in the world (based on dozens of different benchmarks), hence I've moved it to be the first one. Quad9 has about the same response speed as Google DNS, hence I've added Quad9 as a second.

TL;DR: Replace Google DNS by Quad9 in order to address possible privacy concerns.

### DIFF
--- a/hassio/const.py
+++ b/hassio/const.py
@@ -32,7 +32,7 @@ DOCKER_NETWORK = "hassio"
 DOCKER_NETWORK_MASK = ip_network("172.30.32.0/23")
 DOCKER_NETWORK_RANGE = ip_network("172.30.33.0/24")
 
-DNS_SERVERS = ["dns://8.8.8.8", "dns://1.1.1.1"]
+DNS_SERVERS = ["dns://1.1.1.1", "dns://9.9.9.9"]
 DNS_SUFFIX = "local.hass.io"
 
 LABEL_VERSION = "io.hass.version"


nhorvath:
Why not just load the nameservers from DHCP and let people choose on their own?

tjorim:
People can still choose their own if they like (using the CLI).

The main reason for this switch (as I see it), are users with DNS resolution issues.
For example if installing the Pi-hole or AdGuard Home add-on and changing the DNS server on their router, pointing at their Hass.io installation.
This creates a loop and the user would experience a lot of issues.

It's a win-win really, people with the knowledge how to do things correctly can simply change the DNS to their own liking. People that want a quick and easy installation have less troubles.

frenck:
@nhorvath It is meant to be a fallback in case of issues.

Landrash:
Worth thinking about is local coverage of the chosen DNS servers for fallback. Having a DNS server in the US while in the EU results in rather bad loading times.

Also since this should be under "not expected behavior" this really should be documented in the next release patch notes.

frenck:
> Worth thinking about is local coverage of the chosen DNS servers for fallback. Having a DNS server in the US while in the EU results in rather bad loading times.

Both supply local entities all of the world. So this comment doesn't make sense to me.

Landrash:
@frenck Edited my comment during you writing yours. I hope the change to the comment is sufficient.
Comment in question was to address any other potential concerns I could think of. Hopefully it now covers that.

pvizeli:
@nhorvath if we finish the NetworkManager integration, we would also add the local dns server from dhcp to the list. CoreDNS can detect loops or not working settings and ignore that kind of entry.

Issue #1251, opened 2019-08-21
Custom DNS Server not overwriting default servers in Corefile.


elmigbot:
**Home Assistant release with the issue:**

181/97.2
**Operating environment (HassOS/Generic):**

Hassio installed on VM with docker
**Supervisor logs:**

**Description of problem:**

Despite following directions to configure custom DNS servers, I see in my firewall logs that hassio is using my server _and_ the default servers.

Commands: 
```
hassio dns options --servers dns://192.168.90.1
hassio dns restart

```
Upon inspection of the corefile in Hassio_DNS container, I see the issue: My DNS server is _added_ to the available list and prioritized.
```

.:53 {
    log
    errors
    hosts /config/hosts {
        fallthrough
    }
    template ANY AAAA local.hass.io hassio {
        rcode NOERROR
    }
    forward . dns://192.168.90.1 dns://1.1.1.1 dns://9.9.9.9 {
        except local.hass.io
        policy sequential
        health_check 10s
    }
}
```

Is there a command to delete the others? I want just mine :).

...and yes it seems my DNS server fails health check, kicking in the sequential policy. Although it seems fine, I will be looking into that too.

Thanks!

MattXcz:
try reboot

MattiasC:
I would also prefer that the system would not try to go outside my designated DNS server if it temporarily does not respond.

If the other DNS servers are deleted manually, will they be restored when that docker container is updated automatically?

I don't think it is a problem if hass.io cannot get a DNS response if the local DNS server is down since that problem will affect more devices as well and that would make the problem more obvious than if hass.io continues to work in some half-ok state which id did for me now that all my host names stopped working.

elmigbot:
Fixed and released in less than 24 hrs! Thank you!

wonderslug:
#1255 is not really a fix for this for a couple of reasons.

The problem is in [hassio/dns.py ](https://github.com/home-assistant/hassio/blob/937b31d84549aa33c721bf32242526b3cc3ad17d/hassio/dns.py#L218) which Always adds the DNS_SERVERS which are always defined as the [default servers](https://github.com/home-assistant/hassio/blob/937b31d84549aa33c721bf32242526b3cc3ad17d/hassio/const.py#L35)..  There is no way to not have the default servers in the list at this point. 


Like others, sometimes there is a hiccup in dns which forces the sequential policy.  When using internal only DNS names (like I have for cameras etc) this will fail.

[...]

PR #1262, opened 2019-08-24
Allow default DNS servers to be a fallback.

wonderslug:
This change set the priority of the DNS servers.

The logic is as follows:

1. If there are manually defined servers use those.  The end user has
manually set what the DNS servers should be and that should be honored.

2. If there are no manually defined DNS servers, use the local host
defined. Of note, Not all os types use NetworkManager and dbus so this
can show up empty

3. If there are no manual or no local host based DNS servers defined
use the default servers.

This will provide that the DNS should always be set to the users
preferences and only include the external cloud based ones if all other
options have been excluded.

This would address issues brought up in #1251.

### DIFF
--- a/hassio/dns.py
+++ b/hassio/dns.py
@@ -213,9 +213,25 @@ class CoreDNS(JsonConfig, CoreSysAttributes):
             _LOGGER.error("Can't read coredns template file: %s", err)
             raise CoreDNSError() from None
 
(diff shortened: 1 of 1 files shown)


pvizeli:
Please test you change before you start a PR or mark the PR as draft.

wonderslug:
Any particular reason this was closed?

PR #1345, opened 2019-10-23
Fix secrets containing unsupported types

### LINKED ISSUE #1344: Validator for secrets
We did only support the values they are on our current validator. If there are other types, it fails:
https://github.com/home-assistant/hassio/blob/master/hassio/secrets.py#L15

The solution, we use vol and filter not supported values instead of validating the hole list.

frenck:
Fixes #1344

Removed the use of vol at this point (since it is filtered, it became pretty useless I guess 🤷‍♂)

Removed NoneType as supported secrets value, but added in boolean types.

### DIFF
--- a/hassio/secrets.py
+++ b/hassio/secrets.py
@@ -2,17 +2,14 @@
 from datetime import timedelta
 import logging
 from pathlib import Path
-from typing import Dict
+from typing import Dict, Union
 
 from ruamel.yaml import YAML, YAMLError
-import voluptuous as vol
 
 from .coresys import CoreSys, CoreSysAttributes
 from .utils import AsyncThrottle
 
 _LOGGER: logging.Logger = logging.getLogger(__name__)
-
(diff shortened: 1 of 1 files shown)


hassio/secrets.py
```diff
@@ -55,6 +55,10 @@ def get(self, secret: str) -> str:
             yaml = YAML()
             data = await self.sys_run_in_executor(yaml.load, self.path_secrets) or {}
 
+            # Filter to only get supported values
+            # pylint: disable=unidiomatic-typecheck
+            data = {k: v for k, v in data.items() if type(v) in (int, float, str)}
+
```
pvizeli: I'm fine to use bool also. Not sure if `isInstance(v, (int, float, str))` solve you lint issue

frenck: Yes, that will solve it.

frenck: Done.

PR #1359, opened 2019-11-02
Add support for specifying buildargs in the config file

snowzach:
So I'll preface this with, I'm not a python programmer and I'm not really sure how to test this first but I took a stab. 

I want to be able to specify buildargs from the config. Mainly I want the users of my docker add-on to be able to pick from a handful of other images based on hardware/cpu features they may have on their platfrom running hass.io.


pvizeli:
That is not how the system works. We deliver ready to use images. Later we allow users to build images locally to make it simpler that developers don't need an image hub and prebuild all add-ons (is not user-friendly because it can take a long time to install the add-on, but it works). Users should also not edit the config.json because it's will be read once and overwritten.

Different features/images are different add-ons. If you want to edit files, you can also edit build.json they include the args.

pvizeli:
Look like this is stale

PR #1391, opened 2019-12-04
Allow lists in secrets.yaml

danielwelch:
Use case discussed here: https://github.com/danielwelch/hassio-zigbee2mqtt/issues/259

### DIFF
--- a/hassio/secrets.py
+++ b/hassio/secrets.py
@@ -2,7 +2,7 @@
 from datetime import timedelta
 import logging
 from pathlib import Path
-from typing import Dict, Union
+from typing import Dict, List, Union
 
 from ruamel.yaml import YAML, YAMLError
 
@@ -18,7 +18,7 @@ class SecretsManager(CoreSysAttributes):
     def __init__(self, coresys: CoreSys):
         """Initialize secret manager."""
         self.coresys: CoreSys = coresys
-        self.secrets: Dict[str, Union[bool, float, int, str]] = {}
+        self.secrets: Dict[str, Union[bool, float, int, str, List[int]] = {}
 
     @property
     def path_secrets(self) -> Path:
@@ -55,7 +55,7 @@ class SecretsManager(CoreSysAttributes):
 
             # Filter to only get supported values
             self.secrets = {
-                k: v for k, v in data.items() if isinstance(v, (bool, float, int, str))
+                k: v for k, v in data.items() if isinstance(v, (bool, float, int, str, list))
             }
 
         except YAMLError as err:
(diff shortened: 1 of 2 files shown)


hassio/secrets.py
```diff
@@ -18,7 +18,7 @@ class SecretsManager(CoreSysAttributes):
     def __init__(self, coresys: CoreSys):
         """Initialize secret manager."""
         self.coresys: CoreSys = coresys
-        self.secrets: Dict[str, Union[bool, float, int, str]] = {}
+        self.secrets: Dict[str, Union[bool, float, int, str, List[int]] = {}
```
frenck: Why is this only a list of integers?

danielwelch: No reason whatsoever, other than that’s all that my add-on’s specific use case would require. I’m happy to expand it to whatever you all deem acceptable.

pvizeli:
You need also to address the CI issues. As next, the current implementation on `addon/validate.py` does support only single types. If you want list support, you need to extend the validator to work correctly with this.

danielwelch:
I guess my proposal would be to move this: https://github.com/home-assistant/hassio/blob/f5c171e44fd8dd4333b7a94b9b846a707401eb90/hassio/addons/validate.py#L359-L364

out of `_single_validate`, and instead do something like this

```python
def validate(struct):
        """Create schema validator for add-ons options."""
        options = {}

        # read options
        for key, value in struct.items():
            # Ignore unknown options / remove from list
            if key not in raw_schema:
                _LOGGER.warning("Unknown options %s", key)
                continue

           # Lookup secret
           if str(value).startswith("!secret "):
               secret: str = value.partition(" ")[2]
               value = coresys.secrets.get(secret)
           if value is None:
               raise vol.Invalid(f"Unknown secret {secret}")

            typ = raw_schema[key]
            try:
                if isinstance(typ, list):
                    # nested value list
                    options[key] = _nested_validate_list(coresys, typ[0], value, key)
                elif isinstance(typ, dict):
                    # nested value dict
                    options[key] = _nested_validate_dict(coresys, typ, value, key)
                else:
                    # normal value
                    options[key] = _single_validate(coresys, typ, value, key)
            except (IndexError, KeyError):
                raise vol.Invalid(f"Type error for {key}") from None

        _check_missing_options(raw_schema, options, "root")
        return options

    return validate
```

pvizeli:
Could work, please add some tests to verify that it works. Thanks

hassio/secrets.py
```diff
@@ -55,7 +57,9 @@ def get(self, secret: str) -> Union[bool, float, int, str]:
 
             # Filter to only get supported values
             self.secrets = {
-                k: v for k, v in data.items() if isinstance(v, (bool, float, int, str))
+                k: v
+                for k, v in data.items()
+                if isinstance(v, (bool, float, int, str, list))
```
frenck: The contents of lists are not filtered/checked this way...

pvizeli: Since we support complex types and use schemes validation, we don't need the type check here?

pvizeli:
Any future plan on this?

PR #1392, opened 2019-12-05
Fix grammatical issue in log message

issacg:
rebuilded (valid, but archaic) --> rebuilt

### DIFF
--- a/hassio/addons/__init__.py
+++ b/hassio/addons/__init__.py
@@ -245,7 +245,7 @@ class AddonManager(CoreSysAttributes):
             raise AddonsError() from None
         else:
             self.data.update(store)
-            _LOGGER.info("Add-on '%s' successfully rebuilded", slug)
+            _LOGGER.info("Add-on '%s' successfully rebuilt", slug)
 
         # restore state
         if last_state == STATE_STARTED:


issacg:
Thanks for the super-fast response, as usual :)

Quick unrelated question (I know this is probably the wrong place to ask, but clueless as to where the right place is)

I've got a tiny 2-line addon whose sole purpose in life is to help me (as a developer) get the local LAN IP address of my hassos instance.  My use-case is for when I'm on a non-private network (for example shared work space - like WeWork) and so don't have a static IP, but *do* have a static hostname (accessing via nabu casa, for example).  This way I can get my local IP without needing to connect an external display...

I've never bothered even self-publishing it as it's usually just as painful to copy the files to the hassio device in question as it would be to manage my own private add-on repo for just one add-on.  

Do you think that it's something worth publishing to the community repo?  I ask as you're more intimate with the community than I am - maybe you have insight on how helpful (or not!) this would be to other people.  A one-click install would be easier than what I do now :D

Thanks!

frenck:
@issacg Join our Discord chat, we have a #devs_hassio channel where we can discuss and help you out!

https://www.home-assistant.io/join-chat

Issue #1316, opened 2019-09-23
Raspberry is still using public DNS servers


darootler:
Home Assistant release with the issue:
hassos: "2.11"
homeassistant: 0.99.2
machine: raspberrypi3
supervisor: "188"

Operating environment (Hass.io/Docker/Windows/etc.):
Hass.io

Component/platform:
DNS

Description of problem:
The raspberry is still using 1.1.1.1 and 9.9.9.9 as DNS servers:

![image](https://user-images.githubusercontent.com/11849583/65836971-31168300-e2f3-11e9-860d-3f5938c4ed7c.png)

I already set the dns server with the following command:

![image](https://user-images.githubusercontent.com/11849583/65409175-4ff2b200-dde7-11e9-981a-34cd803cb3c1.png)

Diaoul:
It looks like [there](https://github.com/home-assistant/hassio/issues/1232) [are](https://github.com/home-assistant/hassio/issues/1251) [multiple](https://github.com/home-assistant/hassio/pull/1262) [issues](https://github.com/home-assistant/hassio/issues/1239) on this topic.

I've read the code a bit and it seems there is no way to override those settings and have CoreDNS **not** use public DNS servers.
Indeed the [`DNS_SERVERS`](https://github.com/home-assistant/hassio/blob/master/hassio/const.py#L35) configuration [is used no matter what](https://github.com/home-assistant/hassio/blob/master/hassio/dns.py#L221).

Please put *local control and privacy first* :smile: :muscle:

jsiemek:
The commands to overwrite DNS settings are:

hassio dns options --servers dns://192.168.90.1
hassio dns restart

darootler:
> The commands to overwrite DNS settings are:
> 
> hassio dns options --servers dns://192.168.90.1
> hassio dns restart

As i already mentioned above this command does not really work.

Regards
Richard

bart1970:
I have the same experience that it doesn't work. See also:
https://community.home-assistant.io/t/working-of-dns-in-container/150857

matt-FFFFFF:
I also have this issue, the ```corefile``` generated by the supervisor does not prefer the admin-defined servers implemented using the ```--servers dns://w.x.y.z```.

In my case the order of servers is DHCP -> 1.1.1.1 -> AdminDefined -> 9.9.9.9

This seems a very strange way of doing things. A better way would perhaps be:

[...]

PR #1399, opened 2019-12-15
Prefer admin defined DNS

matt-FFFFFF:
Small change to prefer admin defined DNS servers, that is those defined using ```hassio dns options --servers dns://1.2.3.4``` 

See #1316 for discussion.

### DIFF
--- a/hassio/dns.py
+++ b/hassio/dns.py
@@ -199,7 +199,7 @@ class CoreDNS(JsonConfig, CoreSysAttributes):
             raise CoreDNSError() from None
 
     async def reset(self) -> None:
-        """Reset DNS and hosts"""
+        """Reset DNS and hosts."""
         # Reset manually defined DNS
         self.servers.clear()
         self.save_data()
--- a/hassio/dns.py
+++ b/hassio/dns.py
@@ -215,12 +215,12 @@ class CoreDNS(JsonConfig, CoreSysAttributes):
             _LOGGER.error("Can't read coredns template file: %s", err)
             raise CoreDNSError() from None
 
-        # Prepare DNS serverlist: Prio 1 Local, Prio 2 Manual, Prio 3 Fallback
+        # Prepare DNS serverlist: Prio 1 Manual, Prio 2 Local, Prio 3 Fallback
         local_dns: List[str] = self.sys_host.network.dns_servers or ["dns://127.0.0.11"]
         _LOGGER.debug(
-            "local-dns = %s, config-dns = %s, backup-dns = %s",
-            local_dns,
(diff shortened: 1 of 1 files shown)


matt-FFFFFF:
This is especially important for users that block outbound DNS or users that run split DNS, using the same namespace internally as externally.

E.g. LetsEncryot cert for hass.mydomain.com, internally this resolves to the RFC 1918 address of the hassio instance, externally resolves to the public IP.

pvizeli:
The issue is, that this property before 193 is set default to the predefined servers. Next version don't set this server.

That means, all instance use the backup dns server instead the dns server from DHCP/local system. If we solve this issue, I'm fine to change that to prefere manual DNS over local DHCP/local settings from host OS.

matt-FFFFFF:
As 193 is out now can this be included in 194?

pvizeli:
Nope, because only new installs are not affected from this.

However, that means only this PR needs smarter to avoid the issue, that doesn't mean we can not do it.

matt-FFFFFF:
I have added a test to compare ```self.servers``` to ```DNS_SERVERS```. If identical then self.servers is ignored.

pvizeli:
We should add on line `119` or so this check and clear the server list with a note that we remove that code on version 200. Good work, thanks!

matt-FFFFFF:
Hi - I have added comments to the reset method, is that what you meant? I could not see anything on ```119```.

pvizeli:
More like

```
async def load(...):
    ....
   # (After line 115) Fix dns server handling before 194 / Cleanup with version 200
   if DNS_SERVERS == self.servers:
      self.servers.clear()
      self.save_data()
````

**Comment by matt-FFFFFF** [comment]
Ok, done. Thanks

File: hassio/dns.py [hunk 1]
```diff
@@ -215,15 +222,28 @@ def _write_corefile(self) -> None:
             _LOGGER.error("Can't read coredns template file: %s", err)
             raise CoreDNSError() from None
 
-        # Prepare DNS serverlist: Prio 1 Local, Prio 2 Manual, Prio 3 Fallback
+        # Prepare DNS serverlist: Prio 1 Manual, Prio 2 Local, Prio 3 Fallback
+        # If Manual servers are identical to Fallback (was default before 194), then ignore manual
         local_dns: List[str] = self.sys_host.network.dns_servers or ["dns://127.0.0.11"]
+        if DNS_SERVERS == self.servers:
```
pvizeli: That now not happens anymore with your change above. So you can fallback to your first solution

matt-FFFFFF: Got it, thanks - removed condition in 1a8c1d5

PR #1425, opened 2020-01-09
Fix snapshot HA / remove API password

pvizeli:


frenck:
👍 Some Pylint errors to fix, and some mentions of the api_password in the API docs (<https://github.com/home-assistant/hassio-supervisor/blob/dev/API.md>).

PR #1451, opened 2020-01-25
Change rating for ingress add-on

pvizeli:
If an add-on is protected by Home Assistant / let spend +1

### DIFF
--- a/hassio/addons/utils.py
+++ b/hassio/addons/utils.py
@@ -43,6 +43,10 @@ def rating_security(addon: AddonModel) -> int:
     if addon.access_auth_api:
         rating += 1
 
+    # Home Assistant Protect WebUI
+    if addon.with_ingress:
+        rating += 1
+
     # Privileged options
     if any(
         privilege in addon.privileged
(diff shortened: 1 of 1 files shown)


hassio/addons/utils.py
```diff
@@ -43,6 +43,10 @@ def rating_security(addon: AddonModel) -> int:
     if addon.access_auth_api:
         rating += 1
 
+    # Home Assistant Protect WebUI
+    if addon.with_ingress:
```
frenck: Second thought, let's combine this with the `access_auth_api`, so adding both still doesn't get you +3

```suggestion
    if addon.with_ingress or addon.access_auth_api:
```

pvizeli: make sense

pvizeli: or make ingress +2?

frenck: Sure that is fine as well, that adds additional security.

```py
if addon.with_ingress or addon.access_auth_api:
   rating += 1

if addon.with_ingress
   rating += 1
```

Gives +2 for ingress, +1 for auth api, but +2 for both (instead of +3)

PR #1476, opened 2020-02-05
First clean renaming for smooth migration

pvizeli:


hassio/const.py
```diff
@@ -58,11 +58,13 @@
...
-HEADER_HA_ACCESS = "X-Ha-Access"
-HEADER_TOKEN = "X-Hassio-Key"
 COOKIE_INGRESS = "ingress_session"
 
-ENV_TOKEN = "HASSIO_TOKEN"
+HEADER_TOKEN = "X-Supervisor-Token"
+HEADER_TOKEN_OLD = "X-Hassio-Key"
+
+ENV_TOKEN_OLD = "HASSIO_TOKEN"
+ENV_TOKEN = "SYSTEM_TOKEN"
```
frenck: Isn't this just the supervisor token?

```suggestion
ENV_TOKEN = "SUPERVISOR_TOKEN"
```
?

pvizeli: yes thanks

PR #1516, opened 2020-02-19
Add video devices to video mapping

zewelor:
Allow to pass video* devices to containers. Should allow to use usb webcams / uvc tuners.


pvizeli:
Thanks

PR #1532, opened 2020-02-27
Fix homeassistant config check with overlay-s6

pvizeli:


supervisor/docker/homeassistant.py
```diff
@@ -65,7 +65,7 @@ def _run(self) -> None:
             hostname=self.name,
             detach=True,
             privileged=True,
-            init=True,
+            init=False,
```
frenck: Shouldn't we keep the init as we disable S6 with the blank entry point?

pvizeli: that is for HA itself

Issue #1556, opened 2020-03-03
Supervisor not starting after update to 207


Alexxtheonly:
**Operating environment (HassOS/Generic):**

Ubuntu 18.04  Supervisor 207

**Supervisor logs:**

Mar 03 15:30:58 hassio hassio-supervisor[7786]:     raise EnvironmentError('Could not create udev monitor')
Mar 03 15:30:58 hassio hassio-supervisor[7786]: OSError: Could not create udev monitor
Mar 03 15:30:58 hassio hassio-supervisor[7786]: 20-03-03 15:30:58 ERROR (MainThread) [asyncio] Unclosed client session
Mar 03 15:30:58 hassio hassio-supervisor[7786]: client_session: <aiohttp.client.ClientSession object at 0x7f3e87052bd0>
Mar 03 15:30:58 hassio hassio-supervisor[7786]: 20-03-03 15:30:58 ERROR (MainThread) [asyncio] Unclosed client session
Mar 03 15:30:58 hassio hassio-supervisor[7786]: client_session: <aiohttp.client.ClientSession object at 0x7f3e86e16150>
Mar 03 15:30:58 hassio hassio-supervisor[7786]: [cont-finish.d] executing container finish scripts...
Mar 03 15:30:58 hassio hassio-supervisor[7786]: [cont-finish.d] done.
Mar 03 15:30:58 hassio hassio-supervisor[7786]: [s6-finish] waiting for services.
Mar 03 15:30:58 hassio hassio-supervisor[7786]: [s6-finish] sending all processes the TERM signal.

**Description of problem:**
Supervisor doesn't start, see log above.

Just updated the supervisor to 207 via frontend, now it logs the `Could not create udev monitor` error

jsiemek:
Hi,

As i saw your post, I pressed update to 207 with shaking hand .
Works on my side:
![SU 207](https://user-images.githubusercontent.com/36767404/75794173-b6093b00-5d70-11ea-90ec-6a997a254044.png)

bratanon:
Same issue here. Running the generic installation on:
Distributor ID: Ubuntu
Description:    Ubuntu 18.04.4 LTS
Release:        18.04
Codename:       bionic

Alexxtheonly:
I don't know what caused it, but after 
`docker pull homeassistant/amd64-hassio-supervisor:202`
`docker tag homeassistant/amd64-hassio-supervisor:202 homeassistant/amd64-hassio-supervisor:latest`
[...]

PR #1566, opened 2020-03-04
Fix udev error without privileged

pvizeli:
Old generic installation can run SU without privileged and this break udev. Let show an error instead break the startup. Fix partial: #1556

### DIFF
--- a/supervisor/hwmon.py
+++ b/supervisor/hwmon.py
@@ -1,5 +1,6 @@
 """Supervisor Hardware monitor based on udev."""
 from datetime import timedelta
+from contextlib import suppress
 import logging
 from pprint import pformat
 from typing import Optional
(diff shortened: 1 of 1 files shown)


supervisor/hwmon.py
```diff
@@ -19,21 +20,28 @@ def __init__(self, coresys: CoreSys):
...
-        self.observer = pyudev.MonitorObserver(self.monitor, self._udev_events)
-        self.observer.start()
-
-        _LOGGER.info("Start Supervisor hardware monitor")
+        try:
+            self.monitor = pyudev.Monitor.from_netlink(self.context)
+            self.observer = pyudev.MonitorObserver(self.monitor, self._udev_events)
+            self.observer.start()
+        except OSError:
+            _LOGGER.fatal("No privileged to run udev. Update your installation!")
```
balloob:
```suggestion
            _LOGGER.fatal("Not privileged to run udev. Update your installation!")
```

supervisor/hwmon.py
```diff
@@ -19,21 +20,28 @@ def __init__(self, coresys: CoreSys):
...
+        else:
+            _LOGGER.info("Start Supervisor hardware monitor")
```
balloob:
```suggestion
            _LOGGER.info("Started Supervisor hardware monitor")
```

PR #1597, opened 2020-03-26
Use DoT as fallback

pvizeli:


supervisor/dns.py
```diff
@@ -208,24 +215,17 @@ def _write_corefile(self) -> None:
...
-        try:
-            corefile_template: Template = Template(COREDNS_TMPL.read_text())
-        except OSError as err:
-            _LOGGER.error("Can't read coredns template file: %s", err)
-            raise CoreDNSError() from None
-
         # Prepare DNS serverlist: Prio 1 Manual, Prio 2 Local, Prio 3 Fallback
-        local_dns: List[str] = self.sys_host.network.dns_servers or ["dns://127.0.0.11"]
-        servers: List[str] = self.servers + local_dns + DNS_SERVERS
+        local_dns: List[str] = self.sys_host.network.dns_servers or []
```
frenck: This change makes the NetworkManager even more mandatory than before.
We should adjust the hassio-installer to require it and we might run into people without it, to end up with problems in some cases.

(although we fallback, local things might not resolve correctly anymore).

pvizeli: I think we can revert that part for now

PR #1598, opened 2020-03-27
Added support for external DNS records

elad-bar:
Add file `hosts.config` to apply custom records,
Should use same naming convention as in /etc/host 

Custom records will be added to generated host file of CoreDNS

### DIFF
--- a/supervisor/dns.py
+++ b/supervisor/dns.py
@@ -57,6 +57,11 @@ class CoreDNS(JsonConfig, CoreSysAttributes):
         return Path(self.sys_config.path_dns, "hosts")
 
     @property
+    def hosts_custom_records(self) -> Path:
+        """Return Path to hosts_config."""
+        return Path(self.sys_config.path_dns, "hosts.config")    
+    
+    @property
     def servers(self) -> List[str]:
         """Return list of DNS servers."""
         return self._data[ATTR_SERVERS]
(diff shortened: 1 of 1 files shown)


homeassistant:
Hi @elad-bar, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;supervisor#1598).

Thanks!

pvizeli:
No person have access to the internal data structure outside off the development. So you need create a full UI based API around this helper and use our internal datastructure to store the records.

elad-bar:
@pvizeli, not sure I udnerstand you correctly,
How do you expect me to access the DNS records from API?
Is there documentation of how to do that?

I cannot set external DNS records and everytime the supervisor is being updated I lose all manually entered records.

Using that code, I can add an external file that never gets updated, load the records into the host file, if the custom file doesn't exists, it will ignore it.

thanks

pvizeli:
You need to write this API and store it as a record in the internal dns config.

elad-bar:
and then restart the DNS to start affact it, seems not reasonable thing to ask compare to the code written, if you don't want my pull request, don't use but please allow in a convient way to do that

elad-bar:
Or at least, allow using another method instead of CoreDNS

frenck:
> please allow in a convient way to do that

That is what pvizelli is proposing since the change you've made here is not publicly accessible. So your own motivation for this PR is not provided by this PR.

pvizeli:
1) Extend the supervisor API to handle custom dns entry
2) Store this with other dns jsons
3) Use that on the part we generate host
4) Make sure that a custom entry can't overwrite a system one

elad-bar:
Sorry, not familiar with supervisor api, will check it,
Thanks

elad-bar:
just to be aligned, by api you mean:
https://github.com/home-assistant/cli?

thanks

elad-bar:
or that one:
https://github.com/home-assistant/supervisor/blob/dev/supervisor/api/dns.py

pvizeli:
Your question to show that you did not understand how the supervisor work, and that makes it impossible for you to archive the goal. The supervisor is just a backend, and the webui or cli is the frontend of that. That is a pretty common why in software development to do things.

I close this PR now because I don't think there are somethings useful. However, I would love to see a PR with that feature. Maybe after days/months of study, you can make that working 👍 I'm happy to see a future PR.

elad-bar:
I read about it, hopefully will work on that during the weekend,
Need to expose in the API the endpoint that will allow to do that,
I will create an addon that will coomunicate with the supervisor API and publish list of custom records to the API same as each addon is doing during startup

Hope that I got it right

PR #1599, opened 2020-03-27
DNS loop protection

pvizeli:


supervisor/dns.py
```diff
@@ -209,16 +210,36 @@ def need_update(self) -> bool:
...
+
         await self.sys_addons.sync_dns()
 
+    async def loop_detection(self) -> None:
+        """Check if there was a loop found."""
+        log = await self.instance.logs()
+
+        # Check the log for loop plugin output
+        if b"plugin/loop: Loop" in log:
+            _LOGGER.fatal("Detect a DNS loop in local Network!")
```
frenck:
```suggestion
            _LOGGER.error("Detect a DNS loop in local Network!")
```

Since it is not fatal? We recover now...

pvizeli: yeah, is not fatal for us now

PR #1691, opened 2020-05-01
Update panel with new tabs

pvizeli:
Signed-off-by: Pascal Vizeli <pvizeli@syshack.ch>


brucek1642:
A quick question with the new update panel. For Snapshots, where did the refresh button go to? I was able to work around by doing a reboot to access the snapshot I had reloaded and needed to recover from.

PR #1736, opened 2020-05-21
Add Docker info & check

pvizeli:


pvizeli:
Yeah, this is just warnings, all that is logged as critical will be freeze later the system

PR #1741, opened 2020-05-21
Fix S6 supervisor not listening in finish script

frenck:
Noticed during shutdown of the supervisor in my dev environment:

```
20-05-21 10:02:19 INFO (MainThread) [__main__] Close Supervisor
[cont-finish.d] executing container finish scripts...
[cont-finish.d] done.
[s6-finish] waiting for services.
s6-svscanctl: fatal: unable to control /var/run/s6/services: supervisor not listening
[s6-finish] sending all processes the TERM signal.
[s6-finish] sending all processes the KILL signal and exiting.
```

This PR addressed that, by adjusting the S6 Overlay service finish script for the Supervisor.

### DIFF
--- a/rootfs/etc/services.d/supervisor/finish
+++ b/rootfs/etc/services.d/supervisor/finish
@@ -1,5 +1,8 @@
-#!/usr/bin/execlineb -S0
+#!/usr/bin/execlineb -S1
 # ==============================================================================
 # Take down the S6 supervision tree when Supervisor fails
 # ==============================================================================
+if { s6-test ${1} -ne 0 }
+if { s6-test ${1} -ne 256 }
+
 s6-svscanctl -t /var/run/s6/services
--- a/rootfs/etc/services.d/supervisor/finish
+++ b/rootfs/etc/services.d/supervisor/finish
@@ -1,8 +1,5 @@
-#!/usr/bin/execlineb -S1
+#!/usr/bin/execlineb -S0
 # ==============================================================================
 # Take down the S6 supervision tree when Supervisor fails
 # ==============================================================================
-if { s6-test ${1} -ne 0 }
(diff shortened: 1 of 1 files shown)


rootfs/etc/services.d/supervisor/finish
```diff
@@ -1,5 +1,8 @@
-#!/usr/bin/execlineb -S0
+#!/usr/bin/execlineb -S1
 # ==============================================================================
 # Take down the S6 supervision tree when Supervisor fails
 # ==============================================================================
+if { s6-test ${1} -ne 0 }
```
pvizeli, line 5: This prevents the OTA updater to close itself for systemd to create a new container from updated image

frenck: Right, because it closes down with a 0 exit code. 👍 
Will adjust.

frenck: Ok, so since the loop is stopped nicely, it will return exit 0.
Which is nice, but problematic here, as we can't tell the difference if the program is shut down by itself or something else.

The s6-svscan might be shutting down already, causing the error we see. However, this error doesn't make a difference. As we are either shutting down already OR killing it because of a crash. In both cases, the supervision tree needs to be taken down.

As there is no direct way to test the current state of the svscan, simply suppressing the error, in this case, will do. The result is the same either way.

pvizeli: why not just `-eq 0`?

pvizeli: however, I'm fine with that

frenck: Because it can still be a nice exit, causing the svscan already being terminated/shutting down (raising the same error).

PR #1745, opened 2020-05-22
Complete typehinting in CoreSys

frenck:
Completes the type hinting in CoreSys.

Added a None check to all getters to ensure the returned type is correct. While it adds a bit of code, it prevents NoneType checking in the rest of the codebase.

This file now passes MyPy without anything disabled.

### DIFF
--- a/supervisor/coresys.py
+++ b/supervisor/coresys.py
@@ -2,7 +2,7 @@
 from __future__ import annotations
 
 import asyncio
-from typing import TYPE_CHECKING, Optional
+from typing import TYPE_CHECKING, Any, Callable, Coroutine, Optional, TypeVar
 
 import aiohttp
 
@@ -33,6 +33,9 @@ if TYPE_CHECKING:
     from .tasks import Tasks
     from .updater import Updater
     from .plugins import PluginManager
+
+
+T = TypeVar("T")
 
 
 class CoreSys:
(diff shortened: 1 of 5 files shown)


frenck:
Right, Pylint fails. The only thing I didn't check 🤦

supervisor/coresys.py
```diff
@@ -549,10 +594,12 @@ def sys_hassos(self) -> HassOS:
...
 
-    def sys_run_in_executor(self, funct, *args) -> asyncio.Future:
+    def sys_run_in_executor(
+        self, funct: Callable[..., T], *args: Any
+    ) -> Coroutine[Any, Any, T]:
         """Add an job to the executor pool."""
         return self.sys_loop.run_in_executor(None, funct, *args)
 
-    def sys_create_task(self, coroutine) -> asyncio.Task:
+    def sys_create_task(self, coroutine: Coroutine) -> asyncio.Task:
```
pvizeli, line 603: I means this type work like the Callable one?

frenck: Basically, yes, `create_task` expects a Coroutine.

Since no arguments are taken by this method, both coroutines and coroutine functions are accepted.

e.g.,:

```py
if asyncio.iscoroutine(coroutine):
   return self.sys_loop.create_task(coroutine)

if asyncio.iscoroutinefunction(coroutine):
   return self.sys_loop.create_task(coroutine)
```

In other cases, it should be run in the executor.

PR #1754, opened 2020-05-26
Detect lxc

pvizeli:


supervisor/core.py
```diff
@@ -48,6 +48,10 @@ async def connect(self):
                 "Docker version %s is not supported by Supervisor!",
                 self.sys_docker.info.version,
             )
+        elif self.sys_docker.info.inside_lxc:
+            self.healthy = False
+            _LOGGER.critical("Docker inside LXC is not supported by Supervisor!")
```
frenck, line 53:
```suggestion
            _LOGGER.critical("Detected Docker running inside LXC. Running Home Assistant with the Supervisor on LXC is not supported!")
```

rkkoszewski:
Just for the sake of completeness. LXD is not LXC. LXD is a wrapper for LXC. Therefore this change will not detect if if Hass is running inside a LXC container but instead if it's running in a LXC container using the LXD wrapper (The usual on Ubuntu). 

If the LXC container is running without LXD this will not detect anything (For example Debian does not provide a LXD package, and systems like Proxmox use their own wrapper that use LXC directly). 

There is supposedly a way to detect LXC directly by querying the file "/proc/self/cgroup" and detecting if it's running in a container with cgroup. But as far as I understand, it is also possible to run LXC without cgroup enabled, so this is a tricky one to solve.

Tollbringer:
Why is it important to detect LXC in the first place?

rkkoszewski:
I guess in case people have issues and they upload the Logs, so it's visible, as it's a non supported way of installing hass. Also probably because people usually ask regarding some issue without even stating that it's running in a LXC container. LXC is not particularly easy to debug, especially for people with little experience with it. Failing to disclose that you're running Hass in an LXC can make you or whoever wants to help, run in circles debugging all kind of unobvious issues.

Tollbringer:
That would be very valuable information for the devs to save their resources and wash their hands of the problem. 100% in support of this.

If it is used to cripple HA on other "implementations" of this amazing software, I would not support this purpose. Who knows- just offering my limited concerns.

PR #1757, opened 2020-05-28
Fix watchdog & scheduler

pvizeli:


supervisor/misc/scheduler.py
```diff
@@ -51,7 +52,7 @@ def _run_task(self, task_id):
     def _schedule_task(self, interval, task_id):
         """Schedule a task on loop."""
         if isinstance(interval, (int, float)):
-            job = self.loop.call_later(interval, self._run_task, task_id)
+            job = self.sys_loop.loop.call_later(interval, self._run_task, task_id)
```
MartinHjelmare, line 55:
```suggestion
            job = self.sys_loop.call_later(interval, self._run_task, task_id)
```

supervisor/misc/scheduler.py
```diff
@@ -40,8 +41,8 @@ def _run_task(self, task_id):
         """Run a scheduled task."""
         data = self._data[task_id]
 
-        if not self.suspend:
-            self.loop.create_task(data[CALL]())
+        if self.sys_core.state == CoreStates.RUNNING:
+            self.sys_loop.create_task(data[CALL]())
```
MartinHjelmare, line 45: Is the context unspecific here, so we need to specify what loop to use when creating the task, Ie use `loop.create_task` instead of `asyncio.create_task`?

pvizeli: It's `sys_create_task`. asyncio.create_task has more overhead which we can avoid anyway with our structure they we created before with python 3.4

MartinHjelmare: Ok. In the long run I think it's wise to try to move to the recommended higher level APIs. This is a trend in asyncio.

pvizeli: Sure, I wait for executor/callback pool to have a high-level API and we can migrate once all to it. maybe with Python 4 - anyway there are ~53 bigger tasks on ClickUp

PR #1763, opened 2020-06-01
Slowdown docker stop/start to avoid issue on slow IO

pvizeli:


MartinHjelmare:
Looks good!

supervisor/addons/__init__.py
```diff
@@ -80,28 +80,47 @@ async def load(self) -> None:
...
-            tasks.append(addon.start())
+            tasks.append(addon)
 
+        # Evaluate add-ons which need to be started
         _LOGGER.info("Phase '%s' start %d add-ons", stage, len(tasks))
-        if tasks:
-            await asyncio.wait(tasks)
-            await asyncio.sleep(self.sys_config.wait_boot)
+        if not tasks:
+            return
```
balloob, line 92: If there are no tasks, we are not calling `await asyncio.sleep(self.sys_config.wait_boot)`. Is that on purpose?

supervisor/misc/tasks.py
```diff
@@ -140,16 +139,16 @@ async def _update_addons(self):
                 continue
 
             if addon.test_update_schema():
-                tasks.append(addon.update())
+                _LOGGER.info("Add-on auto update process %s", addon.slug)
+                try:
```
balloob, line 143: Since we're doing an `await` in the for-loop, better to reverse the if-statement.

```python
if not addon.test_update_schema():
    _LOGGER.warning(…)
    continue

try:
    …
```

supervisor/addons/__init__.py
pvizeli: It's like before if there is nothing to do, we need also not spend time waiting until services are up

PR #1767, opened 2020-06-02
Docs string and exception message improvements

frenck:
Some improvements to docstrings and error messages. Added the add-on slug to some exceptions thrown, so it becomes more clear which add-on causes it.

### DIFF
--- a/supervisor/api/addons.py
+++ b/supervisor/api/addons.py
@@ -137,10 +137,10 @@ class APIAddons(CoreSysAttributes):
 
         addon = self.sys_addons.get(addon_slug)
         if not addon:
-            raise APIError("Addon does not exist")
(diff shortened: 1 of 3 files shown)


supervisor/api/addons.py
```diff
@@ -137,10 +137,10 @@ def _extract_addon(
 
         addon = self.sys_addons.get(addon_slug)
         if not addon:
-            raise APIError("Addon does not exist")
+            raise APIError(f"Addon {addon_slug} does not exist", addon_slug)
```
MartinHjelmare, line 140: The second argument is a mistake, right?

frenck: Yes, it is!

Adjusting...

PR #1769, opened 2020-06-02
Improve type hinting for add-ons

frenck:


supervisor/api/addons.py
```diff
@@ -315,6 +317,11 @@ async def security(self, request: web.Request) -> None:
         addon: AnyAddon = self._extract_addon(request)
         body: Dict[str, Any] = await api_validate(SCHEMA_SECURITY, request)
 
+        if not isinstance(addon, Addon):
+            raise APIError(
+                f"Cannot set security options, add-on {addon.slug} is not installed"
+            )
```
pvizeli, line 323: That is handle by _extract_addon

supervisor/api/addons.py
```diff
@@ -280,6 +280,8 @@ async def info(self, request: web.Request) -> Dict[str, Any]:
     async def options(self, request: web.Request) -> None:
         """Store user options for add-on."""
         addon: AnyAddon = self._extract_addon(request)
+        if not isinstance(addon, Addon):
+            raise APIError(f"Cannot set options, add-on {addon.slug} is not installed")
```
pvizeli, line 284: same

supervisor/api/addons.py
```diff
@@ -325,6 +332,9 @@ async def security(self, request: web.Request) -> None:
     async def stats(self, request: web.Request) -> Dict[str, Any]:
         """Return resource information."""
         addon: AnyAddon = self._extract_addon(request)
+        if not isinstance(addon, Addon):
+            raise APIError(f"Cannot get stats, add-on {addon.slug} is not installed")
+
```
pvizeli, line 337: too

supervisor/api/addons.py
frenck: Yes, agreed, however, when type hinting, the `_extract_addon` returns `AnyAddon`, which can be an instance of `Addon` or `AddonStore`.

Since these methods (where I've added this check), rely on methods only exposed in `Addon` and are not available in `AddonStore`/`AddonModel`.

If this check is not here, it fails type checking.

I agree the `_extract_addon` is helpful, however, its return type is ambiguous.

pvizeli: Which raises the error before this code happens...

frenck: This is about type checking and hinting, not runtime.
Sure, at runtime it is captured, from a type checking perspective, it isn't, as `_extract_addon` returns `AnyAddon`, it means that any static type checking tool will take into account it may be `AddonStore` as well.

pvizeli: Let's split _extract_addon into two functions to make mypi happy. I don't like to maintain code which never can happen.

frenck: Done

PR #1770, opened 2020-06-02
Handle new API / Frontend early boot

pvizeli:
Avoid also log flooding because we restart Core.

90% of the code is for P2 and the other 10% for P1. Since we need anyway touch the code and change the logic. The P1 adds the 503 for API which makes this PR a lot simpler.

### DIFF
--- a/supervisor/homeassistant.py
+++ b/supervisor/homeassistant.py
@@ -274,7 +274,7 @@ class HomeAssistant(JsonConfig, CoreSysAttributes):
 
         # Start landingpage
         _LOGGER.info("Start HomeAssistant landingpage")
-        with suppress(HomeAssistantError):
+        with suppress(DockerAPIError):
             await self.instance.run()
 
     @process_lock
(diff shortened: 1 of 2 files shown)


supervisor/homeassistant.py
```diff
@@ -559,12 +563,21 @@ async def make_request(
...
+        # Check if API is up
         with suppress(HomeAssistantAPIError):
-            async with self.make_request("get", "api/") as resp:
+            async with self.make_request("get", "api/config") as resp:
                 if resp.status in (200, 201):
-                    return True
-                status = resp.status
-            _LOGGER.warning("Home Assistant API config mismatch: %s", status)
+                    data = await resp.json()
+                    if data.get("state", "RUNNING") == "RUNNING":
```
bramkragten, line 577: Why would we wait on running for the rollback check? The frontend would be up already the user would be able to configure or fix whatever problem there could be, so why not just check if the frontend is up?

pvizeli: Let's merge P2 and we can start to play with that. You can after P2 start to add new integration or remove integration etc while HA is startup? We can later split up the check for block startup and API proxy 👍

PR #1771, opened 2020-06-02
Adjust core code base

pvizeli:


rootfs/etc/cont-init.d/patch-core.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/with-contenv bashio
+# ==============================================================================
+# Patching Home Assistant Core
+# ==============================================================================
+
+# Start with 0.111 there is a global constant for handling max core startup time
+# we can handle that better and avoid migration issue on database.
+sed -i "s|TIMEOUT_EVENT_BOOTSTRAP = .+|TIMEOUT_EVENT_BOOTSTRAP = 0|" /usr/src/homeassistant/homeassistant/bootstrap.py || true
```
MartinHjelmare, line 8: Doesn't this make it hard to troubleshoot the core if we change the source here?

pvizeli: Wrong place anyway...

PR #1772, opened 2020-06-02
Expose all services/discovery with possible add-ons

pvizeli:


supervisor/api/discovery.py
```diff
@@ -53,7 +55,13 @@ async def list(self, request):
...
             )
 
-        return {ATTR_DISCOVERY: discovery}
+        # Get available services/add-ons
+        services = {}
+        for addon in self.sys_addons.all:
+            for name in addon.discovery:
+                services.setdefault(name, []).append(addon.slug)
+
+        return {ATTR_DISCOVERY: discovery, ATTR_SERVICES: services}
```
balloob, line 64: Please update the test to make sure that this works.

PR #1777, opened 2020-06-04
Update the frontend

balloob:
- Remove all old frontend artifacts on rebuild. The old index.html referenced `hassio-app.html` but that file didn't exist anymore. It's really only for 2+ year old HA installations. 
- Update the frontend. This makes it work with 111 beta, and is backwards compatible with older versions of HA (as I tested it on HA 106)


frenck:
Confirmed on my end as well. Fixes my instances on 0.111.0b0

PR #1782, opened 2020-06-04
Clean api doc

MartinHjelmare:
- Fix grammar.
- Indent text and code per bullet to easier see what bullet is targeted.

### DIFF
--- a/API.md
+++ b/API.md
@@ -320,44 +320,44 @@ Return the host log messages (dmesg).
 
 - GET `/hardware/info`
 
-```json
-{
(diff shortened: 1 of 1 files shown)


MartinHjelmare:
The plan is to first do some minor clean up in this PR. Then in a follow up try to improve the structure of how we describe the api, eg by changing the describing name of placeholder variable values.

balloob:
Rebase to fix the tests 👍

pvizeli (approved):

PR #1793, opened 2020-06-15
Collection of minor changes

frenck:
This PR is a collection of minor changes to the codebase.

- Type hinting
- Pythonic changes
- isort
- `DBusInterface` is now abstract

98 MyPy issues left 😉

### DIFF
--- a/supervisor/addons/addon.py
+++ b/supervisor/addons/addon.py
@@ -130,12 +130,9 @@ class Addon(AddonModel):
         return {**self.data[ATTR_OPTIONS], **self.persist[ATTR_OPTIONS]}
 
     @options.setter
-    def options(self, value: Optional[Dict[str, Any]]):
+    def options(self, value: Optional[Dict[str, Any]]) -> None:
         """Store user add-on options."""
-        if value is None:
-            self.persist[ATTR_OPTIONS] = {}
-        else:
-            self.persist[ATTR_OPTIONS] = deepcopy(value)
+        self.persist[ATTR_OPTIONS] = {} if value is None else deepcopy(value)
 
     @property
     def boot(self) -> bool:
@@ -417,7 +414,7 @@ class Addon(AddonModel):
         exists_addon = self.path_apparmor.exists()
 
         # Nothing to do
-        if not exists_local and not exists_addon:
+        if not (exists_local or exists_addon):
             return
 
         # Need removed
--- a/supervisor/addons/addon.py
+++ b/supervisor/addons/addon.py
@@ -414,7 +414,7 @@ class Addon(AddonModel):
         exists_addon = self.path_apparmor.exists()
 
         # Nothing to do
-        if not (exists_local or exists_addon):
+        if not exists_local and not exists_addon:
             return
 
         # Need removed
--- a/supervisor/addons/model.py
+++ b/supervisor/addons/model.py
@@ -539,13 +539,17 @@ class AddonModel(CoreSysAttributes, ABC):
             return False
 
         # Home Assistant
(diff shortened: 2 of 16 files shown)


supervisor/api/hardware.py
```diff
@@ -54,4 +54,4 @@ async def audio(self, request: web.Request) -> Dict[str, Any]:
     @api_process
     def trigger(self, request: web.Request) -> None:
         """Trigger a udev device reload."""
-        return asyncio.shield(self.sys_hardware.udev_trigger())
+        asyncio.shield(self.sys_hardware.udev_trigger())
```
pvizeli, line 57: await?

supervisor/addons/model.py
```diff
@@ -540,12 +540,12 @@ def _available(self, config) -> bool:
...
-        ):
-            return False
-
-        return True
+        return (
+            not version
+            or self.sys_homeassistant.version is None
+            or pkg_version.parse(self.sys_homeassistant.version)
+            >= pkg_version.parse(version)
+        )
```
pvizeli, line 548: That is very hard to understand. All pythonic in love but it should not end up in Perl. Anyway, this code is currently wrong because of the new version handling, which needs force us to handle exception because a HA version can also be "landingpage" which is not a valid version. Good catch

balloob: I think that this code can be simplified even more because we default to `self.sys_homeassistant.version` and then check if that is >= that, which is never true


```python
version = config.get(ATTR_HOMEASSISTANT)

if version is None:
    return True

…
```

Are you adding the 2 extra checks because of typing? In that case consider fixing the type of `self.sys_homeassistant.version` or ensure it using an assert, it should always be there?

supervisor/api/addons.py
```diff
@@ -141,46 +141,43 @@ def _extract_addon(self, request: web.Request) -> AnyAddon:
 
     def _extract_addon_installed(self, request: web.Request) -> Addon:
         addon = self._extract_addon(request)
-        if not isinstance(addon, Addon) or not addon.is_installed:
+        if not (isinstance(addon, Addon) and addon.is_installed):
```
balloob, line 144: I think that this is less readable?

supervisor/docker/interface.py
```diff
@@ -135,10 +135,7 @@ def _is_running(self) -> bool:
             return False
 
         # container is not running
```
balloob, line 137: This comment makes no sense anymore.

supervisor/addons/model.py
frenck: Adjusted, added the Exception handling as well.

supervisor/addons/addon.py
```diff
@@ -417,7 +414,7 @@ async def install_apparmor(self) -> None:
         exists_addon = self.path_apparmor.exists()
 
         # Nothing to do
-        if not exists_local and not exists_addon:
+        if not (exists_local or exists_addon):
```
MartinHjelmare, line 417: Without the parenthesis the expression can be easily read from left to right in one pass. That's not possible with the parenthesis.

supervisor/addons/model.py
```diff
@@ -539,13 +539,17 @@ def _available(self, config) -> bool:
...
-        ):
-            return False
-
-        return True
+        version = config.get(ATTR_HOMEASSISTANT)
+        if version is None:
+            return True
+
+        try:
+            assert self.sys_homeassistant.version is not None
```
pvizeli, line 547: What will happen if that is None, do that don't raise an InvalidVersion error?

frenck: Yes, it will, it will throw a TypeError. I went to much with Paulus his suggestion on asserting there.

Fixed now.

PR #1795, opened 2020-06-17
Bugfix: No such file or directory: '/data/homeassistant/home-assistant_v2.db-shm'

### LINKED ISSUE #779: Hassio Snapshot not taking backup of /homeassistant
*@SalkinDK commented on Oct 18, 2018, 2:55 PM UTC:*

i get this error in the log, when taking the snapshot. If i try to open the snapshot tar file, not all of the homeassistant folder is there? fx the whole packages folder is missing.

`WARNING (SyncWorker_13) [hassio.snapshots.snapshot] Can't snapshot folder homeassistant: [Errno 2] No such file or directory: '/data/homeassistant/home-assistant_v2.db-shm'`

**Home Assistant release with the issue:**

not sure when it started but, happens on both 0.80.0 and 0.80.1

**Operating environment (Hass.io/Docker/Windows/etc.):**

Hassio on a raspberry pi 3

*This issue was moved by [fabaff](https://github.com/fabaff) from [home-assistant/home-assistant#17583](https://github.com/home-assistant/home-assistant/issues/17583).*

boesing:
As the folders might contain files which are being removed temporarily (e.g. shared memory file of sqlite database), relying on `tar_file.add` becomes problematic as it crashes the whole backup process if a file does not exist anymore.
This becomes annoying, if the file which causes the error should be excluded by the filter.

To workaround this issue, we manually iterating over the files/directories and apply filters before passing the file or directory to the `tar_file.add` method. As per [documentation](https://docs.python.org/3/library/pathlib.html#pure-paths), pure path does not access the file system.

Fixes #779


supervisor/snapshots/snapshot.py
```diff
@@ -370,17 +371,49 @@ def _folder_save(name):
...
                 _LOGGER.info("Snapshot folder %s", name)
                 with SecureTarFile(tar_name, "w", key=self._key) as tar_file:
-                    tar_file.add(
-                        origin_dir,
-                        arcname=".",
-                        filter=exclude_filter(MAP_FOLDER_EXCLUDE.get(name, [])),
+                    arcname = "."
+                    tar_file.add(origin_dir, arcname, False)
+                    add_directory_contents(
+                        tar_file, origin_dir, MAP_FOLDER_EXCLUDE.get(name, []), arcname
```
pvizeli, line 377: Don't make multiple calls. Extend the `add_directory_contents` to be the only called function

supervisor/snapshots/snapshot.py
```diff
@@ -370,17 +371,49 @@ def _folder_save(name):
...
                     )
 
                 _LOGGER.info("Snapshot folder %s done", name)
                 self._data[ATTR_FOLDERS].append(name)
             except (tarfile.TarError, OSError) as err:
                 _LOGGER.warning("Can't snapshot folder %s: %s", name, err)
 
+        def add_directory_contents(
```
pvizeli, line 385: Rename this function and export it to `utils/tar` and write tests for that function

supervisor/snapshots/snapshot.py
```diff
@@ -370,17 +371,49 @@ def _folder_save(name):
...
+            tar_file: tarfile.TarFile, origin_dir, excludes, arcname="."
+        ):
+            """Append directories and/or files to the TarFile if excludes wont filter."""
+
+            def excluded_path(file_path: PurePath):
```
pvizeli, line 390: Use the exists and well tested function exclude_filter

supervisor/snapshots/snapshot.py
```diff
@@ -370,17 +371,49 @@ def _folder_save(name):
...
+                        continue
+                    _LOGGER.debug("Ignore %s because of %s", file_path.name, exclude)
+                    return True
+                return False
+
+            origin_path = PurePath(origin_dir)
+            if excluded_path(origin_path):
+                return None
+
+            for directory_item in os.listdir(origin_dir):
```
pvizeli, line 403: Don't mix old style path/file handling with new OO style pathlib. Just use pathlib

boesing:
> The problem is, that the file system change while we run create the tar?

The problem is, that the file is being removed and re-created during tar-creation.
If you do a `watch ls -l home-assistant_v2.db-shm`, you will see that the file is being removed from filesystem and a few seconds later, there is a new one.

supervisor/snapshots/snapshot.py
pvizeli: make sure we use this function also for add-on and other folder snapshots

supervisor/snapshots/snapshot.py
boesing: You mean `exclude_filter` from `utils.tar`?
That file requires a `TarInfo` which actually is the main problem, because the `TarInfo` requires access to the I/O where the file is missing...
Or do I miss something?

supervisor/snapshots/snapshot.py
boesing: Just as mentioned [here](https://github.com/home-assistant/supervisor/pull/1795#discussion_r442111156):

I don't think that this should be part of the tar utils as it does not use any tar-related class/info or what benefit do we want to achieve?

supervisor/snapshots/snapshot.py
pvizeli: you can make an internal function in tar utils and share it between this two helpers

supervisor/snapshots/snapshot.py
boesing: > make sure we use this function also for add-on and other folder snapshots

Can you give me a hint where this could be applied aswell? I only see folder handling `supervisor.snapshots.snapshot::store_folders`? Thus, all folder snapshots pass that method.

All other snapshots are taken from docker.

boesing:
I've applied the suggestions.

Just want to say that I am not a python developer, thus I am not familiar with some stuff.

Is there a way to mock the tar file so I can test `supervisor.utils.tarfile::recursively_add_directory_contents_to_tar_file` without actually creating a real tar file from a real directory/file?

boesing:
> I've applied the suggestions.
> 
> Just want to say that I am not a python developer, thus I am not familiar with some stuff.
> 
> Is there a way to mock the tar file so I can test `supervisor.utils.tarfile::recursively_add_directory_contents_to_tar_file` without actually creating a real tar file from a real directory/file?

supervisor/utils/pathlib.py
```diff
@@ -0,0 +1,18 @@
+"""Pathlib utility for exclusion filtering."""
+import logging
+from pathlib import PurePath
+from typing import List
+
+_LOGGER: logging.Logger = logging.getLogger(__name__)
+
+
+def is_excluded_by_filter(path: PurePath, exclude_list: List[str]) -> bool:
```
pvizeli, line 9: Let's keep it more simple, we use that only inside tar utils. Just copy the function back to tar utils as `_is_exlude_by_filter` and use it inside the function.

supervisor/snapshots/snapshot.py
```diff
@@ -370,10 +374,11 @@ def _folder_save(name):
             try:
                 _LOGGER.info("Snapshot folder %s", name)
                 with SecureTarFile(tar_name, "w", key=self._key) as tar_file:
-                    tar_file.add(
+                    recursively_add_directory_contents_to_tar_file(
```
pvizeli, line 377: Make the same also for the add-ons data folder

supervisor/utils/tar.py
```diff
@@ -140,14 +142,38 @@ def exclude_filter(
...
+def recursively_add_directory_contents_to_tar_file(
+    tar_file: tarfile.TarFile, origin_dir: str, excludes: List[str], arcname: str = "."
+) -> None:
+    """Append directories and/or files to the TarFile if excludes wont filter."""
+
+    origin_path = Path(origin_dir)
+    if is_excluded_by_filter(origin_path, excludes):
+        return None
+
+    tar_file.add(origin_dir, arcname, recursive=False)
```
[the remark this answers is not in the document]
boesing: I tried without adding the directory itself, but that didn't work as expected. 
Thats btw exactly the same what `tar_file` internally does, when you add a directory. It adds itself and all other stuff by iterating over the files/directories. I just wanted to keep this in sync and as I had issues without adding the directory (I had only one file in my tar after snapshotting a directory), I've added it and it worked.
Not sure if we should remove it.

supervisor/utils/tar.py
```diff
@@ -140,14 +142,38 @@ def exclude_filter(
...
+        file_path = PurePath(tar.name)
+        if is_excluded_by_filter(file_path, exclude_list):
             return None
 
         return tar
 
     return my_filter
+
+
+def recursively_add_directory_contents_to_tar_file(
```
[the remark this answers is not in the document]
boesing: And less detailed, but I can rename that. No problem.

supervisor/utils/tar.py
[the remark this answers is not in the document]
boesing: Another problem might be, that an empty directory wont be added to the tar-file at all.
Lets assume we have a `configuration.yaml` like:

```
sensor: !include_dir_list sensors/
```

But the `sensors` directory is empty. 
This wont be an issue for now, but might become an issue after restore? I am not sure about this as I am not that deep in the functionality of homeassistant. Do you know what will happen if the directory is missing?

supervisor/snapshots/snapshot.py
boesing: I still don't get what folder you mean. An addon is snapshotted in `addon.snapshot`.
Please give me a hint which line you mean so I can apply the changes you are requesting.

pvizeli: https://github.com/home-assistant/supervisor/blob/dev/supervisor/addons/addon.py#L574

supervisor/utils/tar.py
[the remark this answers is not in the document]
pvizeli: okay. Let's add a short comment there, so we don't forget this in future

tests/utils/test_tarfile.py
```diff
@@ -31,30 +31,3 @@ def test_not_secure_path():
...
-    filter_funct = exclude_filter(["*.txt", "data/*", "bla/blu/ble"])
-    test_list = [
-        TarInfo("test.txt"),
-        TarInfo("data/xy.blob"),
-        TarInfo("bla/blu/ble"),
-        TarInfo("data/test_files/kk.txt"),
-    ]
-
-    for info in [filter_funct(result) for result in test_list]:
-        assert info is None
```
pvizeli, line 60: Don't remove tests, fix it

supervisor/addons/addon.py
```diff
@@ -572,13 +575,15 @@ def _write_tarfile():
                 """Write tar inside loop."""
                 with tar_file as snapshot:
                     # Snapshot system
-                    snapshot.add(temp, arcname=".")
```
pvizeli, line 575: Revert this, we need it only for data folder

tests/utils/test_tarfile.py
boesing: I removed the test as the function is not used anymore. If you prefer having dead code in your codebase, I can revert that...

supervisor/addons/addon.py
boesing: You pointed me to this line? I dont really get what you want as the directory is being added inside the `atomic_contents_add` function.

Please take a little more time for the review so you dont waste your and my time. I know this is open source and I appreciate the feedback but its quite annoying if you change your requirements from commit to commit.

Thanks.

tests/utils/test_tarfile.py
pvizeli: Make sure that you function are covered by tests. You remove the filter tests without add a new test for you new code. We can not merge PR they lost test coverage.

boesing: I've implemented the same test cases for the new `_is_excluded_by_filter` function.

pvizeli:
Good work, let's test it on dev

boesing:
Thanks. I've changed to dev channel aswell to see if this works as expected 👍

PR #1805, opened 2020-06-28
Add progress bar to hassio update/install

sanyatuning:
This code will emit hassio_progress events to Core while docker pull.
This will allow us to display progress bars when updating Core or installing/updating an addon.

Frontend PR: https://github.com/home-assistant/frontend/pull/6267

Something like this:
progress2


supervisor/docker/progress.py
```diff
@@ -0,0 +1,132 @@
...
+    def done(self):
+        """Mark current pull as done and send this info to HA Core"""
+        self._download.done_all()
+        self._extract.done_all()
+        self.send_event()
+
+    async def _send_event(self, status) -> None:
+        with suppress(HomeAssistantAPIError):
+            async with self._interface.sys_homeassistant.make_request(
+                "post", "api/events/hassio_progress", json=status, timeout=2,
```
pvizeli, line 89: We are not allow to use the event/service bus of the core. I would suggest to make an `/{object}/progress` endpoint which can be called by UI. If they return `None` there is no job running, otherwise we return information to show the progressbar.

sanyatuning: Do you mean we should poll this endpoint?
I think it is not the best way.
What do you think about creating a new WebSocket connection from the UI to the supervisor?

supervisor/docker/progress.py
```diff
@@ -0,0 +1,132 @@
...
+import time
+from contextlib import suppress
+
+import docker
+
+from ..coresys import CoreSysAttributes
+from ..exceptions import HomeAssistantAPIError
+
+
+def pull_with_progress(interface: CoreSysAttributes, image, tag):
```
pvizeli, line 12: Implement this as part of the DockerAPI object and don't forward the interface, lookup it over the coresys

supervisor/docker/progress.py
pvizeli: If you want stream the logs, we need change the library to aiodocker or other asyncio library, otherwise we need change the socket object into nonblock and wrap it with a async streamreader/writter.

However, what you want do is a ton of work over multible repository, 1-3 month. Feel free to give a shot, but maybe make first an RFC to reduce the changes with reviews.

balloob: Pascal and I discussed this offline. We should do both.

We want an event to indicate that something has changed, and then an API endpoint that gives us the details. We shouldn't include all the data in the event.

supervisor/docker/progress.py
sanyatuning: Do you mean the whole file or just this function?

supervisor/docker/progress.py
sanyatuning: Do you mean an event in the core, as I did?
What would be the next step? Help me a little :)

pvizeli: I need write an RFC for an job based API based on `process_lock` and the new contextvars which write/manage jobs global and can send events to the bus. But I'm also in vacation for next weeks and limited what I can do.

balloob: @Sanyatuning next step will be to wait until Pascal is back from parental leave and design a more elaborate job-progress system. We don't want to hack it into just the Docker API as there might be other tasks that we want to report on (backups etc).

sanyatuning: Thanks for the explanation.

supervisor/docker/progress.py
pvizeli: this function, move the states and the other object into docker/utils.py as helper like we do in other modules

sanyatuning: OK, will try it maybe tomorrow

sanyatuning: @pvizeli pls check

sanyatuning:
Any update on this?

pvizeli:
Yep. I started a draft https://github.com/home-assistant/supervisor/issues/1946

I think after this is finalized and implemented, we can move forward with this. 👍

sanyatuning:
Thanks

pvizeli:
You can also pickup this RFC if you want implement somethings or search some work

sanyatuning:
> You can also pickup this RFC if you want implement somethings or search some work

@pvizeli Could you check my updates?
I also ask some questions in the RFC ticket #1946.

Thanks

elbueno222:
Why was this closed? This functionality would be really cool as there is no way to know if an update is happening at the moment.

lsdlsd88:
up ? why is it closed?

pboschi:
year 2023... still no progress bar and no way to understand what's happening during updates... 
image

smartroad:
This is something that is seriously needed. I think I have a backup running but no way to know all I get is an error that the system is in a freeze state :(

PR #1810, opened 2020-07-02
Update frontend to fd07152

ludeeus:


supervisor/api/panel/entrypoint.js
```diff
@@ -1,9 +1,9 @@
 
 try {
-  new Function("import('/api/hassio/app/frontend_latest/entrypoint.js')")();
+  new Function("import('/api/hassio/app/frontend_latest//api/hassio/app/frontend_latest/entrypoint.c64d3ee6.js')")();
```
Misiu, line 3: is this correct?
I think it should be:
``` javascript
new Function("import('/api/hassio/app/frontend_latest/entrypoint.c64d3ee6.js')")();
```

supervisor/api/panel/entrypoint.js
```diff
@@ -1,9 +1,9 @@
...
 } catch (err) {
   var el = document.createElement('script');
-  el.src = '/api/hassio/app/frontend_es5/entrypoint.js';
+  el.src = '/api/hassio/app/frontend_es5//api/hassio/app/frontend_es5/entrypoint.4a2722b2.js';
```
Misiu, line 6: same as above - double slash issue.

ludeeus: double slash is not the issue

ludeeus:
Build scripts needs adjustments

PR #1811, opened 2020-07-02
Update frontend panel to 750e7b

ludeeus:


balloob:
Not sure how the update script works, but you can delete `manifest.json` and `entrypoint.js.map` from the root panel dir.

ludeeus:
The script just copy over everything in the build dir.
https://github.com/home-assistant/supervisor/blob/0fe27088df09915eef7025b3f6185f335c6de21f/scripts/update-frontend.sh#L18

balloob:
It first deletes everything though https://github.com/home-assistant/supervisor/blob/0fe27088df09915eef7025b3f6185f335c6de21f/scripts/update-frontend.sh#L17

ludeeus:
True... not really sure where those came from, tried to run the script again now and it did not produce them :man_shrugging:

balloob:
They are from the old panel. I deleted most of it but forgot those last time. However I would have expected the files to be deleted. This also means that the old panel is not being removed right now. If you see, you remove only 5 lines ! There should be a ton more deleted.

ludeeus:
I got it!

PR #1814, opened 2020-07-05
Add QA/CI to Supervisor

frenck:
Adds GitHub Actions and pre-commit for the Supervisor, including some additions:

- JQ has been replaced by pre-commit, including problem matcher
- Added isort
- Added pyupgrade to prevent old patterns (I've removed all of its warnings in previous PRs)
- Check for files with executable permissions that should not have them
- Add testing for both Python 3.7 & 3.8
- Add coverage handling & codecov.io reporting (comments by codecov disabled after merge to dev)

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -3,6 +3,7 @@
   "context": "..",
   "dockerFile": "Dockerfile",
   "appPort": "9123:8123",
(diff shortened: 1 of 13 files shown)


MartinHjelmare:
How much time does the venv cache restoration save us?

frenck:
since the supervisor is fairly minimal, probably around 2-3 minutes (on each of the tasks).

PR #1824, opened 2020-07-13
Migrate to python 3.8

pvizeli:


.github/workflows/ci.yaml
```diff
@@ -345,7 +345,7 @@ jobs:
     needs: prepare
     strategy:
       matrix:
-        python-version: [3.7, 3.8]
+        python-version: [3.8]
```
pvizeli, line 348: We only run on the version which we use inside our container, so we don't need check multible version

frenck: Agree, however, the tests don't pass on Python 3.8.

I don't know or understand why at this point.

pvizeli: ok

balloob:
It's because `mock_coro` is not awaited, which fails the tests. Using the built-in `patch` will work.

balloob:
Pushed fix

frenck:
Looks good to me now 🎉 

One small thing, the 3.7 can be removed from the first job step matrix.

If we leave the matrix in, we can easily add a new Python version (before migration) in the future.

Issue #1819, opened 2020-07-07
CRITICAL (MainThread) [supervisor.homeassistant] HomeAssistant update fails -> rollback!


Toni981:
**Home Assistant release with the issue:**
0.112.2

**Operating environment (HassOS/Generic):**
HassOS

**Supervisor logs:**

CRITICAL (MainThread) [supervisor.homeassistant] HomeAssistant update fails -> rollback!

**Description of problem:**

Good evening.
I am trying to update version 0.112.2 to version 0.112.3 but I have the error:

CRITICAL (MainThread) [supervisor.homeassistant] HomeAssistant update fails -> rollback!

I have also tried from Terminal but I have the same error.

A few days ago I reprogrammed everything from scratch so I am sure that I have no problems in the few addons that I have. Can you give me some suggestions?

It has already happened to me before going from 0.112.0 to 0.112.2 where I had to install everything from scratch but now I find myself with the same problem ...

balloob:
Please fill in the issue template.

Toni981:
ok sorry...I have correct it. can you help me?

pvizeli:
There should be more inside the log as this one line

fedebruni84:
> There should be more inside the log as this one line

Look at this, seems there's no other lines in the log:
https://community.home-assistant.io/t/error-while-updating-from-0-111-4-to-0-112-3/210086

Toni981:
Hi, yesterday I downloaded 0.112.3 and installed the backup. 
Today I saw the update to version 0.112.4 and I tried to do the update but I have this problem:

`20-07-09 09:05:12 INFO (MainThread) [supervisor.store.git] Update add-on https://github.com/hassio-addons/repository repository
20-07-09 09:05:12 INFO (MainThread) [supervisor.store.git] Update add-on https://github.com/danielwelch/hassio-addons repository
20-07-09 09:05:12 INFO (MainThread) [supervisor.store.git] Update add-on https://github.com/home-assistant/hassio-addons repository
20-07-09 09:05:12 INFO (MainThread) [supervisor.updater] Fetch update data from https://version.home-assistant.io/stable.json
20-07-09 09:05:14 INFO (MainThread) [supervisor.store] Load add-ons from store: 66 all - 0 new - 0 remove
20-07-09 09:06:17 INFO (MainThread) [supervisor.homeassistant] Update Home Assistant to version 0.112.4
[...]

PR #1852, opened 2020-07-23
Fix rollback timeout on new versions with early stage UI

pvizeli:
Fix: #1819


supervisor/homeassistant.py
```diff
@@ -581,9 +580,20 @@ async def check_api_state(self) -> bool:
 
         return False
 
-    async def _block_till_run(self) -> None:
+    async def _block_till_run(self, version: str) -> None:
         """Block until Home-Assistant is booting up or startup timeout."""
+        # Skeep landingpage
```
ludeeus, line 585:
```suggestion
        # Skip landingpage
```

Looks like another typo :)

PR #1864, opened 2020-08-03
Use "stale" instead of "wontfix"

ludeeus:


balloob:
Also rename the label?

ludeeus:
Good point!
Have done that now.

PR #1867, opened 2020-08-03
Make a copy of the logfile before rollback

ludeeus:
Creates `home-assistant-rollback.log` from `home-assistant.log` before rollback, so the reason for rollback is not lost

### DIFF
--- a/supervisor/homeassistant.py
+++ b/supervisor/homeassistant.py
@@ -360,12 +360,16 @@ class HomeAssistant(JsonConfig, CoreSysAttributes):
         if self.error_state and rollback:
             _LOGGER.critical("HomeAssistant update fails -> rollback!")
             # Make a copy of the current log file if it exsist
-            logfile = Path(self.sys_config.path_homeassistant, "home-assistant.log")
+            logfile = self.sys_config.path_homeassistant / "home-assistant.log"
             if logfile.exists():
-                backup = Path(
-                    self.sys_config.path_homeassistant, "home-assistant-rollback.log"
+                backup = (
+                    self.sys_config.path_homeassistant / "home-assistant-rollback.log"
                 )
-                shutil.copy(logfile.as_posix(), backup.as_posix())
+
+                shutil.copy(logfile, backup)
+                _LOGGER.info(
+                    "A backup of the logfile is stored in /config/home-assistant-rollback.log"
+                )
             await _update(rollback)
         else:
             raise HomeAssistantUpdateError()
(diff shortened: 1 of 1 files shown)


supervisor/homeassistant.py
```diff
@@ -359,6 +359,13 @@ async def _update(to_version: str) -> None:
         # Update going wrong, revert it
         if self.error_state and rollback:
             _LOGGER.critical("HomeAssistant update fails -> rollback!")
+            # Make a copy of the current log file if it exsist
+            logfile = Path(self.sys_config.path_homeassistant, "home-assistant.log")
+            if logfile.exists():
+                backup = Path(
+                    self.sys_config.path_homeassistant, "home-assistant-rollback.log"
+                )
+                shutil.copy(logfile.as_posix(), backup.as_posix())
```
balloob, line 368: `shutil.copy` doesn't take `Path` objects?

ludeeus: Looks like it does, I'll change it to use that.

supervisor/homeassistant.py
```diff
@@ -359,6 +359,13 @@ async def _update(to_version: str) -> None:
         # Update going wrong, revert it
         if self.error_state and rollback:
             _LOGGER.critical("HomeAssistant update fails -> rollback!")
+            # Make a copy of the current log file if it exsist
+            logfile = Path(self.sys_config.path_homeassistant, "home-assistant.log")
```
balloob, line 363: If `path_homeassistant` is a Path object, this works too:
```suggestion
            logfile = self.sys_config.path_homeassistant / "home-assistant.log"
```

supervisor/homeassistant.py
```diff
@@ -359,6 +359,17 @@ async def _update(to_version: str) -> None:
...
             _LOGGER.critical("HomeAssistant update fails -> rollback!")
+            # Make a copy of the current log file if it exsist
+            logfile = self.sys_config.path_homeassistant / "home-assistant.log"
+            if logfile.exists():
+                backup = (
+                    self.sys_config.path_homeassistant / "home-assistant-rollback.log"
+                )
+
+                shutil.copy(logfile, backup)
+                _LOGGER.info(
```
balloob, line 370: They will never see this, right?

ludeeus: It will show in the supervisor log

ludeeus:
```
20-08-03 10:40:12 CRITICAL (MainThread) [supervisor.homeassistant] HomeAssistant update fails -> rollback!
20-08-03 10:40:12 INFO (MainThread) [supervisor.homeassistant] A backup of the logfile is stored in /config/home-assistant-rollback.log
```

PR #1868, opened 2020-08-03
Add free disk space log on docker install error

MartinHjelmare:
- Add free disk space info method.
- Add info log about available space on docker image install error.

### DIFF
--- a/supervisor/docker/interface.py
+++ b/supervisor/docker/interface.py
@@ -99,8 +99,13 @@ class DockerInterface(CoreSysAttributes):
                 docker_image.tag(image, tag="latest")
         except docker.errors.APIError as err:
             _LOGGER.error("Can't install %s:%s -> %s.", image, tag, err)
-            free_space = self.coresys.host.info.get_disk_free_space()
-            _LOGGER.info("Available space in /data is: %s GiB", free_space)
+            if err.status_code == 404:
+                free_space = self.coresys.host.info.get_disk_free_space()
+                _LOGGER.info(
+                    "This error is often caused by not having enough disk space available. "
+                    "Available space in /data is: %s GiB",
+                    free_space,
+                )
             raise DockerAPIError() from None
         else:
             self._meta = docker_image.attrs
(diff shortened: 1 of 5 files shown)


supervisor/host/info.py
```diff
@@ -66,6 +67,11 @@ async def get_dmesg(self) -> bytes:
 
         return stdout
 
+    def get_disk_free_space(self, path: str = "/data") -> float:
```
MartinHjelmare, line 70: Is `/data` the path that is used for pulling docker images?

MartinHjelmare: Maybe use this property instead?
https://github.com/home-assistant/supervisor/blob/e1ce19547e0a0d056d645436aa291e86fbabbe9b/supervisor/config.py#L135-L138

supervisor/docker/interface.py
```diff
@@ -99,6 +99,8 @@ def _install(
                 docker_image.tag(image, tag="latest")
         except docker.errors.APIError as err:
             _LOGGER.error("Can't install %s:%s -> %s.", image, tag, err)
+            free_space = self.coresys.host.info.get_disk_free_space()
```
frenck, line 102: We could check `err.response.status_code == 404:` here and add some additional message in the log.

e.g., `This error is often caused by not having enough disk space available.`

timmo001:
I had this exact problem the other night. Caused all sorts of problems on my server. Thanks for this! 👍

supervisor/docker/interface.py
```diff
@@ -99,6 +99,13 @@ def _install(
                 docker_image.tag(image, tag="latest")
         except docker.errors.APIError as err:
             _LOGGER.error("Can't install %s:%s -> %s.", image, tag, err)
+            if err.status_code == 404:
+                free_space = self.coresys.host.info.get_disk_free_space()
```
pvizeli, line 103:
```suggestion
                free_space = self.sys_host.info.get_disk_free_space()
```

supervisor/host/info.py
```diff
@@ -66,6 +68,13 @@ async def get_dmesg(self) -> bytes:
 
         return stdout
 
+    def get_disk_free_space(self, path: Optional[Union[str, Path]] = None) -> float:
```
pvizeli, line 71: Let's make an which use fix the path_supervisor
```
@property
free_space(self) -> float:
```

And put the `get_disk_free_space` into hardware as helper with path.

pvizeli:
perfect 🐬

PR #1869, opened 2020-08-04
Update frontend to bb5f6e8

ludeeus:
This updates the frontend so it can utilize the new theme options that are in core 0.113+


pvizeli:
Work that also with older HA version?

ludeeus:
Good question!

It "works" but it does not work
image

PR #1870, opened 2020-08-04
Add diagnostics support

pvizeli:
If diagnostic is None, user have not take a choise and it's disabled. Otherwise it's False/True.

### DIFF
--- a/supervisor/bootstrap.py
+++ b/supervisor/bootstrap.py
@@ -24,6 +24,8 @@ from .const import (
     ENV_SUPERVISOR_NAME,
     ENV_SUPERVISOR_SHARE,
     SOCKET_DOCKER,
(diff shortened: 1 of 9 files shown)


supervisor/bootstrap.py
```diff
@@ -270,3 +282,54 @@ def supervisor_debugger(coresys: CoreSys) -> None:
...
+            return None
+
+        # Not full startup - missing information
+        if coresys.core.state in (CoreStates.INITIALIZE, CoreStates.SETUP):
+            return event
+
+        # Update information
+        with sentry_sdk.configure_scope() as scope:
+            scope.set_context(
+                "home_assistant",
```
balloob, line 302: Should this be context of `supervisor` ? 
```suggestion
                "supervisor",
```

PR #1872, opened 2020-08-05
Replace debuger with debugpy

pvizeli:
Solve memory leaks


supervisor/bootstrap.py
```diff
@@ -274,14 +274,14 @@ def supervisor_debugger(coresys: CoreSys) -> None:
     if not coresys.config.debug:
         return
     # pylint: disable=import-outside-toplevel
-    import ptvsd
+    import debugpy
```
frenck, line 277: You don't have to import this outside of the top-level anymore.

debugpy can safely be imported top level.

PR #1883, opened 2020-08-10
Allow Core access to dbus for bluetooth

pvizeli:
Fix: #1882

### DIFF
--- a/supervisor/docker/homeassistant.py
+++ b/supervisor/docker/homeassistant.py
@@ -6,7 +6,7 @@ from typing import Awaitable, Dict, Optional
 
 import docker
 
-from ..const import ENV_TIME, ENV_TOKEN, ENV_TOKEN_OLD, LABEL_MACHINE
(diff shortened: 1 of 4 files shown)


supervisor/docker/homeassistant.py
```diff
@@ -48,7 +48,7 @@ def ip_address(self) -> IPv4Address:
     @property
     def volumes(self) -> Dict[str, Dict[str, str]]:
         """Return Volumes for the mount."""
-        volumes = {}
+        volumes = {"/run/dbus": {"bind": "/run/dbus", "mode": "ro"}}
```
balloob, line 51: This gives Home Assistant full dbus access right?

pvizeli: Yes. however, they run already with full rights and on the host network. We should write a custom AppArmor profile for Core which can limit the access & scope also on dbus. But that is a challenge. I'm also fine comment core out of dbus and just forward with the other small code cleanup

balloob: I think it's fine. Core can already talk to the supervisor anyway.

If we find integrations are messing it up, we will revisit.

PR #1887, opened 2020-08-11
Tag installation type & more infos

pvizeli:


supervisor/bootstrap.py
```diff
@@ -301,13 +301,19 @@ def filter_data(event, hint):
...
                     "core": coresys.homeassistant.version,
                     "audio": coresys.plugins.audio.version,
                     "dns": coresys.plugins.dns.version,
                     "multicast": coresys.plugins.multicast.version,
                     "cli": coresys.plugins.cli.version,
                 },
             )
+            scope.set_tag(
+                "installation_type",
+                f"{'generic' if coresys.hassos.available else 'hassos'}",
```
frenck, line 315: Should we use the official names instead?
"Home Assistant OS", "Home Assistant Supervised"?

pvizeli: it's more about sorting inside sentry (short is better because otherwise OS/Supervised is cut)

pvizeli: we can use the ADR number?

PR #1896, opened 2020-08-12
Adds image denylist

ludeeus:
Adds a list that we can add image names we know causes issues.
If one is found, the state of the system will be unhealthy.


supervisor/const.py
```diff
@@ -37,6 +37,7 @@
 DOCKER_NETWORK = "hassio"
 DOCKER_NETWORK_MASK = ip_network("172.30.32.0/23")
 DOCKER_NETWORK_RANGE = ip_network("172.30.33.0/24")
+DOCKER_IMAGE_DENYLIST = ["containrrr/watchtower", "v2tec/watchtower"]
```
frenck, line 40: So... isn't any container, unknown to the Supervisor, making the system unsupported? (according to the ADR)

pvizeli: I think we should be fair and only mark things as unsupported if they make issues.

supervisor/core.py
```diff
@@ -44,6 +44,10 @@ async def connect(self):
             )
         self.sys_docker.info.check_requirements()
 
+        # Check if imagenames from denylist exsist
```
MartinHjelmare, line 47: Check if image names from denylist exist

supervisor/docker/__init__.py
```diff
@@ -230,3 +230,17 @@ def repair(self) -> None:
             _LOGGER.debug("Networks prune: %s", output)
         except docker.errors.APIError as err:
             _LOGGER.warning("Error for networks prune: %s", err)
+
+    def check_denylist_images(self) -> bool:
+        """Retrun a boolen if the host has images in the denylist."""
```
MartinHjelmare, line 235: Return

supervisor/const.py
frenck: If we consider that, we should be splitting this into 2 sections.

1) We've detected unknown, external, containers, this is not officially supported
2) We've detected known, external, containers that are known to cause issues

Both deserve a log message of some sort I think.

ludeeus: I agree, 1 should set it as unsupported, 2 should set it as unhealthy

supervisor/docker/__init__.py
```diff
@@ -230,3 +230,17 @@ def repair(self) -> None:
             _LOGGER.debug("Networks prune: %s", output)
         except docker.errors.APIError as err:
             _LOGGER.warning("Error for networks prune: %s", err)
+
+    def check_denylist_images(self) -> bool:
+        """Return a boolen if the host has images in the denylist."""
```
MartinHjelmare, line 235: Sorry, I missed this one:
boolean

supervisor/const.py
pvizeli: 2) let's show an error but not set unhealthy

supervisor/core.py
```diff
@@ -44,6 +44,10 @@ async def connect(self):
             )
         self.sys_docker.info.check_requirements()
 
+        # Check if image names from denylist exist
+        if self.sys_docker.check_denylist_images():
```
pvizeli, line 48:
```suggestion
        if wait self.sys_run_in_executor(self.sys_docker.check_denylist_images):
            self.coresys.supported = False
```

supervisor/docker/__init__.py
```diff
@@ -230,3 +230,17 @@ def repair(self) -> None:
...
+        for image in self.images.list():
+            for tag in image.tags:
+                imagename = tag.split(":")[0]
+                if imagename in DOCKER_IMAGE_DENYLIST:
+                    has_denied_images = True
+                    _LOGGER.error(
+                        "Found image %s, this is not supported, remove this from the host!",
+                        imagename,
+                    )
+        return has_denied_images
```
pvizeli, line 246: This show for the same image multible times the error log. Let's write all found image into a set() and check if the set have an image inside, if no, return False otherwise show an error which include the list from the set.

supervisor/docker/__init__.py
MartinHjelmare:
```suggestion
        """Return a boolean if the host has images in the denylist."""
```

Google thinks it should be "deny list" or "deny-list". "denylist" seems quite frequently used too though.

ludeeus: I was going by <https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=49decddd39e5f6132ccd7d9fdc3d7c470b0061bb>

supervisor/docker/__init__.py
```diff
@@ -230,3 +230,21 @@ def repair(self) -> None:
...
+                if (
+                    imagename in DOCKER_IMAGE_DENYLIST
+                    and imagename not in denied_images
+                ):
+                    denied_images.add(imagename)
+                    _LOGGER.error(
+                        "Found image '%s', this is not supported, remove this from the host!",
+                        imagename,
+                    )
+                    denied_images.add(imagename)
```
MartinHjelmare, line 249: Words are separated with underscore in variable names.
```suggestion
                image_name = tag.split(":")[0]
                if (
                    image_name in DOCKER_IMAGE_DENYLIST
                    and image_name not in denied_images
                ):
                    denied_images.add(image_name)
                    _LOGGER.error(
                        "Found image '%s', this is not supported, remove this from the host!",
                        image_name,
                    )
                    denied_images.add(image_name)
```

supervisor/docker/__init__.py
```diff
@@ -230,3 +230,21 @@ def repair(self) -> None:
...
+        """Return a boolean if the host has images in the denylist."""
+        denied_images = set()
+        for image in self.images.list():
+            for tag in image.tags:
+                imagename = tag.split(":")[0]
+                if (
+                    imagename in DOCKER_IMAGE_DENYLIST
+                    and imagename not in denied_images
+                ):
+                    denied_images.add(imagename)
```
MartinHjelmare, line 244: A duplicate of row 249.

ludeeus: Ahh.. yes

supervisor/docker/__init__.py
```diff
@@ -230,3 +230,20 @@ def repair(self) -> None:
...
+                if (
+                    image_name in DOCKER_IMAGE_DENYLIST
+                    and image_name not in denied_images
+                ):
+                    _LOGGER.error(
+                        "Found image '%s', this is not supported, remove this from the host!",
+                        image_name,
+                    )
+                    denied_images.add(image_name)
+        return bool(denied_images)
```
pvizeli, line 249: Prevent multible tags / error messages
```
if not denied_images:
   return False
_LOGGER.error(...)
return True
```

ludeeus: Should we not make the user aware of all offending images they have so they can do something about it?
Currently, it logs 1 error pr image in the list

pvizeli: `_LOGGER.error("We found such images: %s which is not acceptable....", denied_images)`

supervisor/docker/__init__.py
```diff
@@ -241,9 +241,13 @@ def check_denylist_images(self) -> bool:
...
-                        image_name,
-                    )
                     denied_images.add(image_name)
-        return bool(denied_images)
+
+        if not denied_images:
+            return False
+
+        _LOGGER.error(
+            "Found images: '%s' which is not  supported, remove these from the host!",
```
MartinHjelmare, line 250:
```suggestion
            "Found images: '%s' which are not supported, remove these from the host!",
```

PR #1898, opened 2020-08-12
Fix validation issue addon options

pvizeli:


pvizeli:
let me write some tests

supervisor/addons/validate.py
```diff
@@ -439,6 +444,11 @@ def _nested_validate_dict(coresys, typ, data_dict, key):
     """Validate nested items."""
     options = {}
 
+    # Make sure it is a list
```
balloob, line 447:
```suggestion
    # Make sure it is a dict
```

PR #1904, opened 2020-08-13
Set not supported with wrong docker configuration

ludeeus:
Set not supported if:

- Wrong docker logger
- Wrong storage for docker
- No access to udev monitor


supervisor/misc/hwmon.py
```diff
@@ -29,6 +29,7 @@ async def load(self) -> None:
             self.observer = pyudev.MonitorObserver(self.monitor, self._udev_events)
         except OSError:
             _LOGGER.critical("Not privileged to run udev monitor!")
+            self.coresys.supported = False
```
pvizeli, line 32: Set this before we log the critical to avoid sending that message :)

PR #1908, opened 2020-08-13
Sanitize event

ludeeus:


supervisor/bootstrap.py
```diff
@@ -319,6 +321,25 @@ def filter_data(event, hint):
...
+        if event.get("request"):
+            if event["request"].get("url"):
+                event["request"]["url"] = sanitize_url(event["request"]["url"])
+
+            for header in event["request"].get("headers", []):
+                if header[0] == "Referer":
+                    header[1] = sanitize_url(header[1])
+
+                if header[0] == "X-Hassio-Key":
+                    header[1] = DUMMY_VALUE
```
pvizeli, line 338: We use that only once, let's just write "xxxxx" without a global

supervisor/bootstrap.py
```diff
@@ -319,6 +321,25 @@ def filter_data(event, hint):
...
+
+        if event.get("request"):
+            if event["request"].get("url"):
+                event["request"]["url"] = sanitize_url(event["request"]["url"])
+
+            for header in event["request"].get("headers", []):
+                if header[0] == "Referer":
+                    header[1] = sanitize_url(header[1])
+
+                if header[0] == "X-Hassio-Key":
```
pvizeli, line 337: All kind of headers which we use are in `const.py`. There are more around

supervisor/bootstrap.py
```diff
@@ -319,6 +321,25 @@ def filter_data(event, hint):
...
+                    header[1] = DUMMY_VALUE
+
+                if header[0] in ["Host", "X-Forwarded-Host"]:
+                    event["request"]["headers"].remove(header)
```
pvizeli, line 341: Let not remove it, we need it for fix ingress, but it into header dummy replacement above

supervisor/utils/__init__.py
```diff
@@ -7,8 +7,9 @@
 import socket
 from typing import Optional
 
+from ..const import RE_STRING, RE_URL
+
 _LOGGER: logging.Logger = logging.getLogger(__name__)
-RE_STRING = re.compile(r"\x1b(\[.*?[@-~]|\].*?(\x07|\x1b\\))")
```
pvizeli, line 11: Do we need this global?

supervisor/const.py
```diff
@@ -16,6 +17,9 @@
 
 SUPERVISOR_DATA = Path("/data")
 
+RE_STRING = re.compile(r"\x1b(\[.*?[@-~]|\].*?(\x07|\x1b\\))")
+RE_URL: re.Pattern = re.compile(r"(\w+:\/\/)(.*\.\w+)(.*)")
```
pvizeli, line 21: See comment above

supervisor/bootstrap.py
ludeeus: These 2 only contains the hostname like "mydomain.com", just displaying "example.com" seems unneeded?

supervisor/bootstrap.py
ludeeus: Good idea

supervisor/utils/__init__.py
ludeeus: Probably not

supervisor/const.py
ludeeus: See comment above

supervisor/bootstrap.py
pvizeli: Is more to have the full header around and not hunting ghosts

supervisor/const.py
```diff
@@ -63,6 +63,9 @@
 CONTENT_TYPE_URL = "application/x-www-form-urlencoded"
 COOKIE_INGRESS = "ingress_session"
 
+HEADER_FORWARDED_HOST = "X-Forwarded-Host"
+HEADER_HOST = "Host"
+HEADER_REFERER = "Referer"
```
pvizeli, line 68: Let's use for the official headers the hdrs module as we do on core and other parts i.e. (https://github.com/home-assistant/supervisor/blob/dev/supervisor/api/ingress.py#L8)

supervisor/utils/filter.py
```diff
@@ -0,0 +1,87 @@
+"""Filter utils."""
```
pvizeli, line 1: let's move this into `misc` because `utils` are unaware of coresys.

PR #1927, opened 2020-08-15
Split extra info and add more metrics

ludeeus:


supervisor/misc/filter.py
```diff
@@ -42,21 +42,29 @@ def filter_data(coresys: CoreSys, event: dict, hint: dict) -> dict:
     event.setdefault("extra", {}).update(
         {
             "supervisor": {
-                "machine": coresys.machine,
+                "channel": coresys.updater.channel,
+                "installed_addons": [x.slug for x in coresys.addons.installed],
```
ludeeus, line 46: Are other things interesting?

Like:
- name
- repository

supervisor/misc/filter.py
```diff
@@ -38,25 +38,39 @@ def filter_data(coresys: CoreSys, event: dict, hint: dict) -> dict:
...
         return event
 
+    # List installed addons
+    installed_addons = [
+        {"slug": addon.slug, "repository": addon.repository, "name": addon.name}
+        for addon in coresys.addons.installed
+    ]
+
     # Update information
     event.setdefault("extra", {}).update(
```
frenck, line 48: We should start using `contexts` instead.

It will render its own boxes (according to the founder of sentry, that will be made searchable in the future as well).

I've just applied it locally on the Core for testing:

image


```py
        event.setdefault("contexts", {}).update(
            {
                "system": system_info,
                "integrations": {
                    "active": "\n".join(sorted(integrations)),
                    "custom_components": "\n".join(sorted(custom_components)),
                },
            }
        )
```

ludeeus: I like that!

PR #1931, opened 2020-08-16
Disable jemalloc via specific ENV variable

Wetzel402:


homeassistant:
Hi @Wetzel402, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;supervisor#1931).

Thanks!

frenck:
@Wetzel402 Could you elaborate on this change? I can understand this change for the Home Assistant Container (as it can be run without the Supervisor on any OCI containerization system), however, this does not apply to the Supervisor.

pvizeli:
Do you have a supported system which need this change? Otherwise it going to be closed because

pvizeli:
We support no system they not support jemalloc. We use it a lot over all pleace, so it need a lot more code as just disable it in env variable. I'm fine to to get a full solution merged, with options on API to disable it which will be routed and managed over all containers.

Wetzel402:
@frenck, @pvizeli ,

Thank you for the clarification.  I found that my QNAP TS-431 (ARM) has problems with jemalloc (unsupported page size error) when running containerized home assistant or supervisor.  I saw that home assistant can have jemalloc disabled with a simple environment variable and thought the same could work in supervisor.  Unfortunately it doesn't seem there is an easy fix for this problem as I understand jemalloc hard codes the page size at build time.

Wetzel402:
@frenck , @pvizeli ,

Looking at the extracted supervisor docker image it appears jemalloc is built from source so the issue of incompatibility might be more than just the hard coded page size issue reported by others.

Issue #1932, opened 2020-08-16
Supervisor version 234 crashes on Ubuntu 18.04.5 LTS


DrBlokmeister:
**Home Assistant release with the issue:**

- Home assistant: 0.113.0
- Supervisor: 234

**Operating environment (HassOS/Generic):**

Installation on a generic Linux host

- Intel NUC Kit NUC8i5BEH
- Ubuntu 18.04.5 LTS

**Supervisor logs:**

```
20-08-16 18:38:48 INFO (SyncWorker_2) [supervisor.docker.interface] Attach to hassioaddons/grafana-amd64 with version 5.1.1
20-08-16 18:38:48 INFO (SyncWorker_1) [supervisor.docker.interface] Attach to homeassistant/amd64-addon-rpc_shutdown with version 2.2
20-08-16 18:38:48 INFO (SyncWorker_2) [supervisor.docker.interface] Attach to hassioaddons/ssh-amd64 with version 7.0.1
20-08-16 18:38:48 INFO (SyncWorker_0) [supervisor.docker.interface] Attach to hassioaddons/mqtt with version 1.2.0
20-08-16 18:38:48 INFO (SyncWorker_3) [supervisor.docker.interface] Attach to hassioaddons/sonweb-amd64 with version 0.11.0
20-08-16 18:38:48 INFO (SyncWorker_1) [supervisor.docker.interface] Attach to homeassistant/amd64-addon-deconz with version 6.2.3
20-08-16 18:38:48 INFO (SyncWorker_5) [supervisor.docker.interface] Attach to hassioaddons/influxdb-amd64 with version 3.7.3
20-08-16 18:38:48 INFO (SyncWorker_4) [supervisor.docker.interface] Attach to homeassistant/amd64-addon-mosquitto with version 5.1
20-08-16 18:38:48 INFO (SyncWorker_6) [supervisor.docker.interface] Attach to homeassistant/amd64-addon-mariadb with version 2.2.1
20-08-16 18:38:48 INFO (SyncWorker_2) [supervisor.docker.interface] Attach to hassioaddons/vscode-amd64 with version 2.4.0
20-08-16 18:38:48 INFO (SyncWorker_7) [supervisor.docker.interface] Attach to homeassistant/amd64-addon-check_config with version 3.4.0
20-08-16 18:38:48 INFO (SyncWorker_8) [supervisor.docker.interface] Attach to hassioaddons/tautulli with version 1.5.0
[...]

PR #1936, opened 2020-08-17
Don't break startup with corrupt docker filesystem

pvizeli:
Fix: #1932


supervisor/core.py
```diff
@@ -144,8 +149,11 @@ async def setup(self):
             _LOGGER.error("Systemd DBUS is not connected")
 
         # Check if image names from denylist exist
-        if await self.sys_run_in_executor(self.sys_docker.check_denylist_images):
-            self.coresys.supported = False
+        try:
+            if await self.sys_run_in_executor(self.sys_docker.check_denylist_images):
+                self.coresys.supported = False
+                self.healthy = False
+        except DockerAPIError:
```
MartinHjelmare, line 156: Possibly use an `else:` block to only wrap the line that can raise.

PR #1937, opened 2020-08-17
MVP: Add Network Manager context

ludeeus:
As a part of the MVP this will be able to do the following:

- Get information about all active connections (interfaces)
- Update basic configuration for an interface:
  - IPv4 address
  - Nameservers (IPv4)
  - Method (DHCP/Static) 
  - The gateway address.

This PR adds the groundwork for future projects (WiFi, Bluetooth, IPv6)

Known missing parts:

- [x] Change and apply settings on interfaces
- [x] Update documentation (API.md)
- [x] Tests
- [x] Tests


supervisor/dbus/network/const.py
```diff
@@ -12,17 +12,19 @@
 ATTR_ADDRESS_DATA = "AddressData"
 ATTR_CONFIGURATION = "Configuration"
 ATTR_CONNECTION = "Connection"
+ATTR_CONNECTIONS = "connections"
 ATTR_DEFAULT = "Default"
 ATTR_DOMAINS = "domains"
 ATTR_FILENAME = "filename"
 ATTR_FLAGS = "flags"
-ATTR_GATEWAY = "gateway"
+ATTR_GATEWAY = "Gateway"
```
pvizeli, line 20: we use lowercase for our API

API.md
```diff
@@ -442,6 +442,31 @@ Proxy to Home Assistant Core websocket.
...
+  "connections": [
+    {
+      "ip_address": "192.168.2.149",
+      "Gateway": "192.168.2.1",
+      "Id": "Wired connection 1",
+      "Type": "802-3-ethernet"
+    }
+  ]
+}
+```
```
pvizeli, line 468: Let's simplify this (also for CLI):
```
{
  "connections": [
    {
      "ip_address": "192.168.2.149",
      "gateway": "192.168.2.1",
      "id": "Wired connection 1",
      "type": "802-3-ethernet",
      "primary": "bool"
    }
  ]
}
```

pvizeli: I know networkmanager handle it different, but we don't need follow exactly his schema, we can map this between our network layer and the networkmanager dbus.

ludeeus: In that case, just drop the "connection" and just serve a list I guess

supervisor/dbus/network/const.py
ludeeus: Yeah, this was fast and dirty to get something working.
I will move attributes used for the api to supervisor/const.py (or maybe create a new one under supervisor/api/const.py), and have attrs for dbus responses here.

API.md
ludeeus: Looks like a dict is required for it to return anything :shrug:

supervisor/api/network.py
```diff
@@ -0,0 +1,81 @@
...
+    async def info(self, request: web.Request) -> Dict[str, Any]:
+        """Return network information."""
+        interfaces = {}
+        for interface in self.sys_dbus.network.interfaces:
+            interfaces[self.sys_dbus.network.interfaces[interface].name] = {
+                ATTR_IP_ADDRESS: self.sys_dbus.network.interfaces[interface].ip_address,
+                ATTR_GATEWAY: self.sys_dbus.network.interfaces[interface].gateway,
+                ATTR_ID: self.sys_dbus.network.interfaces[interface].id,
+                ATTR_TYPE: self.sys_dbus.network.interfaces[interface].type,
+                ATTR_PRIMARY: self.sys_dbus.network.interfaces[interface].primary,
```
pvizeli, line 41: you can use `as_dict` from attrs if you like

ludeeus: Yeah, but I also have to do an "on the fly" conversion, and with this approach, there is more control to what's show.
Since it's used in multiple API responses I have extracted the dict generation to a function that is called with both API's

API.md
```diff
@@ -442,6 +442,77 @@ Proxy to Home Assistant Core websocket.
...
+
+```json
+{
+  "interfaces": {
+    "enp0s31f6": {
+      "ip_address": "192.168.2.149",
+      "gateway": "192.168.2.1",
+      "id": "Wired connection 1",
+      "type": "802-3-ethernet",
+      "nameservers": "192.168.2.1",
```
pvizeli, line 461: Let's make it as list

supervisor/api/network.py
```diff
@@ -31,7 +31,7 @@
         vol.Optional(ATTR_ADDRESS): vol.Coerce(str),
         vol.Optional(ATTR_METHOD): vol.In(ATTR_METHODS),
         vol.Optional(ATTR_GATEWAY): vol.Coerce(str),
-        vol.Optional(ATTR_DNS): vol.Coerce(str),
+        vol.Optional(ATTR_DNS): vol.Schema([vol.Coerce(str)]),
```
pvizeli, line 34:
```suggestion
        vol.Optional(ATTR_DNS): [str],
```

I used the `Coerce` bad for string because it converts null to "None". If we need a string and they need not converted (because it could be a number), let's use only `str`

tests/api/test_network.py
```diff
@@ -47,7 +47,8 @@ async def test_api_network_interface_update(aiohttp_client, coresys):
     """Test network manager api."""
     client = await aiohttp_client(await create_app(coresys))
```
pvizeli, line 48: Let's make a fixture on `tests/conftest.py` for this client

ludeeus: This will change for each part of the API that is tested.
Unless we add _all_ endpoints to the fixtrue

pvizeli: Yes, let's do that

ludeeus: That would be a good amount of duplicate code, I'll see if I find a smart way to do it

pvizeli (approved):

PR #1947, opened 2020-08-18
Fix timeout and reflect s6-overlay

pvizeli:


supervisor/docker/interface.py
```diff
@@ -29,7 +29,7 @@ def __init__(self, coresys: CoreSys):
     @property
     def timeout(self) -> int:
         """Return timeout for Docker actions."""
-        return 30
+        return 10
```
frenck, line 32: I'm confused, isn't this to short?
A shutdown of a container can take longer than 10 seconds...

pvizeli: The internal timeout is 3s for each service + cleanup. This is only the default value that is used for docker. I think 10sec default should be fine for all container which did not overwrite this property.

frenck (approved):

PR #1952, opened 2020-08-20
Fix setup module list

pvizeli:


ludeeus (approved):

setup.py
```diff
@@ -35,10 +35,18 @@
         "supervisor.docker",
         "supervisor.addons",
         "supervisor.api",
+        "supervisor.dbus",
```
MartinHjelmare, line 38: Do we need to list the packages explicitly or can we just find them with `setuptools.find_packages`?

PR #1970, opened 2020-08-25
Watchdog for Add-ons

pvizeli:
Monitor container + allow to monitor an application with port / http requests

### DIFF
--- a/supervisor/addons/addon.py
+++ b/supervisor/addons/addon.py
@@ -35,6 +35,7 @@ from ..const import (
     ATTR_USER,
     ATTR_UUID,
     ATTR_VERSION,
+    ATTR_WATCHDOG,
     DNS_SUFFIX,
     STATE_STARTED,
     STATE_STOPPED,
@@ -156,6 +157,29 @@ class Addon(AddonModel):
         self.persist[ATTR_AUTO_UPDATE] = value
 
     @property
+    def watchdog(self) -> bool:
+        """Return True if watchdog is enable."""
+        return self.persist[ATTR_WATCHDOG]
+
+    @watchdog.setter
+    def watchdog(self, value: bool) -> None:
+        """Set watchdog enable/disable."""
+        self.persist[ATTR_WATCHDOG] = value
+
+    @property
+    def watchdog_network(self) -> Optional[int]:
+        """Return port to monitor or None."""
+        docker_port: str = self.data.get(ATTR_WATCHDOG)
+        if not docker_port:
+            return None
+
+        # Evaluate port
+        port: int = int(docker_port.split("/")[0])
+        if self.host_network:
+            return self.ports.get(docker_port, port)
+        return port
+
+    @property
     def uuid(self) -> str:
         """Return an API token for this add-on."""
         return self.persist[ATTR_UUID]
@@ -244,10 +268,6 @@ class Addon(AddonModel):
             port = t_port
         else:
             port = self.ports.get(f"{t_port}/tcp", t_port)
-
-        # for interface config or port lists
-        if isinstance(port, (tuple, list)):
-            port = port[-1]
 
         # lookup the correct protocol from config
         if t_proto:
--- a/supervisor/addons/addon.py
+++ b/supervisor/addons/addon.py
@@ -524,12 +524,12 @@ class Addon(AddonModel):
             return False
         return True
 
-    def is_fails(self) -> Awaitable[bool]:
-        """Return True if a Docker container is fails state.
+    def is_failed(self) -> Awaitable[bool]:
+        """Return True if a Docker container is failed state.
 
         Return a coroutine.
         """
-        return self.instance.is_fails()
+        return self.instance.is_failed()
 
     async def state(self) -> str:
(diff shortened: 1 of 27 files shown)


supervisor/addons/addon.py
```diff
@@ -462,6 +487,13 @@ def test_update_schema(self) -> bool:
             return False
         return True
 
+    def is_fails(self) -> Awaitable[bool]:
+        """Return True if a Docker container is fails state.
```
MartinHjelmare, line 491:
```suggestion
    def is_failed(self) -> Awaitable[bool]:
        """Return True if a Docker container is in a failed state.
```

supervisor/addons/addon.py
```diff
@@ -462,6 +487,13 @@ def test_update_schema(self) -> bool:
...
+
+        Return a coroutine.
+        """
+        return self.instance.is_fails()
```
MartinHjelmare, line 495: It would be good to rename this method too if possible.

supervisor/misc/tasks.py
```diff
@@ -344,5 +350,37 @@ async def _watchdog_addon_docker(self):
...
+            # if Addon have running actions / Application work
+            if addon.in_progress or await addon.watchdog_application():
+                continue
+
+            # Look like we run into a problem
+            retry_scan += 1
+            if retry_scan == 1:
+                self._cache[addon.slug] = retry_scan
+                _LOGGER.warning(
+                    "Watchdog miss application response from %s", addon.slug
```
MartinHjelmare, line 375:
```suggestion
                    "Watchdog missing application response from %s", addon.slug
```

tests/test_validate.py
```diff
@@ -47,7 +47,7 @@ def test_dns_server_list_combined():
     assert validate.dns_server_list(combined)
     # test max_length is OK still
     assert validate.dns_server_list(combined)
-    # test that it fails when the list is too long
+    # test that it failed when the list is too long
```
MartinHjelmare, line 50:
```suggestion
    # test that it fails when the list is too long
```

supervisor/addons/addon.py
```diff
@@ -353,6 +384,48 @@ def save_persist(self) -> None:
...
+        else:
+            proto = s_prefix
+
+        # Make HTTP request
+        try:
+            url = f"{proto}://{self.ip_address}:{port}{s_suffix}"
+            async with self.sys_websession_ssl.get(
+                url, timeout=WATCHDOG_TIMEOUT
+            ) as req:
+                if req.status < 300:
```
ludeeus, line 422: Should this be < 400, redirects should be valid?

supervisor/addons/addon.py
```diff
@@ -64,8 +67,15 @@
...
 )
 
+RE_WATCHDOG = re.compile(
+    r"^(?:(?P<s_prefix>https?|tcp)|\[PROTO:(?P<t_proto>\w+)\])"
+    r":\/\/\[HOST\]:\[PORT:(?P<t_port>\d+)\](?P<s_suffix>.*)$"
+)
+
 RE_OLD_AUDIO = re.compile(r"\d+,\d+")
 
+WATCHDOG_TIMEOUT = aiohttp.ClientTimeout(total=5)
```
ludeeus, line 77: Is 5 enough, some API's are slow to respond on slow HW.
We should not let the watchdog send an addon in a boot loop because the API of the application is to slow to respond.

supervisor/api/addons.py
```diff
@@ -213,7 +215,7 @@ async def info(self, request: web.Request) -> Dict[str, Any]:
             ATTR_MACHINE: addon.supported_machine,
             ATTR_HOMEASSISTANT: addon.homeassistant_version,
             ATTR_URL: addon.url,
-            ATTR_STATE: STATE_NONE,
+            ATTR_STATE: AddonState.UNKNOWN,
```
ludeeus, line 218: The addon is initalized with `AddonState.UNKNOWN` we can use `addon.state` here

supervisor/api/addons.py
```diff
@@ -255,12 +257,13 @@ async def info(self, request: web.Request) -> Dict[str, Any]:
...
             ATTR_INGRESS_PORT: None,
             ATTR_INGRESS_PANEL: None,
+            ATTR_WATCHDOG: None,
         }
 
         if isinstance(addon, Addon) and addon.is_installed:
             data.update(
                 {
-                    ATTR_STATE: await addon.state(),
+                    ATTR_STATE: addon.state,
```
ludeeus, line 266: No need for this line with the change above.

supervisor/addons/addon.py
pvizeli: we follow redirect

supervisor/api/addons.py
pvizeli: It can be also an store add-on

supervisor/api/addons.py
pvizeli: because of it :)

ludeeus (approved):

PR #1972, opened 2020-08-26
Update frontend to 04df6c3

ludeeus:
This brings in the new system tab which includes the UI dialog for NetworkManager :tada:


pvizeli:
🐬

PR #1982, opened 2020-08-27
Split payload so we can set auto without IP configuration

ludeeus:
- Fixes issue when setting DHCP an keeping the IP.
- Discover all active ETH and WiFi interfaces with a ip4 configuration.


supervisor/dbus/const.py
```diff
@@ -48,6 +48,8 @@
 DBUS_ATTR_STATIC_HOSTNAME = "StaticHostname"
 DBUS_ATTR_STATIC_OPERATING_SYSTEM_CPE_NAME = "OperatingSystemCPEName"
 DBUS_ATTR_TYPE = "Type"
+DBUS_ATTR_TYPE_ETH = "802-3-ethernet"
+DBUS_ATTR_TYPE_WIFI = "802-11-wireless"
```
ludeeus, line 52: Move this to enum

pvizeli (approved):

PR #1991, opened 2020-08-29
Add log about needing to rerun the installation script

ludeeus:
Add log about needing to rerun the installation so users can fix their own issues


frenck (approved):
Nice! That will reduce some questions 👍

pvizeli (commented):
That show a lot wrong comments on ubuntu. Let's check if we run with privileged rights on start, if not - mark system as unsupported. We have a function for this allready. Look like I removed the warning in past for it

PR #1992, opened 2020-08-29
Sentry send env infos

pvizeli:


supervisor/misc/filter.py
```diff
@@ -34,6 +34,8 @@ def filter_data(coresys: CoreSys, event: dict, hint: dict) -> dict:
     if not coresys.config.diagnostics or not coresys.core.supported or dev_env:
         return None
 
+    event.setdefault("extra", {}).update({"os.environ": dict(os.environ)})
```
ludeeus, line 37: Can we add this to the "context" instead as a new entry in that dict?

And maybe also create a new dict from it, so we can filter out any keys containing "token"

pvizeli: Because it's mach with the sys.args feature on the event.

We don't have any kind of trust or secrets on the env for supervisor

pvizeli: image

ludeeus (approved):

PR #1996, opened 2020-08-29
Adds options_validate API

ludeeus:
Add an API endpoint to addons to validate the current options


pvizeli:
Do we realy need an new endpoint? We can change the schema fo options API call to allow all and validate it before we write it to add-ons like this

pvizeli (approved):

PR #2005, opened 2020-09-03
Prevent enable watchdog on startup once

pvizeli:


supervisor/addons/addon.py
```diff
@@ -185,7 +186,10 @@ def watchdog(self) -> bool:
     @watchdog.setter
     def watchdog(self, value: bool) -> None:
         """Set watchdog enable/disable."""
-        self.persist[ATTR_WATCHDOG] = value
+        if value and self.startup == AddonStartup.ONCE:
+            _LOGGER.warning("Ignore watchdog on %s", self.slug)
```
ludeeus, line 190: Maybe add the reason why it's ignored

ludeeus (approved):

supervisor/addons/addon.py
ludeeus, line 190:
```suggestion
            _LOGGER.warning("Ignoring watchdog for %s because startup type is 'once'", self.slug)
```

PR #2006, opened 2020-09-03
Addon api changes

ludeeus:
- Add startup key to api/info
- Don't raise an error when validating configuration for addon, just return it to the user


supervisor/api/addons.py
```diff
@@ -320,10 +322,12 @@ async def options(self, request: web.Request) -> None:
     async def options_validate(self, request: web.Request) -> None:
         """Validate user options for add-on."""
         addon = self._extract_addon_installed(request)
+        data = {"message": ""}
         try:
             addon.schema(addon.options)
```
bramkragten, line 327:
```suggestion
            addon.schema(addon.options)
            data["valid"] = true
```

supervisor/api/addons.py
```diff
@@ -320,10 +322,12 @@ async def options(self, request: web.Request) -> None:
...
         except vol.Invalid as ex:
-            raise APIError(humanize_error(addon.options, ex)) from None
+            data["message"] = humanize_error(addon.options, ex)
```
bramkragten, line 329:
```suggestion
            data["valid"] = false
            data["message"] = humanize_error(addon.options, ex)
```

pvizeli (commented):
Please use ATTR_ variable for API returns and update API.md

pvizeli (approved):

PR #2023, opened 2020-09-07
Add repr for add-ons

pvizeli:


supervisor/addons/addon.py
```diff
@@ -87,6 +87,10 @@ def __init__(self, coresys: CoreSys, slug: str):
         self.instance: DockerAddon = DockerAddon(coresys, self)
         self.state: AddonState = AddonState.UNKNOWN
 
+    def __repr__(self) -> str:
```
ludeeus, line 90: Shouldn't this be in `AddonModel` instead?

pvizeli: No, they own by Addon and Store

ludeeus (approved):

PR #2028, opened 2020-09-08
implify API calls for supervisor

ludeeus:
Currently, when a user changes the "channel" for instance in the UI, we have to make 3 API calls for it to reflect in the UI.
With this change, that drops to 1 call.

### DIFF
--- a/supervisor/api/supervisor.py
+++ b/supervisor/api/supervisor.py
@@ -145,8 +145,11 @@ class APISupervisor(CoreSysAttributes):
 
         self.sys_updater.save_data()
         self.sys_config.save_data()
-        self.reload()
 
+        # Reload the supervisor after options changed
+        await asyncio.shield(self.reload())
+
+        # Return the changed data
         return await self.info()
 
     @api_process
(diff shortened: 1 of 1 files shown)


supervisor/api/supervisor.py
```diff
@@ -146,6 +146,12 @@ async def options(self, request: web.Request) -> None:
         self.sys_updater.save_data()
         self.sys_config.save_data()
 
+        # Reload the supervisor after options changed
+        await asyncio.shield(self.reload())
```
pvizeli, line 150: We do no access on options API. That is up to the frontend to make it right

supervisor/api/supervisor.py
```diff
@@ -146,6 +146,12 @@ async def options(self, request: web.Request) -> None:
...
+
+        # Return the changed data
+        return await self.info()
```
pvizeli, line 153: All `options` API work in same way. If you want that, you need update all API calls with `options` + update the API.md

pvizeli (changes requested):

supervisor/api/supervisor.py
pvizeli: same above

ludeeus:
Let's wait with this until the new API handler.

supervisor/api/supervisor.py
pvizeli: Please looking also into CLI and make sure this fronted don't show that

Issue #1866, opened 2020-08-03
RFC: Snapshot also Home Assistant OS settings


pvizeli:
## Context

Currently, we only snapshot Supervisor settings but no Home Assistant OS data. If people move between devices, they lost settings like hostname or other customization which are possible with Home Assistant OS. The goal for the OS is just to be a layer for the Supervisor and all logic running and controlled by the Supervisor. We need archive this but allow also the minimal customization and address the differences between different devices/systems.

## Decision

We map `/mnt/boot` and `/mnt/overlay` into supervisor as `/os/{boot|overlay}`. We add also a new metadata field into snapshot data to know which machine and version the OS data have. If there is no OS (because of old Snapshot or just a Supervisord installation), we keep it like it is.

The snapshot gets new options for Home Assistant OS. However, the restore of OS data is only possible on the same mayor OS level. Also, we only snapshot `config.txt` and `cmd.txt` which are only restore on the same machine type (We need to maintain a list between compatible machine type).

pvizeli:
Implemented on OS > 5.1 / 4.12

Misiu:
@pvizeli does it mean that after enabling I2C or 1Wire we can use snapshot functionality to restore (enable) the same config on a different machine (assuming it's the same type, for example, pi3/4)?
This would be a massive help because many users have a problem with this, like in this issue: https://github.com/Misiu/argon40/issues/1.

According to the instructions: https://community.home-assistant.io/t/i2c-not-working-on-the-pi4/160489/7?u=misiu to enable I2C 2 files are needed:
hassos-boot\config.txt
hassos-overlay\etc\modules-load.d\rpi-i2c.conf

the first one (`config.txt`) is mentioned by you, but what about files in overlay partition?

pvizeli:
You can use the config feature to modify the modules-load.d for now until there come up a new solution

Misiu:
I need to add a file called `rpi-i2c.conf` to `assos-overlay\etc\modules-load.d`.
Changing only the config.txt won't be enough.
But we are one step closer thanks to your change.
backup and restore of `config.txt` will allow for easy enabling/disabling 1Wire, but I2C also requires creating that extra file.

Are there any security reasons why the rest (`etc\modules-load.d` folder) can't be also added?

pvizeli:
@agners I think we should have that into the OS agent that we can send him a tarfile path for restore somethings. I going to store the Network Settings into snapshot since we have a Layer on it

[...]

PR #2038, opened 2020-09-10
Support for installing add-ons from password protected registries

skateman:
I started playing around with private docker registries and realized that the supervisor can't deal with addons living in them. In my specific case the problem is that the GitHub Packages Docker registry cannot be accessed anonymously.

The idea here is to have a docker client's configfile under `HASSIO_DATA/docker.json`. The registries specified in this file are iterated through and the docker client authenticates against each of these. This allows us to pull addons from password-protected registries.

https://community.home-assistant.io/t/add-ons-from-password-protected-docker-registries/225746

closes https://github.com/home-assistant/supervisor/issues/2039


tests/docker/test_login.py
```diff
@@ -0,0 +1,41 @@
...
+        call(registry="home-assistant.io", username="user_1", password="pass_1"),
+        call(registry="example.com", username="user_2", password="pass_2"),
+    ]
+
+    docker.docker.login.assert_has_calls(calls, any_order=True)
+
+    supervisor.docker.DOCKERCFG = old_dockercfg
+
+
+def test_login_configfile_folder(docker):
```
skateman, line 30: The reason behind this test is that if the mounted file doesn't exist on the host, docker creates a folder in the target container with that name.

pvizeli:
We have here 2 layers:
- OS need store the credential and manage this
- Supervisor can manage the logins

We can provide an API endpoint which allow user to login, but we don't need store this data. On OS we need make sure that docker can write this credential persists and they get snapshot with RFC https://github.com/home-assistant/supervisor/issues/1866

pvizeli:
Otherwise it's difficult to restore addon if we not directly know if we can pull the image. If we want handling this on Supervisor:

- create docker/login.py with a login manager and use our json logic and attach it to DockerAPI
- create an API endpoints to manage that secrets
- on every add-on install, let's give this manager the registry url and he make sure they is logged in

We don't map files into container, all need following the same logic and exposed to API

skateman:
@pvizeli I tried to implement the stuff based on your recommendations, but some stuff weren't clear and I was improvising a lot by looking at existing pieces of code. So I guess there will be some issues and the tests are missing for now, but it does what you described and it's ready for a round of review.

supervisor/docker/validate.py
```diff
@@ -0,0 +1,17 @@
...
+    {
+        vol.Coerce(str): {
+            vol.Required(ATTR_USERNAME, default=None): vol.Coerce(str),
+            vol.Required(ATTR_PASSWORD, default=None): vol.Coerce(str),
+        }
+    }
+)
+
+SCHEMA_DOCKER_CONFIG = vol.Schema(
+    {vol.Optional(ATTR_REGISTRIES, default=dict): SCHEMA_DOCKER_REGISTRY}
```
pvizeli, line 16: Make just one schema

supervisor/docker/interface.py
```diff
@@ -95,6 +95,17 @@ def _install(
 
         _LOGGER.info("Pull image %s tag %s.", image, tag)
         try:
+            # If the image name contains a path to a registry, try to find its credentials and log in
+            if "/" in image:
```
pvizeli, line 99: Every image has an `/` in, but you can use the regex from validation now. Please offload the part above into an internal function `_docker_login` on the interface.

supervisor/api/docker.py
```diff
@@ -0,0 +1,44 @@
...
+            data_registries[hostname] = {
+                ATTR_USERNAME: registry[ATTR_USERNAME],
+            }
+
+        return {ATTR_REGISTRIES: data_registries}
+
+    @api_process
+    async def create_registry(self, request: web.Request):
+        """Create a new docker registry."""
+        body = await api_validate(SCHEMA_DOCKER_REGISTRY, request)
```
pvizeli, line 32: Define the schema in that file, because the file or API can change over time and go forget where it was used. You can look to other API parts how they do it.

pvizeli (changes requested):
Looks fine. Some small adjustments, after that we can merge it 👍

pvizeli:
We need to get this data also into a snapshot and restore it

supervisor/docker/validate.py
```diff
@@ -0,0 +1,17 @@
+"""Validate docker configuration schema."""
+import voluptuous as vol
+
+from ..const import ATTR_PASSWORD, ATTR_REGISTRIES, ATTR_USERNAME
+
+SCHEMA_DOCKER_REGISTRY = vol.Schema(
+    {
+        vol.Coerce(str): {
```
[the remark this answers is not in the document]
skateman: I'm not entirely sure what does this do :confused:

supervisor/docker/interface.py
skateman: So I guess it should be a regex that matches at least two `/` in a string, right? Should the condition also go to the `_docker_login` or just anything below it?

skateman:
@pvizeli I had some questions above, can you please answer them? Thanks!

supervisor/docker/interface.py
pvizeli: The regex they we use for config validation can be use here to detect an host inside image url

supervisor/docker/validate.py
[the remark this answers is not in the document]
pvizeli: https://github.com/home-assistant/supervisor/blob/dev/supervisor/validate.py#L147-L161

supervisor/api/docker.py
pvizeli: https://github.com/home-assistant/supervisor/blob/dev/supervisor/api/supervisor.py#L55-L70

supervisor/docker/validate.py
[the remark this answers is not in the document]
skateman: I mean what does the empty regex should do here...

skateman:
> We need to get this data also into a snapshot and restore it

Gonna figure out how to do this, the rest should be done, except I still don't understand [this](https://github.com/home-assistant/supervisor/pull/2038#discussion_r488608761) :confused:

supervisor/api/docker.py
```diff
@@ -0,0 +1,53 @@
...
+from ..coresys import CoreSysAttributes
+from .utils import api_process, api_validate
+
+_LOGGER: logging.Logger = logging.getLogger(__name__)
+
+SCHEMA_DOCKER_REGISTRY = vol.Schema(
+    {
+        vol.Coerce(str): {
+            vol.Required(ATTR_USERNAME, default=None): vol.Coerce(str),
+            vol.Required(ATTR_PASSWORD, default=None): vol.Coerce(str),
```
pvizeli, line 18:
```suggestion
            vol.Required(ATTR_USERNAME): str,
            vol.Required(ATTR_PASSWORD): str,
```

supervisor/docker/validate.py
```diff
@@ -0,0 +1,21 @@
+"""Validate docker configuration schema."""
+import re
+
+import voluptuous as vol
+
+from ..const import ATTR_PASSWORD, ATTR_REGISTRIES, ATTR_USERNAME
+
+RE_REGISTRY = re.compile(r"")
```
pvizeli, line 8:
```suggestion
RE_REGISTRY = re.compile(r"^([a-z0-9]+(-[a-z0-9]+)*\.)+[a-z]{2,}$")
```

supervisor/docker/interface.py
```diff
@@ -2,21 +2,24 @@
...
-from ..const import LABEL_ARCH, LABEL_VERSION
+from ..const import ATTR_PASSWORD, ATTR_USERNAME, LABEL_ARCH, LABEL_VERSION
 from ..coresys import CoreSys, CoreSysAttributes
 from ..exceptions import DockerAPIError, DockerError, DockerNotFound, DockerRequestError
 from ..utils import process_lock
 from .stats import DockerStats
 
 _LOGGER: logging.Logger = logging.getLogger(__name__)
 
+IMAGE_WITH_HOST = re.compile(r"^([^\/]+)\/[^\/]+\/[^\/]+")
```
pvizeli, line 21:
```suggestion
IMAGE_WITH_HOST = re.compile(r"^((?:[a-z0-9]+(?:-[a-z0-9]+)*\.)+[a-z]{2,})/.+")
```

pvizeli (changes requested):
Looks good, just fix the snapshot part and it's ready to merge.

1) Update snapshot schema.
2) Decrypt/Encrypt passwort like HA token

supervisor/snapshots/__init__.py
```diff
@@ -297,6 +302,7 @@ async def do_restore_partial(
                 if FOLDER_HOMEASSISTANT in folders:
                     await self.sys_homeassistant.core.stop()
                     snapshot.restore_homeassistant()
+                    snapshot.restore_dockerconfig()
```
ludeeus, line 305: Why only restore this `if FOLDER_HOMEASSISTANT in folders` ?

supervisor/snapshots/__init__.py
```diff
@@ -238,6 +239,10 @@ async def do_restore_full(self, snapshot, password=None):
                 _LOGGER.info("Restore %s run Repositories", snapshot.slug)
                 await snapshot.restore_repositories()
 
+                # Restore docker config
+                _LOGGER.info("Restore %s run Docker Config", snapshot.slug)
+                await snapshot.restore_dockerconfig()
```
ludeeus, line 244: This should be done before the restore of repositories

supervisor/snapshots/snapshot.py
```diff
@@ -131,6 +135,16 @@ def homeassistant(self):
         """Return snapshot Home Assistant data."""
         return self._data[ATTR_HOMEASSISTANT]
 
+    @property
+    def docker(self):
+        """Return snapshot Docker config data."""
+        return self._data[ATTR_DOCKER]
```
ludeeus, line 141: This will throw a `KeyError` if  `ATTR_DOCKER` is not in the snapshot?

supervisor/snapshots/snapshot.py
```diff
@@ -481,3 +495,28 @@ def restore_repositories(self):
...
+
+    def restore_dockerconfig(self):
+        """Restore the configuration for Docker."""
+        self.sys_docker.config.registries.update(
+            {
+                registry: {
+                    ATTR_USERNAME: credentials[ATTR_USERNAME],
+                    ATTR_PASSWORD: self._decrypt_data(credentials[ATTR_PASSWORD]),
+                }
+                for registry, credentials in self.docker[ATTR_REGISTRIES].items()
```
ludeeus, line 519: This will throw a `TypeError` if  `self.docker` is not a dict (`None` if not set)?

ludeeus (commented):
Added a few comments, please don't force push after review has started, it makes it hard to follow the process :+1:

supervisor/snapshots/snapshot.py
```diff
@@ -481,3 +495,28 @@ def restore_repositories(self):
         Return a coroutine.
         """
         return self.sys_store.update_repositories(self.repositories)
+
+    def store_dockerconfig(self):
+        """Store the configuration for Docker."""
+        self.docker = {
+            ATTR_REGISTRIES: {
+                registry: {
+                    ATTR_USERNAME: credentials[ATTR_USERNAME],
```
pvizeli, line 504: Should we also encrypt the username?

pvizeli (approved):
Ready to merge after @ludeeus commits are addressed

supervisor/snapshots/snapshot.py
ludeeus: In what scenario would that be needed?

pvizeli: just a Suggestion

supervisor/snapshots/__init__.py
skateman: Should this be restored unconditionally? This is not a folder, but a file, so I can't just pass it to `folders` I guess. I could add a new attribute that would list all the configfiles we want to restore (and back up), but I am afraid that would be a more invasive change than this whole PR itself.

supervisor/snapshots/snapshot.py
skateman: So should this be encrypted as well? TBH I don't see why, but your call...

ludeeus: I think it's fine to skip it for now, we can always add that later if needed

supervisor/snapshots/__init__.py
ludeeus: Yeah, we should always store and restore this configuration, and the restoration of it needs to happen before the restoration of anything else container-related.

supervisor/snapshots/snapshot.py
```diff
@@ -131,6 +135,16 @@ def homeassistant(self):
         """Return snapshot Home Assistant data."""
         return self._data[ATTR_HOMEASSISTANT]
 
+    @property
+    def docker(self):
+        """Return snapshot Docker config data."""
+        return self._data.get(ATTR_DOCKER) is not None
```
ludeeus, line 141: This returns a bool

```suggestion
        return self._data.get(ATTR_DOCKER, {})
```

supervisor/snapshots/__init__.py
```diff
@@ -234,6 +235,10 @@ async def do_restore_full(self, snapshot, password=None):
                     self.sys_homeassistant.core.update(snapshot.homeassistant_version)
                 )
 
+                # Restore docker config
+                _LOGGER.info("Restore %s run Docker Config", snapshot.slug)
+                await snapshot.restore_dockerconfig()
```
ludeeus, line 240: Since we support custom image for homeassistant, move this to before the homeassistant restore

supervisor/snapshots/__init__.py
```diff
@@ -313,6 +318,10 @@ async def do_restore_partial(
                         )
                     )
 
+                # Restore docker config
+                _LOGGER.info("Restore %s run Docker Config", snapshot.slug)
+                snapshot.restore_dockerconfig()
```
ludeeus, line 323: Move this so it's the first action under `async with snapshot:`

supervisor/snapshots/__init__.py
ludeeus: restore_dockerconfig is not awaitable

ludeeus (changes requested):

skateman:
@ludeeus thanks for your help, I fixed the issues. Should I squash my commits before merging, or is it okay as it is?

ludeeus:
We squash all PRs on merge, so no need for you to do that in your branch 👍

supervisor/docker/__init__.py
```diff
@@ -64,6 +72,19 @@ def check_requirements(self) -> None:
...
+    """Home Assistant core object for Docker configuration."""
+
+    def __init__(self):
+        """Initialize the JSON configuration."""
+        super().__init__(FILE_HASSIO_DOCKER, SCHEMA_DOCKER_CONFIG)
+
+    @property
+    def registries(self) -> Dict[str, Any]:
+        """Return credentials for docker registries."""
+        return self._data[ATTR_REGISTRIES]
```
ludeeus, line 85:
```suggestion
        return self._data.get(ATTR_REGISTRIES, {})
```

API.md
```diff
@@ -1233,3 +1233,31 @@ We support:
     "password": "new-password"
   }
   ```
+
+### Docker Registries
+
+You can configure password-protected Docker registries that can be used as a
+source when pulling add-ons.
```
ludeeus, line 1240:
```suggestion
source when pulling docker images.
```

ludeeus (changes requested):

ludeeus (approved):
Looks good :+1: 
Since there are no tests here, have you tested the current code?

skateman:
@ludeeus I did, but I will run one more round with the edge cases you discovered, e.g. old snapshot without a docker config. Give me like an hour...

skateman:
I downloaded the HassOS VM and went with `docker exec` into the supervisor, installed the `patch` command, applied the changes from this PR and ran the `compileall` command I found in the Dockerfile. Then outside the container I ran `docker commit` and `docker tag` on the supervisor image so both the `latest` and `245` had my changes. I immediately created a snapshot after the restart and also between the following steps. I added a [new addon repository](https://github.com/skateman/hh-hassio-repo/) and set its credentials using the API. I was able to install both packages from the repo and I was able to restore back and forth any snapshot and the changes were always there without any error.

ludeeus:
Nice work @skateman! :medal_sports:

Issue #1884, opened 2020-08-10
All addon roles need /host/info endpoint access


gdrapp:
**Home Assistant release with the issue:**
0.114.0.dev20200802

**Operating environment (HassOS/Generic):** 
VSCode devcontainer / HassOS on production instance

**Supervisor logs:**
20-08-10 16:07:03 WARNING (MainThread) [supervisor.api.security] /host/info no role for 9a32d4f4_amazon-s3-backup
20-08-10 16:07:03 ERROR (MainThread) [supervisor.api.security] Invalid token for access /host/info

**Description of problem:**
The base addon [docker container](https://github.com/hassio-addons/addon-base) makes a call to the /host/info endpoint in [00-banner.sh](https://github.com/hassio-addons/addon-base/blob/1b2735f48c16179db80600f13588f660c51b9a56/base/rootfs/etc/cont-init.d/00-banner.sh#L24).

If I give my addon the "backup" role (which is all it needs), I receive the errors listed above. I have to give my addon the "manager" role to fix the issue, which is more access than it needs.

If the base container is making this call, shouldn't "/host/info" be allowed for [all addon roles](https://github.com/home-assistant/supervisor/blob/19ca485c2879c4bb795f3f59c253611839cee9c8/supervisor/api/security.py#L43)?

PR #2044, opened 2020-09-11
Expose more info

pvizeli:


gdrapp:
@pvizeli Why was '/host/info' removed from the ADDONS_API_BYPASS regex? This reverted the fix from #1884. Without this fix any addon that uses the backup role will show errors in the log on startup.

Issue #2051, opened 2020-09-14
Home Assistant UI unresponsive


hwoordes:
**Home Assistant release with the issue:**
Core version: 0.114.4

**Operating environment (HassOS/Generic):**
Running as a VM on Proxmox 6.2-11 (host kernel 5.4.60-1-pve). Used one of the virtual applicances from https://www.home-assistant.io/hassio/installation.
HassOS: 4.13

**Supervisor logs:**
I currently only have console access via Proxmox, so I can't easily attach the entire log. The image below shows the errors that are shown multiple times in the log. No other errors are shown. Any other lines are green information lines.
![su_logs2](https://user-images.githubusercontent.com/14937458/93068371-a7370580-f67c-11ea-9e6f-1cd943770e4e.png)

I'm not able to open Home Assistant in the browser (Unable to connect). 
HA was running fine for a long period of time, but suddenly stopped working. The most probable cause was an update of Proxmox or Home Assistant. No Home Assistant config changes were made.

When running a core start, core stop, core rebuild, core restart, an error is thrown. 
![core_stop](https://user-images.githubusercontent.com/14937458/93069599-25e07280-f67e-11ea-9eb8-6ada0a8c6bb7.png)

PR #2052, opened 2020-09-14
Capture typeerror exception

ludeeus:
Fixes https://github.com/home-assistant/supervisor/issues/2051

### DIFF
--- a/supervisor/utils/log_format.py
+++ b/supervisor/utils/log_format.py
@@ -3,7 +3,7 @@ import re
 
 import sentry_sdk
 
-from ..exceptions import APIError
+from ..exceptions import FormatError
 
 RE_BIND_FAILED = re.compile(r".*Bind for.*:(\d*) failed: port is already allocated.*")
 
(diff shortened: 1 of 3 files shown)


supervisor/utils/log_format.py
```diff
@@ -1,15 +1,21 @@
...
-        return (
-            f"Port '{match.group(1)}' is already in use by something else on the host."
-        )
+    try:
+        match = RE_BIND_FAILED.match(message)
+        if match:
+            return f"Port '{match.group(1)}' is already in use by something else on the host."
+    except TypeError as err:
+        sentry_sdk.capture_exception(err)
+        raise APIError("Type of message is not string") from None
```
pvizeli, line 19: We can't raise an APIError because we use that on the API response level

ludeeus: Would `HassioError` be better?

supervisor/utils/log_format.py
```diff
@@ -16,6 +17,6 @@ def format_message(message: str) -> str:
             return f"Port '{match.group(1)}' is already in use by something else on the host."
     except TypeError as err:
         sentry_sdk.capture_exception(err)
-        raise FormatError("Type of message is not string") from None
+        _LOGGER.error("Type of message is not string - %s", err)
```
pvizeli, line 20: let's print the error and capture after that the exception, so we get all in one sentry

pvizeli (approved):

PR #2057, opened 2020-09-15
Disable auto boot for add-ons with missconfig

pvizeli:
Don't flood sentry and it makes no sense that we try to boot such add-ons over and over if they have a config issue or can't start because the port is already used.

Same for add-ons that need repair. Until the user have fix the issue, we don't need to try to boot it again


ludeeus (approved):
Nice! :100:

PR #2076, opened 2020-09-23
Use multipart for snapshot uploads

ludeeus:


supervisor/api/snapshots.py
```diff
@@ -182,11 +182,15 @@ async def upload(self, request):
...
+            contents = await reader.next()
             try:
                 with tar_file.open("wb") as snapshot:
-                    async for data in request.content.iter_any():
-                        snapshot.write(data)
+                    while True:
+                        chunk = await contents.read_chunk()
+                        if not chunk:
+                            break
+                        snapshot.write(chunk)
```
MartinHjelmare, line 193: This does I/O in the event loop. Probably also true for opening the file and opening and closing the temporary directory.

Not sure what's the best solution here. If we don't use context managers to handle the files we could open and close the files explicitly on the executor.

supervisor/api/snapshots.py
```diff
@@ -180,25 +180,39 @@ async def download(self, request):
...
+        contents = await reader.next()
+        try:
+            while True:
+                chunk = await contents.read_chunk()
+                if not chunk:
+                    break
+                chunks.append(chunk)
+
+        except OSError as err:
+            _LOGGER.error("Can't write new snapshot file: %s", err)
```
MartinHjelmare, line 204: This error should move too, right? Do we need to handle aiohttp errors here?

supervisor/api/snapshots.py
```diff
@@ -180,25 +180,39 @@ async def download(self, request):
...
+            return tar_file
+
+        reader = await request.multipart()
+        contents = await reader.next()
+        try:
+            while True:
+                chunk = await contents.read_chunk()
+                if not chunk:
+                    break
+                chunks.append(chunk)
```
pvizeli, line 201: Don't cache it on the memory, write it directly to the file as a stream

ludeeus: https://github.com/home-assistant/supervisor/pull/2076#discussion_r493724676
If we do that we are doing I/O in the eventloop

pvizeli: Supervisor handle all file request in I/O. File handling is a bit bit complexer and not every file access doing also IO to disk, just the close of the file doing IO.

pvizeli: The supervisor have no issue with micro blocks, unlike HA

pvizeli: Maybe we can switch to an good library with low level file handling for asyncio loop (and not just execute stuff on a executor)

ludeeus: Aiit :+1: 
I'll revert this commit then :)
<[`7f26b43` (#2076)](https://github.com/home-assistant/supervisor/pull/2076/commits/7f26b432cb057d4fb08e11993edf5f3a008e7c5e)>

pvizeli (approved):

Issue #1946, opened 2020-08-18
RFC: Job based API


pvizeli:
## Context

Currently, each API call can end up with using `process_lock` to guard an object against new action while a current action is running. 

## Decision

We create a job monitor that replaces `process_lock` but works similarly and using https://docs.python.org/3/library/contextvars.html which makes it possible that we can call other function of this object without being deadlock or need to guard this.

The API works like before, but for some calls, it creates tasks/jobs and a new API to manage/show these jobs or getting details like progress. The job monitor is allowing us to use the core event bus to inform the UI about start/stop or progress on this job.

```
MyXy(JobGroup):

    job_group = "xy"

   @Job(lock=True, wait=False, name="Optional name if not method")
   def xy():
```

And generate `/xy/update` which is set on the first context and gets inherit to child jobs. So the frontend can get the correct root job. We can also lock future access over this JobGroup. The question is, how we deal with sub progress etc.

API:

| job | id | task | parent | progress |
|-----|---|------|---------|-----------|
| /supervisor/update | 13kj123lk1j2 | Supervisor.update | null | null |
| | jkfjsdwlf23423 | DockerSupervisor.install | 13kj123lk1j2 | 20% |

## Future

We can use this with logging to attach the right error message from deep back over the API to the UI.

sanyatuning:
@pvizeli I realy want to help implement this but I need some hint.
What the new API should look like?
Here the current usage of `process_lock`.
```py
    @process_lock
    def install(self, tag: str, image: Optional[str] = None, latest: bool = False):
        """Pull docker image."""
        return self.sys_run_in_executor(self._install, tag, image, latest)
```

sanyatuning:
I started something in #1805.

pvizeli:
Please make a new PR. Need a lot more code. Like an API and job manager as coresys object. The event is just a small part of this construct and replace also all locks on code

sanyatuning:
Did you hear about MVP or KISS? :)

pvizeli:
For history: https://github.com/home-assistant/supervisor/pull/2348

agners:
Implemented by #4514.

PR #2100, opened 2020-10-06
Extract exception message from chain for API errors

ludeeus:
## Proposed change

While working on #2099 I noticed that many operations would result in a blank error message returned to the CLI/API, this change extracts the message from the exception chain

## Type of change

- [ ] Dependency upgrade

## Additional information

## Checklist

### DIFF
--- a/supervisor/addons/addon.py
+++ b/supervisor/addons/addon.py
@@ -567,7 +567,7 @@ class Addon(AddonModel):
             raise AddonsError() from err
         except DockerError as err:
             self.state = AddonState.ERROR
-            raise AddonsError(err) from err
+            raise AddonsError() from err
         else:
             self.state = AddonState.STARTED
 
--- a/supervisor/api/utils.py
+++ b/supervisor/api/utils.py
@@ -17,7 +17,8 @@ from ..const import (
     RESULT_ERROR,
     RESULT_OK,
 )
-from ..exceptions import APIError, APIForbidden, HassioError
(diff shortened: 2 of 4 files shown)


supervisor/addons/__init__.py
```diff
@@ -169,7 +169,7 @@ async def install(self, slug: str) -> None:
             await addon.instance.install(store.version, store.image)
         except DockerError as err:
             self.data.uninstall(addon)
-            raise AddonsError() from err
+            raise AddonsError(err) from err
```
pvizeli, line 172: Do we need raise from err if we own the error? Important that our exception chain checker work with the new structure as well

ludeeus: Not sure, it should, I'll add a test for it

pvizeli:
Your PR for handle error on API like for port mapping is not needed with https://github.com/home-assistant/supervisor/issues/1946

I approved it anyway, because it's don't change to match. Now we start to change the hole error handling to parse error message with regex on API return. It feel a bit like an hack.

I think instead to move forward with this, we should implement #1946 and allow to set usable error message where they are see (Like on run/start function on docker)

supervisor/addons/__init__.py
ludeeus: Actually, check the current tests for `check_exception_chain` it seems to cover all cases already?

pvizeli (approved):

PR #2125, opened 2020-10-13
Add issues/suggestion to resolution center / start with diskspace

ludeeus:
## Proposed change

If we detect a system with < 1GB free space:

- Create a persistant notification.
- Block supervisor updates
- Block add-on updates

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to documentation pull request: https://github.com/home-assistant/home-assistant.io/pull/15233 / https://github.com/home-assistant/developers.home-assistant/pull/683

## Checklist


supervisor/core.py
```diff
@@ -192,6 +192,9 @@ async def setup(self):
         except DockerError:
             self.healthy = False
 
+        # Check free space
+        await self.sys_host.check_free_space()
+
```
pvizeli, line 197: Let's remove this. We can not do anything and we don't need this information for a startup, because the system was running before.

supervisor/host/__init__.py
```diff
@@ -104,3 +104,27 @@ async def load(self):
...
+                    "message": f"Available space is {free}GB, see https://www.home-assistant.io/more-info/free-space for more information.",
+                    "notification_id": "supervisor_free_space",
+                },
+            ) as resp:
+                if resp.status in (200, 201):
+                    _LOGGER.debug("Sucessfully created persistent_notification")
+                else:
+                    _LOGGER.error("Can't create persistant notification")
+        except HomeAssistantAPIError:
+            _LOGGER.error("Can't create persistant notification")
```
pvizeli, line 130: Let's put this to a resolution object. We need also to think a step forward, how we can inform users on UI, and how we want to get this information to UI with other issues. We should create a `Resolution.system_check` which we run as a task maybe at midnight or morning which checks things like that and perform what we want.

pvizeli (commented):
The part of the tasks is fine but the other needs a bit more streamline and research with a new resolution center. But I think we get that since you are the architect of the nice unsupported feature :)

supervisor/core.py
pvizeli: normally we use the load() function on the sys_object to initialize default values.

pvizeli: See my comment on the review, we use the load function for such things. I will later also offload the supported check from setup into the right load functions because we have now a clean structure with resolution center.

ludeeus:
Converted to draft, the base structure should now be clear.
But tests are missing, and the 2 new lists on the resolution object are not in the API, same with suggestion actions

supervisor/api/resolution.py
```diff
@@ -3,15 +3,39 @@
...
+from ..resolution.const import Suggestion
 from .utils import api_process
 
 
 class APIResoulution(CoreSysAttributes):
     """Handle REST API for resoulution."""
 
     @api_process
-    async def base(self, request: web.Request) -> Dict[str, Any]:
+    async def base(self, _request: web.Request) -> Dict[str, Any]:
```
pvizeli, line 17:
```suggestion
    async def base(self, request: web.Request) -> Dict[str, Any]:
```

supervisor/resolution/__init__.py
```diff
@@ -21,4 +64,58 @@ def unsupported(self) -> List[UnsupportedReason]:
...
+            self._unsupported.append(reason)
+
+    def clear(self, collection: str) -> None:
+        """Clear the contents of the collection."""
+        if collection == "issues":
+            self._issues = []
+        elif collection == "suggestions":
+            self._suggestions = []
+        elif collection == "unsupported":
+            self._unsupported = []
```
pvizeli, line 77:
```suggestion
        if collection == "issues":
            self._issues.clear()
        elif collection == "suggestions":
            self._suggestions.clear()
        elif collection == "unsupported":
            self._unsupported.clear()
```

supervisor/resolution/__init__.py
```diff
@@ -21,4 +64,58 @@ def unsupported(self) -> List[UnsupportedReason]:
...
+
+    async def load(self):
+        """Load the resoulution manager."""
+        # Initialize sub objects
+        self._notify = ResolutionNotify(self.coresys)
+        self._storage = ResolutionStorage(self.coresys)
```
pvizeli, line 83: But this into init

supervisor/resolution/__init__.py
```diff
@@ -21,4 +64,58 @@ def unsupported(self) -> List[UnsupportedReason]:
     @unsupported.setter
     def unsupported(self, reason: UnsupportedReason) -> None:
         """Add a reason for unsupported."""
-        self._unsupported.append(reason)
+        if reason not in self._unsupported:
+            self._unsupported.append(reason)
+
+    def clear(self, collection: str) -> None:
```
pvizeli, line 70: We are going away from str which is not strict to Enum. So we need to use enum like CollectionType for it. Another question is, why not just call on code `self.issue.clear()`

ludeeus: Fair point, clearing should only happen from the resolution manager, so we can clear directly

supervisor/resolution/notify.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def __init__(self, coresys: CoreSys):
+        """Initialize the notify class."""
+        self.coresys = coresys
+
+    async def issue_notifications(self):
+        """Create persistant notifications about issues."""
+        if (
+            not self.sys_resolution.issues
+            or not self.sys_homeassistant.core.is_running()
```
pvizeli, line 27:
```suggestion
            or not self.sys_homeassistant.api.check_api_state()
```

pvizeli (approved):

PR #2144, opened 2020-10-19
Speedup HA core auth

pvizeli:
## Proposed change

The change is less as maybe it can look at first view. It's like the speed dial for wifi. Timing attacks with that are only possible if you know the password and it returns anyway True. The old password maybe still working sometimes but we speak about add-ons and the new password work all time.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR fixes or closes issue: fixes: https://github.com/home-assistant/hassio-addons/issues/1630
- Link to documentation pull request: https://github.com/home-assistant/developers.home-assistant/pull/688

## Checklist

### DIFF
--- a/supervisor/api/__init__.py
+++ b/supervisor/api/__init__.py
@@ -224,7 +224,7 @@ class RestAPI(CoreSysAttributes):
             [
                 web.post("/auth", api_auth.auth),
                 web.post("/auth/reset", api_auth.reset),
-                web.post("/auth/cache", api_auth.cache),
+                web.delete("/auth/cache", api_auth.cache),
             ]
         )
 
--- a/supervisor/api/__init__.py
+++ b/supervisor/api/__init__.py
@@ -221,7 +221,11 @@ class RestAPI(CoreSysAttributes):
         api_auth.coresys = self.coresys
 
         self.webapp.add_routes(
-            [web.post("/auth", api_auth.auth), web.post("/auth/reset", api_auth.reset)]
+            [
+                web.post("/auth", api_auth.auth),
+                web.post("/auth/reset", api_auth.reset),
+                web.post("/auth/cache", api_auth.cache),
+            ]
         )
 
     def _register_supervisor(self) -> None:
--- a/supervisor/auth.py
+++ b/supervisor/auth.py
@@ -81,7 +81,7 @@ class Auth(JsonConfig, CoreSysAttributes):
             return await self._backend_login(addon, username, password)
 
         # Home Assistant Core take over 1-2sec to validate it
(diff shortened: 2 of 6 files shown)


ludeeus:
Wouldn't this make auth be successful, even if the user object has changed in core?

pvizeli:
Sounds worse at it is in context for what this function is used

pvizeli:
I added an API for reset data with cli if user want force it

supervisor/api/__init__.py
```diff
@@ -221,7 +221,11 @@ def _register_auth(self) -> None:
         api_auth.coresys = self.coresys
 
         self.webapp.add_routes(
-            [web.post("/auth", api_auth.auth), web.post("/auth/reset", api_auth.reset)]
+            [
+                web.post("/auth", api_auth.auth),
+                web.post("/auth/reset", api_auth.reset),
+                web.post("/auth/cache", api_auth.cache),
```
ludeeus:
```suggestion
                web.delete("/auth/cache", api_auth.cache),
```

Should be delete

ludeeus (changes requested):

supervisor/api/__init__.py
ludeeus: Lol... you already did that.

ludeeus (approved):

balloob:
How often do we hit this API that it is worth potentially serving an outdated one from the cache instead of accessing the real source?

pvizeli:
I.e. mqtt hit that a lot for every connection. Some devices just connects send state and disconnect.

pvizeli:
Old:
Someone change the password (anyway not often happening on a service account). Everything stop impotently working.

Now:
Sone changes the password. All working until the cache is expired or someone logs in with new service accounts credentials.

The second provides anyway a better user experience and for pro users, they can run `ha auth cache` to force the old experience if they are really imminently needed.

pvizeli:
@balloob I implemented an extended cache handling with background jobs. I think now it's like in old day but faster

supervisor/auth.py
```diff
@@ -61,11 +68,29 @@ async def check_login(self, addon: Addon, username: str, password: str) -> bool:
...
             _LOGGER.debug("Home Assistant not running, checking cache")
-            return self._check_cache(username, password)
+            return cache_hit is True
+
+        # No cache hit
+        if cache_hit is None:
+            return await self._backend_login(addon, username, password)
+
+        # Home Assistant Core take over 1-2sec to validate it
+        # Let's use the cache and if they don't hit, we ask on core
```
ludeeus, line 84: This comment looks out of place

ludeeus (approved):

PR #2214, opened 2020-11-03
Supervisor container startup health function

pvizeli:
## Proposed change

Instead of using container timings to detect corruption on init/python/supervisor - we use a lock file mechanic which allows the OS to recreate the container/image without a false positive quote.

## Type of change

- [ ] Dependency upgrade

## Additional information

## Checklist

### DIFF
--- a/supervisor/__main__.py
+++ b/supervisor/__main__.py
@@ -2,11 +2,14 @@
 import asyncio
 from concurrent.futures import ThreadPoolExecutor
 import logging
+from pathlib import Path
 import sys
 
 from supervisor import bootstrap
 
 _LOGGER: logging.Logger = logging.getLogger(__name__)
+
+CONTAINER_OS_HEALTH = Path("/os/overlay/supervisor.health")
 
 
 # pylint: disable=invalid-name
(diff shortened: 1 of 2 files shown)


supervisor/__main__.py
```diff
@@ -2,12 +2,15 @@
...
 from concurrent.futures import ThreadPoolExecutor
 import logging
+from pathlib import Path
 import sys
 
 from supervisor import bootstrap
 
 _LOGGER: logging.Logger = logging.getLogger(__name__)
 
+CONTAINER_OS_HEALTH = Path("/os/overlay/supervisor.health")
```
agners: This puts the device onto a non-volatile storage. I think it would be better if we have it somewhere volatile.

How about creating a volatile directory shared between OS and the supervisor, e.g. bind mounting host `/run/supervisor/` to supervisor `/run/supervisor`. Then we can use `/run/supervisor/health`

agners: `health` sounds like it should stay there, how about `startup-marker`.

pvizeli: that is a good idea. The path is already used for the watchdog but we can /run/supervisor:/run/os

tests/test_main.py
```diff
@@ -0,0 +1,17 @@
+"""Testing handling with main."""
+from pathlib import Path
+
+import supervisor.__main__ as main
+
+
+def test_write_state(tmp_path):
+    """Test write corestate to /run/supervisor."""
```
agners, line 8:
```suggestion
-    """Test write corestate to /run/supervisor."""
+    """Test startup-marker file cleanup."""
```

agners (approved):

PR #2223, opened 2020-11-06
Add validate session API

balloob:
## Proposed change

Add a new API endpoint `POST /api/ingress/validate_session` that will extend the validation of the current session and returns a 401 if it is no longer valid.

I couldn't get auth working in the API test, so decided to stub it out. Not sure how the other API tests do it. Still verify that we call the check auth function.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to documentation pull request: https://github.com/home-assistant/developers.home-assistant/pull/714

## Checklist

### DIFF
--- a/supervisor/api/ingress.py
+++ b/supervisor/api/ingress.py
@@ -93,8 +93,6 @@ class APIIngress(CoreSysAttributes):
             _LOGGER.warning("No valid ingress session %s", data[ATTR_SESSION])
             raise HTTPUnauthorized()
 
-        return {}
-
     async def handler(
         self, request: web.Request
     ) -> Union[web.Response, web.StreamResponse, web.WebSocketResponse]:
--- a/supervisor/api/ingress.py
+++ b/supervisor/api/ingress.py
@@ -12,6 +12,7 @@ from aiohttp.web_exceptions import (
     HTTPUnauthorized,
 )
 from multidict import CIMultiDict, istr
+import voluptuous as vol
 
 from ..addons.addon import Addon
 from ..const import (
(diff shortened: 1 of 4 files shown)


supervisor/api/ingress.py
```diff
@@ -78,6 +78,19 @@ async def create_session(self, request: web.Request) -> Dict[str, Any]:
...
+    @api_process
+    async def validate_session(self, request: web.Request) -> Dict[str, Any]:
+        """Create a new session."""
+        self._check_ha_access(request)
+
+        # Check Ingress Session
+        session = request.cookies.get(COOKIE_INGRESS)
+        if not self.sys_ingress.validate_session(session):
+            _LOGGER.warning("No valid ingress session %s", session)
+            raise HTTPUnauthorized()
```
ludeeus, line 90: This needs to be excepted in `api_process` so we can return it like all other API calls

balloob: Are you sure? It's already there for the other ingress endpoints.

Also realized that I do it now based on cookies, rewriting it to JSON data like other APIs.

ludeeus: For the other calls that raise those, they are not using `api_process`

balloob: In `security.py`, the security middleware also just raises this error. It's a security incident, not an error.

pvizeli: I don't know how do you want to handle this on the frontend. If you want like /ingress/session then yes, You need a response with an APIError().

supervisor/api/ingress.py
```diff
@@ -78,6 +78,19 @@ async def create_session(self, request: web.Request) -> Dict[str, Any]:
...
+
+        return {}
```
pvizeli, line 96: is not need. No return generate a `{"status": "ok", "data": null}`

ludeeus (approved):

PR #2241, opened 2020-11-11
Use technology neutral "auto"  instead of "dhcp" for autoconfiguration

agners:
IPv6 is often run with SLAAC which does not use a DHCP server but still
does automatic address configuration. Follow wording of NetworkManager
and use "auto" instead.

## Type of change

- [ ] Dependency upgrade

## Additional information

## Checklist


ludeeus:
If we do this, we should just rip out our mapping and just use `NMInterfaceMethod`

pvizeli (commented):
Looks good. You need fix all tests + update documentation + cli

> If we do this, we should just rip out our mapping and just use NMInterfaceMethod

This would remove our wanted abstraction between that interfaces

ludeeus:
> This would remove our wanted abstraction between that interfaces

Correct, but with this, we are halfway between two types, which is weird.
If we still want our mapper, `STATIC` should become `MANUAL` as well

pvizeli (approved):
Fine to merge if docs & cli is ready

agners:
> Correct, but with this, we are halfway between two types, which is weird.
Well the types just happen to have very similar values, but we use clearly different types. I think abstracting from NetworkManager does make sense in case we want to switch to something else in the future...

As for `static` or `manual`, either way is fine for me, but `dhcp` is really misleading/wrong in IPv6 context where DHCPv6 is not used (which is most often the case).

agners:
@pvizeli 

- docs: https://github.com/home-assistant/supervisor/pull/2241 
- cli: https://github.com/home-assistant/cli/pull/260

PR #2245, opened 2020-11-12
Check & fix git repository integrity

pvizeli:
## Type of change

- [ ] Dependency upgrade

## Additional information

## Checklist


supervisor/resolution/const.py
```diff
@@ -37,6 +38,7 @@ class IssueType(str, Enum):
 
     FREE_SPACE = "free_space"
     CORRUPT_DOCKER = "corrupt_docker"
+    CORRUPT_REPOSITORY = "corrupt_repository"
```
agners, line 41:
```suggestion
    CORRUPT_GIT_REPOSITORY = "corrupt_git_repository"
```

supervisor/store/git.py
```diff
@@ -53,12 +59,29 @@ async def load(self):
...
-            return True
+        # Fix possible corruption
+        async with self.lock:
+            try:
+                _LOGGER.debug("Integrity check add-on %s repository", self.path)
+                await self.sys_run_in_executor(self.repo.git.execute, ["git", "fsck"])
+            except git.GitCommandError as err:
+                _LOGGER.error("Integrity check on %s failed: %s.", self.path, err)
+                self.sys_resolution.create_issue(
+                    IssueType.CORRUPT_REPOSITORY,
```
agners, line 77:
```suggestion
                    IssueType.CORRUPT_GIT_REPOSITORY,
```

supervisor/resolution/const.py
pvizeli: The idea behind it is to hold it open and short. Like CORRUPT_DOCKER, possible to split into CORRUPT_OVERLAY_DOCKER, CORRUPT_NETWORK_DOCKER, etc but in the end, it doesn't maters, just CORRUPT_DOCKER is enough to fix it. So to hold it short, we could use `CORRUPT_GIT`, but not all store repo is git repository but all are Repository objects, which route the fix to correct underlay object.

agners: IMHO, the issue type should be rather specific so that we can identify what actually is wrong with the system. The resolution to it can be more generic, as a specific solution (e.g. recreate git repository) can fix multiple issues (corrupt git repository, missing git repository...)

supervisor/store/git.py
```diff
@@ -86,16 +109,19 @@ async def clone(self):
                 git.GitCommandError,
             ) as err:
                 _LOGGER.error("Can't clone %s repository: %s.", self.url, err)
-                await self._remove()
-                return False
-
-            return True
+                self.sys_resolution.create_issue(
```
ludeeus, line 112: Why create an issue instead of just removing the bad repo as before?

pvizeli: Removing repo does not solve/fix anything. After I'm finished with the conference stuff, I create something like evolution but to apply fixes. So I can fix it right after start if the internet is back etc... Would be great stuff. Based on sentry, that code never triggered in last 6month

ludeeus (approved):

supervisor/store/git.py
ludeeus: I was worried that it would impact the `update_repositories` method that is called when a user add a repo in the UI, but it's not :+1:

supervisor/resolution/const.py
ludeeus: When we need to differentiate the different types of issues, I would agree, but that is not the case here, I'm fine with it being generic where it's currently used, if we need more separation in the future we can always add/change it up :+1:

PR #2317, opened 2020-11-30
Use persistent MAC address instead clone / don't touch default

pvizeli:
## Type of change

- [ ] Dependency upgrade

## Additional information

## Checklist

### DIFF
--- a/supervisor/dbus/payloads/interface_update.tmpl
+++ b/supervisor/dbus/payloads/interface_update.tmpl
@@ -61,7 +61,7 @@
     ,
     '802-3-ethernet':
         {
-            'assigned-mac-address': <'stable'>
+            'assigned-mac-address': <'preserve'>
         }
 {% endif %}
 
@@ -78,7 +78,7 @@
     ,
     '802-11-wireless':
         {
-            'assigned-mac-address': <'stable'>,
+            'assigned-mac-address': <'preserve'>,
             'ssid': <[byte {{ interface.wifi.ssid }}]>,
             'mode': <'{{ interface.wifi.mode.value }}'>,
             'powersave': <uint32 1>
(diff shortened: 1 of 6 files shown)


supervisor/host/network.py
```diff
@@ -279,15 +284,15 @@ def from_dbus_interface(inet: NetworkInterface) -> Interface:
                 inet.connection.ipv4.nameservers,
             )
             if inet.connection and inet.connection.ipv4
-            else None,
```
agners, line 282: It feels to me nicer to return None, but if its a hassle for frontend maybe returning an empty is easier? @ludeeus

supervisor/dbus/payloads/interface_update.tmpl
```diff
@@ -61,7 +61,7 @@
     ,
     '802-3-ethernet':
         {
-            'assigned-mac-address': <'stable'>
+            'assigned-mac-address': <'preserve'>
```
agners, line 64: Is there a reason why we set this in first place? Can we not set it at all and leave it up to NM?

pvizeli: Different NM working differently based on his version. The platform should work everywhere exactly in the same way and eliminate customizing. Also if we update NM on OS, it should not change his behavior after the update.

supervisor/host/network.py
ludeeus: Not a hassle, but this is better IMO.

ludeeus (approved):

supervisor/host/network.py
pvizeli: I agree, as we not represent NM on our abstracted interface, it's better to show what it currently is

Issue #2330, opened 2020-12-03
Installation in unhealthy state - ConnectivityCheckEnabled


EmmanuelLM:
**Systems:**
- Odroid-XU4
- OS: DietPi v6.33.3 (based on Debian 9)
- Supervisor: 2020.12.2 (stable channel)
- Home-Assistant: 0.118.4
- All running on docker 19.03.13
- Connected via ethernet

**Info**

System Health

Home Assistant Core Integration

version: 0.118.4
installation_type: Home Assistant Supervised
dev: false
hassio: true
docker: true
virtualenv: false
python_version: 3.8.6
os_name: Linux
os_version: 4.14.180+
arch: armv7l
timezone: Europe/London

Home Assistant Cloud

logged_in: true
subscription_expiration: December 16, 2020, 12:00 AM
relayer_connected: true
remote_enabled: false
remote_connected: false
alexa_enabled: false
google_enabled: true
can_reach_cert_server: ok
can_reach_cloud_auth: ok
can_reach_cloud: ok

Hass.io

host_os: Debian GNU/Linux 9 (stretch)
update_channel: stable
supervisor_version: 2020.12.2
docker_version: 19.03.13
disk_total: 28.1 GB
disk_used: 14.6 GB
healthy: failed to load: Unhealthy
supported: failed to load: Unsupported
supervisor_api: ok
version_api: ok
installed_addons: Node-RED (7.2.11), Let's Encrypt (4.11.0), Mosquitto broker (5.1), ESPHome (1.15.3), MariaDB (2.2.1), InfluxDB (3.7.9), Samba share (9.3.0), Traccar (0.9.2), Grafana (5.3.6), phpMyAdmin (0.1.4), NGINX Home Assistant SSL proxy (3.0.1), File editor (5.2.0), Zigbee2mqtt (1.16.1), TasmoAdmin (0.13.1)

Lovelace

dashboards: 2
mode: storage
views: 11
resources: 2

**Problem**

Installation was carried out with the convenience script install.sh

Apart from the 'Unsupported OS', everything was running fine.

Today, the Supervisor started throwing out the following error:
"20-12-03 14:24:13 ERROR (MainThread) [asyncio] Task exception was never retrieved
future: <Task finished name='Task-374' coro=<HostManager.reload() done, defined at /usr/src/supervisor/supervisor/host/__init__.py:90> exception=KeyError('ConnectivityCheckEnabled')>
Traceback (most recent call last):
[...]

PR #2339, opened 2020-12-04
Revert "Fix KeyError in connectivity_enabled"

pvizeli:
Reverts home-assistant/supervisor#2336


ludeeus (changes requested):
This will reintroduce #2330, we should not do that without other guards (like version check)

PR #2348, opened 2020-12-05
These are not the droids you're looking for

ludeeus:
## Type of change

- [ ] Dependency upgrade

## Additional information

## Checklist


pvizeli:
Looks fine. We need a concept, how the API should look like to be usable. I think maybe we need also something like this:

```
MyXy(JobGroup):

    job_group = "xy"

   @Job(lock=True, wait=False, name="Optional name if not method")
   def xy():
```

And generate `/xy/update` which is set on the first context and gets inherit to child jobs. So the frontend can get the correct root job. We can also lock future access over this JobGroup. The question is, how we deal with sub progress etc.

API:

| job | id | task | parent | progress |
|-----|---|------|---------|-----------|
| /supervisor/update | 13kj123lk1j2 | Supervisor.update | null | null |
| | jkfjsdwlf23423 | DockerSupervisor.install | 13kj123lk1j2 | 20% |

ludeeus:
In due time @github-actions, in due time.

ludeeus:
I think each root activity (started by API access or system), should generate a `JobGroup`
The `JobManager` would manage the `JobGroup` and the `JobGroup` would manage (create/update/remove) and track individual jobs.


So if you call `/supervisor/update`:
1. `JobManager` creates a new `JobGroup` with a uuid and that API path tracked as the name.
2. The `JobGroup` will then spawn individual `SupervisorJob` (jobs) that is needed (update apparmor, install new image, update ) to complete the API call
3. All processes are tracked individually in the jobs, and the `JobGroup` will keep overall progress.
4. The `JobGroup` will remove all associated jobs, and request removal of itself by the `JobManager` when the process is complete (success/fail)

pvizeli:
make sense, yes

PR #2429, opened 2021-01-12
Next generation hardware handling

pvizeli:
## Proposed change

Rewrite the complete udev handling inside Supervisor. Now we only based on the host udev events and all hardware interactions are based on these events/lists.

It also removes the udev handling needs from the add-on. This is now a complete handle over the supervisor. The add-on config has a new option type `device`. This will be lookup and control by Supervisor and make most of the device hot&plug able. Based on device types or USB IDs, we generate a complete list for UI selection. We work internally also with the udev ID to make sure devices get correct mapped, also if the user does not use device-by-id selectors.

These changes aim a full working device handling to the user, almost like he do run Core.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to documentation pull request: https://github.com/home-assistant/developers.home-assistant/pull/784

## Checklist


rootfs/etc/cont-init.d/udev.sh
```diff
@@ -2,9 +2,16 @@
 # ==============================================================================
 # Start udev service
 # ==============================================================================
+
+if bashio::bashio::fs.directory_exists /run/udev; then
```
frenck, line 6:
```suggestion
if bashio::fs.directory_exists /run/udev; then
```

pvizeli:
This would be also requested by Deconz @manup

supervisor/api/hardware.py
```diff
@@ -64,6 +60,6 @@ async def audio(self, request: web.Request) -> Dict[str, Any]:
         }
 
     @api_process
-    def trigger(self, request: web.Request) -> Awaitable[None]:
+    async def trigger(self, request: web.Request) -> Awaitable[None]:
         """Trigger a udev device reload."""
-        return asyncio.shield(self.sys_hardware.udev_trigger())
+        _LOGGER.warning("DEPRICATED hardware trigger function called")
```
agners, line 65:
```suggestion
        _LOGGER.warning("Ignoring DEPRECATED hardware trigger function call.")
```

supervisor/docker/addon.py
```diff
@@ -124,34 +125,57 @@ def environment(self) -> Dict[str, Optional[str]]:
...
+    def devices(self) -> Optional[List[str]]:
         """Return needed devices."""
-        devices = []
+        devices = set()
 
         # Extend add-on config
-        for device in self.addon.devices:
-            if not Path(device.split(":")[0]).exists():
+        for device_path in self.addon.static_devices:
+            if not self.sys_hardware.exists_device_node(device_path):
```
agners, line 134: Didn't we drop `sys_hardware`?

supervisor/docker/addon.py
```diff
@@ -124,34 +125,57 @@ def environment(self) -> Dict[str, Optional[str]]:
...
-
-        # Auto mapping UART devices
-        if self.addon.with_uart:
-            for device in self.sys_hardware.serial_devices:
-                devices.append(f"{device.path.as_posix()}:{device.path.as_posix()}:rwm")
-                if self.addon.with_udev:
+            devices.add(f"{device_path.as_posix()}:{device_path.as_posix()}:rwm")
+
+        # Auto mapping UART devices / LINKS
+        # Depricated: Add-on need in future self create device links based on API data
```
agners, line 140:
```suggestion
        # Deprecated: In the future the add-on needs to create device links based on API data by itself
```

supervisor/docker/addon.py
agners: Ah nevermind, that was in the `CoreSys` class.

supervisor/docker/addon.py
```diff
@@ -284,7 +308,7 @@ def volumes(self) -> Dict[str, Dict[str, str]]:
         # Init other hardware mappings
 
         # GPIO support
-        if self.addon.with_gpio and self.sys_hardware.support_gpio:
+        if self.addon.with_gpio and self.sys_hardware.helper.support_gpio:
```
agners, line 311: The new `libgpiod` uses `/dev/gpiochip*` devices. Do we support that? I guess we could just add that as well if an addon wants GPIO access. See e.g. https://embeddedbits.org/new-linux-kernel-gpio-user-space-interface/

supervisor/exceptions.py
```diff
@@ -296,6 +296,14 @@ class DockerNotFound(DockerError):
 # Hardware
 
 
+class HardwareError(HassioError):
+    """General Hardware Error on Supervisor."""
+
+
+class HardwareNotFound(HardwareError):
+    """Hardware path or device don't exists on Host."""
```
agners, line 304:
```suggestion
    """Hardware path or device doesn't exist on the Host."""
```

supervisor/hardware/data.py
```diff
@@ -0,0 +1,36 @@
...
+        """Return Major cgroups."""
+        return int(self.attributes.get("MAJOR", 0))
+
+    @property
+    def cgroups_minor(self) -> int:
+        """Return Major cgroups."""
+        return int(self.attributes.get("MINOR", 0))
+
+    @property
+    def by_id(self) -> Optional[Path]:
```
agners, line 30: There are other symlinks than `by-id` (e.g. for partitions/disks knows by-uuid etc...). Not sure if we should consider doing this more generic, e.g. just `device_symlinks` and than make it a list...? But I guess this is internals we can extend rather easily in a second step if necessary.

supervisor/addons/model.py
```diff
@@ -532,7 +533,7 @@ def schema_ui(self) -> Optional[List[Dict[str, Any]]]:
 
         if isinstance(raw_schema, bool):
             return None
-        return schema_ui_options(raw_schema)
+        return UiOptions(self.coresys)(raw_schema)
```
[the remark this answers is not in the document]
pvizeli: That is UIOptions, which work different

supervisor/docker/addon.py
pvizeli: Done by cgroup rules. We need only to know if we can apply it. But yeah, this function would be gone after I know the sys endpoint and can do it like for audio.

supervisor/hardware/data.py
pvizeli: We have this list above. But 99% of the links can't be used by docker because of special chars. Only this link and that is what user search

pvizeli: It's is only a helper for UI

ludeeus (commented):
About the deprecation warnings, I think we should only print those for add-ons that are installed.

supervisor/hardware/helper.py
```diff
@@ -0,0 +1,78 @@
...
+        """Return True if the system have audio support."""
+        return len(self.sys_hardware.filter_devices(subsystem=UdevSubsystem.AUDIO))
+
+    @property
+    def support_gpio(self) -> bool:
+        """Return True if device support GPIOs."""
+        return _SOC_DEVICES.exists() and _GPIO_DEVICES.exists()
+
+    @property
+    def last_boot(self) -> Optional[str]:
```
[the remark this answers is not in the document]
pvizeli: there is no block, procfs is not a blocking IO device

ludeeus (approved):

PR #2449, opened 2021-01-18
Add argument to reboot in fire-and-forget mode

agners:
If a reboot is triggered through the HA frontend, we cannot really
send back success information as the HA server is being shutdown during
that processes. In fact, the TCP connection delays shutdown of the core.
Allow to trigger the reboot without waiting for a result.

This allows to prevent an open TCP connection between the frontend and
the HA server. Currently the aiohttp server disconnects after 10s in any
case (and we could shorten that time as well), but rather than shorten
the delay further, just not wait for the operation seems the porper fix.

See also: https://github.com/home-assistant/core/pull/45033

## Type of change

- [ ] Dependency upgrade

## Additional information

## Checklist

- [x] Documentation added/updated for [developers.home-assistant.io][docs-repository]: https://github.com/home-assistant/developers.home-assistant/pull/775


agners:
First I was thinking to *not* wait by default. But when running from CLI it does make sense to report back as we would see if D-Bus errors or similar would happen.

pvizeli:
I think that needs to be on the frontend which causes the issue because it's not only the host reboot even.

agners:
Maybe I understand that wrong, but from what I understand the frontend does a POST request via API to the Supervisor, and the Supervisor. The supervisor starts to do operations and _does not_ complete the POST until HA is shutdown. And HA does not shutdown because the frontend is still waiting for that POST to complete.

Afaik, we can't make a "POST but don't wait for answer" on the frontend...

agners:
Hm, I see, the hassio component in HA core can actually create a post request to the supervisor as a task and not wait for it just like I did here. Since the HA core is what the frontend invokes, and the problem is HA core's aiohttp server doesn't like that the connection stays open, I agree we should fix this in HA core instead here. Closing.

PR #2686, opened 2021-03-05
Set connectivity state on network tasks

ludeeus:
## Proposed change

We have several tasks that use network, we should use the result of those tasks to set the connectivity state.
Changed both connectivity attributes to be setters so we can update the frontend with this information if anything changes.

## Type of change

- [ ] Dependency upgrade

## Additional information

## Checklist

### DIFF
--- a/supervisor/resolution/checks/addon_pwned.py
+++ b/supervisor/resolution/checks/addon_pwned.py
@@ -35,6 +35,7 @@ class CheckAddonPwned(CheckBase):
                     if not await check_pwned_password(self.sys_websession, secret):
                         continue
                 except PwnedError:
+                    self.sys_supervisor.connectivity = False
                     continue
 
                 # Check possible suggestion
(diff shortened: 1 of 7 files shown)


supervisor/resolution/checks/addon_pwned.py
```diff
@@ -35,6 +35,7 @@ async def run_check(self) -> None:
                     if not await check_pwned_password(self.sys_websession, secret):
                         continue
                 except PwnedError:
+                    self.sys_supervisor.connectivity = False
```
pvizeli, line 38: I think all looks fine. But this one, I'm not sure because if there is an error like 400 we set it disabled, it could just be that the service is down or have another issue. We should only set this (like we do with other parts of this PR) with our services fetch. So that only our stuff (which we control) can take the functionality of the system down. If our stuff is not working, then it's fine to freeze services because other will also not work.

pvizeli (approved):

PR #2680, opened 2021-03-04
New error handling

pvizeli:
## Proposed change

image

## Type of change

- [ ] Dependency upgrade
- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (which adds functionality to the supervisor)
- [ ] Breaking change (fix/feature causing existing functionality to break)
- [x] Code quality improvements to existing code or addition of tests

## Additional information

- This PR fixes or closes issue: fixes #
- This PR is related to issue:
- Link to documentation pull request:

## Checklist

- [ ] The code change is tested and works locally.
- [ ] Local tests pass. **Your PR cannot be merged unless tests pass**
- [ ] There is no commented out code in this PR.
- [ ] I have followed the [development checklist][dev-checklist]
- [ ] The code has been formatted using Black (`black --fast supervisor tests`)
- [ ] Tests have been added to verify that the new code works.

If API endpoints of add-on configuration are added/changed:

- [ ] Documentation added/updated for [developers.home-assistant.io][docs-repository]

[dev-checklist]: https://developers.home-assistant.io/docs/en/development_checklist.html
[docs-repository]: https://github.com/home-assistant/developers.home-assistant

PR #2703, opened 2021-03-08
Check management

ludeeus:
## Proposed change

Adds API's for check management.
Needs #2680

## Type of change

- [ ] Dependency upgrade

## Additional information

## Checklist


supervisor/api/__init__.py
```diff
@@ -228,6 +228,12 @@ def _register_resolution(self) -> None:
         self.webapp.add_routes(
             [
                 web.get("/resolution/info", api_resolution.info),
+                web.post(
+                    "/resolution/check/{check}/disable", api_resolution.disable_check
```
balloob, line 232: Do we expect other options to become available and so would a generic `/options` endpoint be better? How do we do it for other APIs in the Supervisor (addons etc)

ludeeus: Checks are "stupid", they run or they don't, if we want options that are not pr. check but on resolution or resolution/check

supervisor/api/resolution.py
```diff
@@ -62,3 +76,29 @@ async def dismiss_issue(self, request: web.Request) -> None:
     def healthcheck(self, request: web.Request) -> Awaitable[None]:
         """Run backend healthcheck."""
         return asyncio.shield(self.sys_resolution.healthcheck())
+
+    @api_process
+    async def enable_check(self, request: web.Request) -> None:
+        """Enable check."""
+        if request[REQUEST_FROM] != self.sys_homeassistant:
```
balloob, line 83: Can we make this a decorator? Seems like it will be used more often. `@require_user`

supervisor/api/resolution.py
```diff
@@ -62,3 +78,17 @@ async def dismiss_issue(self, request: web.Request) -> None:
     def healthcheck(self, request: web.Request) -> Awaitable[None]:
         """Run backend healthcheck."""
         return asyncio.shield(self.sys_resolution.healthcheck())
+
+    @api_process
+    @require_home_assistant
+    async def check_options(self, request: web.Request) -> None:
```
balloob, line 84: Since the word `check` is ambiguous, could we rename this to `set_check_options` ?

supervisor/api/resolution.py
```diff
@@ -62,3 +78,17 @@ async def dismiss_issue(self, request: web.Request) -> None:
...
+        """Set check options."""
+        body = await api_validate(SCHEMA_CHECK_OPTIONS, request)
```
balloob, line 86: Should we make this a decorator? 😄 

https://github.com/home-assistant/core/blob/dev/homeassistant/components/http/data_validator.py#L16

supervisor/api/resolution.py
ludeeus: sure, but since @pvizeli marked it as draft I will not be doing anything until he is done commenting

supervisor/api/resolution.py
ludeeus: not in this PR

supervisor/api/resolution.py
```diff
@@ -4,11 +4,22 @@
...
+    ATTR_UNHEALTHY,
+    ATTR_UNSUPPORTED,
+)
 from ..coresys import CoreSysAttributes
-from ..exceptions import APIError, ResolutionNotFound
-from .utils import api_process
+from ..exceptions import APIError, ResolutionError, ResolutionNotFound
+from .utils import api_process, api_validate, require_home_assistant
+
+SCHEMA_CHECK_OPTIONS = vol.Schema({vol.Required(ATTR_ENABLED): bool})
```
balloob, line 22: Should we set it to `optional` given that more options might be added.

supervisor/api/resolution.py
balloob: btw any reason you're keeping the schema defined on the top of the file? Doesn't that make it harder to follow what's going on because now the schema and the logic are on the other side of the page?

supervisor/api/resolution.py
```diff
@@ -62,3 +78,17 @@ async def dismiss_issue(self, request: web.Request) -> None:
...
+    @api_process
+    @require_home_assistant
+    async def check_options(self, request: web.Request) -> None:
+        """Set check options."""
+        body = await api_validate(SCHEMA_CHECK_OPTIONS, request)
+
+        try:
+            if body[ATTR_ENABLED]:
+                self.sys_resolution.check.enable(request.match_info.get("check"))
+            else:
```
balloob, line 91: When `enabled` becomes `optional`, you should check here that it's not `None`, or else not providing `enabled` will disable it.

supervisor/api/resolution.py
ludeeus: That's the default :shrug:

balloob: Challenge everything.

supervisor/resolution/checks/base.py
```diff
@@ -41,11 +43,16 @@ async def __call__(self) -> None:
...
-        if affected and not self.multiple:
+        if affected and self.context not in (ContextType.ADDON, ContextType.PLUGIN):
             return
         _LOGGER.info("Run check for %s/%s", self.issue, self.context)
         await self.run_check()
 
+    @property
+    def slug(self) -> str:
+        """Return the check slug."""
+        return Path(sys.modules[self.__class__.__module__].__file__).stem
```
balloob, line 54: Wouldn't this be the same:

```suggestion
        return self.__class__.__module__.split('.')[-1]
```

pvizeli: We need the file name, not the module folder

supervisor/api/resolution.py
pvizeli: Yeah, there like the API handling and registration process need a rework

supervisor/resolution/checks/base.py
balloob: `stem` returns the filename without extension. That is the same as the last part of the module.

image

balloob (approved):

supervisor/resolution/checks/base.py
pvizeli: 🤷 let see, we have tests for that

PR #2752, opened 2021-03-25
Allow optional array configs

mdegat01:
## Proposed change

During a [review of a change](https://github.com/home-assistant/addons/pull/1941#pullrequestreview-620834679) to the MariaDB addon, it was raised that it should be possible to make an array config option in an add-on truly optional. Currently if you define the schema of a config as `["str?"]` then the option itself must be present to pass validation, even if its just `[]`. This makes it so if the element inside is optional then the entire config option is optional.

### Example

Given an addon with this schema:

```json
{
    "schema": {
        "log_level": "str",
        "privileges": ["list(x|y|z)?"]
    }
}
```

This configuration was invalid prior to this PR. Now it is valid:

```yaml
log_level: info
```

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR is related to issue: https://github.com/home-assistant/addons/pull/1941#pullrequestreview-620834679
- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/addons/options.py
+++ b/supervisor/addons/options.py
@@ -241,8 +241,15 @@ class AddonOptions(CoreSysAttributes):
         """Check if all options are exists."""
         missing = set(origin) - set(exists)
         for miss_opt in missing:
-            if isinstance(origin[miss_opt], str) and origin[miss_opt].endswith("?"):
+            miss_schema = origin[miss_opt]
+
+            # If its a list then value in list decides if its optional like ["str?"]
+            if isinstance(miss_schema, list) and len(miss_schema) > 0:
+                miss_schema = miss_schema[0]
+
+            if isinstance(miss_schema, str) and miss_schema.endswith("?"):
                 continue
+
             raise vol.Invalid(
                 f"Missing option '{miss_opt}' in {root} in {self._name} ({self._slug})"
             ) from None
--- a/tests/addons/test_options.py
+++ b/tests/addons/test_options.py
@@ -70,6 +70,38 @@ def test_complex_schema_list(coresys):
         )({"name": "Pascal", "password": "1234", "extend": "test"})
 
 
+def test_optional_schema_list(coresys):
+    """Test with an optional list schema."""
+    assert AddonOptions(
+        coresys,
+        {"name": "str", "password": "password", "extend": ["str?"]},
+        MOCK_ADDON_NAME,
+        MOCK_ADDON_SLUG,
+    )({"name": "Pascal", "password": "1234"})
+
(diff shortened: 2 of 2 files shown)


pvizeli (approved):
Thanks

supervisor/addons/options.py
```diff
@@ -241,8 +241,15 @@ def _check_missing_options(
         """Check if all options are exists."""
         missing = set(origin) - set(exists)
         for miss_opt in missing:
-            if isinstance(origin[miss_opt], str) and origin[miss_opt].endswith("?"):
+            miss_schema = origin[miss_opt]
+
+            # If its a list then value in list decides if its optional like ["str?"]
+            if isinstance(miss_schema, list) and len(miss_schema) > 0:
+                miss_schema = miss_schema[0]
```
frenck, line 248: We should instead check each item in the list if they are optional or not.

Now it doesn't work with lists containing dicts with items.

mdegat01: So like if its a list of objects and each key in the object is optional then make the entire object optional? That makes sense although objects might also have other object/list fields so its a bit more complicated. You'd kind of have to recursively look through each object field you found to answer the question of whether its optional or not.

I'll think about it. I feel like it should be a separate PR regardless right? It's a separate gap and I don't think its trivial because of the nested fields problem.

mdegat01: Also just to clarify, the previous code here was this:

```python
 def _check_missing_options(
    self, origin: Dict[Any, Any], exists: Dict[Any, Any], root: str
) -> None:
    """Check if all options are exists."""
    missing = set(origin) - set(exists)
    for miss_opt in missing:
        miss_schema = origin[miss_opt]

        if isinstance(miss_schema, str) and miss_schema.endswith("?"):
            continue

        raise vol.Invalid(
            f"Missing option '{miss_opt}' in {root} in {self._name} ({self._slug})"
        ) from None
```

As you can see, it only had handling for fields which were of type string in the schema. If the schema (`origin`) said a field wasn't type string then it simply assumed it was required. I added the capability to have a config which is an array of strings be optional but there's no change to how configs of type `dict` or `list(dict)` are handled, they will always be considered required.

Note that this doesn't mean we are no longer validating dictionaries in arrays if that's your concern. This method is only responsible for looking at the provided dictionary and ensuring all required fields have a value. Other methods go through each dictionary in each list and field and make sure they meet validation requirements.

frenck: My concern was this fixes one of the possible problems. Would have been nice if we could do the same for dicts in the list.

PR #2909, opened 2021-05-20
Update OS on intel-nuc to generic-x86-64 image

agners:
## Proposed change
Since OS renamed from intel-nuc to generic-x86-64, the new image must be fetched by the supervisor. This change makes sure that when updating on a board named `intel-nuc` to OS releases with major version 6 or higher, the board name `generic-x86-64` is used instead.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR is related to issue: https://github.com/home-assistant/operating-system/issues/1048
- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/hassos.py
+++ b/supervisor/hassos.py
@@ -55,15 +55,23 @@ class HassOS(CoreSysAttributes):
         """Return board name."""
         return self._board
 
-    async def _download_raucb(self, version: AwesomeVersion) -> Path:
-        """Download rauc bundle (OTA) from github."""
+    def _get_download_url(self, version: AwesomeVersion) -> str:
         raw_url = self.sys_updater.ota_url
         if raw_url is None:
             raise HassOSUpdateError("Don't have an URL for OTA updates!", _LOGGER.error)
-        url = raw_url.format(version=str(version), board=self.board)
 
+        update_board = self.board
+
+        # OS version 6 and later renamed intel-nuc to generic-x86-64...
+        if update_board == "intel-nuc" and version.section(0) >= 6:
+            update_board = "generic-x86-64"
+
+        url = raw_url.format(version=str(version), board=update_board)
+        return url
+
+    async def _download_raucb(self, url: str, raucb: Path) -> None:
+        """Download rauc bundle (OTA) from URL."""
         _LOGGER.info("Fetch OTA update from %s", url)
-        raucb = Path(self.sys_config.path_tmp, f"hassos-{version!s}.raucb")
         try:
             timeout = aiohttp.ClientTimeout(total=60 * 60, connect=180)
             async with self.sys_websession.get(url, timeout=timeout) as request:
@@ -82,7 +90,6 @@ class HassOS(CoreSysAttributes):
                         ota_file.write(chunk)
 
             _LOGGER.info("Completed download of OTA update file %s", raucb)
-            return raucb
 
         except (aiohttp.ClientError, asyncio.TimeoutError) as err:
             self.sys_supervisor.connectivity = False
@@ -154,7 +161,9 @@ class HassOS(CoreSysAttributes):
             )
 
         # Fetch files from internet
-        int_ota = await self._download_raucb(version)
+        ota_url = self._get_download_url(version)
+        int_ota = Path(self.sys_config.path_tmp, f"hassos-{version!s}.raucb")
+        await self._download_raucb(ota_url, int_ota)
         ext_ota = Path(self.sys_config.path_extern_tmp, int_ota.name)
 
         try:
--- a/supervisor/hassos.py
+++ b/supervisor/hassos.py
@@ -63,7 +63,7 @@ class HassOS(CoreSysAttributes):
         update_board = self.board
 
         # OS version 6 and later renamed intel-nuc to generic-x86-64...
-        if update_board == "intel-nuc" and version.section(0) >= 6:
+        if update_board == "intel-nuc" and version >= 6:
             update_board = "generic-x86-64"
 
         url = raw_url.format(version=str(version), board=update_board)
--- a/supervisor/hassos.py
+++ b/supervisor/hassos.py
@@ -63,7 +63,7 @@ class HassOS(CoreSysAttributes):
         update_board = self.board
 
         # OS version 6 and later renamed intel-nuc to generic-x86-64...
-        if update_board == "intel-nuc" and version >= 6:
+        if update_board == "intel-nuc" and version >= 6.0:
             update_board = "generic-x86-64"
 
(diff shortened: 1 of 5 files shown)


supervisor/hassos.py
```diff
@@ -55,15 +55,23 @@ def board(self) -> Optional[str]:
...
+    def _get_download_url(self, version: AwesomeVersion) -> str:
         raw_url = self.sys_updater.ota_url
         if raw_url is None:
             raise HassOSUpdateError("Don't have an URL for OTA updates!", _LOGGER.error)
-        url = raw_url.format(version=str(version), board=self.board)
 
+        update_board = self.board
+
+        # OS version 6 and later renamed intel-nuc to generic-x86-64...
+        if update_board == "intel-nuc" and version.section(0) >= 6:
```
ludeeus, line 66:
```suggestion
        if update_board == "intel-nuc" and version >= 6:
```

ludeeus (commented):
Is this the best place for it?
I was thinking it would make sense to do this logic under <https://github.com/home-assistant/supervisor/blob/main/supervisor/hassos.py#L117>

Then it will populate correctly everywhere, not just for this place.

supervisor/hassos.py
agners: That seems not to work with the current rc releases:
```
>>> test = awesomeversion.AwesomeVersion("6.0.rc1")
>>> test > 6
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.9/site-packages/awesomeversion/awesomeversion.py", line 117, in __gt__
    raise AwesomeVersionCompare(
awesomeversion.exceptions.AwesomeVersionCompare: Can't compare unknown
```

ludeeus: I fixed that a few weeks ago, make sure your installed requirements matches the requirements file

agners:
> Is this the best place for it?
> I was thinking it would make sense to do this logic under https://github.com/home-assistant/supervisor/blob/main/supervisor/hassos.py#L117
> 
> Then it will populate correctly everywhere, not just for this place.

In my thinking, the board is at that time still a `intel-nuc` board. Its running on a 5.x OS, and if you want to up/downgrade in the 5.x series of releases, it should stay a `intel-nuc`. It only should pick the update file for `generic-x86-64` to become a `generic-x86-64` board. Note that its the target version we are updating to which is used to decide if the transformation happens.

What you are suggesting would "convert" all existing installations to `generic-x86-64` boards, at least from a Supervisor perspective. For OS update that seems not ideal to me since we don't know to what version we are updating to. For other things it might be beneficial as we can get rid of `intel-nuc` earlier (probably).

ludeeus:
Right! we actually only need this for the updates, once updated it will be generic-x86-64

supervisor/hassos.py
ludeeus: This is with `21.5.0` which is the version pinned for the supervisor
```
>>> test = AwesomeVersion("6.0.rc1")
>>> test > 6
True
>>> test >= 6
True
```

tests/test_hassos.py
```diff
@@ -0,0 +1,27 @@
+"""Test Home Assistant OS functionality."""
+
+from awesomeversion import AwesomeVersion
+import pytest
+
+from supervisor.coresys import CoreSys
+
+URL_TEST = "https://version.home-assistant.io/stable.json"
```
ludeeus, line 8: This is not refrenced

tests/test_hassos.py
```diff
@@ -0,0 +1,27 @@
...
+# pylint: disable=protected-access
+
+
+@pytest.mark.asyncio
+async def test_ota_url_generic_x86_64_rename(coresys: CoreSys) -> None:
+    """Test download URL generated."""
+
+    coresys.hassos._board = "intel-nuc"
+    coresys.hassos._version = AwesomeVersion("5.13")
+    await coresys.updater.fetch_data()
```
ludeeus, line 19: I know the updater tests also do this, but we should patch I/O in tests

ludeeus (changes requested):

tests/test_hassos.py
agners: Agreed. Let's fix it once updater is fixed.

ludeeus (approved):

tests/test_hassos.py
```diff
@@ -0,0 +1,25 @@
...
+
+    version6str = "6.0"
+    version6 = AwesomeVersion(version6str)
+    url = coresys.updater.ota_url.format(
+        version=str(version6str), board="generic-x86-64"
```
pvizeli, line 22:
```suggestion
    version6 = AwesomeVersion("6.0")
    url = coresys.updater.ota_url.format(
        version=str(version6), board="generic-x86-64"
```

tests/test_hassos.py
```diff
@@ -0,0 +1,22 @@
...
+@pytest.mark.asyncio
+async def test_ota_url_generic_x86_64_rename(coresys: CoreSys) -> None:
+    """Test download URL generated."""
+
+    coresys.hassos._board = "intel-nuc"
+    coresys.hassos._version = AwesomeVersion("5.13")
+    await coresys.updater.fetch_data()
+
+    version6 = AwesomeVersion("6.0")
+    url = coresys.updater.ota_url.format(version=str(version6), board="generic-x86-64")
```
ludeeus, line 20: There is no need to wrap it in a str here
```
>>> OTA_URL = "https://github.com/home-assistant/operating-system/releases/download/{version}/hassos_{board}-{version}.raucb"
>>> version6 = AwesomeVersion("6.0")
>>> OTA_URL.format(version=version6, board="generic-x86-64")
'https://github.com/home-assistant/operating-system/releases/download/6.0/hassos_generic-x86-64-6.0.raucb'
```
```suggestion
    url = coresys.updater.ota_url.format(version=version6, board="generic-x86-64")
```

agners: This comes from hassos.py: https://github.com/home-assistant/supervisor/blob/main/supervisor/hassos.py#L69

Does that work in every and all cases? E.g. how does Python knows that format requires a string there? Could it be that if AwesomeVersion implements `__int__`, this is picked over `__str__`?

agners: By experimentation it seems that `__str__` really is the preferred method, over `__int__` and `__repr__`.

ludeeus: It does not work everywhere, it works here because of <https://github.com/ludeeus/awesomeversion/blob/main/awesomeversion/awesomeversion.py#L83-L84>, if it had `__init__` (which it can't), `__str__` would still be used

PR #2940, opened 2021-06-08
Rename snapshot -> backup

ludeeus:
## Proposed change

Rename "snapshot" to "backup" everywhere

Related:
<https://github.com/home-assistant/core/pull/51629>
<https://github.com/home-assistant/frontend/pull/9393>
<https://github.com/home-assistant/my.home-assistant.io/pull/95>
<https://github.com/home-assistant/home-assistant.io/pull/18129>
<https://github.com/home-assistant/developers.home-assistant/pull/963>
<https://github.com/home-assistant/cli/pull/301>
<https://github.com/home-assistant/addons/pull/2088>

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist


ludeeus:
Need update log messages from https://github.com/home-assistant/supervisor/pull/2941

pvizeli (commented):
Looks fine, just a small comment, let's handle it like the os/hassos or homeassistant/core renaming. We don't need to have the def _register_snapshots(self) around with the fake class.

ludeeus:
It's not fake?

pvizeli:
Can we not do it like: https://github.com/home-assistant/supervisor/blob/main/supervisor/api/__init__.py#L315 ?

ludeeus:
we can

supervisor/addons/validate.py
```diff
@@ -282,8 +283,9 @@ def _migrate(config: Dict[str, Any]):
         vol.Optional(ATTR_SERVICES): [vol.Match(RE_SERVICE)],
         vol.Optional(ATTR_DISCOVERY): [valid_discovery_service],
         vol.Optional(ATTR_SNAPSHOT_EXCLUDE): [str],
-        vol.Optional(ATTR_SNAPSHOT_PRE): str,
-        vol.Optional(ATTR_SNAPSHOT_POST): str,
+        vol.Optional(ATTR_BACKUP_EXCLUDE): [str],
```
pvizeli, line 286: Need to be done with the migration function and need test

pvizeli (changes requested):

supervisor/addons/validate.py
ludeeus: Not doing any more with renaming until _everyone_ is on the same page on what we should/shouldn't rename

pvizeli:
If we follow this, we need to replace everything from snapshot to backup. I will not accept a new building-place, we have too many of that started but never finished stuff now.

ludeeus:
We don't _need_ to replace everything, if we use "snapshot" internally, no one will ever know.
This is about the external wording.
There is no reason to keep this open if it will never be accepted, so feel free to close this and all related PRs :+1:

supervisor/addons/validate.py
```diff
@@ -224,6 +227,23 @@ def _migrate(config: Dict[str, Any]):
                 )
             config[ATTR_TMPFS] = True
 
+        # 2021-06 "snapshot" renamed to "backup"
+        for entry in (
+            ATTR_SNAPSHOT_EXCLUDE,
+            ATTR_SNAPSHOT_POST,
+            ATTR_SNAPSHOT_PRE,
+            ATTR_SNAPSHOT,
```
pvizeli, line 235:
```suggestion
            "snapshot_exclude",
            "snapshot_post",
            "snapshot_pre",
            "snapshot",
```

No need to keep it as an attribute if we want to ditch it. Because this code stay maybe for the next 5 years around

supervisor/api/backups.py
```diff
@@ -0,0 +1,220 @@
...
+                        ATTR_HOMEASSISTANT: backup.homeassistant_version is not None,
+                        ATTR_ADDONS: backup.addon_list,
+                        ATTR_FOLDERS: backup.folders,
+                    },
+                }
+            )
+
+        if request.path == "/snapshots":
+            # Kept for backwards compability
+            return {ATTR_SNAPSHOTS: data_backups}
```
pvizeli, line 101: Let's ditch that and we make a blog post and give software developers 2 weeks time to update their add-on. As we use the output on CLI, that confuse a lot to have all things twice

supervisor/backups/validate.py
```diff
@@ -24,14 +24,14 @@
     ATTR_VERSION,
     ATTR_WAIT_BOOT,
     ATTR_WATCHDOG,
+    BACKUP_FULL,
+    BACKUP_PARTIAL,
```
pvizeli, line 28: use that new style as an enum on `backup/const.py`

supervisor/backups/manager.py
```diff
@@ -0,0 +1,399 @@
+"""Backup manager."""
```
pvizeli, line 1: We started to call this `module.py`. However, I'm fine to call this `manager.py` or `core.py`, we should just rename the other exists one. Maybe `core.py` would be a better fit? (I never liked `module.py`, but as you bring that on topic, we can discuss it)

supervisor/addons/addon.py
```diff
@@ -679,23 +679,23 @@ async def write_stdin(self, data) -> None:
...
         except DockerError as err:
             _LOGGER.error(
-                "Failed running pre-/post-snapshot command %s: %s", command, err
+                "Failed running pre-/post backup command %s: %s", command, err
             )
             raise AddonsError() from err
 
-    async def snapshot(self, tar_file: tarfile.TarFile) -> None:
-        """Snapshot state of an add-on."""
+    async def backup(self, tar_file: tarfile.TarFile) -> None:
```
pvizeli, line 697: I think that could stay as `snapshot` to match with `restore` in a backup process?

supervisor/backups/manager.py
ludeeus: Yeah, was thinking about using "module" first, but the class was `BackupManager`  so `manager` felt more correct here, I don't really have strong feelings about it

supervisor/api/backups.py
ludeeus: I don't think 2 weeks is enough, the UI can't use the /backups endpoint until the user is on a core version after https://github.com/home-assistant/core/pull/51629 is released

ludeeus: nevermind, that only applies to uploads, but I still think 2 weeks are too little

supervisor/backups/manager.py
pvizeli: Let's rename it to `core.py`. I did the same now on os agent. I can later rename the others too

ludeeus: "core" might be confusing with core which we use for homeassistant
"manager", "base", "module", "main", "element", "entry" are all better than "core" for this purpose IMO

supervisor/api/backups.py
pvizeli: Right, but we rename also the meta-file inside the backup. So one "big" breaking change. We can also wait 3 weeks after the blog post or 4 weeks, whatever you think makes sense

ludeeus: We can never break loading `snapshot.json` (the metafile), doing that will render all previous backups useless

balloob (approved):
This looks ✅ 

I think that 2 weeks transition time is too little. Let's do at least 2 months and then we can drop the warning code. If it's not too much code, I would even consider just keeping it around longer.

pvizeli (approved):

PR #2954, opened 2021-06-11
Cleanup security layer - Add-on default access role

pvizeli:
## Proposed change

- Cleanup old API call
- Allow every add-on to get network info (on all profiles)

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist


frenck (approved):
Nice 👍

pvizeli:
Hmm, the regex before should also catch that. Why does that not work😐 maybe looking at wrong rules

ludeeus:
This was supposed to fix `[supervisor.api.middleware_security] Invalid token for access /network/interface/default/info` but I don't think that is from an add-on

If it was an addon and the path was not allowed we would see `"%s no role for %s", request.path, addon.slug`
I'm more thinking this is from the banner on the OS

ludeeus:
For the samba add-on I've make a PR to fix that <https://github.com/home-assistant/addons/pull/2070>

pvizeli:
I think that is a fine rework. A new review is needed I guess

supervisor/api/hardware.py
```diff
@@ -58,8 +58,3 @@ async def audio(self, request: web.Request) -> Dict[str, Any]:
                 },
             }
         }
-
-    @api_process
-    async def trigger(self, request: web.Request) -> Awaitable[None]:
-        """Trigger a udev device reload."""
-        _LOGGER.debug("Ignoring DEPRECATED hardware trigger function call.")
```
ludeeus, line 65: Instead of removing this.
Should we change back the level to the warning, and include the "request_from" ?

supervisor/api/middleware/security.py
```diff
@@ -65,22 +64,24 @@
     ROLE_DEFAULT: re.compile(
         r"^(?:"
         r"|/.+/info"
-        r"|/addons"
```
ludeeus, line 68: This would be breaking for any addon that uses this path with the default role?

pvizeli: With `free` info call, that is not in use a longer time. Right @frenck ? (I did that for him before the self slug was burn)

supervisor/api/hardware.py
pvizeli: This feature was never real in the wild, It was just used for deconz and worked too bad for going forward that, we replaced it with a new hardware layer which also forced the user to update deconz (because of a mistake) and on that update, we removed that call.

supervisor/api/middleware/security.py
pvizeli: oh, right they is because /addons have no info call

supervisor/api/hardware.py
ludeeus: Alright 👍

supervisor/api/middleware/security.py
frenck: Hmmm... maybe we should add `/addons/self/info` to the ADDON_API_BYPASS?

pvizeli: Self is all ready bypass (all calls)

frenck: Oh right, good morning ☕️☕️

frenck (approved):

PR #2987, opened 2021-06-30
If any ingress options are set, enable ingress

ludeeus:
## Proposed change

None of these are used if ingress is not enabled, so we should just set ingress to true if any of these are defined.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/addons/validate.py
+++ b/supervisor/addons/validate.py
@@ -225,6 +225,14 @@ def _migrate_addon_config(protocol=False):
                 )
             config[ATTR_TMPFS] = True
 
+        # If any ingress options are set, enable ingress
+        if (
+            config.get(ATTR_INGRESS_ENTRY) is not None
+            or config.get(ATTR_INGRESS_PORT) is not None
+            or config.get(ATTR_INGRESS_STREAM)
+        ):
+            config[ATTR_INGRESS] = True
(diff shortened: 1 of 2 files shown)


supervisor/addons/validate.py
```diff
@@ -225,6 +225,14 @@ def _migrate(config: Dict[str, Any]):
...
             config[ATTR_TMPFS] = True
 
+        # If any ingress options are set, enable ingress
+        if (
+            config.get(ATTR_INGRESS_ENTRY) is not None
+            or config.get(ATTR_INGRESS_PORT) is not None
+            or config.get(ATTR_INGRESS_STREAM)
+        ):
+            config[ATTR_INGRESS] = True
+
```
pvizeli, line 235: We should not actively change user add-on config. We can make a warning in the function above, that some things are not correctly configured. I saw many add-ons which tried somethings to archive and never cleans it up which now just fails

ludeeus: I can add a warning too, but I don't agree that we should not set this.
It does not make sense that I have to add ìngress: true`  in the configuration if I already have `ingress_stream: true`  which only are used for ingress.

PR #3291, opened 2021-11-05
Add enhanced logging REST endpoints using systemd-journal-gatewayd

agners:
## Proposed change

Add `/host/logs/entries`, `/host/logs/{identifier}/entries`, `/host/logs/boot/{bootid}/entries`
and `/host/logs/boot/{bootid}/{identifier}/entries` to expose log entries
from systemd-journald running on the host. Use systemd-journal-gatewayd
which exposes the logs to the Supervisor via Unix socket.

For all endpoints can add `/follow` to the end. This will keep the HTTP
request open and send new log entries as they get added to the systemd-journal.
Assuming they match the bootid and/or identifier (if provided).

Forward the `Range` header to systemd-journal-gatewayd. This allows to
select only a certain amount of log data. The `Range` header is a standard
header to select only partial amount of data. However, the "entries="
prefix is custom for systemd-journal-gatewayd, denoting that the numbers
following represent log entries (as opposed to bytes or other metrics).

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to documentation pull request: https://github.com/home-assistant/developers.home-assistant/pull/1127
- Link to cli pull request: https://github.com/home-assistant/cli/pull/364

## Checklist


pvizeli:
1) Please write a generic RFC on how we can handle logs API call for signaling to use streams or just return a range. 
2) Params are not allowed, use path for this (append /boot or /stream at the end i.e.) however, stream is part of the RFC.
3) /host/logs/entries and /host/logs/{identifier}/entries is not how we handle Paths. use `/host/logs/all/entries` for all

pvizeli:
Maybe it would make sense to offload it to `/journal/` I guess

agners:
Reading up a bit about RESTful API design it seems that multiple URI's for the same resource as well as GET query parameters for filtering/paging are "allowed" and commonly used...  Now the question is: is `boot` or `follow` a "filter" or merely a different URI for the same resource.

I'd say `follow` is more a filter type of thing, and `boot` warrants a separate URI. So I think this might be a good RESTful API design:

```
/host/logs/all/entries
/host/logs/{identifier}/entries
/host/logs/boot/list
/host/logs/boot/{bootid}/entries
```

All of them would support the `follow` query string parameter.

mdegat01:
The query to get boot IDs used in this PR is equivalent to running `journalctl -o json MESSAGE_ID=b07a249cd024414a82dd00cd181378ff | jq '._BOOT_ID'` on the host. A 100% accurate list of boot IDs for a given host can be found by doing `journalcl --list-boots`. We are not using this for two reasons:
1) It was found to be very expensive memory and CPU wise, particularly on devices like rpi3s
2) `systemd-journal-gatewayd` does not support it so we would need some other mechanism for running it on the host and getting the results

The query being used does not have these issues but isn't 100% accurate. In my testing it missed some older boots on some systems. It also would obviously miss any boot that ended before systemd finished starting up since that is when messages with that ID are logged.

I believe it is good enough for our use here. That being said, I would appreciate more evidence to back that up if anyone wants to run those two queries on a host and share the compare their results.

pvizeli (approved):

tests/api/test_host.py
```diff
@@ -1,9 +1,28 @@
...
+    "hassos-config",
+    "hassos-expand",
+    "hassos-overlay",
+    "hassos-persists",
+    "hassos-zram",
+    "kernel",
+    "os-agent",
+    "rauc",
+    "systemd",
+]
```
agners, line 24: I'd add a couple of additional identifiers here, specifically:

```
bluetoothd
bthelper
btuart
dbus-broker
dockerd
fstrim
systemd-coredump
systemd-fsck
systemd-growfs
systemd-hostnamed
systemd-journal-gatewayd
systemd-journald
systemd-logind
systemd-random-seed
systemd-resolved
systemd-time-wait-sync
systemd-timesyncd
systemd-tmpfiles
systemd-udevd
wpa_supplicant
```

supervisor/api/__init__.py
```diff
@@ -103,16 +104,36 @@ def _register_host(self) -> None:
         self.webapp.add_routes(
             [
                 web.get("/host/info", api_host.info),
-                web.get("/host/logs", api_host.logs),
+                web.get("/host/logs", api_host.advanced_logs),
```
agners, line 107: So with that we make the advanced logging (systemd-journal-gatewayd logging) already mandatory today. I think I'd have preferred if we would have gone a two step approach, essentially reverting to the old approach if systemd-journal-gatewayd is not there (especially since the Supervised installation only recently [got that change in](https://github.com/home-assistant/supervised-installer/pull/235/files). OS exposes `systemd-journal-gatewayd` since 7.0 so from a ADR0015 point of view we are well past that. But we still have almost 3k people on pre-7.0 OS versions :cold_sweat: ).

That said, I am ok with moving fully to it, we just should make sure the system behaves somewhat sensible in case it has the `systemd-journal-gatewayd` socket missing. Especially, how will the user realize what is wrong?

From what I can tell the API will throw `HostNotSupportedError` in that case. Is the frontend behaving somewhat sensible in this situation?

supervisor/host/logs.py
```diff
@@ -0,0 +1,138 @@
...
+        if self.boot_ids:
+            # Doesn't change without a reboot, no reason to query again once cached
+            return
+
+        try:
+            async with self.journald_logs(
+                params=BOOT_IDS_QUERY,
+                accept=LogFormat.JSON,
+                timeout=ClientTimeout(total=20),
+            ) as resp:
```
agners, line 95: I am still slightly worried what this might do in practice. From tests on Raspberry Pi 3/4 it takes up to 2s to complete. It is *much* faster than `journalctl --list-boots`, so that is nice. Let's see how this is going.

agners: That said, I do prefer this solution over keeping track ourselfs... So definitely a :+1: from my side :smile:

supervisor/data/syslog-identifiers.json
```diff
@@ -0,0 +1,34 @@
...
+  "btuart",
+  "dbus-broker",
+  "dockerd",
+  "dropbear",
+  "fstrim",
+  "hassos-apparmor",
+  "hassos-config",
+  "hassos-expand",
+  "hassos-overlay",
+  "hassos-persists",
```
agners, line 14: Can we add `hassos-supervisor` here too? It sometimes has interesting information what the systemd service did to the Supervisor container before starting:
```suggestion
  "hassos-persists",
  "hassos-supervisor",
```

supervisor/data/syslog-identifiers.json
```diff
@@ -0,0 +1,34 @@
...
+  "systemd",
+  "systemd-coredump",
+  "systemd-fsck",
+  "systemd-growfs",
+  "systemd-hostnamed",
+  "systemd-journal-gatewayd",
+  "systemd-journald",
+  "systemd-logind",
+  "systemd-random-seed",
+  "systemd-resolved",
```
agners, line 28: Seems to be a thing too:
```suggestion
  "systemd-resolved",
  "systemd-sysctl",
```

supervisor/data/syslog-identifiers.json
mdegat01: I thought pascal said that one just replicates the supervisor logs? so youll end up seeing all supervisor logs in `/host/logs` as well

agners: That used to be the case back in the days, but changed in in HAOS 7.0. Today you only get output of the [hassos-supervisor](https://github.com/home-assistant/operating-system/blob/9.2/buildroot-external/rootfs-overlay/usr/sbin/hassos-supervisor) script, which prints some useful information in certain situations.

mdegat01: 👍

supervisor/data/syslog-identifiers.json
mdegat01: 👍

supervisor/host/logs.py
mdegat01: Switched over to loading boot IDs and identifiers only on demand. So identifiers will be fetching on each call to `/host/logs/identifiers` and not otherwise. Caching was removed here, user will get the current list of identifiers each time (figured repeated calls in a short interval probably wasn't common enough to warrant caching).

Boot IDs will be fetched on first call to `/host/logs/boots` or `/host/logs/boots/{boot_id}` where `boot_id` is an offset number rather then an actual boot identifier. This will be cached on first use and then the query won't be run again until next restart of supervisor.

Hopefully that will be good. Then user will only feel the perf hit here if they actually need this feature.

agners (commented):
LGTM

PR #3305, opened 2021-11-15
Notify HA Core when backup is being executed

agners:
## Proposed change

Notify Home Assistant Core using the Websocket API when the Home Assistant
config directory is about to get backed up. This allows to implement
custom behavior, e.g. to make sure the SQLite data base is not in use
during backup time.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist


pvizeli (commented):
Now it going complexe, we should handle that like on add-ons and create a backup function like with add-ons to homeassistant object and group the logic on one place. We need also handle it different, based on the core version and if they support this or not

supervisor/homeassistant/module.py
```diff
@@ -284,3 +297,57 @@ async def _hardware_events(self, device: Device) -> None:
             return
 
         self.sys_homeassistant.websocket.send_message({ATTR_TYPE: "usb/scan"})
+
+    async def backup(self, secure_tar_file: SecureTarFile) -> None:
+        """Backup Home Assistant Core config/ directory."""
+
+        # Let Home Assistant Core know we are about to backup
+        try:
+            await self.websocket.async_send_command({ATTR_TYPE: "backup/start"})
```
ludeeus, line 306: Add the type to `WSType` and use that

supervisor/homeassistant/module.py
```diff
@@ -284,3 +297,57 @@ async def _hardware_events(self, device: Device) -> None:
...
+                    excludes=HOMEASSISTANT_BACKUP_EXCLUDE,
+                    arcname=".",
+                )
+
+        try:
+            await self.sys_run_in_executor(_write_tarfile)
+        finally:
+            try:
+                await self.sys_homeassistant.websocket.async_send_command(
+                    {ATTR_TYPE: "backup/end"}
```
ludeeus, line 331: Add the type to `WSType` and use that

supervisor/homeassistant/module.py
```diff
@@ -284,3 +297,57 @@ async def _hardware_events(self, device: Device) -> None:
...
+
+            with secure_tar_file as tar_file:
+                tar_file.extractall(path=origin_dir, members=tar_file)
+
+        try:
+            _LOGGER.info("Restore Home Assistant Core config folder")
+            await self.sys_run_in_executor(_restore_tarfile)
+            _LOGGER.info("Restore Home Assistant Core config folder done")
+        except (tarfile.TarError, OSError) as err:
+            _LOGGER.warning("Can't restore Home Assistant Core config folder: %s", err)
```
ludeeus, line 353: Shouldn't this raise? if it fails we don't want to continue with the restore?

supervisor/backups/manager.py
```diff
@@ -233,6 +236,9 @@ async def _do_restore(
 
                 if FOLDER_HOMEASSISTANT in folder_list:
                     backup.restore_homeassistant()
```
ludeeus, line 238: Not really related, but this should on `if homeassistant:` (L250) instead of here

supervisor/homeassistant/module.py
agners: This replaces the behavior so far: `_folder_restore()` in `restore_folders()` uses the same exception handler. That said, I think backup as well as restore needs some love wrt error handing.

supervisor/backups/backup.py
```diff
@@ -485,6 +474,26 @@ def restore_homeassistant(self):
         # save
         self.sys_homeassistant.save_data()
 
+    async def store_homeassistant_config_dir(self):
+        """Backup Home Assitant Core configuration folder."""
+
+        # Backup Home Assistant Core config directory
+        homeassistant_file = SecureTarFile(
+            Path(self._tmp.name, "homeassistant.tar.gz"), "w", key=self._key
+        )
```
pvizeli, line 483: why not just handle that all on sys_homeassistant.backup?

pvizeli (commented):
We should get away from storing the homeassistant config on the backup metadata and doing it like the add-ons, store it inside their tar.

supervisor/backups/manager.py
agners: Good point. I remember asking myself that when I did the backup/restore de-duplication PR, but somehow forgot to follow up on it.

supervisor/backups/backup.py
agners: This aligns it with how add-ons work: The `SecureTarFile` is also handled in backup.py.

I think its good the way it is: The `Backup` class should know the details about the tar file (where its stored/name, etc.)

agners:
> We should get away from storing the homeassistant config on the backup metadata and doing it like the add-ons, store it inside their tar.

Sounds reasonable. I guess we still need to support restoring the old format.

Should we still keep it separate in the selection? E.g. currently you can backup/restore metadata/version of Core and `config/` folder separately.

supervisor/homeassistant/const.py
```diff
@@ -7,7 +7,11 @@
 
 LANDINGPAGE: AwesomeVersion = AwesomeVersion("landingpage")
 
-MIN_VERSION = {"supervisor/event": "2021.2.4"}
+MIN_VERSION = {
+    "supervisor/event": "2021.2.4",
+    "backup/start": "2021.12.0",
+    "backup/end": "2021.12.0",
+}
```
agners, line 14: @ludeeus I think we use WSType here too, thoughts?

ludeeus: We should

agners: Needs a bit an awkward reordering (constants, class, constants) but :man_shrugging: 
https://github.com/home-assistant/supervisor/pull/3305/commits/5270bb3ce506001b49d719d1125f3d7fd7c4ccb1

ludeeus: `MIN_VERSION` is only used in `supervisor/homeassistant/websocket.py` we can move that there

3v1n0:
Would be also the case to notify when the backup is completed?

It could be useful to implement automations to work with the produced backup files (i.e. upload them somewhere)

pvizeli (approved):

PR #3307, opened 2021-11-16
Exclude old log files from backup

agners:
## Type of change

Currently log files are excluded from the backup. Core uses a time or
startup based log rotation mechanism, both append a dot and an
additional; string (`.1`, `.{timestamp}`). There is also a fault
mechanism which creates a log file with the ending `.log.fault`. Exclude
all cases from the backup.

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/backups/backup.py
+++ b/supervisor/backups/backup.py
@@ -58,6 +58,7 @@ MAP_FOLDER_EXCLUDE = {
         "*.db-shm",
         "__pycache__/*",
         "*.log",
+        "*.log.*",
         "OZW_Log.txt",
     ]
 }


supervisor/backups/backup.py
```diff
@@ -58,6 +58,7 @@
         "*.db-shm",
         "__pycache__/*",
         "*.log",
+        "*.log.*",
```
ludeeus, line 61: Why are we even excluding the log in the first place?
That is a part of the configuration directory?

agners: I guess they were not considered valuable enough to preserve? I kinda agree with that. If you restore an old backup on the same device, the logs are still there (and its probably more likely that you want to look at the last logs instead of the logs from the backup, but that is speculation and depends on the situation).

agners: In any case, the current situation seems incoherent. Either we add this line, or we remove the one above. 

@pvizeli thoughts?

ludeeus: My vote is removing the current line

pvizeli: I have zero opinion on that

frenck: I really think there are zero benefits of having them backupped.
I would exclude them all.

pvizeli (approved):

PR #3317, opened 2021-11-18
WS connection fix

ludeeus:
## Proposed change

Currently, our WebSocket connection breaks on core restart and about 1 after startup, this is silent since we never logged anything.

This PR implements a listener and more error handling so when the connection is dropped we can reinstate the client.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist


supervisor/homeassistant/websocket.py
```diff
@@ -22,61 +26,112 @@ class WSClient:
...
 
     @classmethod
     async def connect_with_auth(
-        cls, session: aiohttp.ClientSession, url: str, token: str
+        cls, session: aiohttp.ClientSession, loop, url: str, token: str
     ) -> "WSClient":
         """Create an authenticated websocket client."""
         try:
-            client = await session.ws_connect(url, ssl=False)
+            client = await session.ws_connect(url, timeout=500)
```
pvizeli, line 121: We need disable SSL check for connection to core

supervisor/homeassistant/websocket.py
```diff
@@ -22,61 +26,112 @@ class WSClient:
...
+        client: aiohttp.ClientWebSocketResponse,
     ):
         """Initialise the WS client."""
-        self.ha_version: AwesomeVersion = ha_version
-        self.client: aiohttp.ClientWebSocketResponse = client
-        self.message_id: int = 0
-        self._lock: asyncio.Lock = asyncio.Lock()
+        self.ha_version = ha_version
+        self._client = client
+        self._message_id: int = 0
```
pvizeli, line 37: I think that gets messed up on the restart of the supervisor during the runtime? Maybe would be better to use https://github.com/home-assistant/core/blob/dev/homeassistant/util/uuid.py#L12

ludeeus: I have restarted core and SU multiple times with this, there were no issues detected so far

supervisor/homeassistant/websocket.py
agners: So this sets `Timeout for websocket to close` to 8.3 minutes, any specific reason for that number? Do we have a guarantee to have a message within this time? Should we set it to `None`?

supervisor/homeassistant/websocket.py
```diff
@@ -22,61 +26,112 @@ class WSClient:
...
+        if msg.type in (
+            WSMsgType.CLOSE,
+            WSMsgType.CLOSED,
+            WSMsgType.CLOSING,
+            WSMsgType.ERROR,
+        ):
+            raise HomeAssistantWSConnectionError()
+
+        if msg.type != WSMsgType.TEXT:
+            raise HomeAssistantWSError(f"Received non-Text message: {msg.type}")
```
agners, line 98: This seems an error which which is related to the message itself, I think we should pass it to the caller via `future.set_exception(...)`.

supervisor/homeassistant/websocket.py
```diff
@@ -22,61 +26,112 @@ class WSClient:
...
+
+        try:
+            data = msg.json()
+        except ValueError as err:
+            raise HomeAssistantWSError(f"Received invalid JSON - {msg}") from err
```
agners, line 103: same here, I think this exception should be passed to the future.

supervisor/homeassistant/websocket.py
ludeeus: This is outdated

supervisor/homeassistant/websocket.py
ludeeus: At this point we don't know if we have something waiting, and we can't know until we read the message data further down

supervisor/homeassistant/websocket.py
ludeeus: At this point we don't know if we have something waiting, and we can't know until we read the message data further down

supervisor/homeassistant/websocket.py
ludeeus: Home Assistant also expect an int for "id" and does not allow for any additional fields that we could use for uuid

supervisor/homeassistant/websocket.py
agners: Oh I see, response ID is part of the payload, which we need to get the future. Yeah well then.

supervisor/homeassistant/websocket.py
```diff
@@ -22,41 +26,95 @@ class WSClient:
...
+        if data["type"] == "result":
+            if (future := self._futures.get(data["id"])) is None:
+                return
+
+            if data["success"]:
+                future.set_result(data["result"])
+                return
 
-            raise HomeAssistantWSError(response)
+            _LOGGER.error("Unsuccessful websocket message - %s", data)
```
agners, line 113: But I think in this case we could inform the future about the unsuccessful message. I guess setting an exception would be most fitting here.

ludeeus: Yeah, we could do that :+1:

supervisor/homeassistant/websocket.py
```diff
@@ -22,41 +26,97 @@ class WSClient:
...
+        msg = await self._client.receive()
+        _LOGGER.debug("Received: %s", msg)
+
+        if msg.type in (
+            WSMsgType.CLOSE,
+            WSMsgType.CLOSED,
+            WSMsgType.CLOSING,
+            WSMsgType.ERROR,
+        ):
+            raise HomeAssistantWSConnectionError()
```
agners, line 95: It seems to me we throw graceful connection closes and errors into the same bucket here. Can we differentiate that?

I guess `CLOSE` is the message we get if we are dealing with a regular shutdown.

How about:
```
if msg.type == WSMsgType.CLOSE:
    return False

if msg.type in (
    WSMsgType.CLOSED,
    WSMsgType.CLOSING,
    WSMsgType.ERROR,
):
    raise HomeAssistantWSConnectionError()
```

supervisor/homeassistant/websocket.py
```diff
@@ -22,41 +26,97 @@ class WSClient:
...
 
-            try:
-                response = await self.client.receive_json()
-            except ConnectionError:
-                return
+        try:
+            while self.connected:
+                await self._receive_json()
+        except HomeAssistantWSConnectionError:
+            pass
```
agners, line 76: .. then we can do something better here:

```
            while self.connected:
                if not await self._receive_json():
                    _LOGGER.debug("Websocket connection closed gracefully.");
                    break
        except HomeAssistantWSConnectionError:
             _LOGGER.warning("Websocket connection closed with connection error.");
```

supervisor/homeassistant/websocket.py
agners: (needs also a `return True` at the end of the function)

ludeeus: We can, but that adds logic that is not needed.
In any of these scenarios, we need to start the process of closing down and recreating the client, which should be transparent for the user.

agners: Yes, we need to recreate the client in any case.

But I'd still prefer if we differentiate between expected flow (regular close) and error case. I am mainly interested in having a warning in the log or something in the error case. This helps us to understand if the system is behaving as we expect.

supervisor/homeassistant/websocket.py
agners: It is mainly that `pass` which I don't like. We could also just add a debug message here, but as stated above, I think its worth differentiate these cases. After all this is going to be a ever more important communication channel, having it sold is worth the effort IMHO :)

supervisor/homeassistant/websocket.py
ludeeus: I don't agree, but whatever. I'll add something

agners (approved):

PR #3366, opened 2021-12-15
Log if board reported by HAOS isn't in version file

agners:
## Proposed change

Explicitly check if the board reported by HAOS is in the version file.
Also make sure that we still set the download URL so it will still be
possible to update to old versions in case needed.

Noted here: https://github.com/home-assistant/version/pull/190#pullrequestreview-831964935

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/updater.py
+++ b/supervisor/updater.py
@@ -242,11 +242,17 @@ class Updater(FileConfiguration, CoreSysAttributes):
 
             # Update HassOS version
             if self.sys_os.board:
-                events.append("os")
-                self._data[ATTR_HASSOS] = AwesomeVersion(
-                    data["hassos"][self.sys_os.board]
-                )
                 self._data[ATTR_OTA] = data["ota"]
+                if self.sys_os.board in data["hassos"]:
+                    events.append("os")
+                    self._data[ATTR_HASSOS] = AwesomeVersion(
+                        data["hassos"][self.sys_os.board]
+                    )
+                else:
+                    _LOGGER.warning(
+                        "Board '%s' not found in version file. No OS updates.",
+                        self.sys_os.board,
+                    )
 
             # Update Home Assistant plugins
             self._data[ATTR_CLI] = AwesomeVersion(data["cli"])
--- a/supervisor/updater.py
+++ b/supervisor/updater.py
@@ -243,11 +243,9 @@ class Updater(FileConfiguration, CoreSysAttributes):
             # Update HassOS version
             if self.sys_os.board:
                 self._data[ATTR_OTA] = data["ota"]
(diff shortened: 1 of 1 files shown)


supervisor/updater.py
```diff
@@ -242,11 +242,17 @@ async def fetch_data(self):
...
-                events.append("os")
-                self._data[ATTR_HASSOS] = AwesomeVersion(
-                    data["hassos"][self.sys_os.board]
-                )
                 self._data[ATTR_OTA] = data["ota"]
+                if self.sys_os.board in data["hassos"]:
+                    events.append("os")
+                    self._data[ATTR_HASSOS] = AwesomeVersion(
+                        data["hassos"][self.sys_os.board]
+                    )
```
ludeeus, line 250:
```suggestion
                if version := data["hassos"].get(self.sys_os.board):
                    events.append("os")
                    self._data[ATTR_HASSOS] = AwesomeVersion(version)
```

supervisor/updater.py
```diff
@@ -242,11 +242,17 @@ async def fetch_data(self):
...
+                else:
+                    _LOGGER.warning(
```
ludeeus, line 252: Should we here mark it as unsupported?

agners: Hm, just because we don't support it in OS 7, I think OS 6.6 should remain supported... I'd say we only mark it as unsupported once we consider that particular OS version as unsupported.

ludeeus: But they will never be able to update?
Shouldn't we note that as soon as possible so they can find something else to run it on?

agners: Hence that message as a warning? It just feels to me that marking it as unsupported already today seems somewhat harsh. What implications does it have?

ludeeus: That warning will be missed by most.
We never really supported the missing boards.

The implications will be a note in the UI and logs

agners: How about marking it as unsupported as soon as its removed from the version file but leave the boards in the version file pointing to 6.6 for a while? This way people using <6.6 still get the update to the last supported version, and we have a bit of a transition time.

agners: There is one problem though with that approach, the GH action updates versions for all boards unconditionally. So that would need some fixes there :sweat:

ludeeus: That would be a better approach indeed 👍
But that's out of the scope in this PR, I'd say is fine as is, and we make it better in a follow up when we have adjusted that logic

ludeeus (approved):

PR #3382, opened 2021-12-28
Using CAS for content-trust

pvizeli:
## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist


supervisor/utils/codenotary.py
```diff
@@ -83,7 +82,7 @@ async def cas_validate(
     if _ATTR_ERROR in data_json:
         raise CodeNotaryBackendError(data_json[_ATTR_ERROR], _LOGGER.warning)
 
-    if data_json[_ATTR_VERIFICATION][_ATTR_STATUS] == 0:
+    if data_json[_ATTR_STATUS] == 0:
```
mdegat01, line 85: @pvizeli Tests were failing because it said the `verification` attribute didn't exist. I tried running the `cas authenticate` command directly in terminal with the same hash and signer and got a response like this:
```json
{
  "uid": "1644241102076801109",
  "kind": "file",
  "name": "stable.json",
  "hash": "8e15c35c6e53d216a9465eb1fecbdad214c43ab02a1f08f007820ef9436f1638",
  "size": 2105,
  "timestamp": "2022-02-07T13:38:22.076801109Z",
  "contentType": "text/plain; charset=utf-8",
  "metadata": {
    "CAS_CI_ENV": "github",
    "CI": "true",
    "GITHUB_ACTION": "__home-assistant_actions",
    "GITHUB_ACTIONS": "true",
    "GITHUB_ACTOR": "agners",
    "GITHUB_API_URL": "https://api.github.com",
    "GITHUB_BASE_REF": "",
    "GITHUB_EVENT_NAME": "push",
    "GITHUB_EVENT_PATH": "/home/runner/work/_temp/_github_workflow/event.json",
    "GITHUB_GRAPHQL_URL": "https://api.github.com/graphql",
    "GITHUB_HEAD_REF": "",
    "GITHUB_JOB": "signing",
    "GITHUB_REF": "refs/heads/master",
    "GITHUB_REPOSITORY": "home-assistant/version",
    "GITHUB_RUN_ID": "1806716161",
    "GITHUB_RUN_NUMBER": "1779",
    "GITHUB_SERVER_URL": "https://github.com",
    "GITHUB_SHA": "10f638ac9cc1b9900bca7c2794afc0d2ecbb8ffb",
    "GITHUB_WORKFLOW": "Version",
    "GITHUB_WORKSPACE": "/home/runner/work/version/version"
  },
  "signer": "bm90YXJ5QGhvbWUtYXNzaXN0YW50Lmlv",
  "revoked": "0001-01-01T00:00:00Z",
  "status": 0,
  "ledger": "PublicNotarizationLedger",
  "PublicKey": "",
  "verified": true,
  "Verbose": null
}
```
And yea, that field is missing. So it seems like maybe the underlying API changed since you made this? I was looking in the CAS repo and not really seeing documentation detailing this JSON or an API spec so its hard to tell specifically what changed when. Is there a spec somewhere that I'm failing to find I can use to confirm this is the correct field? Based on the response I was torn between making this condition based on `status` or `verified`. `verified` sounds right but `status` has a `0` which matches the current check so I went with that.

mdegat01:
@pvizeli ok so `test_fetch_versions` is failing because the first time you run `cas authenticate` on a clean system you get a response like this:
Screen Shot 2022-02-08 at 8 30 17 PM

Notice that message at the top. That's what's breaking it because the response isn't JSON with that there. It happens the first time you run `cas authenticate` on a clean system like we're doing in the workflow. The public key is then cached in `~/.cas-trusted-signing-pub-key` and the message doesn't appear again after that. It's unfortunately in `stdout` not `stderr` so no good way to separate it other then string manipulation before parsing as json.

Think we can ship the image with this key? I assume based on [this](https://github.com/codenotary/cas#verify-cas-server-identity) that it's stable.

As for the other errors I'm not entirely sure what to make of them. Neither seems like they could be related to this change. The one on the Build Supervisor workflow actually sounds like a bug with the action itself based on the output. Found [this issue](https://github.com/jitterbit/get-changed-files/issues/11) but I don't see a force push in the history of this PR so not sure how that would've happened.

pvizeli:
> 
> Think we can just ship the image with this key? I assume based on [this](https://github.com/codenotary/cas#verify-cas-server-identity) that it's stable. That would eliminate this from the response.
> 

Yeah, we should ship the image with preinstalled key on dockerfile + install the key:
- https://github.com/home-assistant/actions/blob/master/helpers/cas/action.yml
- https://github.com/home-assistant/devcontainer/blob/main/common/install/cas

Then it should be solved on our workflows as well

supervisor/utils/codenotary.py
pvizeli: Make sense, we can later also use the data from the ledger to verify it a bit better. But for now it should be fine

pvizeli:
Yeah, push the file into is even better over the rootfs, nice catch

rootfs/root/.cas-trusted-signing-pub-key
```diff
@@ -0,0 +1,4 @@
+-----BEGIN PUBLIC KEY-----
+MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAE03LvYuz79GTJx4uKp3w6NrSe5JZI
+iBtgzzYi0YQYtZO/r+xFpgDJEa0gLHkXtl94fpqrFiN89In83lzaszbZtA==
+-----END PUBLIC KEY-----
```
mdegat01, line 4: @pvizeli I'm realizing we're going to have to do this for every container 😩 Probably ok for now but maybe we should just move it to the base image? Over time we want everything to do this right?

ludeeus: It's only needed where the check is running from? And that is only the supervisor.

pvizeli: Yeah is better to have it on SU roots for now

PR #3465, opened 2022-02-25
Passing platform arg on image pull

mdegat01:
## Proposed change

Populating platform arg when doing an image pull by mapping from the determined cpu architecture.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/addons/model.py
+++ b/supervisor/addons/model.py
@@ -504,6 +504,11 @@ class AddonModel(CoreSysAttributes, ABC):
         return self.data.get(ATTR_MACHINE, [])
 
     @property
+    def arch(self) -> str:
+        """Return architecture to use for the addon's image."""
+        return self._arch(self.data)
+
+    @property
     def image(self) -> Optional[str]:
         """Generate image name from data."""
         return self._image(self.data)
@@ -614,15 +619,22 @@ class AddonModel(CoreSysAttributes, ABC):
         except (AwesomeVersionException, TypeError):
             return True
 
+    def _arch(self, config) -> str:
+        """Determine best fit architecture for the addon's image."""
+        if ATTR_IMAGE in config:
+            return self.sys_arch.match(config[ATTR_ARCH])
+
+        return self.sys_arch.default
+
     def _image(self, config) -> str:
         """Generate image name from data."""
+        arch = self.arch
         # Repository with Dockerhub images
         if ATTR_IMAGE in config:
-            arch = self.sys_arch.match(config[ATTR_ARCH])
             return config[ATTR_IMAGE].format(arch=arch)
 
         # local build
-        return f"{config[ATTR_REPOSITORY]}/{self.sys_arch.default}-addon-{config[ATTR_SLUG]}"
+        return f"{config[ATTR_REPOSITORY]}/{arch}-addon-{config[ATTR_SLUG]}"
 
     def install(self) -> Awaitable[None]:
         """Install this add-on."""
--- a/supervisor/addons/model.py
+++ b/supervisor/addons/model.py
@@ -506,7 +506,10 @@ class AddonModel(CoreSysAttributes, ABC):
     @property
     def arch(self) -> str:
         """Return architecture to use for the addon's image."""
-        return self._arch(self.data)
+        if ATTR_IMAGE in self.data:
+            return self.sys_arch.match(self.data[ATTR_ARCH])
+
+        return self.sys_arch.default
 
     @property
     def image(self) -> Optional[str]:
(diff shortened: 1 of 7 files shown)


supervisor/docker/interface.py
```diff
@@ -171,7 +179,10 @@ def _install(
                 self._docker_login(image)
 
             # Pull new image
-            docker_image = self.sys_docker.images.pull(f"{image}:{version!s}")
+            docker_image = self.sys_docker.images.pull(
+                f"{image}:{version!s}",
+                platform=MAP_ARCH[self.sys_arch.default],
```
pvizeli, line 184: That will not work for add-ons. We have to add an parameter for the function and pass it from upper level function call

pvizeli: i.e. https://github.com/home-assistant/supervisor/blob/main/supervisor/addons/model.py#L621-L622

supervisor/core.py
```diff
@@ -116,14 +116,14 @@ async def setup(self):
             self.sys_host.load(),
             # Adjust timezone / time settings
             self._adjust_system_datetime(),
+            # Load CPU/Arch
+            self.sys_arch.load(),
```
mdegat01, line 120: Kind of a scary change moving things around in core setup but I think it makes sense. Looking at `CpuArch` its only dependencies in `coresys` appear to be `sys_supervisor` and `sys_machine`, both of which are set up before this method runs.

Alternatively I can adjust all the plugins to pass in `self.sys_supervisor.arch` when they call `install` but using `self.sys_arch.default` seemed more correct.

supervisor/addons/model.py
```diff
@@ -614,15 +619,22 @@ def _available(self, config) -> bool:
...
+    def _arch(self, config) -> str:
+        """Determine best fit architecture for the addon's image."""
+        if ATTR_IMAGE in config:
+            return self.sys_arch.match(config[ATTR_ARCH])
+
+        return self.sys_arch.default
+
     def _image(self, config) -> str:
         """Generate image name from data."""
+        arch = self.arch
```
pvizeli, line 631: I guess we can just reference directly to `self.arch` on the code above

supervisor/docker/addon.py
```diff
@@ -515,7 +515,11 @@ def _run(self) -> None:
             )
 
     def _install(
-        self, version: AwesomeVersion, image: str | None = None, latest: bool = False
+        self,
+        version: AwesomeVersion,
+        image: str | None = None,
+        latest: bool = False,
+        arch: str | None = None,
```
pvizeli, line 522: Let's create an Enum for Arch. See https://github.com/home-assistant/supervisor/blob/main/supervisor/const.py#L392 and then we can use it on all our Schema (see AddonStage is used). We currently converting all kinda flags into real data types

supervisor/docker/interface.py
```diff
@@ -150,19 +158,29 @@ def _docker_login(self, image: str) -> None:
...
+        latest: bool = False,
+        arch: Optional[str] = None,
     ) -> None:
         """Pull Docker image.
 
         Need run inside executor.
         """
         image = image or self.image
+        # sys_arch.default not set before plugins and may be an unsupported arch
+        arch = arch or self.sys_supervisor.arch
```
pvizeli, line 183: I guess we can load it early in the process?

mdegat01: I tried that earlier, see the resolved comment chain with myself above. Main problem is [this code](https://github.com/home-assistant/supervisor/blob/main/supervisor/arch.py#L58-L63). When on an unsupported arch sets default arch to that unsupported arch and nothing will have an image for that arch so that will fail.

I realize unsupported architectures are well, unsupported, but that would be pretty catastrophic for anyone in that situation. I'm not sure supervisor would even start anymore since it would just fail to install all the plugins every 30 seconds until shut down.

supervisor/addons/model.py
```diff
@@ -614,15 +619,22 @@ def _available(self, config) -> bool:
         except (AwesomeVersionException, TypeError):
             return True
 
+    def _arch(self, config) -> str:
```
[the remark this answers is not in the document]
mdegat01: I don't have a good reason, I was just following patterns I was seeing to keep the code consistent. I had to modify `image` as a part of its change and I saw it had separated the code into `image` and `_image` so I did the same. I meant to ask about that actually since I wasn't sure what the advantage was of doing it that way. I can certainly merge that code into `arch` though if we prefer.

supervisor/docker/interface.py
mdegat01: Also I was less sure about this one but it seemed like that might create some edge cases where we picked an image with one arch for supervisor and an image with a different arch for all the plugins. While I don't think that would break anything it did seem odd.

pvizeli: Then let's fall back on https://github.com/home-assistant/supervisor/blob/main/supervisor/arch.py#L36 - I created the arch class to have the arch handling central and not refer each time to `self.sys_supervisor.arch` as I did before we had the arch object.

supervisor/addons/model.py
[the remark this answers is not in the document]
pvizeli: Yeah, the image is a bit special as he have to return the raw value sometimes or the lookup value

pvizeli (approved):

Issue #1687, opened 2020-04-29
Allow DNS settings through supervisor UI


andriej:
**Home Assistant release with the issue:** 
n/a

**Operating environment (HassOS/Generic):** 
n/a

**Supervisor logs:** 
n/a

**Description of problem:**
One of the most common on our HA users group are DNS-related issues connected with either wrong DNS from system on supervised installation or problems that could occur because users don't know how to check/change DNS-es.

It would be superb if "Supervisor" menu could do more, what can be achieved through ha-cli but it's always easier for final user to just make sure what he sees in one place.

PR #3528, opened 2022-03-23
Apply network settings on start to fix defaults

mdegat01:
## Proposed change

Apply the current network settings on start of supervisor as a change. When network settings change OS recreates the network config with the current defaults along with the users settings. This give us a way to clean up out of date configs still using the old `default` profile and keep network settings up to date going forward.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist


mdegat01:
I still have to add tests but @agners do you have a system available with an out of date network config that can be used to test this fixes it? Or if you want to send me instructions on how to put my dev system in the bad state to confirm it corrects it. Also is there an issue to link here or was this a potential issue you identified?

supervisor/host/network.py
```diff
@@ -100,6 +99,28 @@ def get(self, inet_name: str) -> Interface:
...
+    def _check_dbus(self):
+        """Check available dbus connection."""
+        if not self.sys_dbus.network.is_connected:
+            raise HostNotSupportedError(
+                "No systemd D-Bus connection available", _LOGGER.error
+            )
+
+    async def load(self):
+        """Load network information and reapply defaults over dbus."""
+        self._check_dbus()
```
pvizeli, line 111: That is the old style. Create and use somethings like `@Job(conditions=[JobCondition.HOST_NETWORK])` (pseudo code) - i.e. https://github.com/home-assistant/supervisor/blob/main/supervisor/homeassistant/core.py#L468

supervisor/host/network.py
```diff
@@ -100,6 +103,25 @@ def get(self, inet_name: str) -> Interface:
             self.sys_dbus.network.interfaces[inet_name]
         )
 
+    @Job(conditions=JobCondition.HOST_NETWORK)
```
agners, line 106: We don't have this in update, is it really required?

pvizeli: because why we should do it without a network manager?

supervisor/jobs/decorator.py
```diff
@@ -181,6 +181,14 @@ def _check_conditions(self):
                 f"'{self._method.__qualname__}' blocked from execution, no Home Assistant OS-Agent available"
             )
 
+        if (
```
pvizeli, line 184: Please add a test for this. It should be not too hard, you can test it without load and then after load (which work because of the dbus mockups)

pvizeli (approved):
Only a single comment, after that looks good to merge 👍

supervisor/host/manager.py
```diff
@@ -95,29 +95,38 @@ def supported_features(self) -> list[HostFeature]:
...
 
         _LOGGER.info("Host information reload completed")
         self.supported_features.cache_clear()  # pylint: disable=no-member
 
     async def load(self):
         """Load host information."""
         with suppress(HassioError):
-            await self.reload()
+            await self.reload(network=False)
+            await self.network.load()
```
pvizeli, line 129: maybe we should call first all platform with `load` and on a new suppress block we can reload it to update data. In that way, we need maybe also no selective reload functionality?

agners:
> I still have to add tests but @agners do you have a system available with an out of date network config that can be used to test this fixes it? Or if you want to send me instructions on how to put my dev system in the bad state I can confirm it corrects it. 

Basically any OS 7.x release or older have a default profile shipped. So installing OS on a system (I often use OVA images for testing) should allow you to test the migration path.

> Also is there an issue to link here or was this a potential issue you identified?

This should improve the situation wrt IPv6 address regenerated due to changing profile UUID (see #1687). Although, I wonder if we have no profile (at first start), then generate a profile using Supervisor, if that doesn't end up being the same situation.

mdegat01:
> Although, I wonder if we have no profile (at first start), then generate a profile using Supervisor, if that doesn't end up being the same situation.

@agners This change shouldn't generate a profile at first start. I put a specific check in there so it will only update existing profiles during startup and not create or delete any. You can see I added this `update_only` option [here](https://github.com/home-assistant/supervisor/blob/c6c396bdbb0e81b91fac017617d6dafc1991ffa0/supervisor/host/network.py#L175) which I then use during this startup check.

supervisor/host/manager.py
mdegat01: Yea I thought about that. I was mostly trying to minimize changes within this PR. None of those other components have a `load` right now so I would have to add one to each. Plus it basically felt like I would be copying everything in the `reload` function and pasting it in another place since I have to do all the same steps.

But I guess that's fine if we think there's value in having a `load` in each since it might do more then just call `update` in the future.

Btw is there a wrapper that checks if `is_connected` returns true before calling the method similar to `dbus_connected` for the `DbusInterface` classes? I couldn't find one. Seems like a better pattern then calling `if is_connected` before every `update`/`load` right?

agners:
> @agners This change shouldn't generate a profile at first start. I put a specific check in there so it will only update existing profiles during startup and not create or delete any. You can see I added this `update_only` option [here](https://github.com/home-assistant/supervisor/blob/c6c396bdbb0e81b91fac017617d6dafc1991ffa0/supervisor/host/network.py#L175) which I then use during this startup check.

Ok, I probably understood it wrong then, I thought we are going to migrate the default configs to named configs such as `Supervisor {interface}`.

With the latest OS change, no profile is distributed anymore. With that Network Manager will generate one like `Wired connection 1`. At the moment (before this change), neither Supervisor nor Network Manager stores this connection profile (so I guess it will get a new UUID on every startup?).

image

agners:
Just confirmed my fear: The fact that there is no default profile (only an ephemeral connection) will cause the IPv6 to change on every reboot... :cold_sweat: 

I guess with your change Network Manager will save that generated profile to disk, so the IPv6 should stay the same from that point on.

pvizeli:
> I guess with your change Network Manager will save that generated profile to disk, so the IPv6 should stay the same from that point on.

Correct. With this supervisor, it will get stable again as on an old day with the defaults 👍

PR #3596, opened 2022-04-26
Add host UTS namespace support for Add-Ons

agners:
## Proposed change

Using the UTS host namespace is useful when running a mDNS responder
which learns the hostname from the gethostname syscall. This way the
add-on can use the system's hostname without further doing.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to documentation pull request: https://github.com/home-assistant/developers.home-assistant/pull/1310
- Link to cli pull request:

## Checklist


supervisor/addons/validate.py
```diff
@@ -285,6 +286,7 @@ def _migrate(config: dict[str, Any]):
         vol.Optional(ATTR_HOST_NETWORK, default=False): vol.Boolean(),
         vol.Optional(ATTR_HOST_PID, default=False): vol.Boolean(),
         vol.Optional(ATTR_HOST_IPC, default=False): vol.Boolean(),
+        vol.Optional(ATTR_HOST_UTS, default=False): vol.Boolean(),
```
ludeeus, line 289: All other host namespace access results in a -1 of the add-on score, should that be done here?

ludeeus (commented):
This needs both documentation and frontend updates

supervisor/addons/validate.py
agners: Maybe.

Host [UTS namespace](https://man7.org/linux/man-pages/man7/uts_namespaces.7.html) doesn't give you much, other than the hostname and hostdomain.

Security wise, it only really gets interesting once you also have `CAP_SYS_ADMIN`. `CAP_SYS_ADMIN` is also known as "the new root", as it actually gives you all the power (see [capabilities](https://man7.org/linux/man-pages/man7/capabilities.7.html) for a list of things). It seems that currently `CAP_SYS_ADMIN` gives -1, I'd argue that it should be more penalizing than that, before penalize UTS.

We could also make it conditional: if `CAP_SYS_ADMIN` is set, then UTS gives -1, thoughts?

agners: That said, I am biased, I am protective of OTBR's score :sweat_smile:

ludeeus: Im not sure, was just mentioning that we usually do -1 
`CAP_SYS_ADMIN` might deserve -2 🤷‍♂️

agners: PID namespace `host` allows killing any process immediately without additional privileges (it seems that `CAP_KILL` is given by default). 

I implemented my suggested option.

agners:
> This needs both documentation and frontend updates

Doc's I can tackle. Frontend, I guess this is about the add-on batches? Question is what type of symbol we use for the UTS namespace. Maybe a computer (since its about the "identity of the system"?

ludeeus:
Lets skip frontend, will probably be more confusing than helpful anyway

agners:
Linked the documentation PR.

supervisor/docker/addon.py
```diff
@@ -469,12 +476,13 @@ def _run(self) -> None:
                 self.image,
                 tag=str(self.addon.version),
                 name=self.name,
-                hostname=self.addon.hostname,
+                hostname=None if self.uts_mode else self.addon.hostname,
```
mdegat01, line 479: So we're telling the container it has no hostname but we're still telling the DNS plugin it does here:
https://github.com/home-assistant/supervisor/blob/716885d0475fdd410176f880ae54de0bb4a7d2c3/supervisor/docker/addon.py#L513
I get that addons with `host_uts` can see the hostname of the host but why does that also mean they don't have their own? Think I'm not understanding exactly what this does for an addon.

agners: > So we're telling the container it has no hostname 

Not exaclty, it seems that the API needs `None` if we don't have our own UTS namesapce. If we try to set a hostname, and don't have our own, the API returns with an exception from what I remember.

> I get that addons with host_uts can see the hostname of the host but why does that also mean they don't have their own? 

That is not really a "can see" thing, its really an identity thing. Using the very low level [`gethostname`](https://man7.org/linux/man-pages/man2/gethostname.2.html) returns the hostname of the underlying system instead of a "virtual" host name of your container's UTS namespace.

But agreed, adding the hostname to the DNS system is not required in this case. The underlying host should already resolve.

agners: Sorry, missed that comment.

It is essentially the same as if multiple processes running on the same host: If they fight over control of the hostname, then it gets problematic :D

On Home Assisant OS, the hostname is set by systemd. On init.d based system its typically some init daemon setting the hostname. Neither of those should run in a container! In a container, the hostname is setup by Docker itself.

So: No, this shouldn't be problematic, unless add-ons do nasty things (which they shouldn't anways). Also, they only can change if they have `CAP_SYS_ADMIN` as well. So of all the `host_*` options we have, this is probably one of the safer option :)

agners:
This turns out to be problematic for OTBR if multiple Home Assistant OTBRs are used in the same network: Currently, they all get the hostname of the add-on, which leads to a hostname conflict in the network:

```
Default: mDNSCoreReceiveResponse: Received from 192.168.87.177:5353    4 core-silabs-multiprotocol.local. Addr 192.168.x.y
Default: mDNSCoreReceiveResponse: ProbeCount 2; restarting probing after 1024-tick pause due to possibly spurious multicast conflict (1/1) via interface -773893056 for    4 core-silabs-multiprotocol.local. Addr 192.168.x.y
```

supervisor/docker/addon.py
```diff
@@ -510,14 +521,15 @@ def _run(self) -> None:
...
-            capture_exception(err)
+        if hostname:
+            # Write data to DNS server
+            try:
+                self.sys_plugins.dns.add_host(
+                    ipv4=self.ip_address, names=[self.addon.hostname]
+                )
+            except CoreDNSError as err:
+                _LOGGER.warning("Can't update DNS for %s", self.name)
+                capture_exception(err)
```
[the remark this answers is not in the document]
agners: Yeah that make sense. Otherwise our OTBR integration would fail to talk to the OTBR REST API (which is the use case which drove this PR).

mdegat01 (approved):

frenck:
@agners this has been merged without documentation... could you take a look and make sure we document all available features/options?

Thanks! 👍 

../Frenck

agners:
@frenck yes, thanks for reminding! I've reopened the existing PR https://github.com/home-assistant/developers.home-assistant/pull/1310. Documentation still applies as is.

PR #3630, opened 2022-05-10
Support new s6 overlay

pvizeli:
## Proposed change

Need the new base image! **Don't merge before!!!**

## Type of change

- [x] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist


mdegat01:
Interesting. Is this a necessary change or an improvement? I'm asking mostly because nearly all addons have an identical finish script for their services. So is this a change nearly all addon developers will have to make as well to upgrade to the latest base image?

pvizeli:
Yeah, if we update an add-on, we have to update the finished script if they have one. I will write a blog post on the dev

rootfs/etc/services.d/supervisor/finish
```diff
@@ -5,4 +5,4 @@
 if { s6-test ${1} -ne 100 }
 if { s6-test ${1} -ne 256 }
 
-redirfd -w 2 /dev/null s6-svscanctl -t /var/run/s6/services
+/run/s6/basedir/bin/halt
```
agners, line 8: The old code also redirected exit code from what I understand. Is the new line doing the same?

pvizeli: It does the same; both don't redirect the exit code. If you want that, you can do it but that also change

frenck:
> Interesting. Is this a necessary change or an improvement? 

If we want to keep upgrading s6-overlay, it is necessary (We can't halt upgrading forever though).

There are possible more changes needed, however, the currently s6-overlay is currently backward compatible to an extent that keeps current needed changes minimal.

I think that currently having an upgrade path with minimal changes is good to get the ball rolling; after that, we can look into how (and if) we upgrade/migrate other things (e.g., maybe we should migrate from `services.d` to the s6-rc, consider if we need to migrate `cont-init.d` scripts to oneshot services with dependencies, etc).

mdegat01 (approved):

PR #3649, opened 2022-05-19
APIs for adding/removing an addon repository

mdegat01:
## Proposed change

Improve the store API by adding APIs for individual management of addon repositories. `POST /store/repositories` adds a new addon repository, `DELETE /store/respositories/{slug}` deletes one.

There are two small breaking changes in this:

1. The existing API for managing addon repos (`POST /supervisor/options` with `addons_repositories` in the body) does not create any issues/suggestions anymore. If an addon repository cannot be added for any reason (invalid, git error, bad url, etc.) then it simply isn't added and an error is returned.

Previously the API would create suggestions for all failures and then immediately execute the `EXECUTE_REMOVE` suggestions. These suggestions which were only created for errors on `git clone` or when validating the repository. Errors caused by git loading the data (most likely file corruption) resulted in an `EXECUTE_RESET` suggestion which meant repository was still added and the user didn't see anything. I turned all of these into normal errors if they occur specifically when adding a new repository.

Note that this *does not apply* to restore from backup. If an error is encountered when restoring a repository from a backup (or during startup) the repository is added and an issue/suggestion is created like it was before.

2. `EXECUTE_REMOVE` suggestions are not auto-applied anymore. The only time they are created is when an addon repo restored from a backup or on startup is found to be invalid or non-existent. Since that means it was totally valid at one point I think users should get a chance to review this situation and not have it automatically removed for them.

There's also one additional bug fix included worth calling out:

1. `POST /supervisor/options` now saves data before processing the addon repositories. Previously you could get into an odd state if you submitted something like this:
```json
{ "debug": true, "addons_repositories":["not_a_repo"]}
```
Supervisor would say it was in debug mode after that but wouldn't actually persist that change because it had an error processing the addon repositories. Now settings changes are persisted first to finish that out before moving on to repos.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist


supervisor/api/supervisor.py
```diff
@@ -155,27 +153,15 @@ async def options(self, request: web.Request) -> None:
...
-
+        # Save changes before processing addons in case of errors
         self.sys_updater.save_data()
         self.sys_config.save_data()
 
+        if ATTR_ADDONS_REPOSITORIES in body:
+            await asyncio.shield(
+                self.sys_store.update_repositories(set(body[ATTR_ADDONS_REPOSITORIES]))
+            )
+
```
pvizeli, line 164: Ditch that, it is not needed anymore. We can update the UI + CLI with this update too. Still have the same issue then

pvizeli (commented):
Let's StoreManager use FileConfiguration. Once a store was success full loaded, we store the metadata there.

supervisor/api/supervisor.py
mdegat01: You want to toss the existing API as part of this? That makes this a much bigger breaking change. I guess I was thinking to separate the frontend changes into another PR that included removing/deprecating this API. I can do it all at once though I suppose.

pvizeli (approved):

PR #3337, opened 2021-12-02
Use volume as Docker storage location

agners:
## Proposed change

Use volume as Docker storage location. This allows to use `overlay2` as file system driver and improves performance significantly.

## Type of change

- [ ] Dependency upgrade
- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (which adds functionality to the supervisor)
- [ ] Breaking change (fix/feature causing existing functionality to break)
- [ ] Code quality improvements to existing code or addition of tests

## Additional information

- This PR fixes or closes issue: fixes #
- This PR is related to issue:
- Link to documentation pull request:
- Link to cli pull request:

## Checklist

- [ ] The code change is tested and works locally.
- [ ] Local tests pass. **Your PR cannot be merged unless tests pass**
- [ ] There is no commented out code in this PR.
- [ ] I have followed the [development checklist][dev-checklist]
- [ ] The code has been formatted using Black (`black --fast supervisor tests`)
- [ ] Tests have been added to verify that the new code works.

If API endpoints of add-on configuration are added/changed:

- [ ] Documentation added/updated for [developers.home-assistant.io][docs-repository]

[dev-checklist]: https://developers.home-assistant.io/docs/en/development_checklist.html
[docs-repository]: https://github.com/home-assistant/developers.home-assistant

PR #3650, opened 2022-05-19
Revert "Use volume as Docker storage location"

pvizeli:
This break the hole logic how it works. Since path are all inside devcontainer but now not anymore and all point into wrong folders

Reverts home-assistant/supervisor#3337


mdegat01:
I'm confused, what's wrong with it? I've been using this change for like 2 months now without issue. Have just been manually making the change by stashing and popping it since my devcontainer is basically unusable without it. I haven't noticed any missing folders or anything.

PR #3676, opened 2022-06-08
Never disable interfaces at startup

### LINKED ISSUE #3674: Home Assistant OS loses network connectivity when Ethernet goes down (router reboot)
## TL;DR:

Home Assistant OS completely loses connectivity if the router reboots.

## Detailed description:

My router is "fairly modern" TP-Link Archer AX-50 router and when I setup it a year ago there was this option to nightly reboot, which I selected because I'm a developer and I know memory leak is a real thing and why not let it reboot every night at 4AM.

I never had problems with the nightly reboot. Chromecast, PowerLine network, different raspberry PIs (mini-DLNA, piHole, camera) that I have around, everything would just work. It would reboot and in the morning I wouldn't have known because everything just works.

A month or so ago I got started with Home Assistant, so it's currently a small setup, just to get start and learn configurations with time. rPi 3 that I had already, ordered a Conbee II USB stick and bought some bulbs, sockets and remote on Ikea.

Got the latest 64-bit OS installed on a SansDisk A1 card I had laying around and setup went all fine with 'zha' without issues. But every night I would lose access to Home Assistant. Cannot connect via the Android app, or via homeassistant.local:8123 or via 192.168.2.99:8123, cannot ping from my laptop terminal and cannot see it in the router control panel. The only way to get it back was pulling the plug and turn on again.

Very important to note that Home Assistant and 'zha' was still working as I could click on the remote control all the controls for the lights and sockets were still operational. It is just the Ethernet access that was down.

I tried: a new power supply; a new SD-card; re-install with the 32-bit OS, revert back to 64-bit. Nothing worked.

After a while I noticed that it was consistently at night. I could boot-up the rPi in the early morning or just before sleep, in the morning it would be disconnected again.

Then I finally had the insight to remember that the router is rebooting:

**Tested:** HomeAssistant is working and I enter the router control panel and tapped the Reboot icon. The router reboots my internet and Chromecast and other rPi are all back online after a couple minutes, but the Home Assistant network is out.

**Workaround:** Disabled the router auto-reboot function and HomeAssistant have been working for few weeks without issues. It's a valid workaround but there's clearly some issue on HomeAssistant OS on the raspberryPi that does not affect the Raspberry Pi OS.

### What operating system image do you use?

rpi3-64 (Raspberry Pi 3 64-bit OS)

### What version of Home Assistant Operating System is installed?

7.6

### Did you upgrade the Operating System.

No

1. Connect the HomeAssistant OS rPi to a router with software reboot functionality.
2. Enter the router control panel
3. Reboot the router

```txt
Cannot access the logs because network is down
```

### Anything in the Host logs that might be useful for us?

```txt
Cannot access the logs because network is down
```

## System Health

version | core-2022.5.4
-- | --
dev | false
hassio | true
docker | true
user | root
virtualenv | false
python_version | 3.9.9
os_name | Linux
os_version | 5.10.103-v8
arch | aarch64

logged_in | false
-- | --
can_reach_cert_server | ok
can_reach_cloud_auth | ok
can_reach_cloud | ok

</details>

host_os | Home Assistant OS 7.6
-- | --
update_channel | stable
supervisor_version | supervisor-2022.05.1
docker_version | 20.10.9
disk_total | 58.7 GB
disk_used | 4.4 GB
healthy | true
supported | true
board | rpi3-64
supervisor_api | ok
version_api | ok
installed_addons | File editor (5.3.3)

</details>

dashboards | 1
-- | --
resources | 0
mode | auto-gen

</details>

### Additional information

_No response_

mdegat01:
## Proposed change

If the IPv4 and IPv6 methods are disabled for an enabled interface, do not push an update to the interface at startup. This can be caused a connectivity or DHCP issue at startup and then Supervisor will make this small issue a big one by permanently disabling the interface. After that the user can only fix it by connecting a monitor and keyboard after connectivity has been re-established.

Plus if an interface actually has both interfaces disabled then it really doesn't matter if its defaults are up to date.

### Steps to reproduce issue

It's not obvious how this breaks so here is what I was doing to reproduce this 100% of the time:
1) Have a fully functional, network connected HAOS system with ipv4 or ipv6 method set to `auto`
2) Shut it down
3) While shut down, turn off the DHCP server in the network it will connect to (I switched its port from a VLAN with a DHCP server to one without one)
4) Turn it back on
5) Run `ha network info` - should say `disabled`
6) SSH into the host shell and run `nmcli con show 'Supervisor eth0'`. Check `ipv4.method` or `ipv6.method` (whichever you had set to `auto`)

6 is the key. Prior to this PR that would say `disabled` and the user would be screwed. Network access is permanently shut off and can really only be fixed with a keyboard and monitor.

After this PR it still says `auto` and should come right back up if you turn the DHCP server back on and reboot it. Or if the DHCP server comes on before NetworkManager times out waiting for it.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR fixes or closes issue: fixes #3674 
- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/host/network.py
+++ b/supervisor/host/network.py
@@ -119,6 +119,10 @@ class NetworkManager(CoreSysAttributes):
                     self.apply_changes(interface, update_only=True)
                     for interface in interfaces
                     if interface.enabled
+                    and (
+                        interface.ipv4.method != InterfaceMethod.DISABLED
+                        or interface.ipv6.method != InterfaceMethod.DISABLED
+                    )
                 ]
             )
 
--- a/tests/host/test_network.py
+++ b/tests/host/test_network.py
@@ -2,6 +2,9 @@
 from unittest.mock import Mock, patch
 
 from supervisor.coresys import CoreSys
+from supervisor.dbus.const import InterfaceMethod
+from supervisor.host.const import InterfaceType
+from supervisor.host.network import Interface, IpConfig
 
 
 async def test_load(coresys: CoreSys):
@@ -24,8 +27,32 @@ async def test_load(coresys: CoreSys):
         assert coresys.host.network.interfaces[1].name == "wlan0"
         assert coresys.host.network.interfaces[1].enabled is False
 
-        assert activate_connection.call_count == 1
-        assert activate_connection.call_args.args == (
+        activate_connection.assert_called_once_with(
             "/org/freedesktop/NetworkManager/Settings/1",
             "/org/freedesktop/NetworkManager/Devices/1",
         )
+
+
+async def test_load_with_disabled_methods(coresys: CoreSys):
+    """Test load does not disable methods of interfaces."""
+    with patch(
+        "supervisor.host.network.Interface.from_dbus_interface",
+        return_value=Interface(
+            "eth0",
+            True,
+            False,
+            False,
+            InterfaceType.ETHERNET,
+            IpConfig(InterfaceMethod.DISABLED, [], None, []),
+            IpConfig(InterfaceMethod.DISABLED, [], None, []),
+            None,
+            None,
+        ),
+    ), patch.object(
+        coresys.host.sys_dbus.network,
+        "activate_connection",
+        new=Mock(wraps=coresys.host.sys_dbus.network.activate_connection),
+    ) as activate_connection:
+        await coresys.host.network.load()
+
+        activate_connection.assert_not_called()


agners (approved):
What I wonder why the `interface.ip4.method` is even disabled here in first place? From what I understand, that is not what the NetworkManager setting is at this point. So it seems that reading the configuration from D-Bus in this situation either is not save (as its returning not the actual setting), or we have an error in the D-Bus -> Supervisor conversion.

In any case, this helps to alleviate the problem for now. Fine by me to merge it as is.

mdegat01:
Yea you aren't wrong. I put this code locally on a system to confirm it fixes the issue. I took a before and after snapshot from DBus when I did. As you noted @agners the real culprit here appears to be this bit of code:
https://github.com/home-assistant/supervisor/blob/7a74d77d4314581474b3a0a634925feaeccde3f6/supervisor/host/network.py#L357-L358

When I looked at the properties in dbus it's not actually telling us ipv4 and ipv6 are disabled in this situation, it does not return us that config. That code is what coerces null to disabled.

It does this because before we started submitting updates at start the representation was really just for the UI. We showed the user what the current network config was either in the cli or in the GUI, they changed it, we submitted their update. If we got back null to ipv4 coercing that to disabled for a UI representation makes the most sense.

But now its not just for UI, we automatically post the config back. And so any places where we adjusted it for UI representation are problematic.

I still stand by this fix. If an interface actually has both ipv4 and ipv6 disabled we shouldn't automatically update it at startup. It's defaults might get stale but really it has to be enabled before it matters to us and that should fix the defaults when the user does that. But it is also a good idea to review that representation as it may hide other issues like this.

pvizeli (approved):

pvizeli:
Agree, we try to better detect if it's disabled (from us) or just not in use. Maybe with a better thing as [supervisor/supervisor/host/network.py](https://github.com/home-assistant/supervisor/blob/7a74d77d4314581474b3a0a634925feaeccde3f6/supervisor/host/network.py#L357-L358) or even proper metadata over an i.e. network.json

PR #3725, opened 2022-07-08
Docker events based watchdog and docker healthchecks

mdegat01:
## Proposed change

Watch docker events and push the ones about containers supervisor monitors into event bus. Docker watchdog moved from a polling to a push based system. Instead of having tasks that run on a schedule the components simply listen for events on their container and restart if necessary.

Additionally added support for [docker healthchecks](https://docs.docker.com/engine/reference/builder/#healthcheck). If an image uses it then supervisor will receive those events and restart the corresponding container when it becomes unhealthy. Addons could use this as an alternative to `watchdog_application` when the limited monitoring options there don't meet their needs. In fact I noticed in my testing some already have healthchecks and so would immediately benefit from this. We could consider possibly deprecating `watchdog_application` in the future in favor of this.

We also have several application-level checks built into core for internal components (ex. [observer application watchdog](https://github.com/home-assistant/supervisor/blob/ceed1bc318cc2c539ad2e59ee1b9ba9ce2a85e0e/supervisor/misc/tasks.py#L358-L371), [home assistant application watchdog](https://github.com/home-assistant/supervisor/blob/ceed1bc318cc2c539ad2e59ee1b9ba9ce2a85e0e/supervisor/misc/tasks.py#L201-L240), and [dns loop detection](https://github.com/home-assistant/supervisor/blob/ceed1bc318cc2c539ad2e59ee1b9ba9ce2a85e0e/supervisor/misc/tasks.py#L308-L309)). We could consider removing this logic from supervisor and adding healthchecks to their images instead. And possibly adding healthchecks to the other plugins.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist


supervisor/core.py
```diff
@@ -305,6 +313,9 @@ async def shutdown(self):
         if self.state == CoreState.RUNNING:
             self.state = CoreState.SHUTDOWN
 
+        # Stop docker events monitor
+        await self.sys_docker_monitor.unload()
```
pvizeli, line 317: Would be nice if we can handle that like hardware. Kinda `self.sys_docker.monitor` and then also only call `self.sys_docker.shutdown()` or `stop()` or whatever. Other than that, it looks great!

mdegat01: So we can but the problem is `sys_docker` is a global. `DockerAPI` has no access to the `coresys` object by design and the monitor needs access to `coresys` in order to fire events on the bus. So I can have the monitor be an attribute of `DockerAPI` but `DockerAPI` cannot create it. Something else would have to create it and then add it to the `sys_docker` object.

That felt a little odd to me so I went with a separate `sys_docker_monitor`. But I'm open to changing to that approach, it's not a big change.

pvizeli: You are right, DockerAPI doesn't have access to coresys, I guess history-wise. I would say, time to change that so we streamline the handling

pvizeli (approved):

pvizeli:
once at least all pass (without codecov) we can merge it

PR #3769, opened 2022-08-03
Add auto update option

mdegat01:
## Proposed change

Add an auto update setting for supervisor. There have been many requests about this in the forum and issues so should be valuable to users. The option will work as follows:
1. Supervisor will not auto update while it is disabled
2. Plugins cannot be updated while supervisor is out of date
3. Addons and Home Assistant cannot be updated while supervisor or a plugin is out of date
4. Addons cannot be installed when supervisor or a plugin is not up-to-date
5. Backups track the version of supervisor they were made on. Cannot restore a backup made on a newer version of supervisor then what is running.

What's not covered by this:
1. Plugins will continue to auto update as long as supervisor is up to date. Since plugins are effectively internal and not updated much restricting them from auto updating added too many complications. Users don't even know when a plugin update is available. Also if we do make a big change to a plugin it comes with a supervisor update most of the time so it won't auto update anyway.
1. Manual updates are not restricted by this new option, it only prevents automatic updates
1. Cannot install a particular version of supervisor or rollback. The supervisor version in backup just sets a validation minimum, it is not used during restore.
1. Update checks continue. Users will still see update notifications but they won't be able to update. We may want to update core to improve the experience here as clicking install will simply generate an error.

## Type of change

- [ ] Dependency upgrade

## Additional information

    - API Doc: https://github.com/home-assistant/developers.home-assistant/pull/1434
    - Unsupported Reason: https://github.com/home-assistant/home-assistant.io/pull/23733
- Link to cli pull request: https://github.com/home-assistant/cli/pull/357

## Checklist


mdegat01:
Note besides the TODOS, this shouldn't be merged until current supervisor beta goes stable. One thing at a time.

supervisor/backups/backup.py
```diff
@@ -161,11 +167,12 @@ def tarfile(self):
     def new(self, slug, name, date, sys_type, password=None, compressed=True):
         """Initialize a new backup."""
         # Init metadata
-        self._data[ATTR_VERSION] = 2
+        self._data[ATTR_VERSION] = 3
```
pvizeli, line 170: I guess we can keep it v2 and set supervisor_version default to const.version (Schema validation) - so it's compatible to all v2 backups and because we don't have any compatibility issues, we can keep on v2 then

mdegat01: Yea I was thinking about that. I just wasn't sure if the typical process was to bump the schema version for any change to the backup schema or only for backwards incompatible ones. Sounds like its only for backwards incompatible ones then so I'll revert that.

supervisor/core.py
```diff
@@ -182,8 +182,8 @@ async def start(self):
         # Mark booted partition as healthy
         await self.sys_os.mark_healthy()
 
-        # On release channel, try update itself
-        if self.sys_supervisor.need_update:
+        # On release channel, try update itself if auto update enabled
+        if self.sys_supervisor.need_update and self.sys_updater.auto_update:
```
ludeeus, line 186: We need to change how `need_update` works with this PR.

This is the current implementation:

```
    @property
    def need_update(self) -> bool:
        """Return True if an update is available."""
        if self.sys_dev:
            return False

        try:
            return self.version < self.latest_version
        except (AwesomeVersionException, TypeError):
            return False
```

This needs to return True by default until we know we got a good response for an updated version file

supervisor/supervisor.py
```diff
@@ -158,8 +158,16 @@ async def update_apparmor(self) -> None:
                     "Can't update AppArmor profile!", _LOGGER.error
                 ) from err
 
+    @Job(
+        conditions=[
+            JobCondition.FREE_SPACE,
+            JobCondition.HEALTHY,
+            JobCondition.INTERNET_HOST,
+        ],
```
pvizeli, line 166: We need to be able to ignore this on auto update or add an force option to api. If we rollout ever something that set this flag wrong, we can't update anymore any of our devices. But maybe it's also now fine as we have this technology a long time now and it's seems to work. I don't know

pvizeli: I guess Freeze should be on it as well. There are ways to ignore job conditions. I guess that is fine

pvizeli:
Let's copy/past rename freeze state to snapshot or backup

supervisor/core.py
mdegat01: Hm yea good point. Will that have an affect on the front end though? It shows update notifications anytime that is true right? I see your point though, this check needs to assume there is an update until it's sure there isn't. I'll adjust to do that.

supervisor/supervisor.py
mdegat01: I thought about that. I was basically auditing the conditions and trying to make them consistent across the different update jobs. But I was wondering if this particular one i should leave alone for that exact reason - risk of locking ourselves out of updates. But yea then I remembered job conditions can be ignored.

There is still a small risk. If an issue around these conditions ever makes it out of dev and to beta will be instructing people how to ignore jobsl conditions for a while to get it resolved. Maybe not worth it?

supervisor/core.py
ludeeus: If there is an issue in the frontend I’ll fix that

mdegat01:
> Let's copy/past rename freeze state to snapshot or backup

@pvizeli it's been renamed to auto update within the branch. Do you see it referred to as freeze somewhere still? Other then the branch name

supervisor/core.py
pvizeli: We have the update endpoints for the core. We properly only need adjust this API endpoint to see only supervisor and not others, so they can update supervisor and after that it start to show the other updates?

ludeeus: For newer installations it should not use that attribute at all, this should only be an issue for older core installations before that endpoint and before update entities

mdegat01: Well its not just frontend though. LIke if I make `need_update` return true until we have checked for an update think about what happens for the majority of users that have `auto_update` enabled in this spot. It will actually try and execute an update there. We don't want that either.

I think we have two options here:
1) This method becomes `Optional[bool]` and return `None` when we don't know if there's an update because the version file hasn't been pulled.
2) Leave this method alone and instead the `SUPERVISOR_UPDATED` job condition prevents calls to the method when there is a supervisor update or it sees we haven't pulled the version file yet.

supervisor/supervisor.py
pvizeli: that is the golden question. We have to add JobCondition.HEALTHY once if we can pull the min docker API version for the next release at our release files, so people have first to update the OS before it update the supervisor for not break. So the times come to add it, maybe now is fine but we keep the list small as possible, only HEALTHY

supervisor/backups/manager.py
```diff
@@ -316,6 +316,14 @@ async def do_restore_full(self, backup: Backup, password=None):
             _LOGGER.error("Invalid password for backup %s", backup.slug)
             return False
 
+        if backup.supervisor_version > self.sys_supervisor.version:
+            _LOGGER.error(
+                "Backup was made on supervisor version %s, can't restore on %s. Must update supervisor first.",
+                backup.supervisor_version,
+                self.sys_supervisor.version,
+            )
+            return False
```
pvizeli, line 325: backups are not full ported to the exception error handling. You can just use the new raise/log style

supervisor/supervisor.py
pvizeli: other thing is now, it raise an SupervisorJobError, we have to make sure everywhere who we use it (not on API) we catch that case if needed

mdegat01: I'm going to remove this. It's not really the subject of this PR and it sounds like we do some work around this already so I'll leave changes here for that follow-up work.

supervisor/core.py
mdegat01: After a bit of discussion what we went with is a new evaluation for `supervisor_version`. We only support the latest supervisor version so if you are running on an old version of supervisor and have auto-update disabled then the evaluation will mark the system as unsupported. 

No that doesn't stop someone who blocks the version file and prevents supervisor from knowing about updates. But if a user doing that tries to submit an issue the first thing we'll see is that the supervisor version is out of date. So we'll know its an unsupported system even if the check has been worked around.

This also deals with hacks like ignoring the `SUPERVISOR_UPDATED` job condition then un-ignoring it after an update. The system will be marked unsupported regardless because supervisor is out of date.

supervisor/backups/manager.py
mdegat01: I thought that was weird that it returned false here. I'm not going to do that change-over as part of this PR but I'll add it to the list. Seems pretty straightforward 👍

supervisor/resolution/evaluations/supervisor_version.py
```diff
@@ -0,0 +1,34 @@
...
+
+    @property
+    def on_failure(self) -> str:
+        """Return a string that is printed when self.evaluate is False."""
+        return "Not using latest version of Supervisor and auto update is disabled."
+
+    @property
+    def states(self) -> list[CoreState]:
+        """Return a list of valid states when this evaluation can run."""
+        return [CoreState.RUNNING]
```
pvizeli, line 30:
```suggestion
        return [CoreState.RUNNING, CoreState.STARTUP]
```

pvizeli (approved):
Looks good. I guess we have to iterate over but that seems to be a safe start for this. Just a small comment about the running version evaluation, but other than that, let's test it on dev

supervisor/api/supervisor.py
```diff
@@ -143,6 +146,9 @@ async def options(self, request: web.Request) -> None:
         if ATTR_LOGGING in body:
             self.sys_config.logging = body[ATTR_LOGGING]
 
+        if ATTR_AUTO_UPDATE in body:
+            self.sys_updater.auto_update = body[ATTR_AUTO_UPDATE]
```
ludeeus, line 150: If this changes from false to true we should imidiatly check for updates and only store the value to `self.sys_updater.auto_update` if we were able to check

pvizeli: That is correctly handled today. We can force an updater fetch, but that is only a delta of less than 1h. I don't see an issue or abuse use case for this delta in all workflows. I guess let's fix issues in the flow once they going to be an issue and not over complicate it on the beginning for an feature that you basic let you at your own if you use it

mdegat01: This option only controls whether we do updates of supervisor for them automatically. Setting it to `true` does not unlock any other updates. If supervisor isn't up to date then nothing else can be updated regardless of the value of this field.

And if people are blocking our ability to access the version file then there's not a whole lot we can do about that. And I don't think we should try to do much about it to be honest. Systems with an out of date supervisor are not supported. If a user blocking the version file submits an issue we'll know the system is unsupported because the supervisor version will be out of date, even if they've managed to dodge the unsupported flag.

PR #3779, opened 2022-08-08
Set limits on watchdog retries

mdegat01:
## Proposed change

Set limits on how often watchdog will retry when it finds an error. Limits are as follows:
1. If reanimation fails 5 times in the 10 second retry loop then stop trying
2. If we receive 10 failed/unhealthy events in a half hour period then ignore them

This should prevent issues like what was in sentry where an addon with an problematic config is continuously restarted with no hope of succeeding.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/addons/addon.py
+++ b/supervisor/addons/addon.py
@@ -18,8 +18,6 @@ from securetar import atomic_contents_add, secure_path
 import voluptuous as vol
 from voluptuous.humanize import humanize_error
 
-from supervisor.jobs.const import JobExecutionLimit
-
 from ..const import (
     ATTR_ACCESS_TOKEN,
     ATTR_AUDIO_INPUT,
@@ -68,6 +66,7 @@ from ..exceptions import (
 )
 from ..hardware.data import Device
 from ..homeassistant.const import WSEvent, WSType
+from ..jobs.const import JobExecutionLimit
 from ..jobs.decorator import Job
 from ..utils import check_port
 from ..utils.apparmor import adjust_profile
--- a/supervisor/addons/addon.py
+++ b/supervisor/addons/addon.py
@@ -18,6 +18,8 @@ from securetar import atomic_contents_add, secure_path
 import voluptuous as vol
 from voluptuous.humanize import humanize_error
 
+from supervisor.jobs.const import JobExecutionLimit
+
 from ..const import (
     ATTR_ACCESS_TOKEN,
     ATTR_AUDIO_INPUT,
@@ -58,6 +60,7 @@ from ..docker.stats import DockerStats
 from ..exceptions import (
     AddonConfigurationError,
     AddonsError,
+    AddonsJobError,
     AddonsNotSupportedError,
     ConfigurationFileError,
     DockerError,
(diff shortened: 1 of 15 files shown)


supervisor/addons/addon.py
```diff
@@ -18,6 +18,8 @@
 import voluptuous as vol
 from voluptuous.humanize import humanize_error
 
+from supervisor.jobs.const import JobExecutionLimit
```
pvizeli, line 21: relative import

supervisor/addons/addon.py
```diff
@@ -58,17 +60,25 @@
...
+from ..jobs.decorator import Job
 from ..utils import check_port
 from ..utils.apparmor import adjust_profile
 from ..utils.json import read_json_file, write_json_file
-from .const import WATCHDOG_RETRY_SECONDS, AddonBackupMode
+from .const import (
+    WATCHDOG_MAX_ATTEMPTS,
+    WATCHDOG_RETRY_SECONDS,
+    WATCHDOG_THROTTLE_MAX_CALLS,
+    WATCHDOG_THROTTLE_PERIOD,
```
pvizeli, line 79: I guess you wanted import that from `addons.const`?

supervisor/addons/addon.py
ludeeus, line 21: <https://github.com/microsoft/pylance-release/discussions/2684>
(use relative)

supervisor/addons/addon.py
```diff
@@ -104,6 +114,54 @@ def __init__(self, coresys: CoreSys, slug: str):
...
         self._state: AddonState = AddonState.UNKNOWN
 
+        @Job(
+            name=f"addon_{slug}_restart_after_problem",
+            limit=JobExecutionLimit.THROTTLE_RATE_LIMIT,
+            throttle_period=WATCHDOG_THROTTLE_PERIOD,
+            throttle_max_calls=WATCHDOG_THROTTLE_MAX_CALLS,
+            on_condition=AddonsJobError,
+        )
+        async def restart_after_problem(addon: Addon, state: ContainerState):
```
pvizeli, line 124: I can't follow why this function is inline and not just called `_restart_after_problem`

supervisor/addons/addon.py
mdegat01: Isn't that where they are imported from? That's what I see...

supervisor/addons/addon.py
mdegat01: For reference that is also why I had to do this in every plugin:
https://github.com/home-assistant/supervisor/blob/473aed3f31019e59bd841e239db8a765267c3be8/supervisor/plugins/audio.py#L212-L220

Instead of just putting the method in `PluginBase` and making it a job. Then there was one job and one rate limit for all plugins.

supervisor/addons/addon.py
mdegat01: Or are you really asking "why isn't everything importing the same constants since they all have the same values"? I can do that I just figured we might want to tweak them different for core vs plugins vs addons now or in the future.

pvizeli (approved):

PR #3780, opened 2022-08-09
Support basic IPv6

pvizeli:
## Proposed change

That is still not enough for full ipv6 support as we need the IPv6 plugin for handling the iptables correctly like we have IPv4 support until docker get full support of that and not like now, only partial

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/const.py
+++ b/supervisor/const.py
@@ -38,7 +38,7 @@ SYSTEMD_JOURNAL_VOLATILE = Path("/run/log/journal")
 DOCKER_NETWORK = "hassio"
 DOCKER_NETWORK_MASK = ip_network("172.30.32.0/23")
 DOCKER_NETWORK_RANGE = ip_network("172.30.33.0/24")
-DOCKER_NETWORK_LINK_LOCAL = ip_network("fd00:172:30:32::/64")
+DOCKER_NETWORK_ULA = ip_network("fd00:172:30:32::/64")
 
 # This needs to match the dockerd --cpu-rt-runtime= argument.
 DOCKER_CPU_RUNTIME_TOTAL = 950_000
--- a/supervisor/const.py
+++ b/supervisor/const.py
@@ -38,6 +38,7 @@ SYSTEMD_JOURNAL_VOLATILE = Path("/run/log/journal")
 DOCKER_NETWORK = "hassio"
 DOCKER_NETWORK_MASK = ip_network("172.30.32.0/23")
 DOCKER_NETWORK_RANGE = ip_network("172.30.33.0/24")
+DOCKER_NETWORK_LINK_LOCAL = ip_network("fd00:172:30:32::/64")
 
 # This needs to match the dockerd --cpu-rt-runtime= argument.
 DOCKER_CPU_RUNTIME_TOTAL = 950_000
--- a/supervisor/docker/network.py
+++ b/supervisor/docker/network.py
@@ -93,7 +93,7 @@ class DockerNetwork:
         )
         ipam_pool_v6 = docker.types.IPAMPool(
             subnet=str(DOCKER_NETWORK_LINK_LOCAL),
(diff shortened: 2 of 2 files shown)


pvizeli:
I think we can not just upgrade the existing network. Properly we have to create a function that user can call which disconnect all container and recreate the network. This can be later also used if we allow to change the IP address of internal network

supervisor/const.py
```diff
@@ -38,6 +38,7 @@
 DOCKER_NETWORK = "hassio"
 DOCKER_NETWORK_MASK = ip_network("172.30.32.0/23")
 DOCKER_NETWORK_RANGE = ip_network("172.30.33.0/24")
+DOCKER_NETWORK_LINK_LOCAL = ip_network("fd00:172:30:32::/64")
```
agners, line 41: Note that this is a unique local address (ULA), which is not the same as a link local address (which start with `fe80`). Link local addresses can not be routed, unique local addresses can be routed in private networks (used to be called site).

I am not sure if we could get away with link local addressing only (which gets assigned by default when IPv6 is enabled), but definitely doesn't hurt to add a unique local address.

pvizeli: It will work. It looks like we need add for now `  "experimental": true, "ip6tables": true` to dockerd for support nat6

pvizeli: https://github.com/robbertkl/docker-ipv6nat/issues/65#issuecomment-883485635

agners: Just to be clear, I suggest to change the name of that constant, since its a ULA, e.g.

```suggestion
DOCKER_NETWORK_ULA = ip_network("fd00:172:30:32::/64")
```

> It will work.

With just link local addressing (aka, not setting any address manually) or with unique local addressing?

pvizeli:
Yeah, I will create the function and create an issue/fix about it and then we can merge it

cmsj:
@pvizeli is there anything I might be able to do in order to help this PR move forward? I have built an addon that would benefit greatly from being able to get a working v6 address :)

agners:
> @pvizeli is there anything I might be able to do in order to help this PR move forward? I have built an addon that would benefit greatly from being able to get a working v6 address :)

Can you elaborate a bit more on the use case for IPv6 in your add-on? The intention here is to assign an address from the ULA range. Depending on use case this might not be useful.

cmsj:
> Can you elaborate a bit more on the use case for IPv6 in your add-on? The intention here is to assign an address from the ULA range. Depending on use case this might not be useful.

Sure thing, [my addon](https://github.com/cmsj/addon-bind) is just a simple packaging of the Bind DNS server and it a lot of the major DNS registrars/clouds have v6 addresses for their DNS servers and it would be good to be able to reach those as well as the v4 addresses.

FWIW ULA addressing is fine for this sort of thing in Docker because, as with v4, it sets up NAT for outbound connections on v6.

martymarty004:
Just a question, could this new feature be able to fix the issues I'm experiencing in my dual-stack network with the Add-ons?

For example, I've not been able to configure access policies on NPM to work properly, since all traffic that is coming from IPv6 will seem to be coming from the bridge (`172.30.32.1`). 
This has also made me disable fail2ban, since any failed attempts to access Home Assistant or its add-ons from an IPv6 client would ban automatically every IPv6 connection.

I know that I could easily disable dual-stack globally and run IPv4 only, but I'm starting a networks course next month and dual stack looks like it could be a great playground for my experiments (my ISP was kind enough to give me a static public IPv4 and a static /48 prefix for free).

frenck:
This PR has been pending for quite a long time. We do plan on picking picking this feature up at some point in time (it is on the radar).

For now, I'd like to clean up our queues a little bit, so going to close this for now. We can re-open and pick up when we are ready to work on it again.

../Frenck

PR #3781, opened 2022-08-09
Improve system behavior on low memory situation

agners:
Adjust OOM killer score to prevent crucial services from getting
selected.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/docker/observer.py
+++ b/supervisor/docker/observer.py
@@ -55,6 +55,7 @@ class DockerObserver(DockerInterface, CoreSysAttributes):
             },
             volumes={"/run/docker.sock": {"bind": "/run/docker.sock", "mode": "ro"}},
             ports={"80/tcp": 4357},
+            oom_score_adj=-300,
         )
 
         self._meta = docker_container.attrs
--- a/supervisor/docker/addon.py
+++ b/supervisor/docker/addon.py
@@ -489,6 +489,7 @@ class DockerAddon(DockerInterface):
                 environment=self.environment,
                 volumes=self.volumes,
                 tmpfs=self.tmpfs,
+                oom_score_adj=200,
             )
         except DockerNotFound:
             self.sys_resolution.create_issue(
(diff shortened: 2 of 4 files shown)


supervisor/docker/observer.py
```diff
@@ -55,6 +55,7 @@ def _run(self) -> None:
             },
             volumes={"/run/docker.sock": {"bind": "/run/docker.sock", "mode": "ro"}},
             ports={"80/tcp": 4357},
+            oom_score_adj=-300,
```
pvizeli, line 58: I think we should prevent all plugins

agners: Depends what is more important: I'd rather prefer Pulseaudio gets killed (in case it is miss-behaving) before my database add-on...

pvizeli: but dns and also cli would make sense

mdegat01: > before my database add-on...
Perhaps we should also adjust the score for addons marked as services to prevent them from being killed when possible

mdegat01: I guess in terms of plugin importance to keep from being killed I would put Observer, DNS and CLI as most important. Looks like we agree on Observer, DNS because otherwise DNS simply stops working within HA and CLI because it gives a way to figure out what broke and fix it.  Multicast and Audio do seem less critical.

Although I should note that as long as supervisor is running all of them will immediately restart if killed. They've got an always on watchdog that auto-restarts them when they stop. As does Home Assistant for that matter. Only Supervisor and possibly add-ons (since watchdog can be turned off) will actually stop running when killed this way.

pvizeli: Agree, but then at least DNS

agners: Agreed, that sounds reasonable, set the DNS plug-in to -300 as well.

pvizeli (approved):

PR #3795, opened 2022-08-16
Supervisor updated before addon repositories

mdegat01:
## Proposed change

New versions of supervisor may add config options for addons and addon repositories and updated addons/repos may use them immediately. Therefore its not safe to even check for addon updates or add new repositories while supervisor is out of date.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/store/__init__.py
+++ b/supervisor/store/__init__.py
@@ -80,6 +80,7 @@ class StoreManager(CoreSysAttributes, FileConfiguration):
             self._data[ATTR_REPOSITORIES], add_with_errors=True
         )
 
+    @Job(conditions=[JobCondition.SUPERVISOR_UPDATED], on_condition=StoreJobError)
     async def reload(self) -> None:
         """Update add-ons from repository and reload list."""
         tasks = [repository.update() for repository in self.all]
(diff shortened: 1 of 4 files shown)


supervisor/store/__init__.py
```diff
@@ -194,7 +198,6 @@ async def remove_repository(self, repository: Repository, *, persist: bool = Tru
             await self.data.update()
             self._read_addons()
 
-    @Job(conditions=[JobCondition.INTERNET_SYSTEM])
```
mdegat01, line 197: I realize this change is a bit strange since @pvizeli you suggested adding a `SUPERVISOR_UPDATED` condition here and instead I removed one. I realized a couple things:
1. We only need to add the condition to `add_repository` as this calls that for adding repositories
1. Removing repositories (which this can do too) doesn't require internet or supervisor to be updated
1. We use this method when restoring backups and during startup and we don't really want to any more checks then strictly necessary during those times

mdegat01:
I intentionally did not add `SUPERVISOR_UPDATED` here:
https://github.com/home-assistant/supervisor/blob/275bebca1f02b885af62e512c85b94672612b7da/supervisor/store/repository.py#L104

That's because there is a fixup which calls it directly. I figured in the rare situations where supervisor is out of date but the git repo has been corrupted on disk its better to take our chances and let it update the repo then leave it corrupted.

Although I should note if we decide to add a `/store/repositories/<repository>/reload` API in the future then we should reconsider this and probably add the `JobCondition` here.

pvizeli (approved):

PR #3806, opened 2022-08-19
Identify and handle dhcp issues

mdegat01:
## Proposed change

DHCP issues around startup create an issue because it can cause supervisor to disable part of interface permanently. #3676 made this far less common but not impossible. @agners noticed that if a machine has both ipv6 and ipv4 configured and only one has a DHCP issue on startup then we end up permanently disabling the other.

This PR includes a few changes:
1. Identify and report network connection issue when a managed and connected device does not have [state](https://developer-old.gnome.org/NetworkManager/stable/nm-dbus-types.html#NMActiveConnectionState) `ACTIVATING` or `ACTIVATED` or its [state flags](https://developer-old.gnome.org/NetworkManager/stable/nm-dbus-types.html#NMActivationStateFlags) are missing `IP4_READY` or `IP6_READY` (note that this flag is still present if a protocol is disabled for an interface)
1. Do not post an update for an interface with a network connection issue on startup so we don't accidentally disable things
1. Always pull the ipv4/ipv6 method from interface settings instead of assuming it is disabled when details are missing from active connection

The third one will affect the API but only in edge cases like the DHCP issue. Today the API would return `disabled` for method when the user set it to `auto` but dhcp failed. Now it will return `auto` but address, gateway and nameserver will be blank. Which is exactly what NetworkManager does in this situation and seems less confusing to a user since they didn't disable anything.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist


supervisor/host/network.py
```diff
@@ -345,28 +347,38 @@ class Interface:
...
             inet.settings is not None,
             Interface._map_nm_connected(inet.connection),
             inet.primary,
             Interface._map_nm_type(inet.type),
             IpConfig(
-                Interface._map_nm_method(inet.settings.ipv4.method),
+                ipv4_method,
                 inet.connection.ipv4.address,
                 inet.connection.ipv4.gateway,
                 inet.connection.ipv4.nameservers,
```
agners, line 370: Can't we get those 3 values from the settings object as well? Then the if statement below needs changing too, it would make the separate checking above unnecessary and simplify the code.

supervisor/resolution/checks/dhcp.py
```diff
@@ -0,0 +1,62 @@
...
+        interface = self.sys_dbus.network.interfaces.get(reference)
+
+        return interface and CheckDHCP.check_interface(interface)
+
+    @staticmethod
+    def check_interface(interface: NetworkInterface) -> bool:
+        """Return true if a managed interface has a DHCP issue."""
+        if not (interface.managed and interface.settings):
+            return False
+
```
agners, line 39: While it looks good to me, I am worried that this doesn't capture some state we don't foresee (static not applying? I don't see how that could happen, but who knows). Did you consider looking at the NetworkManager state flags?

https://developer-old.gnome.org/NetworkManager/stable/nm-dbus-types.html#NMActivationStateFlags

The logic would go something like:

```
            interface.settings.ipv4
            and interface.settings.ipv4.method != InterfaceMethod.DISABLED.value
            StateFlag & NM_ACTIVATION_STATE_FLAG_IP4_READY
```

supervisor/host/network.py
agners: I see, this would really make `Interface` class not having the current IP information, which would affect our APIs etc.

IMHO, class `Interface` should have a *current* IPv4/IPv6 field as well as a *configured* IPv4/IPv6 field. This is how NM works under the hood, and would avoid these type of problems we are seeing.

supervisor/resolution/checks/dhcp.py
mdegat01: StateFlag is actually on the Active Connection info. It also considers ip6 and ip4 ready if they are disabled since it has nothing to do there. So the check is simpler now, basically just consider it a network connection error if the state flag doesn't have these two values in it regardless of the method used.

I also added another check of `State` not being in `Activated` or `Activating` as that's a problem too. Supervisor provides no way to deactivate an existing interface so if a managed interface isn't in one of those states then there's an issue the user needs to look at.

supervisor/host/network.py
mdegat01: So to my knowledge there's only one possible issue this could cause. If it's somehow possible for an interface using `method: static` to have different values in active connection info then in its settings. Because here is how this information is actually used to make the update call to NetworkManager:
https://github.com/home-assistant/supervisor/blob/5919bc22526832d84b8cdb6596e57f9ce60535f3/supervisor/dbus/network/setting/generate.py#L63-L113

As you can see, we don't even look at those other fields unless the method is static. If we think this is a risk I can account for it in the code but I don't see how. We can't just change this though because the response here is returned from the API and the user expects to see the current address, gateway and nameserver info when they use `auto`

mdegat01: And when I say account for it I mean here:
https://github.com/home-assistant/supervisor/blob/5919bc22526832d84b8cdb6596e57f9ce60535f3/supervisor/host/network.py#L105-L126

This is the only spot where that can be an issue. In this code after calling `from_dbus_interface` I would replace the ipv4 and ipv6 `addresses`, `gateway` and `nameserver` values with what is in settings if the method is static. Because that is the only time the response of `from_dbus_interface` is passed directly to an update call.

The other time `from_dbus_interface` is used in an update call is indirectly. The user is presented the output of `from_dbus_interface` from the API in the UI. They then can choose to submit what they are seeing to update the network interface. At that point it really doesn't matter what's in settings, the user has just told us what they want the settings to be.

pvizeli (approved):

PR #3836, opened 2022-09-01
Reduce connectivity checks

mdegat01:
## Proposed change

Reviewing [this thread](https://community.home-assistant.io/t/how-to-throttle-supervisors-version-check) I noticed there was quite an excessive number of DNS requests to `checkonline.home-assistant.io` in my own DNS server. After some digging I realized the primary source of these was actually Network Manager.

It turns out when you do `nmcli networking connectivity check` (equivalent to our host internet test) Network Manager does a DNS query to the connectivity URI twice per interface (ipv4 and ipv6). And that includes all the `veths` created for docker containers. This was 30+ identical DNS requests in the same second for the same domain for me. We only do a connectivity check every 10 minutes but this adds up quickly.

On the other hand, Network Manager's automated connectivity test on its defined interval is very reasonable. It seems to simply test the real network interfaces (not the veths). So that is two queries every 5 minutes for me. And based on discussion with @agners even that may be reduced in an upcoming release of Network Manager due to [this commit](https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/commit/c1682b2556b5895b48215c7957173d4fc72530ed) (since I don't currently use ipv6 and we think most of this is due to the fallback). I'm not entirely sure why the automated and manual checks are so different but it seems better to rely on Network Manager's automation around this anyway.

So this PR changes up the logic. We only force Network Manager to do an immediate connectivity test at two times:
1. Call to `POST /network/reload`
2. Call to `POST /network/interface/<interface>/update`

All other times we simply check the result of Network Manager's last connectivity test rather then triggering one ourselves.

In addition we don't have a scheduled connectivity check task. Instead when a job requires host or system connectivity we check it. Supervisor's check remains throttled so it only makes a request to `checkonline.home-assistant.io` if the last one failed or its been more then 10 minutes since the last one as it was before.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR is related to issue: https://community.home-assistant.io/t/how-to-throttle-supervisors-version-check
- Link to documentation pull request: https://github.com/home-assistant/home-assistant.io/pull/23988
- Link to cli pull request:

## Checklist


supervisor/host/network.py
```diff
@@ -82,15 +82,16 @@ def dns_servers(self) -> list[str]:
 
         return list(dict.fromkeys(servers))
 
-    async def check_connectivity(self):
+    async def check_connectivity(self, *, force: bool = False):
         """Check the internet connection."""
         if not self.sys_dbus.network.connectivity_enabled:
+            self.connectivity = None
```
mdegat01, line 88: Is this a bad idea? The value of this field can be `None` but normally only at startup until a connectivity check either succeeds or fails. I wanted to revert it back to `None` if someone disabled the connectivity check but doing so will cause supervisor to fire an event from here withe value set to `None`:
https://github.com/home-assistant/supervisor/blob/dcd0592d44026c0e0aa6223ac306e0ccf9f9c20d/supervisor/host/network.py#L60-L62

I don't think we did that before, I think every time this event fired it was value True/False since it never changed back to None. I don't know who listens to this event so I can't check if that possibility is handled.

pvizeli (commented):
Fine for doing this, but we have to listen also to the dbus events (that property change) if the network go away and fire an event that we later use on Core for disable cloud integration

supervisor/host/network.py
pvizeli: I think with this PR, we can support that we not support to disable that. We supported disable it because of old NM version from debian 9 or 10

supervisor/supervisor.py
```diff
@@ -244,6 +246,18 @@ async def repair(self):
             _LOGGER.error("Repair of Supervisor failed")
 
     async def check_connectivity(self):
+        """Check the connection."""
+        if self.connectivity:
+            await self._check_connectivity_throttled()
+        else:
+            await self._check_connectivity()
```
pvizeli, line 253: Let's not do that, we created functions just to avoid adding a job feature. Let's add this as a feature to the job decoration to skip throttling on a variable

mdegat01:
> Fine for doing this, but we have to listen also to the dbus events (that property change) if the network go away and fire an event that we later use on Core for disable cloud integration

So it's the cloud integration that listens for that event? Ok good to know, will check and make sure it's ok with that change.

supervisor/supervisor.py
mdegat01: Allowed someone to pass in a lambda that returns a timedelta instead of a timedelta for `throttle_period`. Thought that makes it pretty clean, let me know what you think.

supervisor/host/network.py
mdegat01: Ah ok so you're thinking add a check for this property. Yea I can do that

mdegat01: Ok I added an evaluation for disabling the NM connectivity check @pvizeli

pvizeli (approved):

Issue #2564, opened 2021-02-13
RFC: Mount Manager


pvizeli:
## Context

We should not relay on external mounts but in some cases it could be nice to have it like for backups or media.

## Decision

We add udisks as a requirement for OS and find a suitable way to handle external mounts.

denilsonsa:
For completeness, here's a list of use-cases for having a network storage mounted in HA OS: https://community.home-assistant.io/t/nas-mount-on-hass-os/320218/15

pvizeli:
I think a developer who will implement this is more usefully :)

ukrolelo:
I don't know why denilsonsa post is market as spam. Just for future reference:
**@denilsonsa
Denilson Sá Maia**

**_Use-cases for NAS mount (SMB/CIFS or NFS) on Home Assistant OS:_**

- Having an external Media source. (After all, a local /media directory is not good enough for people running the OS on SD card on Raspberry Pi or Odroid.)
- Having a way to store stuff from downloader. (e.g. Saving the videos captured by your Ring Door Bell, which is an example in the official HA documentation)
- Having a proper location for NVR tools (Network Video Recorder), such as Frigate.
- Having MariaDB, InfluxDB (or Graphite, or Prometheus) running inside HA OS, but storing the long-term data onto a NAS.
- Having all these large files outside the HA snapshots (backups), which will help making the snapshots much smaller and faster. (see this issue in influxdb add-on)
- Having fewer writes to the SD card can make it last longer, offloading the writes to a dedicated NAS system that is optimized for storage.
**There are two main needs:**
- Mount stuff to be available to the Home Assistant Core instance (running inside HA OS).
- Mount stuff to be available to Add-ons (also running inside HA OS).

There is also the problem with the credentials. Where should those be stored? I trust HA developers can figure out a good solution.

ukrolelo:
> I think a developer who will implement this is more usefully :)

So we are talking about autofs? The credentials would be in a files under  /etc/ ?

pvizeli:
We need someone wich to have the time to implement uddisk dbus support to the supervisor and future a concept of how the mount points can manage without break functionality and platform independence. If you are a python developer and will doing it, then I will leaf your comments, otherwise, it get removed until someone comes who start to implement this and need this issue for useful communication and track the work

Tuckie:
Really interested in this for Frigate!

ukrolelo:
[...]

PR #3848, opened 2022-09-08
Add udisks2 dbus support

raman325:
## Proposed change
Closes https://github.com/home-assistant/supervisor/issues/2564

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR is related to issue: https://github.com/home-assistant/supervisor/issues/2564
- Link to cli pull request:

## Checklist


supervisor/dbus/udisks2/__init__.py
```diff
@@ -0,0 +1,79 @@
...
+            devices.append(device)
+        return devices
+
+    @dbus_connected
+    async def get_filesystems(self) -> list[UDisks2Filesystem]:
+        """Return list of all block devices containing a mountable filesystem."""
+        filesystem_devices: list[UDisks2Filesystem] = []
+        for block_device in await self.dbus.Manager.GetBlockDevices():
+            filesystem_device = UDisks2Filesystem(block_device)
+            await filesystem_device.connect()
```
mdegat01, line 67:
```suggestion
        for block_device in await self.dbus.Manager.call_get_block_devices():
            filesystem_device = UDisks2Filesystem(block_device)
            await filesystem_device.connect(self.dbus.bus)
```
Similar question to `get_block_devices` for error handling. Do you want one failure to fail the whole thing or should it process what it can?

supervisor/dbus/udisks2/block.py
```diff
@@ -0,0 +1,244 @@
...
+
+    def __init__(self, object_path: str) -> None:
+        """Initialize NetworkConnection object."""
+        self.object_path = object_path
+        self.properties = {}
+
+    async def connect(self):
+        """Connect to D-Bus."""
+        self.dbus = await DBus.connect(DBUS_NAME_UDISKS2, self.object_path)
+        await self.update()
```
mdegat01, line 137:
```suggestion

    bus_name: str = DBUS_NAME_UDISKS2
    properties_interface: str = DBUS_IFACE_BLOCK

    def __init__(self, object_path: str) -> None:
        """Initialize NetworkConnection object."""
        self.object_path = object_path
```

supervisor/dbus/udisks2/block.py
```diff
@@ -0,0 +1,244 @@
...
+        return self.properties[DBUS_ATTR_DEVICE_NUMBER]
+
+    @property
+    @dbus_property
+    def configuration(
+        self,
+    ) -> dict[str, list[FstabConfigDetails | CrypttabConfigDetails]]:
+        """Return device configuration."""
+        configuration = defaultdict(list)
+        for type_, details in self.properties[DBUS_ATTR_CONFIGURATION]:
```
mdegat01, line 178:
```suggestion
        for type_, details in self.properties[DBUS_ATTR_CONFIGURATION].items():
```

supervisor/dbus/udisks2/block.py
```diff
@@ -0,0 +1,244 @@
...
+        type_: str,
+        details: FstabConfigDetails | CrypttabConfigDetails,
+        options: dict[str, Any] = None,
+    ):
+        """Add new configuration item."""
+        if not options:
+            options = {}
+        await self.dbus.Block.AddConfigurationItem(
+            ("(sa{sv})", (type_, details.to_dict())), ("a{sv}", options)
+        )
```
mdegat01, line 204:
```suggestion
        await self.dbus.Block.call_add_configuration_item(
            (type_, details.to_dict()), options
        )
```
Good news and bad news. Good news is the high-level dbus-next client adds the signature to inputs for you. Bad news, the input signature includes `a{sv}`. That means you need to send a dictionary that looks like this:
```
{
  "<key: string>": <value: Variant(signature, real_value: Any>
}
```
Take a look at how we update network settings to see what I mean:
https://github.com/home-assistant/supervisor/blob/cd10b597dd1f34fd044f76450e0ef706415010c0/supervisor/dbus/network/setting/__init__.py#L156

We generate the dictionary that gets passed to that call here: https://github.com/home-assistant/supervisor/blob/main/supervisor/dbus/network/setting/generate.py . So we need to know more about what actually goes in these dictionaries to get the signatures right for each item in the dictionary.

supervisor/dbus/udisks2/block.py
```diff
@@ -0,0 +1,244 @@
...
+        type_: str,
+        details: FstabConfigDetails | CrypttabConfigDetails,
+        options: dict[str, Any] = None,
+    ):
+        """Remove existing configuration item."""
+        if not options:
+            options = {}
+        await self.dbus.Block.RemoveConfigurationItem(
+            ("(sa{sv})", (type_, details.to_dict())), ("a{sv}", options)
+        )
```
mdegat01, line 219:
```suggestion
        await self.dbus.Block.call_remove_configuration_item(
            (type_, details.to_dict()), options
        )
```
Same comment as above, we have to get variants and signatures right in each item in this dictionary, can't just pass normal python dictionaries.

supervisor/dbus/udisks2/block.py
```diff
@@ -0,0 +1,244 @@
...
+        options: dict[str, Any] = None,
+    ):
+        """Add new configuration item."""
+        if not options:
+            options = {}
+        await self.dbus.Block.AddConfigurationItem(
+            ("(sa{sv})", (old_type, old_details.to_dict())),
+            ("(sa{sv})", (new_type, new_details.to_dict())),
+            ("a{sv}", options),
+        )
```
mdegat01, line 238:
```suggestion
        await self.dbus.Block.call_add_configuration_item(
            (old_type, old_details.to_dict()),
            (new_type, new_details.to_dict()),
            options,
        )
```

supervisor/dbus/udisks2/filesystem.py
```diff
@@ -0,0 +1,46 @@
...
+from ..utils import dbus_connected
+from .block import UDisks2Block
+
+
+class UDisks2Filesystem(UDisks2Block):
+    """Handle D-Bus interface for UDisks2 filesystem device object.
+
+    http://storaged.org/doc/udisks2-api/latest/gdbus-org.freedesktop.UDisks2.Filesystem.html
+    """
+
```
mdegat01, line 15:
```suggestion

    properties_interface: str = DBUS_IFACE_FILESYSTEM

```

supervisor/dbus/udisks2/filesystem.py
```diff
@@ -0,0 +1,46 @@
...
+    @property
+    @dbus_property
+    def size(self) -> int:
+        """Return size."""
+        return self.properties[DBUS_ATTR_SIZE]
+
+    @dbus_connected
+    async def mount(self, options: dict[str, Any]):
+        """Mount filesystem."""
+        await self.dbus.Filesystem.Mount(("a{sv}", options))
```
mdegat01, line 34:
```suggestion
        await self.dbus.Filesystem.call_mount(options)
```
Similar to comments above, to use a dictionary as an input each value must be a `Variant` with the right signature.

supervisor/dbus/udisks2/filesystem.py
```diff
@@ -0,0 +1,46 @@
...
+
+    @dbus_connected
+    async def unmount(self, options: dict[str, Any]):
+        """Unmount filesystem."""
+        await self.dbus.Filesystem.Unmount(("a{sv}", options))
```
mdegat01, line 39:
```suggestion
       await self.dbus.Filesystem.call_unmount(options)

File: supervisor/dbus/udisks2/filesystem.py [hunk 10]
```diff
@@ -0,0 +1,46 @@
...
+
+    @dbus_connected
+    async def update(self):
+        """Update Properties."""
+        self.properties = {}
+        await super().update()
+        self.properties.update(await self.dbus.get_properties(DBUS_IFACE_FILESYSTEM))
```
mdegat01, line 46:
```suggestion
```
The `DBusInterfaceProxy` just handles this now.

supervisor/dbus/udisks2/__init__.py
```diff
@@ -0,0 +1,79 @@
...
+            # If we get a key error, this block device is not a filesystem device
+            try:
+                assert filesystem_device.mountpoints
+            except KeyError:
+                continue
```
mdegat01, line 72:
```suggestion
           if filesystem_device.mountpoints is None:
               continue
```
You'll never get a `KeyError` here because `mountpoints` is annoted with `dbus_property`:
https://github.com/home-assistant/supervisor/blob/cd10b597dd1f34fd044f76450e0ef706415010c0/supervisor/dbus/interface.py#L12-L22

Alternatively, can remove the annotation.

requirements.txt
```diff
@@ -23,3 +23,4 @@ securetar==2022.2.0
 sentry-sdk==1.9.8
 voluptuous==0.13.1
 dbus-next==0.2.3
+typing_extensions==4.3.0
```
mdegat01, line 26: What is this needed for?

mdegat01 (changes requested):
Lots of changes due to #3872 , #3860 and #3862 . I will work on this as its basically just a rebase+ and should be straightforward.

Other significant thing is every time you have a dictionary as an input every value in the dictionary needs to be a `dbus_next.Variant` type object with the correct signature. That may require more research with the particular interface/method.

supervisor/dbus/udisks2/block.py
mdegat01: Also realized this should be `call_update_configuration_item`. Was very confused at the two different signatures, had never seen that before.

supervisor/dbus/udisks2/__init__.py
```diff
@@ -0,0 +1,79 @@
...
+
+    @dbus_connected
+    async def get_block_devices(self) -> list[UDisks2Block]:
+        """Return list of all block devices."""
+        devices: list[UDisks2Block] = []
+        for block_device in await self.dbus.Manager.GetBlockDevices():
+            device = UDisks2Block(block_device)
+            await device.connect()
+            devices.append(device)
+        return devices
```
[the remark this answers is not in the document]
mdegat01: Suggested changes are accounted for. Leaving open for thoughts on the error handling aspect.

supervisor/dbus/udisks2/__init__.py
mdegat01: Suggested changes are accounted for. Leaving open for thoughts on the error handling aspect.

supervisor/dbus/udisks2/block.py
mdegat01: Oh wait this is my bad. This isn't a dictionary. Undoing this, sorry about that.

supervisor/dbus/udisks2/block.py
mdegat01: Suggested changes are accounted for. Leaving open because we need to adjust the dictionaries to have a `Variant` as the value for each key.

supervisor/dbus/udisks2/block.py
mdegat01: Suggested changes are accounted for. Leaving open because we need to adjust the dictionaries to have a `Variant` as the value for each key.

supervisor/dbus/udisks2/block.py
mdegat01: Suggested changes are accounted for. Leaving open because we need to adjust the dictionaries to have a `Variant` as the value for each key.

supervisor/dbus/udisks2/filesystem.py
mdegat01: Suggested changes are accounted for. Leaving open because we need to adjust the dictionaries to have a `Variant` as the value for each key.

supervisor/dbus/udisks2/filesystem.py
mdegat01: Suggested changes are accounted for. Leaving open because we need to adjust the dictionaries to have a `Variant` as the value for each key.

mdegat01:
Looking at the typed dictionaries in Block we could add something generic to the dbus utils that looks at a dictionary and converts it into one with `Variant` key values. It wouldn't be difficult to handle those since all the values are just `str` or `int` from the looks of it.

But I don't know what can go in `options` in these various method calls.

supervisor/dbus/udisks2/block.py
```diff
@@ -0,0 +1,231 @@
...
+        details: FstabConfigDetails | CrypttabConfigDetails,
+        options: dict[str, Any] = None,
+    ):
+        """Remove existing configuration item."""
+        if not options:
+            options = {}
+        await self.dbus.Block.call_remove_configuration_item(
+            (type_, details.to_dict()), options
+        )
+        await self.update()
```
mdegat01, line 212: You may not need these `update` calls anymore. After #3872 all instances of `DBusInterfaceProxy` listen for property changes signals automatically and update their properties cache on change. So we should test this. If it fires that signal on change of properties after these method calls then calling `self.update` is redundant.

requirements.txt
raman325: it is needed to import `Required` in the block module which is needed in order to create a TypedDict with a dash in the key

supervisor/dbus/udisks2/__init__.py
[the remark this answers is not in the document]
raman325: at a minimum logging the error would make the most sense to me. The error would only happen during the connect right?

supervisor/dbus/udisks2/block.py
raman325: I believe this is resolved. Please review and resolve if you agree

raman325:
> Looking at the typed dictionaries in Block we could add something generic to the dbus utils that looks at a dictionary and converts it into one with `Variant` key values. It wouldn't be difficult to handle those since all the values are just `str` or `int` from the looks of it.
> 
> But I don't know what can go in `options` in these various method calls.

this is a good idea, and options use basic types too (see my updates) but I think it makes more sense to do in a separate PR and think about it more holistically. There are few enough options here that it was simple to do it by hand

supervisor/dbus/udisks2/__init__.py
```diff
@@ -0,0 +1,99 @@
...
+        """Return list of supported filesystems."""
+        return self.properties[DBUS_ATTR_SUPPORTED_FILESYSTEMS]
+
+    @dbus_connected
+    async def get_block_devices(self) -> list[UDisks2Block]:
+        """Return list of all block devices."""
+        devices: list[UDisks2Block] = []
+        for block_device in await self.dbus.Manager.call_get_block_devices({}):
+            device = UDisks2Block(block_device)
+            await device.connect(self.dbus.bus)
```
mdegat01, line 84: Based on the current code, a signal listener is going to be set up for each of these block devices (and each of the filesystems below). But these appear to be transient, you pull info on each device in the list, return it and then the objects are tossed.

We should make one of the following changes:

1. Pull the list of block devices and filesystems and block devices during `update` and cache those lists. Return the cached list here and during future updates call `connect` on new devices/filesystems, `update` on existing ones and `disconnect` on ones that have disappeared. See here for an example of that approach:
https://github.com/home-assistant/supervisor/blob/30932a83f8dcbb6261823b468aeb6e1d28b4a1cd/supervisor/dbus/network/__init__.py#L177-L227
 
2. Add `sync_properties = false` to the class fields of `UDisks2Block` and `UDisks2Filesystem` so the signal listeners are not set up on `connect`. Then you can freely make and toss these devices every time this method is called. Like what we do for access points currently:
https://github.com/home-assistant/supervisor/blob/30932a83f8dcbb6261823b468aeb6e1d28b4a1cd/supervisor/dbus/network/wireless.py#L54-L65

But we need to do one of those two otherwise we risk leaking the listeners when the objects are garbage collected.

supervisor/dbus/udisks2/block.py
```diff
@@ -0,0 +1,262 @@
...
+        self._additional_interfaces: dict[str, DBusInterfaceProxy] = {}
+        super().__init__()
+
+    async def connect(self, bus: MessageBus) -> None:
+        """Connect to bus."""
+        await super().connect(bus)
+        for key in list(set(self.dbus.proxies) - set(ADDITIONAL_INTERFACES)):
+            iface = self.additional_interfaces[key] = ADDITIONAL_INTERFACES[key](
+                self.object_path
+            )
```
mdegat01, line 154: I'm really confused by this. `ADDITIONAL_INTERFACES` seems to only have one key, `org.freedesktop.UDisks2.Filesystem`. But when I introspect one of these objects it has a bunch of other interfaces including `org.freedesktop.DBus.Properties`, `org.freedesktop.DBus.Introspectable`, etc. Seems like all those are just going to result in a `KeyError` unless I'm missing something?

Did you mean to loop over this by any chance:
```
list(set(self.dbus.proxies).intersection(set(ADDITIONAL_INTERFACES)))
```
And just process the set of interfaces that is in both of these sets? Instead of the ones that are exclusively in `dbus.proxies`?

supervisor/dbus/udisks2/block.py
```diff
@@ -0,0 +1,262 @@
...
+            await iface.connect(bus)
+
+    @property
+    def additional_interfaces(self) -> dict[str, DBusInterfaceProxy]:
+        """Return additional interfaces."""
+        return self._additional_interfaces
```
mdegat01, line 160: I get what you're doing here but it might be easier to understand this interface if we make properties for the specific interfaces. Like a `filesystem` property that returns the filesystem `DBusInterfaceProxy`. I'm fine with the `_additional_interfaces` setup internally just seems like this class would be easier to understand with specific typed properties rather then a dictionary that requires the caller know the exact dbus interface.

mdegat01 (changes requested):

raman325:
Not stale

supervisor/dbus/udisks2/block.py
```diff
@@ -0,0 +1,262 @@
...
+        "name": Required[bytearray],
+        "device": Required[bytearray],
+        "passphrase-path": Required[bytearray],
+        "passphrase-contents": Required[bytearray],
+        "options": Required[bytearray],
+    },
+)
+
+
+@dataclass
```
pvizeli, line 94: we should slot all dataclass that can be sloted

PR #3854, opened 2022-09-09
Dbus refactor and tests

mdegat01:
## Proposed change

A number of changes are planned to our dbus implementation including:
1. Switch to new fork [dbus-fast](https://github.com/Bluetooth-Devices/dbus-fast)
2. Refactor to use dbus-next proxy interfaces instead of direct calls to `MessageBus.call`
3. Live property updates via signals instead of scheduled calls to `get_properties`

Before beginning any of this I wanted to do a sweep of the existing code. This adds a bunch of tests for what we have along with some cleanup. The vast majority of the cleanup is non-functional with a few exceptions I'll call out (since I know this PR is large).

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist


supervisor/api/host.py
```diff
@@ -120,25 +120,25 @@ async def services(self, request):
     def service_start(self, request):
         """Start a service."""
         unit = request.match_info.get(SERVICE)
-        return asyncio.shield(self.sys_host.services.start(unit))
+        return [asyncio.shield(self.sys_host.services.start(unit))]
```
mdegat01, line 123: Whenever possible I adjusted our dbus interfaces to return the actual output of the dbus method and only return an array if there is actually multiple outputs (or if the output itself is an array obviously). Whereas today they frequently returned an `Awaitable` which was the output wrapped in a list.

However the response of these 4 methods:
https://github.com/home-assistant/supervisor/blob/01e24a3e74c54c4b93e4c87196d1a5ebf8d0fb42/supervisor/dbus/systemd.py#L76-L106

Directly became response of the corresponding REST APIs. So I re-wrapped it in an array here for backwards compatibility.

Is that actually necessary though? I have no idea who or what uses these APIs but I can't imagine they care about the output. It's a systemd job ID that looks like this: `/org/freedesktop/systemd1/job/7623`. To my knowledge there is nothing a user can do with this response. There's no API they can pass this identifier to get more info, no place they can look up more info about it, they can't take the response from `start_service` and pass it `restart_service`, etc.

I was being careful but I feel like we can drop the array. Maybe we shouldn't return a response at all? Unless I'm missing something.

supervisor/dbus/network/settings.py
```diff
@@ -29,11 +31,16 @@ async def connect(self) -> None:
...
-    def add_connection(self, settings: Any) -> Awaitable[Any]:
+    async def add_connection(self, settings: Any) -> NetworkSetting:
         """Add new connection."""
-        return self.dbus.Settings.AddConnection(("a{sa{sv}}", settings))
+        obj_con_setting = (
+            await self.dbus.Settings.AddConnection(("a{sa{sv}}", settings))
+        )[0]
+        con_setting = NetworkSetting(obj_con_setting)
+        await con_setting.connect()
+        return con_setting
```
mdegat01, line 41: This seemed more consistent with how we handle settings elsewhere. However the only caller of this method today is here:
https://github.com/home-assistant/supervisor/blob/01e24a3e74c54c4b93e4c87196d1a5ebf8d0fb42/supervisor/host/network.py#L221

Notice it does not use the output. So this does create a new situation where we added the setting without error but got an error on `connect`. We wouldn't have encounted this before. I'm not sure how this could possibly occur but if somehow we can't connect to the new settings that feels like a problem we should report.

supervisor/dbus/network/wireless.py
```diff
@@ -32,14 +35,23 @@ def active(self) -> NetworkWirelessAP | None:
...
+        accesspoints_data = (await self.dbus.Device.Wireless.GetAllAccessPoints())[0]
+        accesspoints = [NetworkWirelessAP(ap_obj) for ap_obj in accesspoints_data]
+
+        for err in await asyncio.gather(
+            *[ap.connect() for ap in accesspoints], return_exceptions=True
+        ):
+            if err:
+                _LOGGER.warning("Can't process an AP: %s", err)
+
+        return accesspoints
```
mdegat01, line 54: This actually isn't a functional change although it looks like one. The only caller of this method today is here:
https://github.com/home-assistant/supervisor/blob/01e24a3e74c54c4b93e4c87196d1a5ebf8d0fb42/supervisor/host/network.py#L272-L293

It works identically even down to the logged warning. Except now the processing of setting object paths into connected `NetworkWirelessAP` objects occurs here and `scan_wifi` just uses the successfully connected ones. I felt it was better to move knowledge of the dbus objects in here and just return python objects.

supervisor/dbus/timedate.py
```diff
@@ -66,20 +66,14 @@ async def connect(self):
...
     @dbus_connected
-    def set_ntp(self, use_ntp: bool):
-        """Turn NTP on or off.
-
-        Return a coroutine.
-        """
-        return self.dbus.SetNTP(use_ntp)
+    async def set_ntp(self, use_ntp: bool) -> None:
+        """Turn NTP on or off."""
+        await self.dbus.SetNTP(use_ntp, False)
```
mdegat01, line 76: I believe this function was actually broken. According to the [introspection](https://www.freedesktop.org/software/systemd/man/org.freedesktop.timedate1.html) the signature of `SetNTP` is this:
```
      SetNTP(in  b use_ntp,
             in  b interactive);
```
There are two parameters, not one. Calling it with one won't work.

We don't actually call this method anywhere in supervisor right now so that's probably how it slipped through.

supervisor/dbus/rauc.py
```diff
@@ -69,36 +70,24 @@ def boot_slot(self) -> str | None:
...
     @dbus_connected
-    def signal_completed(self):
-        """Return a signal wrapper for completed signal.
-
-        Return a coroutine.
-        """
-        return self.dbus.wait_signal(DBUS_SIGNAL_RAUC_INSTALLER_COMPLETED)
+    def signal_completed(self) -> DBusSignalWrapper:
+        """Return a signal wrapper for completed signal."""
+        return self.dbus.signal(DBUS_SIGNAL_RAUC_INSTALLER_COMPLETED)
```
mdegat01, line 85: The way we were doing this wasn't really safe. It exposed us to a race condition where the signal for `install_completed` fired before we started listening to it. I actually removed the `DBus.wait_signal` method entirely because it seems like any time we need to wait for a response by signal we should be doing it this way instead:
https://github.com/home-assistant/supervisor/blob/72292b01a24e0b362c11bd3f63aceb9d4eb2b56d/supervisor/os/manager.py#L198-L203

It also doesn't work well for continuously listening to signals since the monitor is unattached after every signal. I plan to put in a better approach for that in an upcoming PR.

pvizeli (approved):

PR #3898, opened 2022-09-26
Bad introspection causes DBus error not KeyError

mdegat01:
## Proposed change

It seems we get back some bad introspections that don't include `org.freedesktop.DBus.Properties` from devices within NetworkManager. After the latest DBus changes this causes a `KeyError` rather then a `DBusInterfaceMethod` error.

Based on this it appears this is expected, we just need to return to raising the correct type of exception:
https://github.com/home-assistant/supervisor/blob/ff462ae9765157c8d74db724b26baa6b7f08744c/supervisor/dbus/network/__init__.py#L182-L187

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/dbus/interface.py
+++ b/supervisor/dbus/interface.py
@@ -58,7 +58,7 @@ class DBusInterfaceProxy(DBusInterface):
         await super().connect(bus)
         await self.update()
 
-        if self.sync_properties:
+        if self.sync_properties and self.is_connected:
             self.dbus.sync_property_changes(self.properties_interface, self.update)
 
     @dbus_connected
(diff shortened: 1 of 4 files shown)


supervisor/dbus/interface.py
```diff
@@ -58,7 +58,7 @@ async def connect(self, bus: MessageBus) -> None:
         await super().connect(bus)
         await self.update()
 
-        if self.sync_properties:
+        if self.sync_properties and self.is_connected:
```
mdegat01, line 61: Not really related but noticed this in sentry as well. There's a possible race condition where disconnect is called before we get here and then dbus is none. This check should prevent that.

pvizeli (approved):

PR #3969, opened 2022-10-24
Run initialization and Supervisor launch as root

agners:
## Proposed change

It seems the devcontainer runs as vscode user by default today. Use sudo to run the devcontainer initialization and Supervisor launch task as root.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -2,7 +2,7 @@
   "name": "Supervisor dev",
   "image": "ghcr.io/home-assistant/devcontainer:supervisor",
   "appPort": ["9123:8123", "7357:4357"],
-  "postCreateCommand": "bash devcontainer_bootstrap",
+  "postCreateCommand": "sudo bash devcontainer_bootstrap",
   "runArgs": ["-e", "GIT_EDITOR=code --wait", "--privileged"],
   "extensions": [
     "ms-python.python",
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -2,7 +2,7 @@
   "name": "Supervisor dev",
   "image": "ghcr.io/home-assistant/devcontainer:supervisor",
   "appPort": ["9123:8123", "7357:4357"],
-  "postCreateCommand": "sudo -u root -g root bash devcontainer_bootstrap",
+  "postCreateCommand": "bash devcontainer_bootstrap",
   "remoteUser": "vscode",
   "runArgs": ["-e", "GIT_EDITOR=code --wait", "--privileged"],
   "extensions": [
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -2,7 +2,8 @@
   "name": "Supervisor dev",
   "image": "ghcr.io/home-assistant/devcontainer:supervisor",
   "appPort": ["9123:8123", "7357:4357"],
-  "postCreateCommand": "sudo bash devcontainer_bootstrap",
+  "postCreateCommand": "sudo -u root -g root bash devcontainer_bootstrap",
+  "remoteUser": "vscode",
   "runArgs": ["-e", "GIT_EDITOR=code --wait", "--privileged"],
   "extensions": [
     "ms-python.python",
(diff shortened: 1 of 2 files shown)


.devcontainer/devcontainer.json
```diff
@@ -2,7 +2,7 @@
   "name": "Supervisor dev",
   "image": "ghcr.io/home-assistant/devcontainer:supervisor",
   "appPort": ["9123:8123", "7357:4357"],
-  "postCreateCommand": "bash devcontainer_bootstrap",
+  "postCreateCommand": "sudo bash devcontainer_bootstrap",
```
ludeeus, line 5: I think we should do this instead <https://github.com/home-assistant/supervisor/pull/3967>

.vscode/tasks.json
```diff
@@ -4,7 +4,7 @@
     {
       "label": "Run Supervisor",
       "type": "shell",
-      "command": "supervisor_run",
+      "command": "sudo supervisor_run",
```
ludeeus, line 7: All tasks need root, not just this one.

ludeeus (changes requested):

.vscode/tasks.json
agners: Why? That is kinda my point, I'd like them to not as root so the permissions stay proper on my host.

All of them seem to run fine on my system, except `Run Supervisor CLI`, that one needs root since we don't allow Docker access without root, it seems.

ludeeus: That one and frontend needs it, the others I did not see

.devcontainer/devcontainer.json
ludeeus: Then adjust the user groups, don't just slap on a "sudo".
Also set the user directly, as that is what has changed now, so that does not give more issues in the future.

agners:
@ludeeus I understand that my PR goes a different direction here. Your PR would really be the fix to restore the old behavior.

But the old behavior is something which annoyed me quite a bit in the past (because of that permissions mix-up). That said, if there are good reasons to run as root by default, I am not opposed... I'd just prefer if we would try to give it a go with user rights in the dev container :sweat_smile: 

@pvizeli @mdegat01 thoughts?

.vscode/tasks.json
agners: Right, that one starts the Docker daemon and container as well. It does other stuff too though :sweat: 

That might be not as straight forward as I thought.

felipecrs:
+1 to not run the devcontainer as root. This is discouraged by VS Code itself. The drawbacks were mentioned already (root writing back files in a repository owned by some other user).

mdegat01 (approved):
Tested the tasks and seems to work. I'm fine with this approach. I also prefer not running containers as root when possible and as noted it is VSCode's recommended best practice.

.devcontainer/devcontainer.json
felipecrs: Honestly replacing `sudo` with `sudo -u root -g root` sounds like a big non-op to me.

besanur:
this way #3976 not works... only @ludeeus solution worked for me with root in remoteUser

plloppii:
Just ran into this and read through the comments. 
Slapping `sudo` and using `root` all sound like bandage fixes. 
As what [ludeeus](https://github.com/ludeeus) eluded to, is there anyway we can change the usergroup permissions for user, vscode?

agners:
> Slapping sudo and using root all sound like bandage fixes.

Some of the commands we run need to run privileged. So it's either running as root or adding sudo...

IMHO, running only what really requires root with sudo is the better approach here (as this PR implements).

.devcontainer/devcontainer.json
agners: > Honestly replacing `sudo` with `sudo -u root -g root` sounds like a big no-op to me.

Right, I interpreted @ludeeus message wrong above. I reverted that change.

Instead, this change now relies on [#32](https://github.com/home-assistant/devcontainer/pull/32), which only selectively uses `sudo`. The current implementation creates `test_supervisor` still root owned (as the actual Supervisor will anyways create things as root). But since that is not part of the git repository, its not problematic.

agners:
With rootless support for the supervisor scripts as provided by the devicontainer repository (see https://github.com/home-assistant/devcontainer/pull/32), this PR basically becomes a noop.

PR #4152, opened 2023-02-20
Allows the supervisor to send a session's user to addon with header X-Remote-User

baflo:
## Related PRs
https://github.com/home-assistant/core/pull/88472
https://github.com/home-assistant/frontend/pull/15505
https://github.com/home-assistant/supervisor/pull/4152

## Proposed change

This implements a feature that allows the supervisor to send a session's username to an addon. It stores the name of a user that created a session in session data accompanying the session ID.

Everytime a user uses the ingress API, the supervisor checks whether the addon is allowed to receive the username (from addon user config) and if it is, it adds the header X-Remote-User with the username.

So the addons can make of the home assistant authentication for a kind of single-sign-on.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to documentation pull request: https://github.com/home-assistant/developers.home-assistant/pull/1691
- Link to cli pull request:

## Checklist


pvizeli (commented):
We not have to make it as add-on option. The header can't produce a conflict. So let's just send it per default if we know the user. It have to be backwards compatible

baflo:
I thought of it as of a privacy feature. Currently addons don't know _which_ ist is authenticated. Might be, that some addons really ob that fact. 
That's why I kept the compatible anonymous way as default.

pvizeli:
Does it still work with the ID? Do we need to expand the /auth API to lookup ID to username?

baflo:
I changed it to send the user's ID now.
The username could be requested using the auth API.

baflo:
> Does it still work with the ID? Do we need to expand the /auth API to lookup ID to username?
> 
> We could allow to config users IDs that allow to use the add-on ingress for privacy. But I think that is over complicated right now

So do you suggest I just strip down the PR to the functionality without measures for configuration? I would
- remove the config.yaml option `remote_user_requested`
- remove the UserAddonOption `remote_user` 
- remove respective checks and code
- close my PR for the frontend

I would keep:
- The new config SCHEMA_INGRESS_CONFIG_SESSION_DATA which should be passed as payload on session creation. Currently, it only expects the user-ID, but it is extensible.
- Always adding the HEADER_REMOTE_USER_ID to the request to the addon, if the user-ID is given

pvizeli:
Yes.

Additionally, Supervisor could ask for the current username on core and send the username with X-Remote-User (it will be easier for add-on developer to match on the username without additional logic. It makes total sense that core only send the ID for create the session, but we can do something different for the add-on.

For now, I would just fetch the username during the create session, later we can see what we can improve with cache to update the username if needed (but ingress session are mostly short-living, I would not see an issue with that).

supervisor/const.py
```diff
@@ -270,7 +271,11 @@
 ATTR_SERVERS = "servers"
 ATTR_SERVICE = "service"
 ATTR_SERVICES = "services"
+ATTR_REMOTE_USER = "remote_user"
+ATTR_REMOTE_USER_REQUESTED = "remote_user_requested"
 ATTR_SESSION = "session"
+ATTR_SESSION_DATA = "session_data"
+ATTR_SESSION_DATA_USER_ID = "user-id"
```
pvizeli, line 278:
```suggestion
ATTR_SESSION_DATA_USER_ID = "user_id"
```

same as in core, we use `_` for json body data

baflo:
> Yes.
> 
> Additionally, Supervisor could ask for the current username on core and send the username with X-Remote-User (it will be easier for add-on developer to match on the username without additional logic. It makes total sense that core only send the ID for create the session, but we can do something different for the add-on.
> 
> For now, I would just fetch the username during the create session, later we can see what we can improve with cache to update the username if needed (but ingress session are mostly short-living, I would not see an issue with that).

Do you have a idea how to quickly fetch the username based on the user ID?
As far as I can see there's no API to get the specific user by its ID (there's a WS-API though to get the list of users).

pvizeli:
> Do you have a idea how to quickly fetch the username based on the user ID? As far as I can see there's no API to get the specific user by its ID (there's a WS-API though to get the list of users).

Yes, over the WS-API. You can create a function kinda get_users and add a throttle job condition for (I don't know, 5min) and fetch it from websocket. I would do that on session creation and store it in the session data as well for now. We can later improve this.

tests/api/test_ingress.py
```diff
@@ -20,7 +20,7 @@ async def test_validate_session(api_client, coresys):
         "aiohttp.web_request.BaseRequest.__getitem__",
         return_value=coresys.homeassistant,
     ):
-        resp = await api_client.post("/ingress/session")
+        resp = await api_client.post("/ingress/session", json={"user-id": "some-id"})
```
pvizeli, line 23: please make also a new test without user_id for make sure old core version still works for now

supervisor/ingress.py
```diff
@@ -35,6 +35,11 @@ def sessions(self) -> dict[str, float]:
         """Return sessions."""
         return self._data[ATTR_SESSION]
 
+    @property
+    def sessions_data(self) -> dict[str, float]:
```
pvizeli, line 39: let's parse the session data from a dict into a DataClass (OOP Style) -> i.e. https://github.com/home-assistant/supervisor/blob/main/supervisor/discovery/__init__.py#L29 but today we can use data class (https://docs.python.org/3/library/dataclasses.html)

baflo:
@pvizeli I updated the PR to use dataclasses

supervisor/validate.py
```diff
@@ -180,10 +181,24 @@ def validate_repository(repository: str) -> str:
 SCHEMA_AUTH_CONFIG = vol.Schema({sha256: sha256})
 
 
+@dataclass
+class UserInfo:
```
pvizeli, line 185: Let's only add things that we now need. Move this also into const.py and we should call it IngressSessionData.

supervisor/validate.py
```diff
@@ -193,7 +208,7 @@ def validate_repository(repository: str) -> str:
             {token: vol.Coerce(float)}
         ),
         vol.Required(ATTR_SESSION_DATA, default=dict): vol.Schema(
-            {token: SCHEMA_INGRESS_CONFIG_SESSION_DATA}
+            {token: vol.Schema({ATTR_SESSION_DATA_USER: vol.Any()})}
```
pvizeli, line 211: We have alway to verify it correctly and not as `Any`

supervisor/api/ingress.py
```diff
@@ -42,6 +42,11 @@
 class APIIngress(CoreSysAttributes):
     """Ingress view to handle add-on webui routing."""
 
+    def __init__(self, coresys: CoreSys) -> None:
```
pvizeli, line 45: Let's not change the style of our API views. It will work also if we not set coresys over init

baflo:
@pvizeli Took a while until I had time to work on this again... but now there's an update 😊

supervisor/api/ingress.py
```diff
@@ -71,7 +85,19 @@ async def panels(self, request: web.Request) -> dict[str, Any]:
     @require_home_assistant
     async def create_session(self, request: web.Request) -> dict[str, Any]:
         """Create a new session."""
-        session = self.sys_ingress.create_session()
+        schema_ingress_config_session_data = await api_validate(
+            SCHEMA_INGRESS_CREATE_SESSION_DATA, request
+        )
+        data: IngressSessionData = None
```
pvizeli, line 91:
```suggestion
        data: IngressSessionData | None = None
```

supervisor/ingress.py
```diff
@@ -97,12 +112,14 @@ def _update_token_list(self) -> None:
         for addon in self.addons:
             self.tokens[addon.ingress_token] = addon.slug
 
-    def create_session(self) -> str:
+    def create_session(self, data: IngressSessionData = None) -> str:
```
pvizeli, line 115:
```suggestion
    def create_session(self, data: IngressSessionData | None = None) -> str:
```

supervisor/validate.py
```diff
@@ -176,18 +179,40 @@ def validate_repository(repository: str) -> str:
 
 SCHEMA_AUTH_CONFIG = vol.Schema({sha256: sha256})
 
+"""Expected optional payload of create session request"""
+SCHEMA_INGRESS_CREATE_SESSION_DATA = vol.Schema(
+    {
+        vol.Optional(ATTR_SESSION_DATA_USER_ID): str,
+    }
+)
```
pvizeli, line 187: This can be directly on top of api/ingress.py - See i.e. api/supervisor.py

supervisor/ingress.py
```diff
@@ -35,6 +41,11 @@ def sessions(self) -> dict[str, float]:
         """Return sessions."""
         return self._data[ATTR_SESSION]
 
+    @property
+    def sessions_data(self) -> dict[str, float]:
```
pvizeli, line 45: typing is wrong

supervisor/api/ingress.py
```diff
@@ -99,21 +125,26 @@ async def handler(
         # Process requests
         addon = self._extract_addon(request)
         path = request.match_info.get("path")
+        session_data = self.sys_ingress.sessions_data[session]
```
pvizeli, line 128: that is wrong, this only return a dict. I think you should create an ``sys_ingress.get_data(session_id: str) -> SessionIngressData` and trow and NotFound exception (there are some others at exception.py).

pvizeli (changes requested):

supervisor/api/ingress.py
baflo: @pvizeli You suggested (far) above, that the tests should be compatible with older versions of home assistant where the user_id is not sent with session creation https://github.com/home-assistant/supervisor/pull/4152#discussion_r1124163083.
Because in that case session_data is `None`, I left this as it is and changed the signature of `_handle_websocket/_handle_request` to expect a `None` IngressSessionData.

pvizeli: That is fine, we can suppress the error until we force it

pvizeli (approved):

pvizeli:
Please address the linter issues

supervisor/api/ingress.py
```diff
@@ -217,11 +258,33 @@ async def _handle_request(
...
+    request: web.Request, addon: Addon, session_data: IngressSessionData | None
+) -> CIMultiDict | dict[str, str]:
     """Create initial header."""
     headers = {}
 
+    if session_data is not None:
+        headers[HEADER_REMOTE_USER_ID] = session_data.user.id
+        headers[HEADER_REMOTE_USER_NAME] = session_data.user.username
+        headers[HEADER_REMOTE_USER_DISPLAY_NAME] = session_data.user.display_name
+
```
pvizeli, line 287: Can you ditch those fields from the header if not needed? Otherwise we have an security issue that we forward the information as it is from the client

baflo: @pvizeli Could you give me a hin how you imagine this, please? Initially my PR had methods to activate this for every addon in the addon configuration. However, I was told to remove that. From the security perspective I could imagine:

* send only user ID (this obviously not secures the ID, but the user's name)
* enable sending the data per addon in the addon configuration
* do not send any information initially, but provide an API where the addon can retrieve user information by its session ID

baflo: Oh, I see... I misunderstood the requirement :)

pvizeli: Meh, I see now, we should remove it on lin 293 which included the headers from remote

pvizeli (approved):

supervisor/api/ingress.py
baflo: I updated it in that regard

PR #4160, opened 2023-02-24
Mock services on session dbus in tests

mdegat01:
## Proposed change

Recently we've had several events where a supervisor update had catastrophic failures despite CI passing. The reason in each case was because we don't actually use dbus in our unit tests. Instead we mock all calls to dbus and return fixtures based on naming patterns.

This setup has a few problems:
1. DBus responses aren't actually JSON-compliant. Specifically dbus has a `byte` type and can return tuples in addition to arrays, neither of which can be represented in JSON. This forces us to do things like [this](https://github.com/home-assistant/supervisor/blob/165af46f54b6f592bbd144417012bf72370c9a60/tests/conftest.py#L123-L142) or [this](https://github.com/home-assistant/supervisor/blob/165af46f54b6f592bbd144417012bf72370c9a60/tests/dbus/test_resolved.py#L30-L49) to post-process fixtures
2. There is currently no testing of signatures. Since the calls to the methods are mocked and the fixtures are based on the object and method name we do not test that the input is accurate to the signature and that the fixture response matches what was in the introspection. This is the primary source of most of the recent failures referenced above.
3. Creating scenario tests with dbus as our setup is difficult. Because our `coresys` and `dbus` fixtures mock dbus calling to always return the same responses based on bus, object and interface, its not really setup to allow testing of specific scenarios of what happens when a method returns a particular response or error (like [this](https://github.com/home-assistant/supervisor/blob/165af46f54b6f592bbd144417012bf72370c9a60/tests/dbus/network/test_network_manager.py#L116) or [this](https://github.com/home-assistant/supervisor/blob/165af46f54b6f592bbd144417012bf72370c9a60/tests/dbus/agent/boards/test_board.py#L14-L39)).
4. Occaisionally we [need the variants](https://github.com/home-assistant/supervisor/blob/165af46f54b6f592bbd144417012bf72370c9a60/tests/dbus/network/setting/test_init.py#L18-L71) in the response. This is rare but obviously there's no `Variant` type in JSON so this does not fit the fixture model.

This PR suggests a new model for DBus testing. We drop the fixtures entirely and no longer mock our methods within DBus. Instead we initialize a session dbus and use dbus-fast to create mock objects on it that match the introspections of the services we actually use. This solves all the problems above since we are truly talking over dbus in our tests, its just a local session dbus with mocked responses. It also adds a few other benefits:
1. Can test signal handling. Currently we have [this setup](https://github.com/home-assistant/supervisor/blob/165af46f54b6f592bbd144417012bf72370c9a60/tests/common.py#L30) which emulates signalling by using the name to find the listener method in the cache and calling it with the fixture. With this we can actually fire dbus signals to test handling as well as ensure things aren't handling them if we wish. And can properly test sequences, like having a call to `Rauc.Install` fire the `Completed` signal in the mock rauc service.
2. Can be confident of dbus-fast dependabot bumps. Since CI will be truly using Dbus if it passes the update should work fine. Currently each has to be manually tested since the library is essentially not used in CI.

I rewrote the tests for 5 dbus services to show how this could work. I would like to do all of them this way as well as change the `coresys` fixture to spin up a pre-specified list of these mock services and not rely on the json fixtures anymore.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -13,7 +13,7 @@ env:
   DEFAULT_CAS: v1.0.2
 
 concurrency:
-  group: '${{ github.workflow }}-${{ github.ref }}'
+  group: "${{ github.workflow }}-${{ github.ref }}"
   cancel-in-progress: true
 
 jobs:
@@ -370,7 +370,7 @@ jobs:
       - name: Install additional system dependencies
         run: |
           sudo apt-get update
-          sudo apt-get install -y --no-install-recommends libpulse0 libudev1
+          sudo apt-get install -y --no-install-recommends libpulse0 libudev1 dbus dbus-x11
       - name: Register Python problem matcher
         run: |
           echo "::add-matcher::.github/workflows/matchers/python.json"
--- a/setup.cfg
+++ b/setup.cfg
@@ -27,3 +27,5 @@ ignore =
     E203,
     D202,
     W504
+per-file-ignores =
+    tests/dbus_service_mocks/*.py: F821,F722
--- a/supervisor/resolution/validate.py
+++ b/supervisor/resolution/validate.py
@@ -6,9 +6,9 @@ import voluptuous as vol
 from ..const import ATTR_CHECKS, ATTR_ENABLED
 
 
-def get_valid_modules(folder) -> list[str]:
+def get_valid_modules(folder, *, base=__file__) -> list[str]:
     """Validate check name."""
-    module_files = Path(__file__).parent.joinpath(folder)
+    module_files = Path(base).parent.joinpath(folder)
     if not module_files.exists():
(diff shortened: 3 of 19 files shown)


mdegat01:
FYI for reviewers - to run the new unit tests locally in devcontainer you will first need to install dbus in the container like this:
```
apt-get update
apt-get install dbus dbus-x11
```
If we like this approach I'll put in a PR for devcontainer to add those packages. Until then there's a manual step.

mdegat01:
@pvizeli I'm going to say this is ready for review. It's already quite large, I don't want to do all the rewriting of dbus tests in one PR. Plus just having this in means we have some unit tests truly using dbus in the tests

pvizeli (approved):

PR #4188, opened 2023-03-13
Improve addon slug validation

mdegat01:
## Proposed change

Addon slugs currently have no validation on what can and cannot be in the string. This is a problem because
1. it shouldn't be that open-ended
1. the slug is in URLs so it needs to be URL safe

I went to our [addons analytics](https://analytics.home-assistant.io/addons.json) and came up with a regex that restricts slugs to what will work where we need it to without breaking backwards compatibility for any of the known addons.

It's mostly what you would expect (`-_a-z0-9`). Although turns out some addons used upper case letters and periods. Those probably wouldn't be included if this regex had existed from the beginning. Since it didn't and they are url safe better to include them and keep backwards compatibility.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/addons/const.py
+++ b/supervisor/addons/const.py
@@ -26,3 +26,5 @@ ADDON_UPDATE_CONDITIONS = [
     JobCondition.PLUGINS_UPDATED,
     JobCondition.SUPERVISOR_UPDATED,
 ]
+
+RE_SLUG = r"[-_.A-Za-z0-9]+"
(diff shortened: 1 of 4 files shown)


pvizeli (approved):

mdegat01:
Actually small update here in case anyone comes here from the release notes. This PR is *not* a breaking change after all. Addon slugs are used as docker container names and this is the validation used for what can be a name in docker:
https://github.com/moby/moby/blob/be97c66708c24727836a22247319ff2943d91a03/daemon/names/names.go#L5-L6

Therefore nothing outside of `[-_.A-Za-z0-9]+` ever worked. Or at least not since 2017. So this is just matching what docker was already doing for us.

PR #4222, opened 2023-03-31
Prevent multiple data disks with name hassos-data-external

mdegat01:
## Proposed change

Covers the last two situations required for doing formatting from supervisor rather then os agent during data disk migration.
1. If moving from one external data disk to another, the original is renamed to `hassos-data-external-old` prior to reboot
2. If another partition happens to have the name `hassos-data-external` which is neither the source nor the target of the migration, error and ask the user to rename or remove it

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist


supervisor/os/data_disk.py
```diff
@@ -216,6 +244,22 @@ async def migrate_disk(self, new_disk: str) -> None:
                     _LOGGER.error,
                 ) from err
 
+            # If migrating from one external data disk to another, rename the old one to prevent conflicts
```
agners, line 247: Any specific reason to do this here? I'd suggest to do it before `mark_data_move()` and even before `_format_device_with_single_partition()`... That way we got that out of the way, and if it fails, we don't enter the race...

mdegat01: oh right because only the filesystem label matters during normal startup. so even though your system will be a bit wonky (external data disk but no partition `hassos-data-external`) it will work fine. Whereas if we do the rename after and it fails then a reboot causes havoc.

Ok will re-arrange, thanks!

pvizeli (approved):

PR #4317, opened 2023-05-26
Make issue for problem with config for containers

mdegat01:
## Proposed change

Make an issue when supervisor detects that a container is running under an old config and needs to be rebuilt. Specifically in this case we're looking for containers that mount media and have propagation set to `rprivate` since they will not be able to see and use media mounts.

The issue setup is a bit odd. The check manages an issue and suggestion per component (core, addon, or plugin) which has a docker config issue. It then is actually tied to a system-context docker config issue that just shows up whenever any of the others exist. The reason its like this is so the system level one can be made into a repair that lists all the things which need to be restarted. The other issues are basically just a way of providing this list to HA.

We may want to revisit this at some point. It might be useful to have issues contain more then just a single string reference as data. For now though we go with this approach since
1. This is the first time we've had this problem/need
2. I can't rewrite resolution manager to handle this by HA beta (next week)

Issue details are basically internal so it can be changed later without creating a breaking change anyway.

Additionally this PR removes mount from the supported feature list if supervisor's container does not have propagation set correctly for its data mount. For OS, that will mean the user has to update to 10.2 to use this feature. For supervised, an update will be required to the installer to change how the container is started.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist


supervisor/resolution/fixups/system_execute_rebuild.py
```diff
@@ -0,0 +1,57 @@
...
+
+    @property
+    def issues(self) -> list[IssueType]:
+        """Return a IssueType enum list."""
+        return [IssueType.DOCKER_CONFIG]
+
+    @property
+    def auto(self) -> bool:
+        """Return if a fixup can be apply as auto fix."""
+        return False
```
ludeeus, line 57:
```suggestion
```
Not needed, `False` is the default in the base class <https://github.com/home-assistant/supervisor/blob/0fda5f6c4b8d9bdbf7bfc7b295ae5571371161a0/supervisor/resolution/fixups/base.py#L65-L68>

supervisor/resolution/checks/docker_config.py
```diff
@@ -0,0 +1,102 @@
...
+
+def setup(coresys: CoreSys) -> CheckBase:
+    """Check setup function."""
+    return CheckDockerConfig(coresys)
+
+
+class CheckDockerConfig(CheckBase):
+    """CheckDockerConfig class for check."""
+
+    async def run_check(self) -> None:
```
ludeeus, line 29: Should this be executed the first time you add a media mount over the API as well?

mdegat01: I mean tbh it will already have run in all likelihood. It's not limiting itself to when media mounts exist. Its just going to tell you that the default config has changed and the following things need a restart

ludeeus: Fair point

PR #4346, opened 2023-06-07
Add backup location to info/list API calls

ludeeus:
## Proposed change

I want to know if my backup exist on the local filesystem or in my backup mount, so I can determine what to cleanup.
Currently there is no indication in the API that say if the backup is local or remote.

This PR adds a new `"location"` key to both the list and info API calls that includes the mount name if its on a mount or `None` if it's local.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/api/backups.py
+++ b/supervisor/api/backups.py
@@ -10,6 +10,7 @@ from aiohttp import web
 from aiohttp.hdrs import CONTENT_DISPOSITION
 import voluptuous as vol
 
+from ..backups.backup import Backup
 from ..backups.validate import ALL_FOLDERS, FOLDER_HOMEASSISTANT, days_until_stale
 from ..const import (
     ATTR_ADDONS,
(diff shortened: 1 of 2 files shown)


supervisor/api/backups.py
```diff
@@ -91,6 +92,14 @@ def _extract_slug(self, request):
             raise APIError("Backup does not exist")
         return backup
 
+    def _backup_location(self, backup: Backup) -> str | None:
```
pvizeli, line 95: I think we should LRU this

pvizeli: or even let backup make a property kinda `backup.location`

ludeeus: Cached property is probably the best option, what do you think @mdegat01 ?

mdegat01: Yeah good idea. Expensive to keep looking it up since we have to do path checking. And I don't really know of a better way to do it.

Small note though, we can use [`is_relative_to`](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.is_relative_to) instead of string checking

pvizeli (approved):

PR #4400, opened 2023-06-22
Allow return of all addon info in one api

mdegat01:
## Proposed change

@bdraco noted that the hassio integration in core can be quite expensive at startup. This is in part because it makes 3 calls per installed addon every time the data refreshes:
1. Get complete addon info (`/addons` only returns a partial view of each)
2. Get docker stats per addon
3. Get changelog per addon

https://github.com/home-assistant/core/blob/b8de7df6098e257cf6d0a347d141e5f40cae66ca/homeassistant/components/hassio/__init__.py#L903C19-L919

This API allows a client to get all of 1 and 2 in a single API call with new query string parameters for `/addons` so we can reduce IO in core. 3 is unchanged for now as changelogs can be quite large.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/api/addons.py
+++ b/supervisor/api/addons.py
@@ -106,7 +106,7 @@ from ..exceptions import (
     PwnedSecret,
 )
 from ..validate import docker_ports
-from .const import ATTR_SIGNED, CONTENT_TYPE_BINARY
+from .const import ATTR_SIGNED, CONTENT_TYPE_BINARY, AddonView
 from .utils import api_process, api_process_raw, api_validate, json_loads
 
 _LOGGER: logging.Logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


mdegat01:
On hold. Think we want to refactor hassio integration in core first. Then see if any supervisor changes required

pvizeli:
We can not do that. Some Information like stats take a lot of resource and time to response. If we now add this each time we fetch information, the API call will block for 30-40second and make the hole UI slow. Stats have to be like logs, a own endpoint

bdraco:
I think we will see a huge difference if we can get core to only request the data for entities that are enabled instead of unconditionally as it does now.

PR #4425, opened 2023-07-05
Separate startup event from update check event

mdegat01:
## Proposed change

Currently users see this a lot in core:
```
Logger: homeassistant.components.hassio.issues
Source: components/hassio/issues.py:312
Integration: Home Assistant Supervisor (documentation, issues)
First occurred: 17:31:34 (1 occurrences)
Last logged: 17:31:34

Failed to update supervisor issues: HassioAPIError('System is not ready with state: setup')
```
This is because when the hassio integration sees a message with type `supervisor_update` and key `supervisor` it assumes supervisor has just finished startup and asks for issues. The problem is we actually fire this message twice:
https://github.com/home-assistant/supervisor/blob/48e9e1c4f94cf89d3f2dbbebe93de7d5744fb6ae/supervisor/core.py#L269
https://github.com/home-assistant/supervisor/blob/48e9e1c4f94cf89d3f2dbbebe93de7d5744fb6ae/supervisor/updater.py#L297

The second one in particular is the issue since it occurs before setup is complete. Therefore every time you restart supervisor without restarting core you see this error.

Nothing breaks so this isn't urgent but it is log noise that should be removed. This PR allows us to distinguish between these two messages and only listen for the first one.

It also queues messages sent to core before startup has completed and sends them after to prevent situations like this where the client reacts by asking for more information before supervisor's API is ready. Commands (where supervisor tells core to do something and blocks waiting for a response) are still allowed as that should be ok. Plus we can't just put those in a queue the caller is blocked and waiting.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/const.py
+++ b/supervisor/const.py
@@ -451,6 +451,7 @@ class BusEvent(str, Enum):
     HARDWARE_NEW_DEVICE = "hardware_new_device"
     HARDWARE_REMOVE_DEVICE = "hardware_remove_device"
     DOCKER_CONTAINER_STATE_CHANGE = "docker_container_state_change"
+    SUPERVISOR_STATE_CHANGE = "supervisor_state_change"
 
 
 class CpuArch(str, Enum):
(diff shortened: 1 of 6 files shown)


mdegat01:
Going to add a message queue to this so messages fired during startup are queued until after. That is safer and will stop this error without requiring a follow-up change in core.

pvizeli (approved):

PR #4428, opened 2023-07-07
Allow API calls from HA core to supervisor when core state is "setup" (HA core / supervisor race condition)

sbyx:
## Proposed change

There seems to be a race condition between HA supervisor and HA core which occasionally leads to the first HA core startup failing after OS boot especially on powerful machines. This has been documented by multiple users:
https://github.com/home-assistant/core/issues/86622
https://github.com/home-assistant/core/issues/95887

When the issue happens HA supervisor API is responding to HA core with:
`homeassistant.components.hassio.handler.HassioAPIError: System is not ready with state: setup`
which if I'm looking in the code is actually counter-intuitive as the check seems to check core state, not supervisor state and is commented with 'Check if core is ready to response.' which it obviously is.

I'm proposing (for lack of a better understanding) to allow core to call supervisor API when core.state is still "setup". Alternatively the core state detection could possibly be fixed but I didn't have the time to dig into that.

## Type of change

- [ ] Dependency upgrade
- [X] Bugfix (non-breaking change which fixes an issue)

## Additional information

- This PR is related to issues: https://github.com/home-assistant/core/issues/86622
https://github.com/home-assistant/core/issues/95887

## Checklist

- [X] The code change is tested and works locally.
- [X] Local tests pass. **Your PR cannot be merged unless tests pass**
- [X] There is no commented out code in this PR.
- [X] I have followed the [development checklist][dev-checklist]
- [X] The code has been formatted using Black (`black --fast supervisor tests`)

### DIFF
--- a/supervisor/api/middleware/security.py
+++ b/supervisor/api/middleware/security.py
@@ -190,6 +190,7 @@ class SecurityMiddleware(CoreSysAttributes):
     ) -> Response:
         """Check if core is ready to response."""
         if self.sys_core.state not in (
+            CoreState.SETUP,
             CoreState.STARTUP,
             CoreState.RUNNING,
             CoreState.FREEZE,


supervisor/api/middleware/security.py
```diff
@@ -190,6 +190,7 @@ async def system_validation(
     ) -> Response:
         """Check if core is ready to response."""
         if self.sys_core.state not in (
+            CoreState.SETUP,
```
epenet, line 193: Would it not make sense to also change the message?
`message=f"System is not ready with core state: {self.sys_core.state.value}"`
or
`message=f"System is not ready with state of core: {self.sys_core.state.value}"`

sbyx: I don't really see the difference here.

pvizeli:
The API is only stable after SETUP

sbyx:
@pvizeli it's not really useful just closing this with a meaningless comment. So what is your suggestion instead, to fix this race condition?

TimoHocker:
I agree with @pvizeli, it doesn't make sense to allow calls to an unstable api. But closing the pr without allowing the requestor to make adjustments and completely disregarding the issue isn't acceptable. You could have just posted the comment and asked them to find another solution.

Without having looked into the code myself, it would make sense to have the supervisor wait for some time and recheck the state. Even having it retry every five minutes forever would be fine since all those services should be able to recover if one isn't available at some point.

PR #4438, opened 2023-07-13
Reduce executor code for docker

mdegat01:
## Proposed change

Currently nearly everything in the core `DockerInterface` class everything shares is done in the executor. This is problematic for a few reasons:

1. Not all of it needs to be. I/O interactions with docker should be done there but interactions with coresys objects don't need to be
2. We occasionally modify coresys objects (like when creating issues). These objects are shared across threads so this isn't safe
3. #1946 is quite difficult to implement. The goal is to allow one job to call another and track that interaction. But if entire jobs are occurring in the executor any place we did that tracking would have to be mutable across threads. If jobs are asyncio tasks that call into the executor as needed for I/O this becomes much more straightforward.

The docker interface is the biggest offender here, this PR changes that.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/tests/dbus/network/test_network_manager.py
+++ b/tests/dbus/network/test_network_manager.py
@@ -144,16 +144,16 @@ async def test_handling_bad_devices(
 
     with patch.object(DBus, "init_proxy", side_effect=DBusFatalError()):
         await network_manager.update(
-            {"Devices": ["/org/freedesktop/NetworkManager/Devices/100"]}
+            {"Devices": [device := "/org/freedesktop/NetworkManager/Devices/100"]}
         )
-        assert not caplog.text
+        assert f"Can't process {device}" not in caplog.text
 
     await network_manager.update()
     with patch.object(DBus, "properties", new=PropertyMock(return_value=None)):
         await network_manager.update(
-            {"Devices": ["/org/freedesktop/NetworkManager/Devices/101"]}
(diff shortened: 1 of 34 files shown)


tests/dbus/network/test_network_manager.py
```diff
@@ -144,16 +144,16 @@ async def test_handling_bad_devices(
 
     with patch.object(DBus, "init_proxy", side_effect=DBusFatalError()):
         await network_manager.update(
-            {"Devices": ["/org/freedesktop/NetworkManager/Devices/100"]}
+            {"Devices": [device := "/org/freedesktop/NetworkManager/Devices/100"]}
```
agners, line 147: I think this test update doesn't belong here.

agners (approved):
Otherwise nice cleanup, LGTM!

tests/dbus/network/test_network_manager.py
mdegat01: I had to change the test because previously it was asserting that the log was completely empty. I forget exactly what I changed as part of this that caused the log to not be completely empty but that was an overtest anyway. The test just wanted to confirm that we weren't generating log noise for particular types of devices so it was better to check that the particular log line I wanted was missing.

PR #4449, opened 2023-07-20
Allow discovery messages for unknown services with a warning

mdegat01:
## Proposed change

Supervisor is too protective when it comes to discovery messages. It won't even let the user see an addon if it lists an unknown type of service in discovery. Also removing a service from the list of known types of discovery can cause the discovery json file to fail validation on startup.

This PR makes the following changes:
1. Supervisor will not check the name of the service in discovery messages stored on disk at startup. Any string will pass
2. ~~If an addon lists an unknown type of service in discovery it will still appear in the store. A message will be logged about this at debug level and an event sent to sentry with basic addon details. This event will only be sent once per system per restart of supervisor to limit noise (as we fetch and process addon config regularly)~~
3. If an addon attempts to send a discovery message with an unknown service it will be allowed and a warning logged asking the user to contact the add-on developer about this.

~~The difference between 2 and 3 is significant IMO hence the difference in log levels. 2 will occur regularly due to scheduled task for addons the user has never installed. Maybe they don't even see them if they are marked advanced. We are likely better able to follow-up then most users, hence the sentry event. 3 though will only occur for an addon the user has installed and started, that name won't be foreign to them.~~

After discussing with @pvizeli using sentry will be too noisy. The reason this PR exists is because we wanted to add  a new type of discovery to an addon in a core repo. The addon was merged with that change and then a beta pushed so users could test it. But that meant every user on stable was hitting an error for discovery of an unknown type of service, we can't send all those events to sentry.

So for now, we always just log a warning in all cases. Can revisit if/when we add support for beta and dev channels of an addon repo so changes to addon config can be pushed through a similar pipeline.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/addons/validate.py
+++ b/supervisor/addons/validate.py
@@ -100,7 +100,6 @@ from ..const import (
 )
 from ..discovery.validate import valid_discovery_service
 from ..docker.const import Capabilities
-from ..utils.sentry import capture_event
 from ..validate import (
     docker_image,
     docker_ports,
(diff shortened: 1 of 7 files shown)


mdegat01:
Personally I'm still of the opinion that supervisor should stay out of discovery validation entirely. Discovery seems like a conversation between addons and core with supervisor just being a dumb proxy to handle authentication. But that's a separate discussion. This should help with add-on development in the meantime, so add-ons can publish with new types of discovery without waiting on a supervisor PR and release.

supervisor/addons/validate.py
```diff
@@ -175,6 +176,32 @@ def _warn_addon_config(config: dict[str, Any]):
...
+            invalid_services.append(service)
+
+    if invalid_services:
+        # Logging at debug level as addon config is processed regularly and user may not have even installed this addon
+        _LOGGER.debug(
+            "Add-on lists the following unknown services for discovery: %s. Please report this to the maintainer of %s",
+            (services_list := ", ".join(invalid_services)),
+            name,
+        )
+        capture_event(
```
pvizeli, line 193: Do we need that on sentry?

frenck:
Nice! Thanks for this one ❤️

PR #4457, opened 2023-07-27
Add job group execution limit option

mdegat01:
## Proposed change

Replace `process_lock` decorator job decorator by adding a new `JOB_GROUP` execution limitation option. This option allows a thread to grab the lock for the group and then freely make calls to methods which share this common lock while still locking other callers out. Instead of having public methods using a lock and private methods that route around the lock to allow one method to call another after obtaining the singular lock on the class.

Lock ownership is tracked via a contextvar which is cloned per asyncio task. This way we can easily determine what task has grabbed hold of a job group lock and which haven't.

In terms of one job calling another, each job defines what it has access to irrespective of its parent. So if a job in a job group calls another job in that job group, that works and the chain can continue. If a job calls another job in a different job group or not in a job group at all then the inner job cannot make more calls in the original job group until control returns to the parent.

This is the start of https://github.com/home-assistant/supervisor/issues/1946 . It implements job groups and leverages them. Progress/stage options exist but are not currently used and there is no way to list all running jobs with the API. Those things can come in a subsequent PR.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/exceptions.py
+++ b/supervisor/exceptions.py
@@ -36,6 +36,14 @@ class JobConditionException(JobException):
     """Exception happening for job conditions."""
 
 
+class JobStartException(JobException):
+    """Exception occurred starting a job on thread."""
+
+
+class JobGroupExecutionLimitExceeded(JobException):
+    """Exception when job group execution limit exceeded."""
+
+
 # HomeAssistant
 
 
@@ -476,6 +484,10 @@ class DockerTrustError(DockerError):
 
 class DockerNotFound(DockerError):
     """Docker object don't Exists."""
+
+
+class DockerJobError(DockerError, JobException):
+    """Error executing docker job."""
 
 
 # Hardware
--- a/supervisor/exceptions.py
+++ b/supervisor/exceptions.py
@@ -37,7 +37,7 @@ class JobConditionException(JobException):
 
 
 class JobStartException(JobException):
-    """Exception occurred starting a job on thread."""
+    """Exception occurred starting a job on in current asyncio task."""
 
 
 class JobNotFound(JobException):
--- a/supervisor/exceptions.py
+++ b/supervisor/exceptions.py
@@ -40,6 +40,10 @@ class JobStartException(JobException):
     """Exception occurred starting a job on thread."""
 
 
+class JobNotFound(JobException):
+    """Exception for job not found."""
+
+
 class JobGroupExecutionLimitExceeded(JobException):
     """Exception when job group execution limit exceeded."""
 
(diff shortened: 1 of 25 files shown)


pvizeli (approved):

supervisor/jobs/__init__.py
```diff
@@ -70,18 +77,28 @@ def ignore_conditions(self, value: list[JobCondition]) -> None:
...
-        if name not in self._jobs:
-            self._jobs[name] = SupervisorJob(self.coresys, name)
+    def new_job(self, name: str, initial_stage: str | None = None) -> SupervisorJob:
+        """Create a new job."""
+        job = SupervisorJob(name, stage=initial_stage)
+        self._jobs[job.uuid] = job
+        return job
+
+    def get_job(self, uuid: UUID | None = None) -> SupervisorJob | None:
+        """Return a job by uuid if it exists. Returns the current job of the thread if uuid omitted."""
```
agners, line 87:
```suggestion
        """Return a job by uuid if it exists. Returns the current job of the asyncio task if uuid omitted."""
```

supervisor/exceptions.py
```diff
@@ -36,6 +36,18 @@ class JobConditionException(JobException):
     """Exception happening for job conditions."""
 
 
+class JobStartException(JobException):
+    """Exception occurred starting a job on thread."""
```
agners, line 40:
```suggestion
    """Exception occurred starting a job in current asyncio task."""
```

supervisor/jobs/__init__.py
```diff
@@ -1,49 +1,56 @@
...
-        self._progress: int = 0
-        self._stage: str | None = None
 
-    @property
-    def progress(self) -> int:
-        """Return the current progress."""
-        return self._progress
+@define
+class SupervisorJob:
+    """Supervisor running job class."""
```
agners, line 26: Can you add a bit of documentation what this class exactly does?

E.g. Tracks a single Supervisor Job. Ensures that only one job is started on a single asyncio task globally, but allows child jobs even on the same asyncio task.

Ideally we should document a bit the big picture how the Job system exactly works, but I don't think that should be here in code.

agners (approved):
Other than that, LGTM.

supervisor/jobs/__init__.py
mdegat01: I added some more info to the `start` method as that made more sense to me. I also put some comments explaining why the contextvar works. Let me know if that works.

PR #4495, opened 2023-08-17
Add job names and references everywhere

mdegat01:
## Proposed change

Since we expect jobs to be included in the front end, job names will likely be used in translations. Therefore require a name be specified for all jobs and never use the python `__qualname__` as that can break on a refactor.

Additionally allow specification of a `reference` on jobs when a job is acting on a specific instance of something (e.g. a particular backup, addon or mount). Allow `JobGroup` to predefine this when possible else the method will need to add it as early as possible within the job.

Finally since job groups now provide a unique identifier for jobs defined on a particular instance of a class, leverage that to refactor `Addon._restart_after_problem` out of the constructor. We can now expand our throttling options to `GROUP_THROTTLE_<x>` so that methods can be throttled by instance and not by class.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/addons/__init__.py
+++ b/supervisor/addons/__init__.py
@@ -152,11 +152,15 @@ class AddonManager(CoreSysAttributes):
                 capture_exception(err)
 
     @Job(
+        name="addon_manager_install",
         conditions=ADDON_UPDATE_CONDITIONS,
         on_condition=AddonsJobError,
     )
     async def install(self, slug: str) -> None:
         """Install an add-on."""
+        if job := self.sys_jobs.get_job():
+            job.reference = slug
+
         if slug in self.local:
             raise AddonsError(f"Add-on {slug} is already installed", _LOGGER.warning)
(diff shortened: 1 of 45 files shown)


pvizeli (approved):
Tests seems to fail, but the PR looks good

mdegat01:
@pvizeli yea it looks like my idea of enforcing uniqueness of group names in the constructor isn't going to work with our unit tests. I also realized it would break if someone uninstalled and reinstalled the same addon so I guess its good the unit tests stop that.

Instead I'm going to switch to declaring all my job group names in the same constant file. So that way the developer can easily see that theirs is unique. It's not a foolproof solution but should be good enough I think.

pvizeli (approved):

pvizeli:
Would say, let merge it on Monday 👍

agners (approved):

PR #4503, opened 2023-08-24
Backup and restore track progress in job

mdegat01:
## Proposed change

Backup and restore jobs now record progress and stage information in the job as they go. In addition any time a job updated Home Assistant is informed via websockets. This will allow us to show information about currently running jobs in the frontend in the future.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist


supervisor/backups/backup.py
```diff
@@ -340,7 +340,9 @@ def _create_backup():
         finally:
             self._tmp.cleanup()
 
-    async def store_addons(self, addon_list: list[str]) -> list[Awaitable[None]]:
+    async def store_addons(
+        self, addon_list: list[str], *, on_complete: Callable[[], None] | None = None
```
pvizeli, line 344: Instead having an `on_complete` callback, can we not create jobs out of `_addon_save` and `_save` and `_restore` (could be still inline function) - I guess we need create an `job_inline` helper decorator as the `job` is more top-level. Then we can add key for step directly as parameter to `job` and `job_inline`. I believe it would be possible to handle this job tracking complete inside the job handler/declerator without modify any other parts of the code, but of course, we need declare more function with the job decorator as now we have an use case for doing this

pvizeli: maybe to track the state (or what is ongoing), we even not need a new key as parameter, I guess we can use the `name` as step

mdegat01:
After chatting with pascal I'm going to overhaul this a bit. Most of this code is about trying to set the progress percentage because I really wanted it to always go forwards and not hit 100% until it was done. Which is really tough to do when you don't know exactly how many steps there will be and how long each will take. Which in truth makes tracking progress a bad idea since its also frustrating for a user when sometimes 10% takes 30 seconds and sometimes it takes a half an hour.

So I'm going to drop the progress. Backup and restore will just track stages as a more user friendly form of logging. We can plan to show a spinner in the front end.

I think also going forward that should be the plan. If we don't have a way to get accurate progress we should not try. So like for docker and git tasks we can get an accurate progress from the library so that'll work. But for ones like backup and restore where we can't do that, we only track stages.

supervisor/backups/manager.py
```diff
@@ -366,19 +387,24 @@ async def _do_restore(
...
+                    await self.sys_homeassistant.core.start()
 
-            # Check If we can access to API / otherwise restart
-            if not await self.sys_homeassistant.api.check_api_state():
-                _LOGGER.warning("Need restart HomeAssistant for API")
-                await self.sys_homeassistant.core.restart()
+                # Check If we can access to API / otherwise restart
+                if not await self.sys_homeassistant.api.check_api_state():
+                    _LOGGER.warning("Need restart HomeAssistant for API")
+                    await self.sys_homeassistant.core.restart()
```
mdegat01, line 407: This code was causing the "Run Supervisor" test to fail. The step to restore the SSH addon only from the backup was getting stuck trying to start Home Assistant. Which makes sense since HA was never supposed to be running in the supervisor CI.

Wrapping it like this fixed it. It also makes much more sense, there's no reason to do anything with HA core when it wasn't even part of the restore.

In general I also question the existence of this code. When you restore HA from a backup the code for updating HA already waits for it to come back online and does a rollback if it doesn't. So what use cases does this bit at the end handle? And is there a particular reason its in the finally and not here (or part of the `task_hass` task):
https://github.com/home-assistant/supervisor/blob/b1380d46ecd99a088ffd74f906359c518b803fcf/supervisor/backups/manager.py#L380-L382

mdegat01: It also causes an edge case bug. If someone has stopped HA core and restores a backup from the cli it will start HA core again. If a restore changes the version of HA we have logic to prevent us from starting HA when it was stopped by the user prior to the update here:
https://github.com/home-assistant/supervisor/blob/b1380d46ecd99a088ffd74f906359c518b803fcf/supervisor/homeassistant/core.py#L244-L246

But that code won't actually matter during a restore because we just end up starting HA later anyway due to the code above.

This is definitely a niche edge case. But still it is a small bug. And since I'm not sure what purpose the code does serve I'd like to remove it possible.

supervisor/backups/manager.py
```diff
@@ -315,28 +328,36 @@ async def _do_restore(
         folder_list: list[str],
         homeassistant: bool,
         replace: bool,
-    ):
+    ) -> bool:
         addon_start_tasks: list[Awaitable[None]] | None = None
+        job = self.sys_jobs.get_job()
```
pvizeli, line 333: let's make it like asyncio with an helper `self.sys_jobs.current`

pvizeli (commented):
other than that 2 comments above, looks good. I think with the comments above, we can simplify the handling a bit, also for future usage of that functionality

supervisor/backups/manager.py
```diff
@@ -366,19 +387,24 @@ async def _do_restore(
...
-            # Do we need start Home Assistant Core?
-            if not await self.sys_homeassistant.core.is_running():
-                await self.sys_homeassistant.core.start()
+            # Leave Home Assistant alone if it wasn't part of the restore
+            if homeassistant:
+                change_stage(RestoreJobStage.CHECK_HOME_ASSISTANT)
+
+                # Do we need start Home Assistant Core?
+                if not await self.sys_homeassistant.core.is_running():
+                    await self.sys_homeassistant.core.start()
```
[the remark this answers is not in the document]
mdegat01: Tbh this entire section is an edge case IMO. When you actually restore Home Assistant it stops and then starts it. And that includes a lot of logic to check and make sure its actually running after that and rollback to the last known good version on failure.

So really there's only two times this code would do anything:
1. If HA restore totally failed (rollback too). Then this will be one last attempt to get it started again (which we probably don't have high hopes for at that point). And there's already been a ton of logging to this point so the user knows something has gone quite wrong
2. The user had stopped core manually (like an `ha core stop` in the CLI) and then restored a backup including home assistant via the cli (not via the UI). If HA was stopped manually by the user before restore began then the earlier steps won't try to start HA but this one will. This is definitely an edge case, restores are generally done via the UI (so HA is definitely running) or without stopping HA first.

So I don't think any more detail is required here. This will rarely come up. The logging would only matter in (2) above and I'm not even fully convinced we should be starting HA again in that edge case. I mentioned that earlier in my comment.

pvizeli (approved):

PR #4510, opened 2023-08-27
feat: add ipv6 addr gen mode selection

novag:
## Proposed change

This change introduces support for selecting the IPv6 address generation mode of NetworkManager when the configuration method is set to "auto". By default the "addr-gen-mode" is set to "stable-privacy" which is not suitable for home users with dynamic IPv6 prefixes. In "stable-privacy" mode the interface ID is only guaranteed to be stable per prefix/subnet as per [RFC7217](https://datatracker.ietf.org/doc/html/rfc7217).

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR fixes or closes issue: fixes https://github.com/home-assistant/operating-system/issues/1687
- This PR is related to issue: https://github.com/home-assistant/operating-system/issues/1687
- Link to cli pull request:
- Link to frontend pull request: https://github.com/home-assistant/frontend/pull/17722

## Checklist


supervisor/dbus/network/configuration.py
```diff
@@ -63,6 +63,7 @@ class IpProperties:
     """IP properties object for Network Manager."""
 
     method: str | None
+    addr_gen_mode: int | None
```
mdegat01, line 66: This isn't really accurate typing. It is always an `int` for ipv6 and always `None` for ipv4. I think we should make a new `Ip6Properties` which inherits from `IpProperties` and adds this field. And then use `IpProperties` in ipv4 places and `Ip6Properties` in ipv6 places.

supervisor/host/const.py
```diff
@@ -14,6 +14,13 @@ class InterfaceMethod(str, Enum):
     AUTO = "auto"
 
 
+class InterfaceAddrGenMode(str, Enum):
```
mdegat01, line 17:
```suggestion
class InterfaceAddrGenMode(StrEnum):
```
After you rebase this suggestion will become more obvious, all the other ones are changed this way as well. `StrEnum` is the new and improved way of making an Enum where every instances is also a string added in 3.11 as opposed to inheriting from both like you used to have to do.

supervisor/dbus/const.py
```diff
@@ -198,6 +198,13 @@ class InterfaceMethod(str, Enum):
     LINK_LOCAL = "link-local"
 
 
+class InterfaceAddrGenMode(int, Enum):
```
mdegat01, line 201:
```suggestion
class InterfaceAddrGenMode(IntEnum):
```
Also will be more obvious after a rebase as all the others are changed this way. Inheriting from `IntEnum` is the new improved alternative to inheriting from `int` and `Enum` as of python 3.11.

mdegat01 (changes requested):
A few things are still needed to get this merged besides the comments below:
1. A rebase (there are some conflicts)
2. Add test cases
3. A PR to update the api doc [here](https://developers.home-assistant.io/docs/api/supervisor/endpoints#network)
4. A PR to update the `network update` command in the CLI [here](https://github.com/home-assistant/cli/blob/master/cmd/network_update.go)

Also could you perhaps explain what issues having `stable_privacy` causes? I'm not sure I understand. Reading the linked issue it appears that it was resolved after we ensured there was always a stable UUID in the config. Is that not the case?

PR #4555, opened 2023-09-12
Switch from ruamel.yaml to pyyaml

mdegat01:
## Proposed change

ruamel.yaml seems to be the source of a lot of yaml parse errors that really don't make any sense since the YAML appears fine. Core has been using pyyaml for a while with a lot less of these issues. The only benefit ruamel.yaml gives us is YAML 1.2 support. Since this isn't something we need or want since its inconsistent with YAML in core (which uses pyyaml and therefore YAML 1.1). And YAML 1.1 is less restrictive then 1.2 so we don't expect this swap to create issues.

Therefore we try swapping it to see if we can reduce the amount of log noise and be consistent with core in YAML handling.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/.github/workflows/builder.yml
+++ b/.github/workflows/builder.yml
@@ -112,7 +112,7 @@ jobs:
           tag: musllinux_1_2
           arch: ${{ matrix.arch }}
           wheels-key: ${{ secrets.WHEELS_KEY }}
-          apk: "libffi-dev;openssl-dev"
+          apk: "libffi-dev;openssl-dev;yaml-dev"
           skip-binary: aiohttp
           env-file: true
           requirements: "requirements.txt"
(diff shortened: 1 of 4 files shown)


frenck (approved):
Looks the same as Core now 👍 

../Frenck

mdegat01:
I checked devcontainer, looks like the libyaml package is already installed there. Its not in our [devcontainer dockerfile](https://github.com/home-assistant/devcontainer/blob/main/supervisor/Dockerfile) so I checked the vscode python base image, that seems to include it:
```
apt list --installed | grep yaml

WARNING: apt does not have a stable CLI interface. Use with caution in scripts.

libyaml-0-2/now 0.2.2-1 arm64 [installed,local]
libyaml-dev/now 0.2.2-1 arm64 [installed,local]
```
I can add it to ours anyway to be safe but doesn't appear supervisor's devcontainer needs a corresponding update.

PR #4556, opened 2023-09-12
Add support for green LEDs to API

mdegat01:
## Proposed change

Add APIs for changing Green LEDs per its DBus interface (https://github.com/home-assistant/os-agent/pull/164)

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to documentation pull request: https://github.com/home-assistant/developers.home-assistant/pull/1905
- Link to cli pull request: https://github.com/home-assistant/cli/pull/423

## Checklist


pvizeli:
I guess @agners changed a bit how LED would work in future. The LED are now get restored from Supervisor on first start and the dbus command directly control the state. Like I understand it, the Yellow will get the same changes for LED. All other options still would need a reboot

mdegat01:
@pvizeli ok I changed it. Supervisor now stores changes to LEDs on disk like the others. It then posts the expected value for LEDs on startup when we connect to the dbus interface for the board.

Although all boards have a different config only one can be in use at the same time so I had them all share `board.json`. The one potential consequence of this is if a user migrates to a different board it will keep options which share the same name (like `power_led` for yellow and green) and drop the rest. This seemed like a good thing for the most part (keep what we can) but if you forsee an unexpected consequence I can adjust to make sure that doesn't happen.

I updated yellow with this PR as well in anticipation of the change you said. I kept the `REBOOT_REQUIRED` issue there for now and only dropped it from green.

agners (approved):

PR #4619, opened 2023-10-11
Eliminate possible addon data race condition during update

mdegat01:
## Proposed change

Currently addon update has a possible race condition with the global addon data. This can be triggered as follows:

1. An addon update is initiated
2. While waiting for the new image to download, store data is updated (possibly due to scheduled task)
3. The config of the addon being updated has changed. Specifically if a new version came out or the image was changed

In this situation Supervisor ends up pulling one image but telling the addon a completely different image is in use. And then promptly removing the image it just pulled as it believes it is stale.

The unit test reproduces this race condition by simulating with asyncio events prior to the code change. It is not reproducible afterwards. I believe this change to be the solution to #4610 . Although we can't be certain as we have no logs or steps to reproduce clearly showing what is occurring on these systems. Still it seems likely as the steps I outlined above would result in that exact error.

## Type of change

- [ ] Dependency upgrade
- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (which adds functionality to the supervisor)
- [ ] Breaking change (fix/feature causing existing functionality to break)
- [ ] Code quality improvements to existing code or addition of tests

## Additional information

- This PR fixes or closes issue: fixes #4610
- This PR is related to issue:
- Link to documentation pull request:
- Link to cli pull request:

## Checklist

- [ ] The code change is tested and works locally.
- [ ] Local tests pass. **Your PR cannot be merged unless tests pass**
- [ ] There is no commented out code in this PR.
- [ ] I have followed the [development checklist][dev-checklist]
- [ ] The code has been formatted using Black (`black --fast supervisor tests`)
- [ ] Tests have been added to verify that the new code works.

If API endpoints of add-on configuration are added/changed:

- [ ] Documentation added/updated for [developers.home-assistant.io][docs-repository]

[dev-checklist]: https://developers.home-assistant.io/docs/en/development_checklist.html
[docs-repository]: https://github.com/home-assistant/developers.home-assistant

Issue #4610, opened 2023-10-07
Image homeassistant/amd64-addon-zwave_js:0.1.96 does not exist for addon_core_zwave_js


louisostrowski:
### The problem

Failed setup will retry error with none in error description 

### What version of Home Assistant Core has the issue?

Core-2023-10-1

### What was the last working version of Home Assistant Core?

2023-09-3

### What type of installation are you running?

Home Assistant OS

### Integration causing the issue

Zwave JS

### Diagnostics information

![IMG_3755](https://github.com/home-assistant/core/assets/48015857/96c296c2-f5c4-431f-bf29-a5f319d84a76)

hendrix04:
Do we have an RCA on this issue?

From someone who isn't familiar with how HA handles these docker add-ons, it seems like the add-on upgraded its docker dependency, but there is no process in place for HA to pull that new dependency.

Is this an issue of a lack "package lock" / package management within HA?

Is there any process in place for this?

Did the developer not follow some standard process that would have avoided this?

What do we need to do in the future to have appropriate checks the prevent this issue from happening again?

iloveicedgreentea:
I had an issue with both the js add-on and the integration after upgrading to 10.1, and random failures with some devices like a zen17. The add-on refused to start (no error) after I restarted it. It loaded the container just fine it just did not update the start status. It also added a duplicate integration with a ton of disabled entities. 

Restoring from the 9.x backup fixed my issues. 

I did not have any error regarding the docker image and it was outputting logs without being marked as started, so it seemed to be starting okay. My devices were just unavailable at that time.

agners:
Can someone who experience this issue please attach the full Supervisor logs from Settings > System > Logs, then choose Supervisor on the top right.

FirewizzNL:
> Can someone who experience this issue please attach the full Supervisor logs from Settings > System > Logs, then choose Supervisor on the top right.

[supervisor_2023-10-09T07-48-52.043Z.log](https://github.com/home-assistant/supervisor/files/12843655/supervisor_2023-10-09T07-48-52.043Z.log)

agners:
@FirewizzNL thank you very much :pray: . Unfortunately the log is a bit too short to show the crucial problems (right after the udpate).

Can you get the logs through a terminal with the following command?

```
ha host logs --identifier hassio_supervisor --lines 1000
```

FirewizzNL:
@agners That is still of no use, nothing else then that there.
[...]

PR #4621, opened 2023-10-11
Capture exception if image is missing on run

mdegat01:
## Proposed change

Follow-up to #4619 . Although we believe #4610 should be fixed now, we don't have explicit confirmation that the race condition is the source. In addition there are edge cases where the image can go missing unexpectedly (failure/outtage in the middle of an `ha supervisor repair`, docker update clearing out images, user pruning images with docker cli, etc.)

~~"{x} cannot be started because the image doesn't exist" is not an error we should ever really be showing users. There is nothing they can do about this except submit an issue here and be told to run `ha supervisor repair`. There is however, more supervisor can do to prevent them from ever seeing this error. Namely we can do exactly what `docker run` does - if the image with the specified tag cannot be found locally, pull it.~~

~~This PR adds a retry step to the `run` for all docker objects to handle `ImageNotFound`. Since this should not happen it first reports the error to sentry so we know there is an issue to address. But it tries to handle this issue for users and only let them know if we also fail to pull the image at that time.~~

EDIT: No longer re-pulling and retrying on run. For now we are just capturing the exception and reporting it to sentry. Although the fixup for image missing is still implemented and marked autofix. So while users will see the error, it will at least try to fix itself afterwards.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/docker/interface.py
+++ b/supervisor/docker/interface.py
@@ -377,6 +377,31 @@ class DockerInterface(JobGroup):
         """Run Docker image."""
         raise NotImplementedError()
 
+    async def _run(self, **kwargs) -> None:
+        """Run Docker image with retry inf necessary."""
+        if await self.is_running():
+            return
+
+        # Cleanup
+        await self.stop()
+
+        # Create & Run container
+        try:
+            docker_container = await self.sys_run_in_executor(
+                self.sys_docker.run, self.image, **kwargs
+            )
+        except DockerNotFound as err:
+            # If image is missing, capture the exception as this shouldn't happen
+            # Try to keep things working for user by pulling image and retrying once
+            capture_exception(err)
+            await self.install(self.version)
+            docker_container = await self.sys_run_in_executor(
+                self.sys_docker.run, self.image, **kwargs
+            )
+
+        # Store metadata
+        self._meta = docker_container.attrs
+
     @Job(
         name="docker_interface_stop",
         limit=JobExecutionLimit.GROUP_ONCE,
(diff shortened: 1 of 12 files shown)


supervisor/docker/interface.py
```diff
@@ -377,6 +377,31 @@ async def run(self) -> None:
...
+
+        # Cleanup
+        await self.stop()
+
+        # Create & Run container
+        try:
+            docker_container = await self.sys_run_in_executor(
+                self.sys_docker.run, self.image, **kwargs
+            )
+        except DockerNotFound as err:
```
pvizeli, line 393: Let's not do that, but the rest of the PR looks good

supervisor/docker/interface.py
```diff
@@ -377,6 +377,31 @@ async def run(self) -> None:
...
+            # If image is missing, capture the exception as this shouldn't happen
+            # Try to keep things working for user by pulling image and retrying once
+            capture_exception(err)
+            await self.install(self.version)
```
agners, line 397: @pvizeli what do you mean by "Let's not do that" exactly?

IMHO adding the sentry capture is fine here, but maybe remove the automatic install for now, until we have a good understand in which cases the image can be missing and they warrant adding it automatically?

```suggestion
```

pvizeli: right, that is what I mean

pvizeli (approved):

agners (approved):

PR #4635, opened 2023-10-18
Remove race with watchdog during backup, restore and update

mdegat01:
## Proposed change

Currently backup, restore and update has a race condition with watchdog. Watchdog will see that the addon has stopped and it will not be marked as a manual stop so it believes it is unexpected. This will cause it to try and reanimate it which can cause issues with the process being run. Block this from happening by treating it as a manual stop as it was initiated by the user.

This situation can also occur with plugin updates currently. The PR stops that as well by having watchdog ignore successful stops of the container as that will only occur during plugin updates or system shutdowns, watchdog should not try to reanimate during either of these events.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/addons/__init__.py
+++ b/supervisor/addons/__init__.py
@@ -315,12 +315,13 @@ class AddonManager(CoreSysAttributes):
             await addon.install_apparmor()
 
         finally:
-            # restore state
-            return (
+            # restore state. Return awaitable for caller if no exception
+            out = (
                 await addon.start()
                 if last_state in {AddonState.STARTED, AddonState.STARTUP}
                 else None
             )
+        return out
 
     @Job(
         name="addon_manager_rebuild",
--- a/supervisor/addons/__init__.py
+++ b/supervisor/addons/__init__.py
@@ -288,33 +288,39 @@ class AddonManager(CoreSysAttributes):
             )
 
         # Update instance
-        last_state: AddonState = addon.state
         old_image = addon.image
         # Cache data to prevent races with other updates to global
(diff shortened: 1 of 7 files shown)


pvizeli:
I don't get why `restore` is not a job and lock down the add-on for other operation?

pvizeli:
We should also make sure, that the update and install process completely lock down the add-on for other operations until it's done

mdegat01:
> I don't get why restore is not a job and lock down the add-on for other operation?

@pvizeli I noted this in discord but there's at least one case I can think of where that wouldn't help:

If a user is using the new `freeze` and `thaw` APIs to take a backup that includes an addon with backup mode `COLD` it will break. The `freeze` job will end which will release the `in_progress` lock and watchdog will try to reanimate the downed addon. Potentially messing up their external backup.

Also current handling of watchdog when there is a job in progress is to wait for that job to complete before continuing to try reanimation. That probably won't matter (since it should be back up by that time) but it seemed like undesirable behavior.

IMO `update`, `backup` and `restore` should not by bypassing the `stop` API on `self` and calling the one on the `DockerAddon` directly. And doing that fixes the issue since it marks it as a manual (or user-initiated) stop, which it is in this case. So this seemed the simplest and most correct fix.

If we want I can also mark those as Jobs with the `GROUP_ONCE` limit. But to me this was the simplest path to getting the bug fixed and is something we should've been doing anyway.

agners (approved):

PR #4650, opened 2023-10-25
Add a public config folder per addon

mdegat01:
## Proposed change

Add a public config folder per addon. This way addons can make internal files for config, dbs, logs, etc. available to users but still have them backed up with the addon (unlike when these files are put into `/share` or `/config` today).

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist


supervisor/addons/addon.py
```diff
@@ -170,6 +172,15 @@ def in_progress(self) -> bool:
 
     async def load(self) -> None:
         """Async initialize of object."""
+        if MAP_CONFIG in self.map_volumes:
```
pvizeli, line 175: Let's add this to: https://github.com/home-assistant/supervisor/blob/9951ceda7b7b234e884d43f562a89ee9457b5e68/supervisor/addons/validate.py#L199

supervisor/docker/addon.py
```diff
@@ -350,6 +352,38 @@ def mounts(self) -> list[Mount]:
                 )
             )
 
+        # Map addon's public config folder if not using deprecated config option
+        else:
```
pvizeli, line 356: I guess the idea was to use a config option for enable that feature, so you see only addon folders inside configs which also use that feature?

supervisor/addons/addon.py
mdegat01: I can add a warning here but I can't migrate it. If using `config` in this option then the home assistant config folder gets mounted to `/config`. Which means the addon's new public config folder cannot get mounted there and is therefore not mounted at all into the container. I can't just migrate one setting to the other or the addon will break as it will be expecting the Home Assistant folder at `/config`, not at `/homeassistant`

supervisor/docker/addon.py
agners: As discussed, I agree we should have a configuration to enable the per add-on config, so that add-ons which don't use it don't get one (avoids unnecessary directories, which comes especially into play for add-ons which will show all this add-on config folders such as VS Code add-on).

agners:
While the naming is mostly exposed to developers, I think we should be careful not mixing too many names for this feature and be as explicit as possible.

Public can be a bit misleading as it is still local only (not shared or similar), so I think I'd not use that term.

Suggestion: Generally, I'd suggest to use "add-on config folder" as term for this feature (as opposed to config folder, or Home Assistant config folder).

Since mappings are typically one per line in the conifg, length is not really an issue here. So I rather opt for verbose naming.

For the mapping names, I'd go for:
* `config` (already exists):  Will eventually be the deprecated config for the Home Assistant config folder
* `homeassistant_config`: New mapping for the Home Assistant config folder
* `addon_config`: New mapping in case the add-on wishes to have such a new add-on config folder.
* `all_addon_configs`: New mapping in case the add-on wishes to have the add-on config folder of all the add-ons mapped.

pvizeli (approved):

PR #4677, opened 2023-11-07
Make check_port an async function

agners:
## Proposed change
This makes `check_port()` an async function. This also requires to change the ingress_port property to a async method.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/addons/addon.py
+++ b/supervisor/addons/addon.py
@@ -387,15 +387,14 @@ class Addon(AddonModel):
 
         return f"{proto}://[HOST]:{port}{s_suffix}"
 
-    @property
-    def ingress_port(self) -> int | None:
+    async def get_ingress_port(self) -> int | None:
         """Return Ingress port."""
         if not self.with_ingress:
             return None
 
         port = self.data[ATTR_INGRESS_PORT]
         if port == 0:
-            return self.sys_ingress.get_dynamic_port(self.slug)
+            return await self.sys_ingress.get_dynamic_port(self.slug)
         return port
 
     @property
@@ -539,7 +538,7 @@ class Addon(AddonModel):
 
         # TCP monitoring
         if s_prefix == "tcp":
-            return await self.sys_run_in_executor(check_port, self.ip_address, port)
+            return await check_port(self.ip_address, port)
 
         # lookup the correct protocol from config
         if t_proto:
--- a/supervisor/addons/addon.py
+++ b/supervisor/addons/addon.py
@@ -387,14 +387,15 @@ class Addon(AddonModel):
 
         return f"{proto}://[HOST]:{port}{s_suffix}"
 
-    async def get_ingress_port(self) -> int | None:
+    @property
+    def ingress_port(self) -> int | None:
         """Return Ingress port."""
         if not self.with_ingress:
             return None
 
         port = self.data[ATTR_INGRESS_PORT]
         if port == 0:
-            return await self.sys_ingress.get_dynamic_port(self.slug)
+            _LOGGER.critical("No port set for add-on %s", self.slug)
         return port
 
     @property
@@ -601,6 +602,16 @@ class Addon(AddonModel):
             _LOGGER.info("Removing add-on data folder %s", self.path_data)
             await remove_data(self.path_data)
 
+    async def _check_ingress_port(self):
+        """Assign a ingress port if dynamic port selection is used."""
+        if not self.with_ingress:
+            return
+
+        if self.data[ATTR_INGRESS_PORT] == 0:
+            self.data[ATTR_INGRESS_PORT] = await self.sys_ingress.get_dynamic_port(
+                self.slug
+            )
+
     @Job(
         name="addon_install",
         limit=JobExecutionLimit.GROUP_ONCE,
@@ -628,6 +639,8 @@ class Addon(AddonModel):
         except DockerError as err:
             self.sys_addons.data.uninstall(self)
             raise AddonsError() from err
+
+        await self._check_ingress_port()
 
         # Add to addon manager
         self.sys_addons.local[self.slug] = self
@@ -715,6 +728,7 @@ class Addon(AddonModel):
         try:
             _LOGGER.info("Add-on '%s' successfully updated", self.slug)
             self.sys_addons.data.update(store)
+            await self._check_ingress_port()
 
             # Cleanup
             with suppress(DockerError):
@@ -755,6 +769,7 @@ class Addon(AddonModel):
                 raise AddonsError() from err
 
             self.sys_addons.data.update(self.addon_store)
+            await self._check_ingress_port()
             _LOGGER.info("Add-on '%s' successfully rebuilt", self.slug)
 
         finally:
--- a/supervisor/addons/addon.py
+++ b/supervisor/addons/addon.py
@@ -1213,6 +1213,7 @@ class Addon(AddonModel):
                     _LOGGER.info("Restore/Update of image for addon %s", self.slug)
                     with suppress(DockerError):
                         await self.instance.update(version, restore_image)
+                self._check_ingress_port()
 
                 # Restore data and config
                 def _restore_data():
--- a/supervisor/addons/addon.py
+++ b/supervisor/addons/addon.py
@@ -395,7 +395,7 @@ class Addon(AddonModel):
 
         port = self.data[ATTR_INGRESS_PORT]
         if port == 0:
-            _LOGGER.critical("No port set for add-on %s", self.slug)
+            raise RuntimeError(f"No port set for add-on {self.slug}")
         return port
 
     @property
--- a/supervisor/addons/addon.py
+++ b/supervisor/addons/addon.py
@@ -73,7 +73,7 @@ from ..homeassistant.const import WSEvent, WSType
 from ..jobs.const import JobExecutionLimit
 from ..jobs.decorator import Job
 from ..store.addon import AddonStore
-from ..utils import async_check_port
+from ..utils import check_port
 from ..utils.apparmor import adjust_profile
 from ..utils.json import read_json_file, write_json_file
 from ..utils.sentry import capture_exception
@@ -539,7 +539,7 @@ class Addon(AddonModel):
 
         # TCP monitoring
         if s_prefix == "tcp":
-            return await async_check_port(self.coresys.loop, self.ip_address, port)
+            return await check_port(self.coresys.loop, self.ip_address, port)
 
         # lookup the correct protocol from config
         if t_proto:
--- a/supervisor/addons/addon.py
+++ b/supervisor/addons/addon.py
@@ -539,7 +539,7 @@ class Addon(AddonModel):
 
         # TCP monitoring
         if s_prefix == "tcp":
-            return await check_port(self.coresys.loop, self.ip_address, port)
+            return await check_port(self.ip_address, port)
 
         # lookup the correct protocol from config
         if t_proto:
--- a/supervisor/addons/addon.py
+++ b/supervisor/addons/addon.py
@@ -73,7 +73,7 @@ from ..homeassistant.const import WSEvent, WSType
 from ..jobs.const import JobExecutionLimit
 from ..jobs.decorator import Job
 from ..store.addon import AddonStore
-from ..utils import check_port
+from ..utils import async_check_port
 from ..utils.apparmor import adjust_profile
 from ..utils.json import read_json_file, write_json_file
 from ..utils.sentry import capture_exception
@@ -538,7 +538,7 @@ class Addon(AddonModel):
 
         # TCP monitoring
         if s_prefix == "tcp":
-            return await check_port(self.ip_address, port)
+            return await async_check_port(self.coresys.loop, self.ip_address, port)
 
         # lookup the correct protocol from config
         if t_proto:
--- a/supervisor/api/addons.py
+++ b/supervisor/api/addons.py
@@ -250,7 +250,7 @@ class APIAddons(CoreSysAttributes):
             ATTR_WEBUI: addon.webui,
             ATTR_INGRESS_ENTRY: addon.ingress_entry,
             ATTR_INGRESS_URL: addon.ingress_url,
-            ATTR_INGRESS_PORT: addon.ingress_port,
+            ATTR_INGRESS_PORT: await addon.get_ingress_port(),
             ATTR_INGRESS_PANEL: addon.ingress_panel,
             ATTR_AUDIO_INPUT: addon.audio_input,
             ATTR_AUDIO_OUTPUT: addon.audio_output,
--- a/supervisor/api/addons.py
+++ b/supervisor/api/addons.py
@@ -250,7 +250,7 @@ class APIAddons(CoreSysAttributes):
             ATTR_WEBUI: addon.webui,
             ATTR_INGRESS_ENTRY: addon.ingress_entry,
             ATTR_INGRESS_URL: addon.ingress_url,
-            ATTR_INGRESS_PORT: await addon.ingress_port,
+            ATTR_INGRESS_PORT: addon.ingress_port,
             ATTR_INGRESS_PANEL: addon.ingress_panel,
             ATTR_AUDIO_INPUT: addon.audio_input,
             ATTR_AUDIO_OUTPUT: addon.audio_output,
--- a/supervisor/api/addons.py
+++ b/supervisor/api/addons.py
@@ -250,7 +250,7 @@ class APIAddons(CoreSysAttributes):
             ATTR_WEBUI: addon.webui,
             ATTR_INGRESS_ENTRY: addon.ingress_entry,
             ATTR_INGRESS_URL: addon.ingress_url,
-            ATTR_INGRESS_PORT: await addon.get_ingress_port(),
+            ATTR_INGRESS_PORT: await addon.ingress_port,
             ATTR_INGRESS_PANEL: addon.ingress_panel,
             ATTR_AUDIO_INPUT: addon.audio_input,
             ATTR_AUDIO_OUTPUT: addon.audio_output,
--- a/supervisor/api/ingress.py
+++ b/supervisor/api/ingress.py
@@ -92,7 +92,7 @@ class APIIngress(CoreSysAttributes):
 
         return addon
 
-    async def _create_url(self, addon: Addon, path: str) -> str:
+    async def _async_create_url(self, addon: Addon, path: str) -> str:
         """Create URL to container."""
         return f"http://{addon.ip_address}:{await addon.get_ingress_port()}/{path}"
 
@@ -190,7 +190,7 @@ class APIIngress(CoreSysAttributes):
         await ws_server.prepare(request)
 
         # Preparing
-        url = await self._create_url(addon, path)
+        url = await self._async_create_url(addon, path)
         source_header = _init_header(request, addon, session_data)
 
         # Support GET query
@@ -224,7 +224,7 @@ class APIIngress(CoreSysAttributes):
         session_data: IngressSessionData | None,
     ) -> web.Response | web.StreamResponse:
         """Ingress route for request."""
-        url = self._create_url(addon, path)
+        url = self._async_create_url(addon, path)
         source_header = _init_header(request, addon, session_data)
 
         # Passing the raw stream breaks requests for some webservers
--- a/supervisor/api/ingress.py
+++ b/supervisor/api/ingress.py
@@ -92,7 +92,7 @@ class APIIngress(CoreSysAttributes):
 
         return addon
 
-    async def _async_create_url(self, addon: Addon, path: str) -> str:
+    def _create_url(self, addon: Addon, path: str) -> str:
         """Create URL to container."""
         return f"http://{addon.ip_address}:{addon.ingress_port}/{path}"
 
@@ -190,7 +190,7 @@ class APIIngress(CoreSysAttributes):
         await ws_server.prepare(request)
 
         # Preparing
-        url = await self._async_create_url(addon, path)
+        url = self._create_url(addon, path)
         source_header = _init_header(request, addon, session_data)
 
         # Support GET query
@@ -224,7 +224,7 @@ class APIIngress(CoreSysAttributes):
         session_data: IngressSessionData | None,
     ) -> web.Response | web.StreamResponse:
         """Ingress route for request."""
-        url = await self._async_create_url(addon, path)
+        url = self._create_url(addon, path)
         source_header = _init_header(request, addon, session_data)
 
         # Passing the raw stream breaks requests for some webservers
--- a/supervisor/api/ingress.py
+++ b/supervisor/api/ingress.py
@@ -224,7 +224,7 @@ class APIIngress(CoreSysAttributes):
         session_data: IngressSessionData | None,
     ) -> web.Response | web.StreamResponse:
         """Ingress route for request."""
-        url = self._async_create_url(addon, path)
+        url = await self._async_create_url(addon, path)
         source_header = _init_header(request, addon, session_data)
 
         # Passing the raw stream breaks requests for some webservers
--- a/supervisor/api/ingress.py
+++ b/supervisor/api/ingress.py
@@ -94,7 +94,7 @@ class APIIngress(CoreSysAttributes):
 
     async def _async_create_url(self, addon: Addon, path: str) -> str:
         """Create URL to container."""
-        return f"http://{addon.ip_address}:{await addon.get_ingress_port()}/{path}"
+        return f"http://{addon.ip_address}:{addon.ingress_port}/{path}"
 
     @api_process
     async def panels(self, request: web.Request) -> dict[str, Any]:
--- a/supervisor/api/ingress.py
+++ b/supervisor/api/ingress.py
@@ -92,9 +92,9 @@ class APIIngress(CoreSysAttributes):
 
         return addon
 
-    def _create_url(self, addon: Addon, path: str) -> str:
+    async def _create_url(self, addon: Addon, path: str) -> str:
         """Create URL to container."""
-        return f"http://{addon.ip_address}:{addon.ingress_port}/{path}"
+        return f"http://{addon.ip_address}:{await addon.get_ingress_port()}/{path}"
 
     @api_process
     async def panels(self, request: web.Request) -> dict[str, Any]:
@@ -190,7 +190,7 @@ class APIIngress(CoreSysAttributes):
         await ws_server.prepare(request)
 
         # Preparing
-        url = self._create_url(addon, path)
+        url = await self._create_url(addon, path)
         source_header = _init_header(request, addon, session_data)
 
         # Support GET query
--- a/supervisor/utils/__init__.py
+++ b/supervisor/utils/__init__.py
@@ -6,7 +6,6 @@ import logging
 import os
 from pathlib import Path
 import re
-import socket
 from tempfile import TemporaryDirectory
 from typing import Any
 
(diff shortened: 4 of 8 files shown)


bdraco:
Tweaked it a bit https://github.com/home-assistant/supervisor/pull/4678

supervisor/utils/__init__.py
```diff
@@ -36,18 +35,12 @@ async def wrap_api(api, *args, **kwargs):
...
+    """Check if port is mapped asynchronously with a timeout."""
     try:
-        result = sock.connect_ex((str(address), port))
-        sock.close()
-
-        # Check if the port is available
-        if result == 0:
-            return True
-    except OSError:
+        await asyncio.wait_for(asyncio.open_connection(str(address), port), 0.5)
```
bdraco, line 41: `wait_for` is not recommended because it has some bad cancellation semantics

supervisor/utils/__init__.py
```diff
@@ -36,18 +35,12 @@ async def wrap_api(api, *args, **kwargs):
     return wrap_api
 
 
-def check_port(address: IPv4Address, port: int) -> bool:
-    """Check if port is mapped."""
-    sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
-    sock.settimeout(0.5)
+async def check_port(address: IPv4Address, port: int) -> bool:
```
bdraco, line 38:
```python
async def async_check_port(
    loop: asyncio.AbstractEventLoop, address: IPv4Address, port: int
) -> bool:
    """Check if port is mapped."""
    sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
    sock.setblocking(False)
    try:
        async with asyncio.timeout(0.5):
            await loop.sock_connect(sock, (str(address), port))
    except (OSError, TimeoutError):
        return False
    finally:
        if sock is not None:
            sock.close()
    return True
```

supervisor/utils/__init__.py
bdraco: cpython 3.12 will change the underlying implementation of `asyncio.wait_for` to use `asyncio.wait` but that is still a long way off for many people:

https://github.com/python/cpython/pull/98518

agners: I see, thanks for the insight! Used your suggestion instead.

agners:
I am not very happy how I have to handle the `ingress_port` port property here. The property currently implicitly gathers a port. When that exactly happens is really depending on when the Add-on info API is being used or the ingress URL is being generated. This code also did I/O (socket connection) in the asyncio event loop, which is not ideal.

The current change converts the property to a async get method which takes the I/O out of the asyncio event loop. But when the port is exactly determinated is still very implicit.

There is a tradeoff of reserving ports unnecessarily early and doing port reservation as late as possible. Ideally I think the port should be determined on first add-on startup. I don't think we need to clear the port on stop.

But this would mean that the URL will not be valid until the add-on gets started, not sure if we handle that properly.

In any case, this can be done in a follow-up PR I think.

bdraco:
> I am not very happy how I have to handle the `ingress_port` port property here. The property currently implicitly gathers a port. When that exactly happens is really depending on when the Add-on info API is being used or the ingress URL is being generated. This code also did I/O (socket connection) in the asyncio event loop, which is not ideal.

Its definitely not ideal. The property having the side effect of allocating a port was a bit scarier to me since it was non-obvious what was going on under the hood.
> 
> The current change converts the property to a async get method which takes the I/O out of the asyncio event loop. But when the port is exactly determinated is still very implicit.

It seems like it should be refactored into a PortManager class or something similar that knows about all the ports
> 
> There is a tradeoff of reserving ports unnecessarily early and doing port reservation as late as possible. Ideally I think the port should be determined on first add-on startup. I don't think we need to clear the port on stop.
> 
That makes more sense to me, I don't think there is a risk we run out of ports.


I think this is a good change but a followup to refactor the ingress_port area would make me feel a lot better about it

bdraco (approved):

bdraco:
I'll throw this on my production now to make sure nothing breaks

bdraco:
```
Nov 08 14:47:40 homeassistant hassio_supervisor[540]: 23-11-08 08:47:40 ERROR (MainThread) [aiohttp.server] Error handling request
Nov 08 14:47:40 homeassistant hassio_supervisor[540]: Traceback (most recent call last):
Nov 08 14:47:40 homeassistant hassio_supervisor[540]:   File "/usr/local/lib/python3.11/site-packages/aiohttp/web_protocol.py", line 452, in _handle_request
Nov 08 14:47:40 homeassistant hassio_supervisor[540]:     resp = await request_handler(request)
Nov 08 14:47:40 homeassistant hassio_supervisor[540]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Nov 08 14:47:40 homeassistant hassio_supervisor[540]:   File "/usr/local/lib/python3.11/site-packages/sentry_sdk/integrations/aiohttp.py", line 139, in sentry_app_handle
Nov 08 14:47:40 homeassistant hassio_supervisor[540]:     reraise(*_capture_exception(hub))
Nov 08 14:47:40 homeassistant hassio_supervisor[540]:   File "/usr/local/lib/python3.11/site-packages/sentry_sdk/_compat.py", line 115, in reraise
Nov 08 14:47:40 homeassistant hassio_supervisor[540]:     raise value
Nov 08 14:47:40 homeassistant hassio_supervisor[540]:   File "/usr/local/lib/python3.11/site-packages/sentry_sdk/integrations/aiohttp.py", line 129, in sentry_app_handle
Nov 08 14:47:40 homeassistant hassio_supervisor[540]:     response = await old_handle(self, request)
Nov 08 14:47:40 homeassistant hassio_supervisor[540]:                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Nov 08 14:47:40 homeassistant hassio_supervisor[540]:   File "/usr/local/lib/python3.11/site-packages/aiohttp/web_app.py", line 543, in _handle
Nov 08 14:47:40 homeassistant hassio_supervisor[540]:     resp = await handler(request)
Nov 08 14:47:40 homeassistant hassio_supervisor[540]:            ^^^^^^^^^^^^^^^^^^^^^^
Nov 08 14:47:40 homeassistant hassio_supervisor[540]:   File "/usr/local/lib/python3.11/site-packages/aiohttp/web_middlewares.py", line 114, in impl
Nov 08 14:47:40 homeassistant hassio_supervisor[540]:     return await handler(request)
Nov 08 14:47:40 homeassistant hassio_supervisor[540]:            ^^^^^^^^^^^^^^^^^^^^^^
Nov 08 14:47:40 homeassistant hassio_supervisor[540]:   File "/usr/src/supervisor/supervisor/api/middleware/security.py", line 186, in block_bad_requests
Nov 08 14:47:40 homeassistant hassio_supervisor[540]:     return await handler(request)
Nov 08 14:47:40 homeassistant hassio_supervisor[540]:            ^^^^^^^^^^^^^^^^^^^^^^
Nov 08 14:47:40 homeassistant hassio_supervisor[540]:   File "/usr/src/supervisor/supervisor/api/middleware/security.py", line 202, in system_validation
Nov 08 14:47:40 homeassistant hassio_supervisor[540]:     return await handler(request)
Nov 08 14:47:40 homeassistant hassio_supervisor[540]:            ^^^^^^^^^^^^^^^^^^^^^^
Nov 08 14:47:40 homeassistant hassio_supervisor[540]:   File "/usr/src/supervisor/supervisor/api/middleware/security.py", line 221, in token_validation
Nov 08 14:47:40 homeassistant hassio_supervisor[540]:     return await handler(request)
Nov 08 14:47:40 homeassistant hassio_supervisor[540]:            ^^^^^^^^^^^^^^^^^^^^^^
Nov 08 14:47:40 homeassistant hassio_supervisor[540]:   File "/usr/src/supervisor/supervisor/api/middleware/security.py", line 280, in core_proxy
Nov 08 14:47:40 homeassistant hassio_supervisor[540]:     return await handler(request)
Nov 08 14:47:40 homeassistant hassio_supervisor[540]:            ^^^^^^^^^^^^^^^^^^^^^^
Nov 08 14:47:40 homeassistant hassio_supervisor[540]:   File "/usr/src/supervisor/supervisor/api/ingress.py", line 141, in handler
Nov 08 14:47:40 homeassistant hassio_supervisor[540]:     return await self._handle_request(request, addon, path, session_data)
Nov 08 14:47:40 homeassistant hassio_supervisor[540]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Nov 08 14:47:40 homeassistant hassio_supervisor[540]:   File "/usr/src/supervisor/supervisor/api/ingress.py", line 217, in _handle_request
Nov 08 14:47:40 homeassistant hassio_supervisor[540]:     async with self.sys_websession.request(
Nov 08 14:47:40 homeassistant hassio_supervisor[540]:   File "/usr/local/lib/python3.11/site-packages/aiohttp/client.py", line 1187, in __aenter__
Nov 08 14:47:40 homeassistant hassio_supervisor[540]:     self._resp = await self._coro
Nov 08 14:47:40 homeassistant hassio_supervisor[540]:                  ^^^^^^^^^^^^^^^^
Nov 08 14:47:40 homeassistant hassio_supervisor[540]:   File "/usr/local/lib/python3.11/site-packages/aiohttp/client.py", line 445, in _request
Nov 08 14:47:40 homeassistant hassio_supervisor[540]:     url = self._build_url(str_or_url)
Nov 08 14:47:40 homeassistant hassio_supervisor[540]:           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
Nov 08 14:47:40 homeassistant hassio_supervisor[540]:   File "/usr/local/lib/python3.11/site-packages/aiohttp/client.py", line 372, in _build_url
Nov 08 14:47:40 homeassistant hassio_supervisor[540]:     url = URL(str_or_url)
Nov 08 14:47:40 homeassistant hassio_supervisor[540]:           ^^^^^^^^^^^^^^^
Nov 08 14:47:40 homeassistant hassio_supervisor[540]:   File "/usr/local/lib/python3.11/site-packages/yarl/_url.py", line 179, in __new__
Nov 08 14:47:40 homeassistant hassio_supervisor[540]:     raise TypeError("Constructor parameter should be str")
Nov 08 14:47:40 homeassistant hassio_supervisor[540]: TypeError: Constructor parameter should be str
Nov 08 14:47:41 homeassistant hassio_supervisor[540]: 23-11-08 08:47:41 ERROR (MainThread) [aiohttp.server] Error handling request
Nov 08 14:47:41 homeassistant hassio_supervisor[540]: Traceback (most recent call last):
Nov 08 14:47:41 homeassistant hassio_supervisor[540]:   File "/usr/local/lib/python3.11/site-packages/aiohttp/web_protocol.py", line 452, in _handle_request
Nov 08 14:47:41 homeassistant hassio_supervisor[540]:     resp = await request_handler(request)
Nov 08 14:47:41 homeassistant hassio_supervisor[540]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Nov 08 14:47:41 homeassistant hassio_supervisor[540]:   File "/usr/local/lib/python3.11/site-packages/sentry_sdk/integrations/aiohttp.py", line 139, in sentry_app_handle
Nov 08 14:47:41 homeassistant hassio_supervisor[540]:     reraise(*_capture_exception(hub))
Nov 08 14:47:41 homeassistant hassio_supervisor[540]:   File "/usr/local/lib/python3.11/site-packages/sentry_sdk/_compat.py", line 115, in reraise
Nov 08 14:47:41 homeassistant hassio_supervisor[540]:     raise value
Nov 08 14:47:41 homeassistant hassio_supervisor[540]:   File "/usr/local/lib/python3.11/site-packages/sentry_sdk/integrations/aiohttp.py", line 129, in sentry_app_handle
Nov 08 14:47:41 homeassistant hassio_supervisor[540]:     response = await old_handle(self, request)
Nov 08 14:47:41 homeassistant hassio_supervisor[540]:                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Nov 08 14:47:41 homeassistant hassio_supervisor[540]:   File "/usr/local/lib/python3.11/site-packages/aiohttp/web_app.py", line 543, in _handle
Nov 08 14:47:41 homeassistant hassio_supervisor[540]:     resp = await handler(request)
Nov 08 14:47:41 homeassistant hassio_supervisor[540]:            ^^^^^^^^^^^^^^^^^^^^^^
Nov 08 14:47:41 homeassistant hassio_supervisor[540]:   File "/usr/local/lib/python3.11/site-packages/aiohttp/web_middlewares.py", line 114, in impl
Nov 08 14:47:41 homeassistant hassio_supervisor[540]:     return await handler(request)
Nov 08 14:47:41 homeassistant hassio_supervisor[540]:            ^^^^^^^^^^^^^^^^^^^^^^
Nov 08 14:47:41 homeassistant hassio_supervisor[540]:   File "/usr/src/supervisor/supervisor/api/middleware/security.py", line 186, in block_bad_requests
Nov 08 14:47:41 homeassistant hassio_supervisor[540]:     return await handler(request)
Nov 08 14:47:41 homeassistant hassio_supervisor[540]:            ^^^^^^^^^^^^^^^^^^^^^^
Nov 08 14:47:41 homeassistant hassio_supervisor[540]:   File "/usr/src/supervisor/supervisor/api/middleware/security.py", line 202, in system_validation
Nov 08 14:47:41 homeassistant hassio_supervisor[540]:     return await handler(request)
Nov 08 14:47:41 homeassistant hassio_supervisor[540]:            ^^^^^^^^^^^^^^^^^^^^^^
Nov 08 14:47:41 homeassistant hassio_supervisor[540]:   File "/usr/src/supervisor/supervisor/api/middleware/security.py", line 221, in token_validation
Nov 08 14:47:41 homeassistant hassio_supervisor[540]:     return await handler(request)
Nov 08 14:47:41 homeassistant hassio_supervisor[540]:            ^^^^^^^^^^^^^^^^^^^^^^
Nov 08 14:47:41 homeassistant hassio_supervisor[540]:   File "/usr/src/supervisor/supervisor/api/middleware/security.py", line 280, in core_proxy
Nov 08 14:47:41 homeassistant hassio_supervisor[540]:     return await handler(request)
Nov 08 14:47:41 homeassistant hassio_supervisor[540]:            ^^^^^^^^^^^^^^^^^^^^^^
Nov 08 14:47:41 homeassistant hassio_supervisor[540]:   File "/usr/src/supervisor/supervisor/api/ingress.py", line 141, in handler
Nov 08 14:47:41 homeassistant hassio_supervisor[540]:     return await self._handle_request(request, addon, path, session_data)
Nov 08 14:47:41 homeassistant hassio_supervisor[540]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Nov 08 14:47:41 homeassistant hassio_supervisor[540]:   File "/usr/src/supervisor/supervisor/api/ingress.py", line 217, in _handle_request
Nov 08 14:47:41 homeassistant hassio_supervisor[540]:     async with self.sys_websession.request(
Nov 08 14:47:41 homeassistant hassio_supervisor[540]:   File "/usr/local/lib/python3.11/site-packages/aiohttp/client.py", line 1187, in __aenter__
Nov 08 14:47:41 homeassistant hassio_supervisor[540]:     self._resp = await self._coro
Nov 08 14:47:41 homeassistant hassio_supervisor[540]:                  ^^^^^^^^^^^^^^^^
Nov 08 14:47:41 homeassistant hassio_supervisor[540]:   File "/usr/local/lib/python3.11/site-packages/aiohttp/client.py", line 445, in _request
Nov 08 14:47:41 homeassistant hassio_supervisor[540]:     url = self._build_url(str_or_url)
Nov 08 14:47:41 homeassistant hassio_supervisor[540]:           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
Nov 08 14:47:41 homeassistant hassio_supervisor[540]:   File "/usr/local/lib/python3.11/site-packages/aiohttp/client.py", line 372, in _build_url
Nov 08 14:47:41 homeassistant hassio_supervisor[540]:     url = URL(str_or_url)
Nov 08 14:47:41 homeassistant hassio_supervisor[540]:           ^^^^^^^^^^^^^^^
Nov 08 14:47:41 homeassistant hassio_supervisor[540]:   File "/usr/local/lib/python3.11/site-packages/yarl/_url.py", line 179, in __new__
Nov 08 14:47:41 homeassistant hassio_supervisor[540]:     raise TypeError("Constructor parameter should be str")
Nov 08 14:47:41 homeassistant hassio_supervisor[540]: TypeError: Constructor parameter should be str

```

Looks like there might be an issue with the url construction

bdraco:
`/usr/local/lib/python3.11/site-packages/awesomeversion/awesomeversion.py:455: RuntimeWarning: coroutine 'APIIngress._create_url' was never awaited`

bdraco (changes requested):
Looks like there is a missing await
```diff
diff --git a/supervisor/api/ingress.py b/supervisor/api/ingress.py
index 75179361..1817071a 100644
--- a/supervisor/api/ingress.py
+++ b/supervisor/api/ingress.py
@@ -201,7 +201,7 @@ class APIIngress(CoreSysAttributes):
         session_data: IngressSessionData | None,
     ) -> web.Response | web.StreamResponse:
         """Ingress route for request."""
-        url = self._create_url(addon, path)
+        url = await self._create_url(addon, path)
         source_header = _init_header(request, addon, session_data)
 
         # Passing the raw stream breaks requests for some webservers
```

supervisor/api/ingress.py
```diff
@@ -201,7 +201,7 @@ async def _handle_request(
         session_data: IngressSessionData | None,
     ) -> web.Response | web.StreamResponse:
         """Ingress route for request."""
-        url = self._create_url(addon, path)
+        url = self._async_create_url(addon, path)
```
bdraco, line 204:
```suggestion
        url = await self._async_create_url(addon, path)
```

bdraco:
testing this on my production now

bdraco (approved):
Tested on production, all good now 👍

bdraco:
Screenshot 2023-11-10 at 10 05 33 AM

bdraco (approved):

supervisor/utils/__init__.py
```diff
@@ -39,20 +39,21 @@ async def wrap_api(api, *args, **kwargs):
...
-        sock.close()
-
-        # Check if the port is available
-        if result == 0:
-            return True
-    except OSError:
-        pass
-    return False
+        async with asyncio.timeout(0.5):
+            await loop.sock_connect(sock, (str(address), port))
```
pvizeli, line 50: It seems to be an expensive replacement since it make now the full handshake 🤔

bdraco:
strace of `check_port`

```
socket(AF_INET, SOCK_STREAM|SOCK_CLOEXEC, IPPROTO_IP) = 3
ioctl(3, FIONBIO, [1])                  = 0
connect(3, {sa_family=AF_INET, sin_port=htons(8123), sin_addr=inet_addr("127.0.0.1")}, 16) = -1 EINPROGRESS (Operation in progress)
poll([{fd=3, events=POLLOUT|POLLERR}], 1, 500) = 1 ([{fd=3, revents=POLLOUT}])
getsockopt(3, SOL_SOCKET, SO_ERROR, [0], [4]) = 0
close(3)                                = 0
```


strace of `async_check_port`
```
connect(6, {sa_family=AF_INET, sin_port=htons(8123), sin_addr=inet_addr("127.0.0.1")}, 16) = -1 EINPROGRESS (Operation in progress)
epoll_ctl(3, EPOLL_CTL_ADD, 6, {events=EPOLLOUT, data={u32=6, u64=140595754434566}}) = 0
epoll_pwait(3, [{events=EPOLLOUT, data={u32=6, u64=140595754434566}}], 2, 500, NULL, 8) = 1
getsockopt(6, SOL_SOCKET, SO_ERROR, [0], [4]) = 0
epoll_pwait(3, [{events=EPOLLOUT, data={u32=6, u64=140595754434566}}], 2, 0, NULL, 8) = 1
epoll_ctl(3, EPOLL_CTL_DEL, 6, 0x7ffc0d5cfedc) = 0
close(6)                                = 0
```

bdraco:
Both do a 0.5s poll, the async version does a 0s poll (returns right away as well), but the async version doesn't have the executor overhead

bdraco:
I guess another question is why do the check here https://github.com/home-assistant/supervisor/pull/4677/files#diff-063212876158cc6d68b7582cea3335ca047c893162509e7f1ee1a1bfa89abf2eR143

Maybe it would make sense to do the api calls with a shorter https://docs.aiohttp.org/en/stable/client_reference.html#aiohttp.ClientTimeout ?

pvizeli (approved):
thx @bdraco for the callstack

supervisor/api/addons.py
```diff
@@ -250,7 +250,7 @@ async def info(self, request: web.Request) -> dict[str, Any]:
             ATTR_WEBUI: addon.webui,
             ATTR_INGRESS_ENTRY: addon.ingress_entry,
             ATTR_INGRESS_URL: addon.ingress_url,
-            ATTR_INGRESS_PORT: addon.ingress_port,
+            ATTR_INGRESS_PORT: await addon.get_ingress_port(),
```
pvizeli, line 253: maybe we can call that once on install and afterward we just use the property with the value from cache

pvizeli (changes requested):

supervisor/api/addons.py
pvizeli: What I mean, we register the port once and the add-on use over the full live circle the same port. Means we only need call this async once we install the add-on (or update maybe as well) afterwards it's just a simple property from cached data in memory

pvizeli: Having `check_port` async is fine, just that part with ingress port handling of the PR should get handle better

agners:
> I guess another question is why do the check here https://github.com/home-assistant/supervisor/pull/4677/files#diff-063212876158cc6d68b7582cea3335ca047c893162509e7f1ee1a1bfa89abf2eR143
> 
> Maybe it would make sense to do the api calls with a shorter https://docs.aiohttp.org/en/stable/client_reference.html#aiohttp.ClientTimeout ?

I guess the intention of the code was to fail faster when the port is still closed? But I'd expect that aiohttp returns quickly in that case too no?

I do agree I think we should remove the port check in that case.

agners:
Could we maybe do that change in a separate PR? I am a bit scared that the change will subtle change the behavior of `check_api_state()` which is used in quite some places :thinking:

supervisor/utils/__init__.py
```diff
@@ -39,20 +39,21 @@ async def wrap_api(api, *args, **kwargs):
     return wrap_api
 
 
-def check_port(address: IPv4Address, port: int) -> bool:
+async def async_check_port(
```
mdegat01, line 42: We don't have to change the name of these functions just because we made them `async`.

It's also a bit confusing to do so actually. Core has a standard practice to prefix names of utility methods with `async` when they must be called in the event loop. That doesn't always mean they return a coroutine, non-async methods that have to be called in the event loop follow this naming convention as well, see https://github.com/home-assistant/core/blob/cd9ad32e329a2106b303ed30c661d668347ce02d/homeassistant/config_entries.py#L767 for example.

Supervisor hasn't been following this convention. We can but we have a lot of refactoring to do if we want to. I don't want to introduce this prefix randomly in a few places and confuse people who are coming over from core. We should either adopt the same convention in a refactoring PR or continue to not use it.

supervisor/addons/addon.py
```diff
@@ -395,7 +395,7 @@ def ingress_port(self) -> int | None:
 
         port = self.data[ATTR_INGRESS_PORT]
         if port == 0:
-            return self.sys_ingress.get_dynamic_port(self.slug)
+            _LOGGER.critical("No port set for add-on %s", self.slug)
```
mdegat01, line 398: This isn't right. If this code is actually called in practice things are going to go poorly. The caller will get back `0` and not realize anything is wrong since no exception was raised and the user won't have a clue what to do with this log. I think we have to take one of the following options:

1. Change from a `property` to `async def get_ingress_port` and adjust callers accordingly. Code inside remains the same other then awaiting `get_dynamic_port`
2. This property can return `int | None | asyncio.Task`. Change this code to `return self.sys_create_task(self.sys_ingress.get_dynamic_port(self.slug))`. Callers will need to be adjusted to await the task if that is what they are returned
3. This code raises an exception here. Callers will need to be adjusted to catch that exception and handle it by doing `await self.sys_ingress.get_dynamic_port(addon.slug)`

1 feels the most clean to me even if its not a property. I like 3 the least because it puts this work and knowledge of how to get an ingress port on the caller when it feels like it should be part of the `Addon` class.

mdegat01 (changes requested):

supervisor/addons/addon.py
agners: 1. That was what I had previously, but based on discussions above I changed it to get initialized earlier to avoid making this a get method (see https://github.com/home-assistant/supervisor/pull/4677#discussion_r1391436425).
2. Seems quite an awkward implementation and puts burden on the caller.
3. Similarly.

I agree that if we want dynamic port allocation when this property is used, then 1 is the cleanest solution.

But we took the direction to allocate statically at installation time. With the new implementation, this if statement is more sanitary. My intention was to get it to show up in Sentry in case, contrary to all expectations, it still happens. Maybe better to just throw an exception?

mdegat01: Ok I see pascal's point. In that case I feel like we should probably just make it a fatal exception if its unset rather then returning 0. If it should always be set its ok (and probably preferable) if it fails catastrophically if somehow it happens to get in that state. Similar to how all the properties work in `CoreSys`:
https://github.com/home-assistant/supervisor/blob/176b63df52c54ccccf47c21144a22f271b5e61b4/supervisor/coresys.py#L320-L322

agners: Ok, I've made it a runtime error.

supervisor/utils/__init__.py
pvizeli: > We don't have to change the name of these functions just because we made them `async`.
> 
> It's also a bit confusing to do so actually. Core has a standard practice to prefix names of utility methods with `async` when they must be called in the event loop. That doesn't always mean they return a coroutine, non-async methods that have to be called in the event loop follow this naming convention as well, see https://github.com/home-assistant/core/blob/cd9ad32e329a2106b303ed30c661d668347ce02d/homeassistant/config_entries.py#L767 for example.
> 
> Supervisor hasn't been following this convention. We can but we have a lot of refactoring to do if we want to. I don't want to introduce this prefix randomly in a few places and confuse people who are coming over from core. We should either adopt the same convention in a refactoring PR or continue to not use it.

The supervisor is async per default, if a function is not, they have it normally written in the function description

supervisor/api/ingress.py
```diff
@@ -92,7 +92,7 @@ def _extract_addon(self, request: web.Request) -> Addon:
 
         return addon
 
-    def _create_url(self, addon: Addon, path: str) -> str:
+    async def _async_create_url(self, addon: Addon, path: str) -> str:
```
pvizeli, line 95:
```suggestion
    def _create_url(self, addon: Addon, path: str) -> str:
```
It's even not async

supervisor/utils/__init__.py
```diff
@@ -39,20 +39,21 @@ async def wrap_api(api, *args, **kwargs):
     return wrap_api
 
 
-def check_port(address: IPv4Address, port: int) -> bool:
+async def check_port(
+    loop: asyncio.AbstractEventLoop, address: IPv4Address, port: int
```
pvizeli, line 43: getting the loop is old school, you can call asyncio directly to get current loop back

agners: This came from the implementation proposed by bdraco in https://github.com/home-assistant/supervisor/pull/4678.

@bdraco any particular reason you added the loop as an argument?

supervisor/api/ingress.py
agners: Right, that came from an earlier implementation where it had to be. Will fix 👍

supervisor/utils/__init__.py
agners: So reason is typically performance, however it got optimized e.g. in https://github.com/python/cpython/issues/100363.

I quickly checked, it is still measurable (`get_running_loop()` Mean +- std dev: 1.39 ms +- 0.02 ms`, property lookup Mean +- std dev: 297 us +- 6 us). But that is for 10000 calls, so rather negligible :sweat_smile:  

I'll go for `asyncio.get_running_loop()` to make the function simpler :+1:

mdegat01 (approved):

PR #4694, opened 2023-11-10
Add support for setting target path in map config

jgoakley:
## Type of change
This change is to add support for configuring the target path for custom addons. In order to persist data outside of the addon container certain paths need to be bound. Currently there are several options to bind path into an addon (data, config, share, etc) but the paths inside the container are set.

Some applications and existing docker images have default storage paths that are not easily changed (if at all). Being able to configre these paths will allow simple integration with these applications. I have created a feature request in the Home Assistant Community Forum that provides some more detail:

https://community.home-assistant.io/t/ability-to-configure-container-target-path-for-custom-addon-such-as-wordpress/635615

- [ ] Dependency upgrade
- [X] New feature (which adds functionality to the supervisor)

## Additional information

- Link to documentation pull request: https://github.com/home-assistant/developers.home-assistant/pull/1974
- Link to cli pull request:

## Checklist

- [X] The code change is tested and works locally.
- [X] Local tests pass. **Your PR cannot be merged unless tests pass**
- [X] There is no commented out code in this PR.
- [X] I have followed the [development checklist][dev-checklist]
- [X] The code has been formatted using Black (`black --fast supervisor tests`)
- [X] Tests have been added to verify that the new code works.

- [X] Documentation added/updated for [developers.home-assistant.io][docs-repository]


jgoakley:
Hi!
This is my first time creating a pull request for home assistant. I am not sure how to local tests for the supervisor, but if that is something that is required for me to do I would be glad to do so! I might need some help in that area though.

If there is anything else I can help with or questions I can answer please feel free to let me know!

Thanks!

pvizeli:
Please make any changes to the mapping. Instead, add a new option, `data_path` that allows to change the location inside the container for the data folder.

supervisor/addons/validate.py
```diff
@@ -336,7 +338,17 @@ def _migrate(config: dict[str, Any]):
...
+            vol.Any(
+                vol.Match(RE_VOLUME),
+                vol.Schema(
+                    {
+                        vol.Required('name'): str,
+                        vol.Optional('target'): str
+                    }
+                )
+            )
+        ],
```
pvizeli, line 351: Only allow one format (dict) and convert the other over to this format

supervisor/addons/model.py
```diff
@@ -538,15 +538,22 @@ def need_build(self) -> bool:
         return ATTR_IMAGE not in self.data
 
     @property
-    def map_volumes(self) -> dict[str, bool]:
-        """Return a dict of {volume: read-only} from add-on."""
+    def map_volumes(self) -> dict[str, dict]:
```
pvizeli, line 541: Create a dataclass to handle this:
```
FolderMapping:
  path: Path
  mode: RW/RO
  type: MappingType.XY
```

supervisor/addons/validate.py
jgoakley: Would the existing format (config:rw) be considered deprecated?

supervisor/addons/model.py
jgoakley: Thanks for the feedback! I updated the pull request to use a FolderMapping dataclass

jgoakley:
I added the FolderMapping dataclass and made the corresponding changes. I tried for the last couple of hours to get a local supervisor running against the main branch with no success. Here is the error message in case it is helpful:

`ERROR: failed to solve: process "/bin/ash -o pipefail -c set -x     && apk add --no-cache         coreutils         eudev         eudev-libs         git         libffi         libpulse         musl         openssl         yaml         && curl -Lso /usr/bin/cosign \"https://github.com/home-assistant/cosign/releases/download/${COSIGN_VERSION}/cosign_${BUILD_ARCH}\"     && chmod a+x /usr/bin/cosign" did not complete successfully: exit code: 6`

Anyway, the reason I say that is because there may be issues that I'm not able to test so if you encounter any issues with my changes please let me know and I will do my best to resolve them.

Thanks!

pvizeli:
> I added the FolderMapping dataclass and made the corresponding changes. I tried for the last couple of hours to get a local supervisor running against the main branch with no success. Here is the error message in case it is helpful:
> 
> `ERROR: failed to solve: process "/bin/ash -o pipefail -c set -x && apk add --no-cache coreutils eudev eudev-libs git libffi libpulse musl openssl yaml && curl -Lso /usr/bin/cosign \"https://github.com/home-assistant/cosign/releases/download/${COSIGN_VERSION}/cosign_${BUILD_ARCH}\" && chmod a+x /usr/bin/cosign" did not complete successfully: exit code: 6`
> 
> Anyway, the reason I say that is because there may be issues that I'm not able to test so if you encounter any issues with my changes please let me know and I will do my best to resolve them.
> 
> Thanks!

You need a new version of devcontainer

supervisor/const.py
```diff
@@ -423,6 +412,21 @@ class AddonState(StrEnum):
     ERROR = "error"
 
 
+class MappingType(StrEnum):
```
pvizeli, line 415: Copy this class into `addons/const.py` - we migrate away from this big single const.py file

supervisor/addons/configuration.py
```diff
@@ -0,0 +1,14 @@
+"""Confgiuration Objects for Addon Config."""
```
pvizeli, line 1: let's rename `addons/configuration.py` to `addons/data.py` like we use on other place

pvizeli (commented):
Looks good, thank you!

supervisor/addons/configuration.py
jgoakley: Looks like `addons/data.py` already exists and it depends on `store.addon.AddonStore` which depends on `addons.model.AddonModel` which is where we need our `FolderMapping` dataclass. Moving `FolderMapping` to data.py creates a circular reference:

`ImportError: cannot import name 'AddonStore' from partially initialized module 'supervisor.store.addon' (most likely due to a circular import) (/usr/src/supervisor/supervisor/store/addon.py)`

Is there somewhere better for the `FolderMapping` dataclass or does `addons/data.py` need to be refactored? I left it in `configuration.py` for now.

jgoakley:
> > I added the FolderMapping dataclass and made the corresponding changes. I tried for the last couple of hours to get a local supervisor running against the main branch with no success. Here is the error message in case it is helpful:
> > `ERROR: failed to solve: process "/bin/ash -o pipefail -c set -x && apk add --no-cache coreutils eudev eudev-libs git libffi libpulse musl openssl yaml && curl -Lso /usr/bin/cosign \"https://github.com/home-assistant/cosign/releases/download/${COSIGN_VERSION}/cosign_${BUILD_ARCH}\" && chmod a+x /usr/bin/cosign" did not complete successfully: exit code: 6`
> > Anyway, the reason I say that is because there may be issues that I'm not able to test so if you encounter any issues with my changes please let me know and I will do my best to resolve them.
> > Thanks!
> 
> You need a new version of devcontainer -> https://github.com/home-assistant/devcontainer

I was able to figure out the devcontainer issue in case it helps anyone in the future. I'm using Docker Desktop and for some reason it was not able to resolve objects.githubusercontent.com when running the `curl` command in the container. I added the following to my `daemon.json` file and that resolved the issue:

```
"dns": [
    "8.8.8.8"
]
```

After that I was able to debug the supervisor changes I made and run the tests and it looks to be working great!

Feel free to let me know if there's anything else I need to do. Thanks!

supervisor/addons/configuration.py
pvizeli: Yeah, you have to rename the `__init__.py` inside `addons` to `manager.py` and create an empty new `__init__.py`. We did that already for some module but only if it was needed because of the shifting from const.py

supervisor/addons/model.py
```diff
@@ -538,14 +539,13 @@ def need_build(self) -> bool:
...
+        """Return a dict of {MappingType: FolderMapping} from add-on."""
         volumes = {}
         for volume in self.data[ATTR_MAP]:
-            result = RE_VOLUME.match(volume)
-            if not result:
-                continue
-            volumes[result.group(1)] = result.group(2) != "rw"
+            volumes[MappingType(volume.get("type"))] = FolderMapping(
+                volume.get("path"), volume.get("read_only")
+            )
```
mdegat01, line 548:
```suggestion
            volumes[MappingType(volume[ATTR_TYPE)] = FolderMapping(
                volume.get(ATTR_PATH), volume[ATTR_READ_ONLY]
            )
```
Since type is required, should not use `get`. Also `MappingType(None)` isn't valid and would fail anyway so using `get` doesn't stop an error here. And the schema should default `read_only` to `True` as suggested in my other comment so that ensures it always exists as well. Plus if `read_only` was actually set to `None` it would break here:
https://github.com/home-assistant/supervisor/blob/40bf7c081b4ad38fdfe661ce0ca88da2a770e09b/supervisor/docker/addon.py#L348

supervisor/addons/configuration.py
```diff
@@ -0,0 +1,11 @@
...
+
+from dataclasses import dataclass
+
+
+@dataclass(slots=True)
+class FolderMapping:
+    """Represent folder mapping configuration."""
+
+    path: str
```
mdegat01, line 10:
```suggestion
    path: str | None
```
The way these instances are created elsewhere in this PR the value for `path` may be `None`, the type should reflect that.

supervisor/addons/validate.py
```diff
@@ -336,7 +359,15 @@ def _migrate(config: dict[str, Any]):
         vol.Optional(ATTR_DEVICES): [str],
         vol.Optional(ATTR_UDEV, default=False): vol.Boolean(),
         vol.Optional(ATTR_TMPFS, default=False): vol.Boolean(),
-        vol.Optional(ATTR_MAP, default=list): [vol.Match(RE_VOLUME)],
+        vol.Optional(ATTR_MAP, default=list): [
+            vol.Schema(
+                {
+                    vol.Required("type"): str,
+                    vol.Optional("read_only"): bool,
```
mdegat01, line 366:
```suggestion
                    vol.Optional(ATTR_READ_ONLY, default=True): bool,
```
If this value is omitted, it should default to `True`. This keeps us from making `FolderMapping` instances with `None` for `read_only` and causing breakages here:
https://github.com/home-assistant/supervisor/blob/40bf7c081b4ad38fdfe661ce0ca88da2a770e09b/supervisor/docker/addon.py#L348

`ATTR_READ_ONLY` doesn't exist currently, it will need to be defined in `supervisor.const` or `supervisor.addons.const` so it can be reused.

supervisor/addons/validate.py
```diff
@@ -336,7 +359,15 @@ def _migrate(config: dict[str, Any]):
...
+                    vol.Optional("path"): str,
```
mdegat01, line 367:
```suggestion
                    vol.Optional(ATTR_PATH): str,
```
As mentioned above we prefer constants when the value is reused. `ATTR_PATH` doesn't currently exist in `supervisor.const` or `supervisor.addons.const` so it'll have to be defined.

And simply `str` for `path` seems a bit loose. At a minimum it has to start with `/` since we want absolute paths not relative ones. Also multiple of these cannot have the same value for path, that would need be validated in a separate validator function. And I'm not sure we should allow choosing absolutely any path within the container. @pvizeli and @agners are there any paths we don't want to allow here?

mdegat01 (changes requested):
This is a good idea. A few small changes and I think we should think whether there are any particular values for `path` to validate against.

supervisor/addons/addon.py
```diff
@@ -464,7 +464,7 @@ def path_extern_data(self) -> PurePath:
     @property
     def addon_config_used(self) -> bool:
         """Add-on is using its public config folder."""
-        return MAP_ADDON_CONFIG in self.map_volumes
+        return self.map_volumes.get(MappingType.ADDON_CONFIG) is not None
```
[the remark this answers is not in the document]
jgoakley: Your approach (like the original) is definitely better. I am very new to python so thank you for the feedback!

supervisor/addons/validate.py
```diff
@@ -265,26 +262,52 @@ def _migrate(config: dict[str, Any]):
...
+            )
+
+        if volumes:
+            config[ATTR_MAP] = volumes
+
         # 2023-10 "config" became "homeassistant" so /config can be used for addon's public config
-        volumes = [RE_VOLUME.match(entry) for entry in config.get(ATTR_MAP, [])]
-        if any(volume and volume.group(1) == MAP_CONFIG for volume in volumes):
+        if any(
+            volume and volume.get("type") == MappingType.CONFIG for volume in volumes
```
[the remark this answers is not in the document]
jgoakley: I agree! Thank you for these improvement!

supervisor/addons/validate.py
```diff
@@ -336,7 +359,15 @@ def _migrate(config: dict[str, Any]):
         vol.Optional(ATTR_DEVICES): [str],
         vol.Optional(ATTR_UDEV, default=False): vol.Boolean(),
         vol.Optional(ATTR_TMPFS, default=False): vol.Boolean(),
-        vol.Optional(ATTR_MAP, default=list): [vol.Match(RE_VOLUME)],
+        vol.Optional(ATTR_MAP, default=list): [
+            vol.Schema(
+                {
+                    vol.Required("type"): str,
```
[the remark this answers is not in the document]
jgoakley: Thank you for showing me this! I was wondering how to validate this in the schema.

supervisor/addons/validate.py
```diff
@@ -265,26 +262,52 @@ def _migrate(config: dict[str, Any]):
...
+
+        for entry in [
+            RE_VOLUME.match(entry)
+            for entry in config.get(ATTR_MAP, [])
+            if isinstance(entry, str)
+        ]:
+            if not entry:
+                continue
+            volumes.append(
+                {"type": entry.group(1), "read_only": entry.group(2) != "rw"}
```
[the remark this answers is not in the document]
jgoakley: I agree! Thank you for your patience as I get more acquainted with best practices.

supervisor/addons/validate.py
```diff
@@ -265,26 +262,52 @@ def _migrate(config: dict[str, Any]):
...
 
+        # 2023-11 "map" entries can also be dict to allow path configuration
+        volumes = []
+        for entry in [
+            entry for entry in config.get(ATTR_MAP, []) if isinstance(entry, dict)
+        ]:
+            result = RE_VOLUME.match(entry.get("type"))
+            if not result:
+                continue
+            volumes.append(entry)
```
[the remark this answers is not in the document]
jgoakley: Thanks for the feedback @mdegat01! To support both `str` and `dict` entries in the `config[ATTR_MAP]` array I have two loops that filter for each and append to the `volumes` array. At the end, I overwrite `config[ATTR_MAP]` with the `volumes` array that was built.

I agree that we can get rid of this validation:

```
result = RE_VOLUME.match(entry.get("type"))
    if not result:
            continue
```

And both of the loops can also be simplified to this:

```
for entry in map:
    if isinstance(entry, dict):
        volumes.append(entry)
    if isinstance(entry, str):
        result = RE_VOLUME.match(entry)
        if not result:
            continue
        volumes.append(
            { ATTR_TYPE:result.group(1), ATTR_READ_ONLY: result.group(2) != "rw" }
        )
```

But if we get rid of the whole loop then they won't be in the final result. Or is there a better approach to transforming the `str` and `dict` entries into one format?

supervisor/addons/validate.py
jgoakley: That's a great point. I did some local testing with several conditions and the containers fail to start. The error messages from Docker make sense in each case:

image
image
image

I'm not sure what Supervisor best practices are, should these conditions also be validated at the config level so as to catch these issues earlier? Or is it sufficient that these are handled when trying to start the Addon?

supervisor/addons/configuration.py
jgoakley: I'm so sorry, I'm new to this repo so please bear with me. Should the `FolderMapping` dataclass be moved into the new `addons/manager.py`? Or should the logic in `addons/data.py` be moved to `addons/manager.py` and then `FolderMapping` go into `addons/data.py`?

Thanks for you help!

supervisor/addons/validate.py
agners: I don't seem to come up with a good reason to filter certain paths. I mean you can do silly things, but it will affect your add-on only, so :man_shrugging:

agners: > I'm not sure what Supervisor best practices are, should these conditions also be validated at the config level so as to catch these issues earlier? Or is it sufficient that these are handled when trying to start the Addon?

I think I would not try to "reimplement" these checks on config levels. An add-on developer certainly starts his add-on once before deploying :sweat_smile: 

Not sure if we can make the Docker errors a bit nicer, e.g. strip the API/http related part. But that is something which bubbles up also in other places, and might be worth addressing separately.

pvizeli: We can just make an validation function to load the string as Path (and return also a Path) - but also check if that is an absolute Path. Pathib have 2 Path object - virtual or physic - we just can use the virtual version to validate

jgoakley: Hi! Sorry @pvizeli, I was out of town on vacation last week. That sounds like a good plan! I can add the path validation logic. I should be able to get to that tomorrow.

Any thoughts on the discussion above?
https://github.com/home-assistant/supervisor/pull/4694#discussion_r1399723085

supervisor/addons/configuration.py
mdegat01: FolderMapping goes into data.py. Essentially you just rename `addons/__init__.py` to `addons/manager.py` and then make a new `addons/__init__.py` that only has a docstring. It's a small change but can be annoying since you have to refactor all existing references to it (well VSCode does most of the work but patches in test sometimes need manual fixing).

To keep it simple, I made a PR just for that. @pvizeli let's get #4760 merged so @jgoakley can rebase off it and not have the circular import issue.

supervisor/addons/validate.py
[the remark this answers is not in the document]
mdegat01: That works. I was only highlighting the loop that specifically went through the entries which are dictionaries and saying that was unnecessary. I agree that you do have to go through and make the string entries into dictionary ones, I wasn't suggesting to remove that.

mdegat01 (approved):
Once #4760 goes this can be rebased off that and then the dataclass moved to `addons/data.py` as per pvizeli's suggestion. Looks like its all ready to go at that point.

supervisor/addons/validate.py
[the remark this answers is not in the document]
jgoakley: Sounds great! Thanks for clarifying!

supervisor/addons/configuration.py
jgoakley: Thanks @mdegat01 for creating that pull request! After it was merged I updated my branch to include those changes but I'm still getting the circular reference error when moving the `FolderMapping` dataclass to `addons/data.py`

I made the change anyway and pushed it to the branch so everyone can see the issue. I may be adding the dataclass incorrectly so please let me know if there is anything I should do differently. Thanks!

mdegat01:
Oh shoot. @pvizeli we can't put this in `addons/data.py` as that is already taken by something entirely different. The class for managing config data of an addon is in there and so yea we now get a circular import issue trying to use it for common dataclasses.

How about `addons/configuration.py` like we do in `resolution` instead?

jgoakley:
> Oh shoot. @pvizeli we can't put this in `addons/data.py` as that is already taken by something entirely different. The class for managing config data of an addon is in there and so yea we now get a circular import issue trying to use it for common dataclasses.
> 
> How about `addons/configuration.py` like we do in `resolution` instead?

@mdegat01 I moved the FolderMapping dataclass to configuration.py as you suggested. If there are any other changes I should make just let me know!

mdegat01 (approved):

mdegat01:
@jgoakley it looks like there's ocnflicts with the base branch. Can you rebase and resolve that? Then this is good to get merged.

jgoakley:
> @jgoakley it looks like there's conflicts with the base branch. Can you rebase and resolve that? Then this is good to get merged.

Hi @mdegat01, I think I did a merge commit instead of a rebase when clicked "update branch" in GitHub. Sorry about that, I'm not very familiar with rebasing. I also synced my fork and performed a merge in VS Code but I didn't see any conflicts. Is "main" the correct base branch I should be merging into?

mdegat01 (approved):

mdegat01:
@jgoakley all good on the merge vs. rebase. We prefer rebases since it makes the PR easier to review as the only commits are your changes, not the merges. But we squash and merge in the end so it doesn't matter after its merged.

Thanks for the contribution 👍

PR #4702, opened 2023-11-13
Revert "Revert Home Assistant configuration to `/config`"

frenck:
Reverts home-assistant/supervisor#4697

As discussed, reverting this change.

Reasoning:

- `/config` will always point to the active config in the matching container. The Home Assistant container will have its config in `/config`, all plugins have their config in `/config` and add-ons will now have their config in `/config` too.
- `/homeassistant` is where other containers can access the Home Assistant configuration (if needed, but in general, their should not need it).

The above will add consistency to the developer experience (as part of the original intention was for this specific case).

For user-facing add-ons (like ssh, samba, vscode, file editor, e.d.), they will be adopted that they map `config` to `/homeassistant`, so that the change is transparent/non-breaking from the user experience.

### DIFF
--- a/supervisor/docker/addon.py
+++ b/supervisor/docker/addon.py
@@ -360,18 +360,18 @@ class DockerAddon(DockerInterface):
                     Mount(
                         type=MountType.BIND,
                         source=self.addon.path_extern_config.as_posix(),
-                        target="/addon_config",
+                        target="/config",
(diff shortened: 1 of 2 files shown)


pvizeli (approved):

agners (approved):
Granted, it makes things consistent technically... But I think this adds more confusion than necessary. 

But if this is preferred way to go, I am ok with it!

Issue #4711, opened 2023-11-16
Partial restore wipes Network storage mounted under /share


phellarv:
### Describe the issue you are experiencing

I have the Immich-addon installed.
Under /share/Immich I have mounted a remote cifs-directory from another server I have.
When I did a partial restore (including /share), the restore completely wiped the remote storage.
This is not expected behavior.

### What type of installation are you running?

Home Assistant OS

### Which operating system are you running on?

Home Assistant Operating System

### Steps to reproduce the issue

1. Take full backup with remote storage attached to a directory under /share
2. Do a partial restore including /share
3. Observe that existing files under /share/<remote-storage> is wiped

### Anything in the Supervisor logs that might be useful for us?

```txt
I do not have any logs right now.
```

### System Health information

"There are currently no repairs available"

### Supervisor diagnostics

[config_entry-hassio-6fc8ace0e1a6b3eb9eaaa3455ff54969.json.txt](https://github.com/home-assistant/supervisor/files/13376656/config_entry-hassio-6fc8ace0e1a6b3eb9eaaa3455ff54969.json.txt)

phellarv:
Forgot this:
<div class="card-header"><h3>Home Assistant Supervisor</h3>  </div>  
Host Operating System | Home Assistant OS 11.1
-- | --
Update Channel | stable
Supervisor Version | supervisor-2023.11.3
Agent Version | 1.6.0
Docker Version | 24.0.6
Disk Total | 116.7 GB
Disk Used | 40.6 GB
Healthy | true
Supported | true
Board | generic-x86-64
Supervisor API | ok
Version API | ok
Installed Add-ons | Terminal  & SSH (9.8.1), MariaDB (2.6.1), InfluxDB (4.8.0), Home Assistant  Google Drive Backup (0.111.1), ESPHome (2023.11.1), Node-RED (16.0.2),  EMQX (0.2.12), Zigbee2MQTT (1.33.2-1), Studio Code Server (5.14.2),  Prometheus Node Exporter (0.7.0), Matter Server (4.10.2), Silicon Labs  Multiprotocol (2.3.2), Whisper (1.0.0), Piper (1.4.0), openWakeWord  (1.8.2), File editor (5.7.0)

agners:
> 3. Observe that existing files under /share/ is wiped

The contents under /share are expected to be wiped, since you restore that folder. But the remote content (in your case under `/share/Immich`) should not get wiped! 

[...]

Issue #4726, opened 2023-11-25
Restoring HAOS backup on Yellow wiped the attached network drives over SAMBA.


Siress:
### The problem

TL;DR - or how to replicate: Using HA Yellow with OS v. 2023.11.3. Connect a network drive over SAMBA title "media". Create backups using HA Google Drive Backup v0.111.1 add-on. Restore the locally-generated backup using HAOS. Watch contents of 'media' network drive vanish.

Can anyone replicate this? I'm going off of log files and correlation with n=1.

It took me quite a while to piece together why my Media folder went from 8TB to a few empty folders. Checking the ZFS logs gave me the timestamp that the change happened. My NAS has been a rock solid appliance and nothing else was out of the ordinary, so I was sure it had to be something with credentials to execute a deletion. That helped limit it to just a few devices. Foolishly, I had given that privilege to HAOS. Originally, I was giving it access to a share to store my HAOS backups and database. In all of my excitement for HAOS's capabilities, I foolishly added my Media share just to experiment with HAOS's media features.

Given the timestamp and the limited devices with the authority to make the change, the only devices that were active at that time were my HA Yellow and my workstation - as I was integrating various things into HA. After one of the integrations went sideways, I decided to restore a backup. I created a backup just before restoring, just in case. I'm using the Home Assistant Google Drive Backup (0.111.1) add-on to create backups, so they go to the cloud as well as my NAS. Keep in mind that it does not restore backups. I then went through HAOS and restored the backup from the NAS. The time stamp on that backup coincided with the time stamp in the ZFS logs. I then realized that it was authorized to delete the Backups dataset as well, but it hadn't deleted it's own backups. Maybe that's useful knowledge for whomever can fix this issue.

Shame on me for granting excess privilege to HAOS. I take full responsibility for that part of this story. I love HAOS, but this is one of the worst bugs I've come across. This needs immediate action.

### What version of Home Assistant Core has the issue?

core-2023.11.3

### What was the last working version of Home Assistant Core?

core-2023.11.3

### What type of installation are you running?

Home Assistant OS

agners:
This very much sounds like #4652, however we added additional safeguards so that would not happen anymore. I try to reproduce your issue on my end again.

agners:
> Shame on me for granting excess privilege to HAOS. I take full responsibility for that part of this story. I love HAOS, but this is one of the worst bugs I've come across. This needs immediate action.

I am very sorry what happened, and I agree this is a really bad bug!

[...]

PR #4733, opened 2023-11-27
Wait until mount unit is deactivated on unmount

agners:
## Proposed change
The current code does not wait until the (bind) mount unit has been actually deactivated (state "inactive"). This is especially problematic when restoring a backup, where we deactivate all bind mounts before restoring the target folder. Before the tarball is actually restored, we delete all contents of the target folder. This lead to the situation where the "rm -rf" command got executed before the bind mount actually got unmounted.

The current code polls the state using an exponentially increasing delay. Wait up to 30s for the bind mount to actually deactivate.

Fixes: #4711, #4726

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/mounts/mount.py
+++ b/supervisor/mounts/mount.py
@@ -201,11 +201,11 @@ class Mount(CoreSysAttributes, ABC):
         ):
             self.sys_resolution.dismiss_issue(self.failed_issue)
 
-    async def _update_state_await(self, expected_state: UnitActiveState) -> None:
+    async def _update_state_await(self, expected_states: list[UnitActiveState]) -> None:
         """Update state info about mount from dbus. Wait up to 30 seconds for the state to appear."""
         for i in range(5):
             await self.update_state()
-            if self.state == expected_state:
+            if self.state in expected_states:
                 return
             await asyncio.sleep(i**2)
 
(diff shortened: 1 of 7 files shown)


pvizeli:
Only for public history, we looking with @mdegat01 if that is possible to use property change events instead polling

mdegat01:
@pvizeli you're right and we can but I don't think we should do that in this PR. I'll add a refactoring task to switch this point and the existing point using polling to listen for a signal

supervisor/mounts/mount.py
```diff
@@ -197,6 +201,20 @@ async def update(self) -> None:
         ):
             self.sys_resolution.dismiss_issue(self.failed_issue)
 
+    async def _update_state_await(self, expected_state: UnitActiveState) -> None:
+        """Update state info about mount from dbus. Wait up to 30 seconds for the state to appear."""
```
mdegat01, line 205: Since we're doing exponential waits now, this comment is incorrect. I could be up to about a minute

supervisor/mounts/mount.py
```diff
@@ -273,6 +291,8 @@ async def unmount(self) -> None:
                 await self.sys_dbus.systemd.reset_failed_unit(self.unit_name)
             else:
                 await self.sys_dbus.systemd.stop_unit(self.unit_name, StopUnitMode.FAIL)
+
+            await self._update_state_await(UnitActiveState.INACTIVE)
```
mdegat01, line 295: It may not go to `INACTIVE`, it could also fail. We really want to know when it has transitioned to any state that is considered stopped.

supervisor/mounts/mount.py
agners:
```
>>> total = 0
>>> for i in range(5):
...     total = total + i**2
... 
>>> print(total)
30
```

supervisor/mounts/mount.py
agners: Yeah I was thinking that we would just timeout, no big deal :man_shrugging:

> We really want to know when it has transitioned to any state that is considered stopped.

That begs the question what states there are. Systemd sources list these:

https://github.com/systemd/systemd/blob/v249.12/src/basic/unit-def.c#L99-L107

There is also "maintenance" and "reloading" :thinking: 

We should probably rely on these high level categorizations:
https://github.com/systemd/systemd/blob/v249.12/src/core/unit.h#L38-L52

But yeah, instead of timing out, we really can do it smarter rather easily, so why not. Updated the code to also wait for "failed" state.

mdegat01:
@agners I implemented the change we discussed for mocking to eliminate some of the really long lists of state transitions in the tests. I left the ones testing failure conditions largely alone as those weren't as easily replaceable but it is a bit nicer for the success cases now.

mdegat01 (approved):
ok @agners I think this is ready now if you want to take a look

agners (commented):
LGTM! Thanks!

PR #4770, opened 2023-12-13
Migrate to pyproject.toml where possible

mdegat01:
## Proposed change

Migrate as much as possible from `setup.py`, `setup.cfg` and the various other config files to `pyproject.toml`. This basically covers everything except `version` (because we set it dynamically from `supervisor.const` where a GitHub action fills in the new value) and `flake8` (because it appears it has no `pyproject.toml` support).

The `flake8` gap is irrelevant because the follow-up to this is to replace `black`, `isort`, `flake8` and most of `pylint` with `ruff`. For now this PR just seeks to largely maintain the status quo in a new format. Although I did use [core's pyproject.toml](https://github.com/home-assistant/core/blob/dev/pyproject.toml) as a reference and borrowed from there at times where it seemed to make sense.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist


pyproject.toml
```diff
@@ -0,0 +1,112 @@
+[build-system]
+requires = ["setuptools~=68.0.0", "wheel~=0.40.0"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "Supervisor"
+dynamic = ["version", "dependencies"]
```
mdegat01, line 7: @frenck I like your approach of generating `requirements.txt` in github actions better then the `dynamic` one I'm using here. The problem I have though is that `requirements.txt` is used in a few places that require more thinking. The two I noticed right away were:
1. The [CI workflow](https://github.com/home-assistant/supervisor/blob/fd3c995c7c9b73d24dce90f3a1371330b89965c0/.github/workflows/ci.yaml#L47). Normally when our workflows change files in the repo that is done on merge of a PR in the builder workflow, not the CI workflow that runs on PR change to check it.
2. [Devcontainer setup](https://github.com/home-assistant/devcontainer/blob/dee6e882b7eeadd45e54fbf8dd9544cc3068557c/supervisor/rootfs/usr/bin/devcontainer_bootstrap#L8). Since that will look at the local requirements file, forcing the user to change it locally to test if supervisor builds

This approach keeps the status quo while still migrating to `pyproject.toml`. Then I can come back and figure out how to get everything into this one file in a separate PR.

pvizeli (approved):
I'm fine with it, if it's working. We not publish anything to pipy - it's only installed during the image generation

PR #4832, opened 2024-01-23
Addon devs can block auto update for breaking versions

mdegat01:
## Proposed change

Provide addon developers with a way to list breaking versions of their addons. This way we can block auto-updates which would cross one of these breaking versions and prevent bad situations like what happened recently with Nginx Proxy Manager.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to documentation pull request: https://github.com/home-assistant/developers.home-assistant/pull/2056
- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/addons/validate.py
+++ b/supervisor/addons/validate.py
@@ -112,6 +112,7 @@ from ..validate import (
 )
 from .const import (
     ATTR_BACKUP,
+    ATTR_BREAKING_VERSIONS,
     ATTR_CODENOTARY,
     ATTR_PATH,
     ATTR_READ_ONLY,
@@ -422,6 +423,7 @@ _SCHEMA_ADDON_CONFIG = vol.Schema(
             vol.Coerce(int), vol.Range(min=10, max=300)
         ),
         vol.Optional(ATTR_JOURNALD, default=False): vol.Boolean(),
+        vol.Optional(ATTR_BREAKING_VERSIONS, default=list): [version_tag],
     },
     extra=vol.REMOVE_EXTRA,
 )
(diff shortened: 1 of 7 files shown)


supervisor/addons/validate.py
```diff
@@ -422,6 +423,7 @@ def _migrate(config: dict[str, Any]):
             vol.Coerce(int), vol.Range(min=10, max=300)
         ),
         vol.Optional(ATTR_JOURNALD, default=False): vol.Boolean(),
+        vol.Optional(ATTR_BREAKING_VERSIONS, default=list): [version_tag],
```
frenck, line 426: I am not fully understand, why this is a list?

As in, you can only upgrade to a latest version, right? Would't that mean only the higher version number that is breaking would matter?

pvizeli: User can still have an old version and this information prevent from auto update if they turn it on while people on never version run as it should.

pvizeli: And the API support handling multiple versions, only git not. In theory we can use backups to provide ways to having multiple version. I think that way is flexible without having update the code + support a wide range of defragmentation

mdegat01: Well tbf @pvizeli the feature is about auto-update. Auto-update is always going to go to the latest version. Manual update may not in the future but manual update won't check this list. We won't stop you from manually updating over a breaking change.

I guess I figured if a user was woefully out of date it could be useful to see all the versions with breaking changes they had to cross over. Also if the version scheme does change we might not be able to compare the addon's version with the latest versions but could still compare against old ones. It also just seemed like a useful record for users and the addon dev.

But I can change it to a simple string if we don't think its helpful.

pvizeli (approved):

Issue #355, opened 2018-02-11
Enable support for IPv6


alexbarcelo:
The `hassio` bridge network doesn't has the IPv6 network stack enabled.

"Culprit" seems to be the line at `hassio/docker/network.py`, in particular the parameter `enable_ipv6=False` in the method `_get_network`.

Enabling is not as straightforward as puting there `True` because to set up correctly IPv6 a subnet should be specified. And that subnet is typically provided by the user (if the user is doing globally routable addresses, which is a likely scenario in IPv6 environments).

I haven't come up with a proposal on how to address that. I was open to suggestions (if a PR is feasible and appreciated, I can work on that) or some workaround tips otherwise (if there are any).

Fubi95:
Go to the HomeAssistant Config file under component http and add server_host: "::0"

alexbarcelo:
@Fubi95 yeah, I added that to the documentation. Note that the homeassistant container is *not* part of the `hassio` bridge, as that container uses host network mode and thus works correctly with IPv6 if the host has IPv6.

However, the problem I found is for the addons, which typically use the `hassio` bridge which --when created-- has an explicit disable of ipv6.

I have abandoned this route, as it started to seem (at least to me) painful and cumbersome. And Docker itself has no clean semantics and features for IPv6 stack of containers IMHO. So not sure if it is worth having this issue opened, as I don't know if it will benefit anybody in the short term.

donnib:
@alexbarcelo I would love to expose my hassio on IPv6 but as far as i understand it's not possible now, correct ?

alexbarcelo:
@donnib You can expose hassio to IPv6 because, as I said, that container is *not* part of the `hassio` bridge. According to my comment (which I honestly don't remember at all) I updated something in the documentation, probably related to the @Fubi95 comment.

That will expose hassio to IPv6, but will not work for any other container in the hassio environment. That was my use case, and was not supported. ATM I am using vanila dockerized hass, so you should try it yourself /shrug

alexbarcelo:
(see https://www.home-assistant.io/components/http/ > `server_host`)

donnib:
@alexbarcelo the doc says explicitly : Don’t use option server_host on a Hass.io installation!

alexbarcelo:
So... either I did not see it, or it was not there when I was toying around.

If somebody wants to try it, go ahead. I have stopped using Hass.io a while ago so I will keep an eye on the project but I am not doing anything with it at the moment.

[...]

PR #4834, opened 2024-01-24
Expand docker version evaluation.

donaldguy:
**EDIT:** apparently contrary to my impression of their plans, they cut a docker 25.0.1 this morning, so this is kinda for no one now. The broader idea _could_ still be helpful in future, but lmk if just want to close

## Proposed change

1. Expand the `EvaluateDockerVersion` to include both a minimum version check and a list of known bad versions for poor supervised-installer users. (I would also accept, even perhaps prefer e.g. a constraint in https://github.com/home-assistant/supervised-installer/blob/main/homeassistant-supervised/DEBIAN/control but this is already here)

2. expand test suite accordingly

3. do away with the apparently otherwise unused `supported_version` property of DockerInfo, concentrating that logic in the evaluation itself

## Type of change

- [ ] Dependency upgrade
- [X] Code quality improvements to existing code or addition of tests

## Additional information

- This PR is related to issues: 
    - #4827 ; and by extension at least
    - home-assistant/supervised-installer#349
    - home-assistant/supervised-installer#351
    - home-assistant/supervised-installer#352

As I discussed at https://github.com/home-assistant/supervisor/issues/4827#issuecomment-1907090382

there is already a fix landed upstream in moby for this ~bug but they are not interested in fast-tracking a patch release apparently.

Ditto is perhaps foolish to suspect a supervisor release would be forthcoming either, but I'd already spent ~5 hours struggling with debugging this issue myself (errantly over-focalizing on pyudev getting denied on `udev_monitor_new_from_netlink`... for some reason), so figured might as well spend 2 hours more closing the loop

- Link to documentation pull request: [... there should perhaps be one for https://www.home-assistant.io/more-info/unsupported/docker_version, but I'll let reviewer have say first ]

## Checklist

    I have verified that the additions do not hamper supervisor (and enough else for the setup wizard to showup at `http://localhost:9123`) startup on docker 24.0.7 in devcontainer (running under orbstack on an M1 MacBook air). 
    
    I have not yet investigated exact behavior/ordering of failures during startup under docker-ce 25.0.0

### DIFF
--- a/tests/resolution/evaluation/test_evaluate_docker_version.py
+++ b/tests/resolution/evaluation/test_evaluate_docker_version.py
@@ -1,26 +1,69 @@
-"""Test evaluation base."""
-# pylint: disable=import-error,protected-access
+"""Test evaluation of docker version."""
+
+import random
 from unittest.mock import patch
+
+# pylint: disable=import-error,protected-access
+from awesomeversion import AwesomeVersion
 
 from supervisor.const import CoreState
 from supervisor.coresys import CoreSys
 from supervisor.resolution.evaluations.docker_version import EvaluateDockerVersion
 
 
-async def test_evaluation(coresys: CoreSys):
-    """Test evaluation."""
+async def test_evaluation_supported(coresys: CoreSys):
+    """Test real evaluation with a current docker daemon."""
+    docker_version = EvaluateDockerVersion(coresys)
+    coresys.core.state = CoreState.INITIALIZE
+    coresys.docker.info.version = AwesomeVersion("24.0.7")
+
+    assert docker_version.reason not in coresys.resolution.unsupported
+
+    await docker_version()
+
+    assert docker_version.reason not in coresys.resolution.unsupported
+
+
+async def test_evaluation_below_min_version(coresys: CoreSys):
+    """Test handling of too-low docker versions."""
     docker_version = EvaluateDockerVersion(coresys)
     coresys.core.state = CoreState.INITIALIZE
 
     assert docker_version.reason not in coresys.resolution.unsupported
 
-    coresys.docker.info.supported_version = False
+    minv = docker_version.min_supported_version
+    coresys.docker.info.version = AwesomeVersion(f"{int(minv.major) - 1}.0.0")
     await docker_version()
     assert docker_version.reason in coresys.resolution.unsupported
 
-    coresys.docker.info.supported_version = True
+    coresys.resolution.unsupported = []
+    coresys.docker.info.version = AwesomeVersion(
+        f"{minv.major}.{minv.minor}.{int(minv.patch) - 1}"
+    )
+    await docker_version()
+    assert docker_version.reason in coresys.resolution.unsupported
+
+
+async def test_evaluation_at_min_version(coresys: CoreSys):
+    """Test handling of just-too-low docker versions."""
+    docker_version = EvaluateDockerVersion(coresys)
+    coresys.core.state = CoreState.INITIALIZE
+
+    coresys.docker.info.version = docker_version.min_supported_version
     await docker_version()
     assert docker_version.reason not in coresys.resolution.unsupported
+
+
+async def test_evaluation_known_broken_version(coresys: CoreSys):
+    """Test handling of a known-bad docker versions."""
+    docker_version = EvaluateDockerVersion(coresys)
+    coresys.core.state = CoreState.INITIALIZE
+
+    coresys.docker.info.version = random.choice(
+        list(docker_version.broken_versions.keys())
+    )
+    await docker_version()
+    assert docker_version.reason in coresys.resolution.unsupported
 
 
 async def test_did_run(coresys: CoreSys):
--- a/supervisor/docker/manager.py
+++ b/supervisor/docker/manager.py
@@ -7,7 +7,7 @@ from pathlib import Path
 from typing import Any, Final
 
 import attr
-from awesomeversion import AwesomeVersion, AwesomeVersionCompareException
+from awesomeversion import AwesomeVersion
 from docker import errors as docker_errors
 from docker.api.client import APIClient
 from docker.client import DockerClient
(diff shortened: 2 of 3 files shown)


donaldguy:
To expound a little/summarize #4827:

1. the supervisor [bootstraps](https://github.com/home-assistant/supervisor/blob/480b383782323532d06b1334fdb293bf26460316/supervisor/docker/network.py#L71) (or finds from previous run) a docker bridge network `hassio` that passes through these CIDRS:

    https://github.com/home-assistant/supervisor/blob/480b383782323532d06b1334fdb293bf26460316/supervisor/const.py#L43-L44

2. the apparent expected behavior is for addons to populate the upper range (`172.30.33.0/24`) while home-assistant itself (supervisor/observer/multicast/etc) grab (fixed?) low IPs at the bottom of the subnet, generally supervisor at `172.30.32.2`, etc.

3. Docker/moby upstream (inadvertently) landed a change that tightened restrictions on how container network attachment could work, limiting it only to the IPRange, not the entire subnet.

4. As such after upgrading to Docker 25.0.0 and attempting to restart the supervisor and/or any other major component, these quietly got errored out of an assignment in the `hassio` network at all and landed all in the default `bridge` docker network instead (`172.17.0.0/16`, so supervisor as first run docker container at `172.17.0.2`)

6. For unclear reasons, this failed/mis-assignment is not clear to other components, so e.g. the /etc/hosts inside the `hassio_cli` container still hardcodes/computes? `supervisor` to `172.30.32.2` - as such no `ha supervisor` commands work 
7. The supervisor is also broadly unhappy and prints loads of errors, before settling on `blocked from execution, system is not healthy - privileged` ; despite the container in fact being privileged  (I went down a deep rabbit hole around the netlink udev refusing to initiate which I though might be the core issue)


As such the hardcoding of a check on docker version in this specific case is a rather course instrument, but I suspect it may come up again in the future.

A more appropriate specific solution might be to (simply) bubble this error:

https://github.com/home-assistant/supervisor/blob/480b383782323532d06b1334fdb293bf26460316/supervisor/docker/network.py#L116-L122

more aggressively as ~fatal in the case of any essential component (supervisor, observer, dns?, etc.)

or more aggressively attempt to do discovery and routing to ~misplaced components

tests/resolution/evaluation/test_evaluate_docker_version.py
```diff
@@ -1,28 +1,71 @@
...
+
 from supervisor.const import CoreState
 from supervisor.coresys import CoreSys
 from supervisor.resolution.evaluations.docker_version import EvaluateDockerVersion
 
 
-async def test_evaluation(coresys: CoreSys):
-    """Test evaluation."""
+async def test_evaluation_supported(coresys: CoreSys):
+    """Test real evaluation with a current docker daemon."""
```
donaldguy, line 15:
```suggestion
    """Test evaluation with a known good docker version."""
```

I kinda thought at one point it would be good to just let it hit the real thing, but I could not finagle getting pytest not to mock it out.

codyc1515:
From an architecture side, would it not make more sense to just decide on a few main version and support those only? Same as what happens with Python versions today.

donaldguy:
> From an architecture side, would it not make more sense to just decide on a few main version and support those only? Same as what happens with Python versions today.

I wouldn't disagree, and obvs the defacto over in hassos land is that there is a single pin at a time

I would say broadly that:
1. the docker API is nominally versioned, along with the software, like this: https://docs.docker.com/engine/api/#api-version-matrix so in theory that is the thing-to-pin-to that is correct 
2. but then you see abberant version-specific-fails like the one in 25.0.0 ; not sure how commonly such are apt to arise as would suggest something like this PR still is warranted nevertheless

and/but then speaking as a `supervised-installer` user, I would say (again, as in PR descrip) **that this pin/constraint should absolutely be in the debian package metadata** (https://github.com/home-assistant/supervised-installer/blob/main/homeassistant-supervised/DEBIAN/control) or at least the https://github.com/home-assistant/supervised-installer/blob/main/homeassistant-supervised/DEBIAN/postinst should call the relevant `apt-mark hold`s

I am willing to have HASSIO holding-back this machine from taking on newer docker engine, but I am pretty peeved to have it break on `apt upgrade`

---

That said, my suspicion is that HA et al are facing down a bigger container runtime reckoning sooner-rather-than later, with cgroups v1 **_officially_** officially deprecated (for sometime-this-year removal) in systemd [255](https://github.com/systemd/systemd/releases/tag/v255)

and with that a question of e.g. jumping ship(/"jumping whale") from docker/moby entirely to podman, or systemd-nspawn, or whatever. 

It did not escape my notice that the `/usr/sbin/hassio-supervisor` run command [recently ](https://github.com/home-assistant/supervised-installer/commit/13ffd23c363efdae0b690abfdb2e1860a7d8d86f) grew a `containerd.sock` mount 

So ... there's that.

agners:
> **EDIT:** apparently contrary to my impression of their plans, they cut a docker 25.0.1 this morning, so this is kinda for no one now. The broader idea _could_ still be helpful in future, but lmk if just want to close

Wrt close: A bit on the fence weather we should add this code, since with 25.0.1 this is kinda already in the past at this point.

On the other hand, maybe it indeed helps for the future. Question: Did you test that with 25.0.0? Did the message appear early enough so it would have been useful for people?

mdegat01:
The issue with this approach is in order for anyone to see this message we have to update supervisor. Like if Docker has a bad version in the future, using this as a mechanism for messaging means we need to urgently get a new version of supervisor to stable with this evaluation updated and the new bad version listed. We can't anticipate future bad versions of docker so existing versions of supervisor won't be aware to warn people.

Even then it would still only help people that had not updated docker yet. People that have updated docker already have unhealthy systems and so are pretty stuck and unlikely to be able to update supervisor to receive this message.

I think #355 is the better approach. The supervised installer should probably pin to a known good version and not simply install the latest. In addition if a known bad version of docker remains out as latest for a longer time we do actually have a way to get messages out to users without needing to update Supervisor or Home Assistant - this is why https://alerts.home-assistant.io/ was created. We can create a new alert and ensure anyone with a supervised type installation sees it as alert in their Home Assistant so they are aware of the potential issue. This seems like a better approach then handling it in supervisor here to me.

PR #4838, opened 2024-01-25
Auto updates to new version delay for 24 hours

mdegat01:
## Proposed change

Part of the reason auto-updates are so dangerous is that users which have not looked at or evaluated the new version in any way are the first to update to it. This means if there was a surprise breaking change in the underlying software, the impact is exacerbated. Since it may take a while before auto-updating users even notice something broke and they won't know why or what happened.

Delaying auto-updates 24 hours increases the likelihood that surprise issues will be caught before auto-updates fire. Since users manually updating will likely know right away that something went wrong and revert. And then addon devs can push another update marking the version as breaking with more info on why. Or fix bugs if there was something wrong on their end.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist


codyc1515 (commented):
Does it make more sense to stagger the auto-updates, i.e. some may be done immediately, some on day one, some of day three and so on? This further reduces the likelihood that users experience an unexpected issue. My concern is less about known breaking changes (features) and more about unexpected breaking changes (bugs).

mdegat01:
@codyc1515 well its going to be somewhat staggered since all the repeating jobs are on timers that start when supervisor starts. So every system will be different based on when that particular system has last restarted.

As for staggering them explicitly at a larger scale, we could only randomize it. We can't control when these systems see the updates as they are distributed from github. And we can't group machines into buckets at scale because none of these systems talk to each other. So all we could do is pick a random number and delay that many days.

I don't really think we should put any more logic here though. Long term this feature should probably leave supervisor entirely and become either a part of update entities or a blueprint. Ideally we'd remove the auto-update feature from addons since its extremely risky and causes a ton of problems but the ship may have sailed on that one.

codyc1515:
I hadn't considered the timer effect. I think this mitigates the risk a little. Though, as you said, a randomised delay may be better.

While I understand that sometimes an add-on could encounter issues because of a breaking change, consider that many HA systems (for better or worse) are exposed to the internet, in many cases directly (with ports forwarded).

By making auto-updating slower, or removing that functionality entirely, we run the risk that if there was a critical security vulnerability (zero-day, for example) we may have a worse situation wherein we leave people's systems vulnerable for longer. Potentially this could be unnoticed for a long time. In a worst case, we could be looking at a full system compromise (including ancillary systems, like cloud credentials).

We could add a flag to versions somehow so authors could flag breaking changes and these could be blocked from auto-updating. This seems the most straightforward way to alleviate concerns around auto-updating, without compromising security and user choice.

As an aside, I'm not sure if we have the following security measures already, but we should (maybe an issue for a separate ADR):

* Add-on security blacklist - same concept as with HACS, there could be a centralised file that is downloaded periodically that could blacklist add-ons for security reasons (compromised add-on author, for example)

* Add-on security whitelist - inverse of the above, wherein there is a centralised file that is downloaded periodically that forces an auto-update (regardless of user choice) to patch a critical security issue, considering that a broken add-on is better than a critically vulnerable one

codyc1515 (approved):
LGTM

agners (approved):

PR #4843, opened 2024-01-27
Significantly speed up creating backups with isal via zlib-fast

bdraco:
## Proposed change

[`isal`](https://github.com/intel/isa-l) via [`python-isal`](https://github.com/pycompression/python-isal) has a drop in replacement for zlib with the cavet that the compression level mappings are different. `zlib-fast` is a tiny piece of middleware to convert the standard zlib compression levels to `isal` compression levels to allow for drop-in replacement

I considered a few other approaches here such as using `pigz` or switching to `zstd` but this seems like the simplest and most compatible.

Wheels are built for all supported platforms https://wheels.home-assistant.io/musllinux/ 

https://github.com/bdraco/zlib-fast/releases/tag/v0.2.0 https://github.com/pycompression/python-isal

Compression for backups is ~5x faster than the baseline. 

https://github.com/powturbo/TurboBench/issues/43

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -27,3 +27,4 @@ setuptools==69.0.3
 voluptuous==0.14.1
 dbus-fast==2.21.1
 typing_extensions==4.9.0
+zlib-fast==0.1.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -27,4 +27,4 @@ setuptools==69.0.3
 voluptuous==0.14.1
 dbus-fast==2.21.1
 typing_extensions==4.9.0
-zlib-fast==0.1.0
+zlib-fast==0.2.0
--- a/supervisor/__main__.py
+++ b/supervisor/__main__.py
@@ -10,8 +10,10 @@ import zlib_fast
 # Enable fast zlib before importing supervisor
 zlib_fast.enable()
 
-from supervisor import bootstrap  # noqa: E402
-from supervisor.utils.logging import activate_log_queue_handler  # noqa: E402
+from supervisor import bootstrap  # pylint: disable=wrong-import-position # noqa: E402
+from supervisor.utils.logging import (  # pylint: disable=wrong-import-position  # noqa: E402
+    activate_log_queue_handler,
+)
 
 _LOGGER: logging.Logger = logging.getLogger(__name__)
 
--- a/supervisor/__main__.py
+++ b/supervisor/__main__.py
@@ -5,8 +5,13 @@ import logging
 from pathlib import Path
 import sys
 
-from supervisor import bootstrap
-from supervisor.utils.logging import activate_log_queue_handler
+import zlib_fast
+
+# Enable fast zlib before importing supervisor
+zlib_fast.enable()
+
+from supervisor import bootstrap  # noqa: E402
+from supervisor.utils.logging import activate_log_queue_handler  # noqa: E402
 
 _LOGGER: logging.Logger = logging.getLogger(__name__)
 


bdraco:
I tested this locally and verified it was loaded
```
28e75fc68efe:/usr/src/supervisor# ulimit -n 1000
28e75fc68efe:/usr/src/supervisor# lsof -p 78
```
```
python3  78 root  mem       REG              259,8            3677071 /usr/local/lib/python3.11/site-packages/isal/igzip_lib.cpython-311-x86_64-linux-musl.so (path dev=0,51)
python3  78 root  mem       REG              259,8            3677074 /usr/local/lib/python3.11/site-packages/isal/isal_zlib.cpython-311-x86_64-linux-musl.so (path dev=0,51)
python3  78 root  mem       REG              259,8            3677070 /usr/local/lib/python3.11/site-packages/isal/_isal.cpython-311-x86_64-linux-musl.so (path dev=0,51)
```

bdraco:
On my local its much faster but I tried it on production and there is no change so I've got to be doing something wrong/different

bdraco:
I see the issue. my test uses zlib and I need to map the gzip path to IGzipFile as well

codyc1515 (approved):
Does this pose any issue with restores? Otherwise, LGTM.

bdraco:
> Does this pose any issue with restores? Otherwise, LGTM.

Restores work the same. decompression is faster as well. ~3x faster

codyc1515:
2024.2 is the month of the back-up! Loooking forward to seeing this, alongside the other changes we have made to reliability.

agners:
> YMMV on the compression ratio a bit depending on the content. On one system it was 2% better, and on another it was 7% worse. securetar uses compress level 6 by default so it seems like the goal there was to balance speed and compress ratio.

I've changed that somewhat recently with https://github.com/pvizeli/securetar/pull/22. It helped already dropping backup time by almost 2x. I remember I considered going to a even lower default setting as the compression numbers in gz do not influence compression ratio drastically (at least not as drastic as others, at least according to [this somewhat older benchmark](https://community.centminmod.com/threads/round-4-compression-comparison-benchmarks-zstd-vs-brotli-vs-pigz-vs-bzip2-vs-xz-etc.18669/)). However, I then thought let's move the needle slowly to not run into surprises. I've not heard any negative feedback, so from my point of view we can continue to change the standard.

That said, I was actually consider starting to look into another compression algorithm, e.g. zstd. It would mean to introduce a backup format version, but that shouldn't be that big of a deal.

One thing I was wondering: Is this library making use of newer instruction set extensions/is the code written such that it still will work on old x86-64 systems without those extensions?

bdraco:
> One thing I was wondering: Is this library making use of newer instruction set extensions/is the code written such that it still will work on old x86-64 systems without those extensions?

I've test the library and it works on older x86_64 systems as well as 32 bit arm. Its just not as fast because it uses fallbacks

bdraco:
I didn't test it on every old x86_64 chip (only the synology box with the old chip we had problem with other libs) Given its an intel origin https://github.com/intel/isa-l I expect they have tested it a bit more on older hardware as they have a bit more access to it

codyc1515:
> That said, I was actually consider starting to look into another compression algorithm, e.g. zstd. It would mean to introduce a backup format version, but that shouldn't be that big of a deal.

Messaging was added to the supervisor recently stating that backups are not backwards compatible (instead of just discreetly failing), so this should not be a blocking issue.

agners:
> I didn't test it on every old x86_64 chip (only the synology box with the old chip we had problem with other libs) Given its an intel origin https://github.com/intel/isa-l I expect they have tested it a bit more on older hardware as they have a bit more access to it

I mean if they use hand written assembler, they probably just thought about it and implemented fallbacks.

Sometimes it is also the compilers optimization steps which select some instructions from an extension, but that then will depend on flags at build time.

Since you already tested on older systems I am not too concerned indeed. Just something to keep an eye out.

I do have a Fujitsu machine with a AMD Geode here which also has a quite limited set of instructions available. I'll do a backup once this is on dev.

agners (approved):

bdraco:
> I do have a Fujitsu machine with a AMD Geode here which also has a quite limited set of instructions available. I'll do a backup once this is on dev.

Nice. Thankfully this is trivial to revert if there is an issue. I'm going to merge it now and update everything I have once dev is built

bdraco:
Bumped everyone of my production and dev systems.  Backups completed successfully on all of them

agners:
Just tested on my Fujitsu AMD GX-222GC based system, works like a charm! :+1:

bdraco:
> A future improvement for I/O constrained system could be to use [docs.python.org/3/library/tarfile.html#tarfile.TarFile.addfile](https://docs.python.org/3/library/tarfile.html#tarfile.TarFile.addfile) to add each gziped tar file to the final archive, rewind the stream, and rewrite the size over it, and jump back to the end so we wouldn't have to write all the tgz files to disk and they can be streamed into the final result. This would cut the writes in half (and increase storage lifetime) and likely make more difference for these systems than this change.

I implemented that in https://github.com/home-assistant/supervisor/pull/4884

PR #4852, opened 2024-01-30
Migrate to Ruff for lint and format

mdegat01:
## Proposed change

Migrate away from flake8, black and other linters to ruff. Since it does all of that and more in much less time.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist


mdegat01:
@pvizeli refactored so that instead of using setters with dbus properties we have normal methods that return an awaitable to the caller. That's probably the pattern we should use going forward with dbus properties instead of trying to make them python properties. It may look cleaner but its not worth the risk of exposing us to [this](https://docs.astral.sh/ruff/rules/asyncio-dangling-task/)

pvizeli (approved):

agners (approved):

tests/jobs/test_job_decorator.py
```diff
@@ -956,7 +963,9 @@ async def execute_default(self) -> bool:
             return True
 
     test1 = TestClass(coresys)
+    # pylint: disable=protected-access
     client = coresys.homeassistant.websocket._client
+    # pylint: enable=protected-access
```
frenck, line 968: This could be a single line

```suggestion
    # pylint: disable-next=protected-access
    client = coresys.homeassistant.websocket._client
```

frenck (commented):
With the move from async_timeout to asyncio, in this PR, I think you can now remove that dependency:

https://github.com/home-assistant/supervisor/blob/1861d756e9f141b2f43941962cfa731942424f63/requirements.txt#L4

frenck (approved):
Two very small comments, otherwise, looks really nice!

Some nice improvements overall in the code as a result of this 👍 

../Frenck

pyproject.toml
```diff
@@ -71,6 +71,136 @@ disable = [
...
+    "too-many-return-statements",        # PLR0911
+    "too-many-statements",               # PLR0915
+    "trailing-comma-tuple",              # COM818
+    "unnecessary-comprehension",         # C416
+    "use-a-generator",                   # C417
+    "use-dict-literal",                  # C406
+    "use-list-literal",                  # C405
+    "useless-object-inheritance",        # UP004
+    "useless-return",                    # PLR1711
+    # "no-self-use", # PLR6301  # Optional plugin, not enabled
```
frenck, line 203: You could consider not doing this... 

Like, on the Core codebase it matters (due to its size, it reduces the CI a little). But on Supervisor, I don't think that difference is noticeable.

tests/jobs/test_job_decorator.py
mdegat01: I did not know about `disable-next`. I will definitely remember that as it helps in a lot of other places too 👍

pyproject.toml
mdegat01: Yea I thought about that. But this list seemed very well-researched, even listing out exactly which rule it overlapped with. So I figured why not leverage that information even if the difference is small.

I also think in general we should keep the linting config of the two projects in sync as much as possible. I don't want contributors to feel like each Home Assistant project has radically different codestyle and quality rules they have to learn as that then becomes a barrier to entry.

frenck (approved):

PR #4872, opened 2024-02-02
Improve error handling when mounts fail

### LINKED ISSUE #4856: For discussion/architecture: Network Storage Issues: Proposed fix for: home-assistant/core #106446

# Proposed fix for: https://github.com/home-assistant/supervisor/issues/4866

### When writing a backup

- Use `findmnt` or similar to make sure that the network mount is in a good state when creating/writing a backup.
- **If the mount is in a good state:** proceed.
- **If the mount is in a bad state, and a fallback location is configured:** write to the fallback location and raise a notice to the user that the last backup was written to the fallback location.
- **If the mount is in a bad state, and a fallback location is not configured:** raise a notice to the user that the last backup failed to write.

### Follow-up enhancement
- Add UI and/or background job to copy/sync backups from the fallback location when the mount is restored.

## Recovering from the error condition
- In the event that the mount point is a regular directory and contains valid backups, it should become the fallback directory, and a notice should be raised to the user that this has occurred.

## Follow-up enhancement
- Raise notice that backups are present in the fallback location but the option to copy them automatically has not been set up.

# UI Fixes
- Show fallback mount configuration guidance when network storage is the default destination and the mount fails (instead of showing the blank page). Explain how to setup robust backups with a transient storage server.

Home Assistant OS

1. Use network storage for backups
2. Have a backup occur when the storage mount is unavailable
3. The next time the mount reconnects, it will fail due to a non-empty directory
...

```txt
N/A
```

## System Information

version | core-2024.1.5
-- | --
dev | false
hassio | true
docker | true
user | root
virtualenv | false
python_version | 3.11.6
os_name | Linux
os_version | 6.1.71-haos
arch | x86_64
timezone | America/New_York
config_dir | /config

GitHub API | ok
-- | --
GitHub Content | ok
GitHub Web | ok
GitHub API Calls Remaining | 5000
Installed Version | 1.33.0
Stage | running
Available Repositories | 1384
Downloaded Repositories | 10

</details>

logged_in | true
-- | --
subscription_expiration | December 31, 2017 at 7:00 PM
relayer_connected | false
relayer_region | null
remote_enabled | true
remote_connected | false
alexa_enabled | true
google_enabled | true
remote_server | null
certificate_status | null
instance_id | a070e6992e114b5397069af990d6ae9c
can_reach_cert_server | ok
can_reach_cloud_auth | ok
can_reach_cloud | ok

</details>

host_os | Home Assistant OS 11.4
-- | --
update_channel | stable
supervisor_version | supervisor-2023.12.1
agent_version | 1.6.0
docker_version | 24.0.7
disk_total | 30.8 GB
disk_used | 21.7 GB
healthy | true
supported | true
board | ova
supervisor_api | ok
version_api | ok
installed_addons | Cloudflared (5.1.1), File editor (5.7.0), Glances (0.21.0), Studio Code Server (5.15.0), Frigate (Full Access) (0.12.1), Double Take (1.13.1), ESPHome (2023.12.9), openWakeWord (1.8.2)

</details>

dashboards | 2
-- | --
resources | 0
views | 2
mode | storage

</details>

oldest_recorder_run | January 21, 2024 at 6:31 PM
-- | --
current_recorder_run | January 25, 2024 at 5:14 PM
estimated_db_size | 991.45 MiB
database_engine | sqlite
database_version | 3.41.2

</details>

### Supervisor diagnostics

_No response_

### Additional information

_No response_

mdegat01:
## Proposed change

When we fail to mount a media or share type mount it is supposed to mount a read-only file in that spot. This ensures that attempts to write to that folder will fail and report rather then essentially silently fail by writing locally (and then creating other issues later). This logic did not work because it relied on user permissions and all docker containers in the Home Assistant ecosystem run as `root`. This changes the bind mount itself to use the read-only option which causes write failures even for `root`.

Since there was a feature request to create mounts of type media and share as read-only this was extended into a feature. Users can now make read-only mounts of type media and share.

As for backup type mounts, we expected that if the user tried to make a backup on a downed mount they would get a timeout. It appears that instead it writes the file locally instead. This causes impossible to fix errors next startup as we cannot mount over a non-empty folder and users can't see these folders without going into the supervisor container.

So if the user tries to make a backup in a mount we first check that the folder is in fact a mountpoint using [Path.is_mount](https://docs.python.org/3/library/pathlib.html#pathlib.Path.is_mount). In testing this only returns true if the folder is a mountpoint and if that connection is active. If this returns false we raise an error immediately and exit.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR fixes or closes issue: fixes #4856
- Link to documentation pull request: https://github.com/home-assistant/developers.home-assistant/pull/2071
- Link to cli pull request: https://github.com/home-assistant/cli/pull/451

## Checklist

### DIFF
--- a/supervisor/backups/manager.py
+++ b/supervisor/backups/manager.py
@@ -15,7 +15,12 @@ from ..const import (
     CoreState,
 )
 from ..dbus.const import UnitActiveState
-from ..exceptions import BackupError, BackupInvalidError, BackupJobError
+from ..exceptions import (
+    BackupError,
+    BackupInvalidError,
+    BackupJobError,
+    BackupMountDownError,
+)
 from ..jobs.const import JOB_GROUP_BACKUP_MANAGER, JobCondition, JobExecutionLimit
 from ..jobs.decorator import Job
 from ..jobs.job_group import JobGroup
@@ -74,11 +79,15 @@ class BackupManager(FileConfiguration, JobGroup):
 
     def _get_base_path(self, location: Mount | type[DEFAULT] | None = DEFAULT) -> Path:
         """Get base path for backup using location or default location."""
-        if location:
-            return location.local_where
-
         if location == DEFAULT and self.sys_mounts.default_backup_mount:
-            return self.sys_mounts.default_backup_mount.local_where
+            location = self.sys_mounts.default_backup_mount
+
+        if location:
+            if not location.local_where.is_mount():
+                raise BackupMountDownError(
+                    f"{location.name} is down, cannot back-up to it", _LOGGER.error
+                )
+            return location.local_where
 
         return self.sys_config.path_backup
 
--- a/supervisor/exceptions.py
+++ b/supervisor/exceptions.py
@@ -607,6 +607,10 @@ class BackupInvalidError(BackupError):
     """Raise if backup or password provided is invalid."""
 
 
+class BackupMountDownError(BackupError):
+    """Raise if mount specified for backup is down."""
+
+
 class BackupJobError(BackupError, JobException):
     """Raise on Backup job error."""
 
--- a/supervisor/mounts/const.py
+++ b/supervisor/mounts/const.py
@@ -8,6 +8,7 @@ FILE_CONFIG_MOUNTS = PurePath("mounts.json")
 ATTR_DEFAULT_BACKUP_MOUNT = "default_backup_mount"
 ATTR_MOUNTS = "mounts"
 ATTR_PATH = "path"
+ATTR_READ_ONLY = "read_only"
 ATTR_SERVER = "server"
 ATTR_SHARE = "share"
 ATTR_USAGE = "usage"
--- a/supervisor/mounts/manager.py
+++ b/supervisor/mounts/manager.py
@@ -297,6 +297,7 @@ class MountManager(FileConfiguration, CoreSysAttributes):
                 name=f"{'emergency' if emergency else 'bind'}_{mount.name}",
                 path=path,
                 where=where,
+                read_only=emergency,
             ),
             emergency=emergency,
         )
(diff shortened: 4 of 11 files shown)


supervisor/backups/manager.py
```diff
@@ -74,12 +79,16 @@ def get(self, slug: str) -> Backup:
...
     def _get_base_path(self, location: Mount | type[DEFAULT] | None = DEFAULT) -> Path:
         """Get base path for backup using location or default location."""
+        if location == DEFAULT and self.sys_mounts.default_backup_mount:
+            location = self.sys_mounts.default_backup_mount
+
         if location:
+            if not location.local_where.is_mount():
+                raise BackupMountDownError(
+                    f"{location.name} is down, cannot back-up to it", _LOGGER.error
+                )
```
agners, line 89: I guess we could debate over if we should fallback to the backup path here. The Supervisor should have raised a repair issue for the mount at this point. And reading issues, I think most people prefer that the backup fails in this case. Otherwise chances are high that we fill up the local disk.

agners (approved):

supervisor/backups/manager.py
mdegat01: I don't think we should do that, at least not silently. If there were no errors, people will expect the backup to be in the location specified. If we want to do a fallback setup I think we'd need to either one of two things:

1. Tell the user via repair or notification of some kind that we made the backup but it is in the wrong place and they need to correct it once their mount is back online
2. Automatically move the backup for them to the right place once the backup is online

I also think this behavior would need to be configurable. Since users with SD cards would likely prefer backups fail in this situation rather then having us write the whole thing to their data disk.

We can do some of this if we want. But its a new feature and should be a separate PR. This PR is mostly about making the existing feature work the way we intended (writes should fail when the mount is down)

agners: Yeah sounds good to me. I think most important is that the user has clear indication that something is off (in this case the mount did not work). The consequences he most likely will realize on his own. I am fine with this :+1:

jmealo:
@mdegat01 thanks for taking care of this! 🙏

PR #4934, opened 2024-02-28
Add an admin only device wipe API

mdegat01:
## Proposed change

Home Assistant should have a factory reset option: https://community.home-assistant.io/t/wth-where-is-factory-reset-option/468659

This isn't exactly a factory reset since we can't really do that. We don't maintain a copy of the state of the device when first turned on and we sit too far above the hardware anyway. But a complete wipe of the datadisk is close enough to serve our purposes here. It provides an easy way to not only wipe all Home Assistant and Add-on config but also settings of docker and downloaded images and reset the user to a nearly clean slate.

~~It notably will miss some user-configurable things such as network settings. Which would definitely confuse people if advertised as a factory reset, should try to clarify the difference in documentation.~~ Actually it will clear out pretty much everything. The only real difference between this and a factory reset will be that the version of OS remains the same.

Note that this API will require an admin token. That means although it will be added to the CLI, it will only work on the host shell. In addition, clients calling it must first ask users for manual confirmation that they want to wipe their data and inform them the process is irreversible.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to documentation pull request: https://github.com/home-assistant/developers.home-assistant/pull/2097
- Link to cli pull request: https://github.com/home-assistant/cli/pull/464

## Checklist

### DIFF
--- a/supervisor/api/__init__.py
+++ b/supervisor/api/__init__.py
@@ -182,6 +182,7 @@ class RestAPI(CoreSysAttributes):
                 web.post("/os/config/sync", api_os.config_sync),
                 web.post("/os/datadisk/move", api_os.migrate_data),
                 web.get("/os/datadisk/list", api_os.list_data),
+                web.post("/os/datadisk/wipe", api_os.wipe_data),
             ]
         )
 
--- a/supervisor/api/middleware/security.py
+++ b/supervisor/api/middleware/security.py
@@ -118,7 +118,7 @@ ADDONS_ROLE_ACCESS: dict[str, re.Pattern] = {
         r"|/multicast/.+"
         r"|/network/.+"
         r"|/observer/.+"
-        r"|/os/.+"
+        r"|/os/(?!datadisk/wipe).+"
         r"|/refresh_updates"
         r"|/resolution/.+"
         r"|/security/.+"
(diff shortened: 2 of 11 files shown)


sairon (commented):
> It notably will miss some user-configurable things such as network settings. Which would definitely confuse people if advertised as a factory reset, should try to clarify the difference in documentation.

Are you sure? The os-agent WipeDevice function also [wipes the overlay](https://github.com/home-assistant/os-agent/blob/c8721c8eee5ae95e5be7218c49cae8cf37b72b14/system/system.go#L68), so the only potential changes in writable partitions could be in the `hassoss-boot` partition.

agners (approved):

agners:
> Are you sure? The os-agent WipeDevice function also [wipes the overlay](https://github.com/home-assistant/os-agent/blob/c8721c8eee5ae95e5be7218c49cae8cf37b72b14/system/system.go#L68), so the only potential changes in writable partitions could be in the `hassoss-boot` partition.

Uh right, completely forgot about this.

So yeah, for almost all use cases this essentially means that using the data disk wipe they'll start with a system just as vanilla as a new installation.

mdegat01:
Oh ok. Was going off my convo with stefan. I'll update the other doc then, good catch!

PR #4937, opened 2024-03-01
Remove discovery config validation from supervisor

mdegat01:
## Proposed change

Currently if an addon developer wants to be able to setup a corresponding integration in Home Assistant they have to do this process:
1. Modify their addon to list the service it wants to send discovery messages for in its config and send those on startup
2. Modify Supervisor to ensure the service is known to it and validate its config (if any is pased in the discovery message)
3. Modify the integration in core to set itself up when a discovery message is received for the new service

This PR eliminates step 2 entirely. Supervisor shouldn't really be involved in this process as more then just a passthrough. Validation of discovery messages should be in the integration in core since that is where the discovery message is actually understood and used (and to my understanding, integration developers generally do this already).

Supervisor will continue to validate that the discovery message received lists a service that is in the associated add-ons config. And that `config` is a dictionary if provided. But that's all. It will have no understanding of what services are out there for Home Assistant to discover and what config looks like for each particular service.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist


supervisor/api/utils.py
```diff
@@ -129,12 +129,15 @@ def api_return_error(
         JSON_RESULT: RESULT_ERROR,
         JSON_MESSAGE: message or "Unknown error, see supervisor",
     }
-    if isinstance(error, APIError) and error.job_id:
-        result[JSON_JOB_ID] = error.job_id
+    status = 400
+    if isinstance(error, APIError):
+        status = error.status
+        if error.job_id:
+            result[JSON_JOB_ID] = error.job_id
```
mdegat01, line 136: This is technically unrelated to the current PR, just something I discovered while writing the test cases. But it felt really wrong to me that the `APIForbidden` exception did not actually return `403 Forbidden` and returned `400 Bad Request`. Especially since we seemed to use it as a way to convey to clients that they did something not allowed and that wasn't being translated to the status code.

Although this does change the status code in a few places, I don't think this is really a backwards compatibility issue. It's still a 400 range error indicating the client did something wrong and its a status code we do use they should already be looking for.

agners: I fully agree with this finding and the change. However, I'd prefer if this is a separate PR, even though it is very small. Mostly about having a proper entry for this change in the release changelog. Also in case we have to revert (even though rather unlikely).

mdegat01:
Waiting on https://github.com/home-assistant/supervisor/pull/4943 so I don't have to change tests twice to check for 400 and then 403

supervisor/api/utils.py
mdegat01: @agners ok, created https://github.com/home-assistant/supervisor/pull/4943

PR #4675, opened 2023-11-07
Cache stat of addon assets

bdraco:
## Proposed change

Since we poll the addons quite a bit we did a lot of stat()s in the event loop

I wasn't 100% sure if the cache of these ever needed to be cleared though.

## Type of change

- [ ] Dependency upgrade
- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (which adds functionality to the supervisor)
- [ ] Breaking change (fix/feature causing existing functionality to break)
- [x] Code quality improvements to existing code or addition of tests

## Additional information

- This PR fixes or closes issue: fixes #
- This PR is related to issue:
- Link to documentation pull request:
- Link to cli pull request:

## Checklist

- [ ] The code change is tested and works locally.
- [ ] Local tests pass. **Your PR cannot be merged unless tests pass**
- [ ] There is no commented out code in this PR.
- [ ] I have followed the [development checklist][dev-checklist]
- [ ] The code has been formatted using Black (`black --fast supervisor tests`)
- [ ] Tests have been added to verify that the new code works.

If API endpoints of add-on configuration are added/changed:

- [ ] Documentation added/updated for [developers.home-assistant.io][docs-repository]

[dev-checklist]: https://developers.home-assistant.io/docs/en/development_checklist.html
[docs-repository]: https://github.com/home-assistant/developers.home-assistant

PR #4944, opened 2024-03-06
Cache existence of addon paths

mdegat01:
## Proposed change

Reduce stat calls by caching the existence of various addon paths. We clear the cache when we see there has been a change in the repository.

As an added performance bonus, we now only do a git reset if we see there has been a new commit in the fetch. If the commit we see matches what was returned for a fetch on the branch we leave it as is.

For the local repository we clear the cache every 3 hours when a store reload is called or when the user clicks rebuild since we can't ask git if there's been a new commit. Not ideal but the local repo is intended for dev and testing anyway and not a place to put addons long term so it seems like an acceptable trade off.

Supercedes #4675

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/addons/addon.py
+++ b/supervisor/addons/addon.py
@@ -180,6 +180,8 @@ class Addon(AddonModel):
 
     async def load(self) -> None:
         """Async initialize of object."""
+        await self.refresh_cache()
+
         self._listeners.append(
             self.sys_bus.register_event(
                 BusEvent.DOCKER_CONTAINER_STATE_CHANGE, self.container_state_changed
@@ -797,7 +799,7 @@ class Addon(AddonModel):
         if it was running. Else nothing is returned.
         """
         last_state: AddonState = self.state
-        self.clear_cache()
+        await self.refresh_cache()
         try:
             # remove docker container but not addon config
             try:
--- a/supervisor/addons/addon.py
+++ b/supervisor/addons/addon.py
@@ -797,6 +797,7 @@ class Addon(AddonModel):
         if it was running. Else nothing is returned.
         """
         last_state: AddonState = self.state
+        self.clear_cache()
         try:
             # remove docker container but not addon config
             try:
--- a/supervisor/addons/addon.py
+++ b/supervisor/addons/addon.py
@@ -181,7 +181,7 @@ class Addon(AddonModel):
     async def load(self) -> None:
         """Async initialize of object."""
         if self.is_detached:
-            await self.refresh_cache()
+            await self.refresh_path_cache()
 
         self._listeners.append(
             self.sys_bus.register_event(
@@ -1431,8 +1431,8 @@ class Addon(AddonModel):
         ]:
             await self._restart_after_problem(event.state)
 
-    def refresh_cache(self) -> Awaitable[None]:
+    def refresh_path_cache(self) -> Awaitable[None]:
         """Refresh cache of existing paths."""
         if self.is_detached:
-            return super().refresh_cache()
-        return self.addon_store.refresh_cache()
+            return super().refresh_path_cache()
+        return self.addon_store.refresh_path_cache()
--- a/supervisor/addons/addon.py
+++ b/supervisor/addons/addon.py
@@ -181,7 +181,7 @@ class Addon(AddonModel):
     async def load(self) -> None:
         """Async initialize of object."""
         if self.is_detached:
-            await self.refresh_path_cache()
+            await super().refresh_path_cache()
 
         self._listeners.append(
             self.sys_bus.register_event(
--- a/supervisor/addons/addon.py
+++ b/supervisor/addons/addon.py
@@ -180,7 +180,8 @@ class Addon(AddonModel):
 
     async def load(self) -> None:
         """Async initialize of object."""
-        await self.refresh_cache()
+        if self.is_detached:
+            await self.refresh_cache()
 
         self._listeners.append(
             self.sys_bus.register_event(
@@ -231,6 +232,34 @@ class Addon(AddonModel):
     def is_detached(self) -> bool:
         """Return True if add-on is detached."""
         return self.slug not in self.sys_store.data.addons
+
+    @property
+    def with_icon(self) -> bool:
+        """Return True if an icon exists."""
+        if self.is_detached:
+            return super().with_icon
+        return self.addon_store.with_icon
+
+    @property
+    def with_logo(self) -> bool:
+        """Return True if a logo exists."""
+        if self.is_detached:
+            return super().with_logo
+        return self.addon_store.with_logo
+
+    @property
+    def with_changelog(self) -> bool:
+        """Return True if a changelog exists."""
+        if self.is_detached:
+            return super().with_changelog
+        return self.addon_store.with_changelog
+
+    @property
+    def with_documentation(self) -> bool:
+        """Return True if a documentation exists."""
+        if self.is_detached:
+            return super().with_documentation
+        return self.addon_store.with_documentation
 
     @property
     def available(self) -> bool:
@@ -799,7 +828,6 @@ class Addon(AddonModel):
         if it was running. Else nothing is returned.
         """
         last_state: AddonState = self.state
-        await self.refresh_cache()
         try:
             # remove docker container but not addon config
             try:
--- a/supervisor/addons/addon.py
+++ b/supervisor/addons/addon.py
@@ -1430,3 +1430,9 @@ class Addon(AddonModel):
             ContainerState.UNHEALTHY,
         ]:
             await self._restart_after_problem(event.state)
+
+    def refresh_cache(self) -> Awaitable[None]:
+        """Refresh cache of existing paths."""
+        if self.is_detached:
+            return super().refresh_cache()
+        return self.addon_store.refresh_cache()
--- a/supervisor/addons/model.py
+++ b/supervisor/addons/model.py
@@ -1,7 +1,7 @@
 """Init file for Supervisor add-ons."""
 from abc import ABC, abstractmethod
 from collections import defaultdict
-from collections.abc import Callable
+from collections.abc import Awaitable, Callable
 from contextlib import suppress
 from datetime import datetime
 import logging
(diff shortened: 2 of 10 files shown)


pvizeli (approved):

supervisor/addons/model.py
```diff
@@ -511,22 +515,30 @@ def url(self) -> str | None:
     @property
     def with_icon(self) -> bool:
         """Return True if an icon exists."""
-        return self.path_icon.exists()
+        if self._path_icon_exists is None:
+            self._path_icon_exists = self.path_icon.exists()
```
agners, line 519: In the end this still will do I/O in the asyncio event loop. :cry: 

We had a similar problem when making check_port async, with the dynamic ingress port assignment in #4677.

I think we could make `_check_ingress_port()` more generic (e.g. `_check_addon()`), and update the these things there in an executor. Thoughts? :thinking:

mdegat01: I mean we could just make `_clear_cache` into `_refresh_cache` and call it during `load` too

mdegat01: Ok I decided to go with the `refresh_cache` model, where the value is always cached and we update it any time the repo does. As noted when we talked there's a risk with this since it means we'll redo our existence checks for all addons in a repo anytime the repo updates. But hopefully that shouldn't be too bad compared to doing this stat check in the event loop for every addon anytime we need to show a list of addons.

The other reason I settled on this model as opposed to making the python properties into methods that run in the executor is because then each `with_<x>` method would run as a separate task. With a `refresh_cache` option I can do all 4 of these at once. It seemed inefficient to do it as 4 separate ones and I didn't want every consumer to have to know the caching details.

I'm still open to other ideas though.

mdegat01: Actually I guess I'm kind of double caching these. Really the installed addons should just look at the store to see the value and we cache it there. I'll see if that works

supervisor/addons/addon.py
```diff
@@ -1399,3 +1430,9 @@ async def watchdog_container(self, event: DockerContainerStateEvent) -> None:
             ContainerState.UNHEALTHY,
         ]:
             await self._restart_after_problem(event.state)
+
+    def refresh_cache(self) -> Awaitable[None]:
```
agners, line 1434: Do we need this method still? Afaik, we can directly rely on `super().refresh_cache()` here now, as we update the store on store update :thinking:

supervisor/addons/model.py
agners: Yeah I think combining the update make sense instead of having 4 tasks. I see you went the "reuse store if possible" route :+1:

supervisor/store/repository.py
```diff
@@ -103,9 +105,17 @@ async def load(self) -> None:
...
+        if not self.validate():
             return
-        await self.git.pull()
+
+        if self.type == StoreType.LOCAL or await self.git.pull():
+            await asyncio.gather(
+                *[
+                    addon.refresh_cache()
+                    for addon in self.sys_addons.all
+                    if addon.repository == self.slug
```
agners, line 116: The `Repository` class so far did not really deal with individual add-ons, maybe it would be better if we update cache from `StoreManager` instead? :thinking:  Then we have the `refresh_cache` calls at the same place. :thinking:

agners (changes requested):

supervisor/addons/addon.py
mdegat01: We don't need to do this its true. I just didn't like having a public method in the class that couldn't be called. Because if someone called it then it would refresh the cache in `Addon` not `StoreAddon` which is wrong for like 99% of addons

supervisor/addons/model.py
```diff
@@ -635,6 +639,17 @@ def breaking_versions(self) -> list[AwesomeVersion]:
         """Return breaking versions of addon."""
         return self.data[ATTR_BREAKING_VERSIONS]
 
+    def refresh_cache(self) -> Awaitable[None]:
```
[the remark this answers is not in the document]
mdegat01: I kind of assumed this might not be the only thing we cache in the future. But I can change it

supervisor/store/repository.py
mdegat01: Well we don't always call it from `StoreManager` currently. We also call `Repository.update` from here:
https://github.com/home-assistant/supervisor/blob/749d0e6c27d2a4ed88e9018ad657b26a4a6bb1d2/supervisor/resolution/fixups/store_execute_reload.py#L44

Also the `update` method is pretty standard on our classes and always returns `None`. This would have to return something else to let the caller know they need to do more. I feel like the repository is the thing that updated so it should just handle it.

supervisor/addons/addon.py
agners: Hm, I see your point.

I guess it is a bit confusing that `load()` now checks for `self.is_detached`, and here we do it again. But it makes sense since cache refreshing for add-ons from store is taken care from store side.

Maybe we should name the two functions to make it less confusing? :thinking: 

Where is `self.addon_store.refresh_cache()` actually implemented? :thinking:

supervisor/addons/model.py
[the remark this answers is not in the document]
agners: Yeah we still can rename it if we need to cache other things. Right now it sounds as if we are updating all metadata or something.

supervisor/addons/addon.py
mdegat01: > I guess it is a bit confusing that load() now checks for self.is_detached, and here we do it again. But it makes sense since cache refreshing for add-ons from store is taken care from store side.

@agners The reason for this is because although it would fine if I removed the check from `load`, it would cause the cache in the corresponding store addon to refresh twice. Once on load of `StoreManager`, once on load of the installed `Addon`. The alternative is filtering out installed addons in `StoreManager` when doing that refresh but this seemed cleaner.

> Where is self.addon_store.refresh_cache() actually implemented? 🤔

In `AddonModel`. It's shared between `StoreAddon` and `Addon` because installed addons may use it as well, if they are detached.

> Maybe we should name the two functions to make it less confusing? 🤔

I don't think this is a good idea. There's a lot of places in supervisor where we get a list of `AnyAddon`, which could be a `StoreAddon` and could be an `Addon`. It's generally better to keep the interface the same when possible IMO.

agners (approved):

supervisor/addons/addon.py
agners: > @agners The reason for this is because although it would fine if I removed the check from load, it would cause the cache in the corresponding store addon to refresh twice. Once on load of StoreManager, once on load of the installed Addon. 

Yes after following code paths, I do see that. But it isn't immediately apparent from reading the code. Which is kinda my reason of suggesting a rename: It makes reading the code a easier.

Looking at the code again, I think it make sense to use `super().self.refresh_path_cache()` in `load()`, this way it is congruent with the `with_ *` properties. In the end it is all minor details. It just is when stumbling as a reviewer, whoever is touching the code next, will probably get confused as well. And potentially introduce a bug because he tries to "optimize" and removes the if in `load()` :see_no_evil:

supervisor/addons/addon.py
```diff
@@ -180,6 +180,9 @@ def in_progress(self) -> bool:
 
     async def load(self) -> None:
         """Async initialize of object."""
+        if self.is_detached:
+            await self.refresh_path_cache()
```
agners, line 184:
```suggestion
            await super().refresh_path_cache()
```

supervisor/store/repository.py
agners: > Well we don't always call it from StoreManager currently. We also call Repository.update from here:

Right, which makes me wonder if that works correctly there:  It doesn't update `StoreManager`, so what if the repository has new/deleted/updated add-ons compared to before the fixups have become necessary? :thinking: 

Anyways, this would be a different PR.

> Also the `update` method is pretty standard on our classes and always returns `None`.

Sure, standardization is nice if wee have good reasons. But we also deviate from it (e.g. in `MountManager`/`Mount` implementation). And I think we should deviate, if we have good reasons.

Actually, looking a bit more into this code, I think we should not over-optimize these four stat calls here.

`StoreManager`'s `reload()` calls `self.data.update()` through `self.load()`, which causes a complete read of all the add-ons in `_read_addons_folder()`.

Couldn't we just add `addon_config[ATTR_WITH_LOGO]` etc. to  `_read_addons_folder()`? :thinking: 

I really like the idea of avoiding updating the whole store when git pull doesn't change something. But I think we should apply this to the complete store update *somehow*. But this would be material for a different PR.

agners (changes requested):
After looking a bit more into how the store updates are handled, I think we should paddle back a bit on those updates. See comment in https://github.com/home-assistant/supervisor/pull/4944#discussion_r1525208585.

supervisor/addons/addon.py
mdegat01: I'll just block it from addon's interface. We can call the superclass one from load in the edge case and the one in addon will be marked as non-functional so the interface is intelligble

supervisor/store/repository.py
mdegat01: We can tweak this in a small way if we route everything through store manager. Since we'll know when a git repo has updated, we can skip all the other stuff if there were no updates. In future we could also be more selective about only re-reading in the specific git repos that updated.

agners: > Well no, it has to be the same PR actually. This PR relies on the fact that whenever we actually detect a change in commit in the repo, we refresh our cache. 

My suggestion is, in a first step, just always update the cache (or rather store them as `addon_config` in `_read_addons_folder()` just like all the other add-on related stuff.

In a second step, we then can opt out of refresh these data as well. Which will lead to even less I/O.

But I am also ok doing it in one step. :smile:

mdegat01: @agners take a look, I think I incorporated this in a reasonable way for the time being. The logic is all in `StoreManager` as discussed. Also as it turns out, we were swallowing any errors that occurred during `Repository.update` so now we're at least logging those. Hazards of `asyncio.wait` vs `asyncio.gather` and their differences 😬

supervisor/addons/addon.py
mdegat01: I left the method as it, its just not used anywhere. There's not really a good pythonic way to block a method at a descendant I can see and it seemed like the same amount of code to just have it implemented in case.

agners (approved):
LGTM, thanks! :rocket:

supervisor/store/repository.py
agners: Awesome thanks! LGTM!

> Hazards of asyncio.wait vs asyncio.gather and their differences 😬

Ugh I see, nice catch!

PR #4945, opened 2024-03-07
Allow client to change boot slot via API

mdegat01:
## Proposed change

Every time OS is updated we keep the old version in a separate boot slot. We use this for automatic rollbacks in case of catastrophic failure but there's no reason users couldn't use this to do fast and easy manual rollbacks in cases of smaller errors they noticed which aren't catastrophic. This new API allows users to change the active boot slot between A and B

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to documentation pull request: https://github.com/home-assistant/developers.home-assistant/pull/2107
- Link to cli pull request: https://github.com/home-assistant/cli/pull/469

## Checklist

### DIFF
--- a/supervisor/api/const.py
+++ b/supervisor/api/const.py
@@ -1,5 +1,7 @@
 """Const for API."""
 
+from enum import StrEnum
+
 CONTENT_TYPE_BINARY = "application/octet-stream"
 CONTENT_TYPE_JSON = "application/json"
 CONTENT_TYPE_PNG = "image/png"
@@ -14,6 +16,8 @@ ATTR_APPARMOR_VERSION = "apparmor_version"
 ATTR_ATTRIBUTES = "attributes"
 ATTR_AVAILABLE_UPDATES = "available_updates"
 ATTR_BACKGROUND = "background"
+ATTR_BOOT_NAME = "boot_name"
+ATTR_BOOT_SLOTS = "boot_slots"
 ATTR_BOOT_TIMESTAMP = "boot_timestamp"
 ATTR_BOOTS = "boots"
 ATTR_BROADCAST_LLMNR = "broadcast_llmnr"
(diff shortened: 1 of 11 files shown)


supervisor/api/const.py
```diff
@@ -60,3 +65,10 @@
 ATTR_USE_NTP = "use_ntp"
 ATTR_USERS = "users"
 ATTR_VENDOR = "vendor"
+
+
+class BootName(StrEnum):
```
agners, line 70: I think I'd go for BootSlot instead but either way works for me

```suggestion
class BootSlot(StrEnum):
```

agners (commented):
Cool, nice feature! :+1:

supervisor/api/const.py
mdegat01: @agners I was trying to stick with the names I was seeing in Rauc I guess. Rauc seemed to be ID'ing boot slots with names like `kernel.0` and `kernel.1` and referred to `A` and `B` as boot names in the status.

pvizeli: Technical it's called boot slot over all implementation. Embedded Developers vs. Application Developers - I guess on Rauc, there where more application developers.

agners: Reading documentation such as https://rauc.readthedocs.io/en/latest/using.html#manually-switch-to-a-different-slot seems to me that RAUC is using slot for the whole active boot "configuration" as well. Also chapter 2 is mostly talking about slots https://rauc.readthedocs.io/en/latest/basic.html 🤔

agners (approved):

mdegat01:
@agners I realized the API still asked for `boot_name`. I renamed that to `boot_slot` since that's what we went with elsewhere. Let me know if you agree.

PR #4963, opened 2024-03-15
Use Journal Export Format for host (advanced) logs

sairon:
Add methods for handling Journal Export Format and use it for fetching of host logs. This is foundation for colored streaming logs for other endpoints as well.

## Proposed change

So far we used the Systemd Journal gateway daemon only for getting the host logs, and while it can be used also for other system components as well, we were still using Docker logs, particularly because they were not stripping colors from the logs. By leveraging the [Journal Export Format](https://systemd.io/JOURNAL_EXPORT_FORMATS/#journal-export-format), we can get the raw messages including ANSI control messages via a line-based protocol, which doesn't have the same overhead (on both sides) as the JSON format considered for this goal before. This unblocks streaming (follow) for all log endpoints, getting logs from individual boots, and other nice things that Docker logs don't have.

There are two formats possible for the output coming from the log reader - either the raw message with MIME type `text/plain`, or a slightly more verbose one that also has timestamps and unit names included, using _de-facto standard_  `text/x-log` (it's what the `mimetype` tool reports for text files suffixed with `.log` - any maybe others). The verbose is by default used for host logs, but can be used for others by requesting the alternative mimetype which can be useful for detailed debugging from downloaded logs.

One of the considerations was also the performance. Basic testing shown the whole pipeline is able to retrieve logs at rate in magnitude of MB/s on my slow-ish x86 target - which should be plenty for our use case. I will do some more testing on RPi 3 too, just to be sure.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist


supervisor/utils/systemd_journal.py
```diff
@@ -0,0 +1,156 @@
...
+            formatter_ = journal_plain_formatter
+        case LogFormatter.VERBOSE:
+            formatter_ = journal_verbose_formatter
+        case _:
+            raise ValueError(f"Unknown log format: {log_formatter}")
+
+    async with journal_logs as resp:
+        incomplete = b""
+        while not resp.content.at_eof():
+            entry = await resp.content.readuntil(b"\n\n")
```
agners, line 146: Uh I see, this is tripping in binary data, splitting up messages which belong together :thinking: 

It is a bit weird that you need to go "back and forth" logically to complete the message. Either side (this code nor `parse_journal_export_entry`) has all the information to reliably parse a complete log entry.

Looking at https://systemd.io/JOURNAL_EXPORT_FORMATS/, I think I would try parse this slightly differently:

```python
entries = []
entry = {}
stream: aiohttp.StreamReader = resp.content

while not stream.at_eof():
    entry = await stream.readuntil(b"\n")
    line = entry.decode("utf-8")
    # If empty line, means end of message:
    if not line:
        entries.append(entry)
        entry = {}

    idx = line.find("=")
    if idx == -1:
        # Binary message follows
        length_raw = stream.read(8)
        length = int.from_bytes(length_raw, byteorder='little')
        name = line
        data = stream.read(length)
        if stream.read(1) != b"\n":
	    raise NotAtTheEndOfBinaryData
    else:
        # Regular message
        name = line[:idx]
        data = line[idx+1:]

    entries[name] = data
```

It separates things a bit less, so that is the downside. But you won't need to raise exceptions to complete messages etc.

agners: Btw, I went `.decode("utf-8")` here compared to your approach. I am guessing `b"="` might be faster, but I am not sure if there are UTF-8 characters which have a binary `b"="` in it (which are not the equal sign itself). The code relies on reliably detect the equals sign, so I went rather safe and strictly by spec: Check for the separator `=` after decoding UTF-8 ...

sairon: > I think I would try parse this slightly differently

Nice, thank you! I had to do some adjustments of the pseudocode to the correct form but it seems to work great! I started with a dumb implementation I had to adjust on the way as I added more test cases and haven't gotten back to optimize the core function. This should be definitely more efficient and is quite straightforward in the end too!

> It separates things a bit less, so that is the downside. But you won't need to raise exceptions to complete messages etc.

Yeah, and rewrite the tests quite a bit. But thanks to that, I discovered a better way how to mock the `StreamResponse` API, so all good!

>  I am not sure if there are UTF-8 characters which have a binary b"=" in it

As we only discussed it privately, adding it here for the public record. Should not be an issue, so I'll use bytes instead and decode to UTF only when adding to dict. Citing you:
> Right it seems that Byte 2 through 4 always start with the first bit set, so should never conflict with the equal sign.
> https://en.wikipedia.org/wiki/UTF-8#Encoding

sairon: Putting it here again for public record - after rewriting the algorithm to one suggested here the performance seems a bit worse (per your testing - I will also do some more tests myself) but still fast enough. Likely the overhead of the frequent reads from the stream is bigger than going back and forth occasionally.   

I added some more fixes, rewrote the the tests to better mock the gateway API and created two builds on DockerHub: `sairon/amd64-hassio-supervisor:logs-v1` with my original algo, and `sairon/amd64-hassio-supervisor:logs-v2` with yours (both also exist as aarch64 image). Everything should be ready for a final review now, we just need to decide between the faster or more straightforward parsing method :)

supervisor/utils/systemd_journal.py
```diff
@@ -0,0 +1,112 @@
...
+        case LogFormatter.VERBOSE:
+            formatter_ = journal_verbose_formatter
+        case _:
+            raise ValueError(f"Unknown log format: {log_formatter}")
+
+    async with journal_logs as resp:
+        entries: dict[str, str] = {}
+        while not resp.content.at_eof():
+            entry = await resp.content.readuntil(b"\n")
+            line = entry
```
agners, line 81:
```suggestion
            line = await resp.content.readuntil(b"\n")
```

supervisor/utils/systemd_journal.py
```diff
@@ -0,0 +1,112 @@
...
+            # newline means end of message:
+            if line == b"\n":
+                yield formatter_(entries)
```
agners, line 84: We have to clear the dictionary here, otherwise we leak entries into the next log entry (if they are not populated there).
```suggestion
                yield formatter_(entries)
                entries = {}
```

supervisor/utils/systemd_journal.py
```diff
@@ -0,0 +1,112 @@
...
+                # field name, followed by newline
+                name = line[:-1].decode("utf-8")  # strip \n
+                # followed by a binary 64-bit little endian size value,
+                length_raw = await resp.content.readexactly(8)
+                length = int.from_bytes(length_raw, byteorder="little")
+                # followed by the binary field data,
+                data = await resp.content.readexactly(length)
+                # followed by a newline as separator to the next field.
+                if (await resp.content.readexactly(1)) != b"\n":
+                    raise MalformedBinaryEntry(line)
```
agners, line 109: It seems we do lose some time due to the repeated read calls, we can safe this one:

```suggestion
                data = await resp.content.readexactly(length + 1)
                # followed by a newline as separator to the next field.
                if data[len(data-1)] != "\n":
                    raise MalformedBinaryEntry(line)
```

Now data will be too long though. Making an unconditional copy using `data = data[:-1]` is bad however :cry: 

But it seems not to fully recoop the time.

We probably can also safe a copy of 

```python
            if name not in formatter_.required_fields:
                continue

            if idx != -1:
                data = line[idx + 1 : -1]  # strip \n
            else:
                data = data[:-1]
            entries[name] = data.decode("utf-8")
```

It all doesn't seem to make a massive difference in a few tests here :cry:

agners (changes requested):

supervisor/utils/systemd_journal.py
sairon: Ah, remnant of code before I moved the conversion from bytes :+1:

supervisor/utils/systemd_journal.py
sairon: Oh, good catch!

supervisor/utils/systemd_journal.py
agners: > we still do ~40 reads for every line for a single journal entry, so it's not surprising it doesn't make much difference.

Right :cry:  And from tracing, it seems that awaiting for the next bytes is really what slows the whole thing down.

I guess we'd need to make this "more synchronous" if we want to speed up. But it gets more complicated as we have reads in several places.

> I'll rather duplicate the if...continue in each block for readability if you don't mind.

Hacking a bit further I came up with this:

```python
            required_fields = {s.encode('utf-8'): s for s in formatter_.required_fields}
            while not resp.content.at_eof():
                ...
                name, sep, data = line.partition(b'=')
                if not sep:
                    # followed by a binary 64-bit little endian size value,
                    length_raw = await resp.content.readexactly(8)
                    length = int.from_bytes(length_raw, byteorder="little")
                    # followed by the binary field data,
                    data = await resp.content.readexactly(length + 1)
                    # followed by a newline as separator to the next field.
                    if data[len(data-1)] != "\n":
                        raise MalformedBinaryEntry(line)

                name = required_fields.get(name)
                if not name:
                    continue

                if not sep:
                    data = data[:-1]

                entries[name] = data.decode("utf-8")
```

My goal was to safe decoding name for each entry, and make less copy. I am not sure if `line.partition()` really leads to less copy, but it kinda ended up kinda neat code :sweat_smile:  But all didn't really make it faster :see_no_evil:  So not sure if the `required_fields` look up is really worth it.

supervisor/utils/systemd_journal.py
agners: Given the performance difference isn't that drastic (we went from 543kbytes/s -> ~450kbytes/s) it doesn't seem worth to me to take the additional complexity for the gain. The more straightforward path is probably easier to further performance optimize, so I'd go with that as a start.

Probably making the inner most parsing synchronous would fix the speed difference, but it is just PITA as we don't know how much bytes we have to read ahead of time :cry: I tried to find if there is a size limit for a single log entry. If so, we could fill a buffer asynchrounsly to MAX_LOG_ENTRY size, and then parse the individual entries synchronously.

Or implement the thing in Rust :sweat:

supervisor/utils/systemd_journal.py
```diff
@@ -0,0 +1,117 @@
...
+                if name not in formatter_.required_fields:
+                    # we must read to the end of the entry in the stream, so we can
+                    # only continue the loop here
+                    continue
+
+            entries[name] = data.decode("utf-8")
```
agners, line 117: We can combine the continue like so:

```suggestion
            name, sep, data = line.partition(b'=')
            if not sep:
                # Other journal fields are serialized in a special binary safe way:
                # field name, followed by newline
                name = name[:-1]  # strip \n
                # followed by a binary 64-bit little endian size value,
                length_raw = await resp.content.readexactly(8)
                length = int.from_bytes(length_raw, byteorder="little")
                # followed by the binary field data,
                data = await resp.content.readexactly(length)
                # followed by a newline as separator to the next field.
                if (await resp.content.readexactly(1)) != b"\n":
                    raise MalformedBinaryEntry(line)

            name = name.decode("utf-8")
            if name not in formatter_.required_fields:
                # we must read to the end of the entry in the stream, so we can
                # only continue the loop here
                continue
            entries[name] = data.decode("utf-8")
```

sairon: Committed with minor adjustments.

mdegat01 (approved):
This looks great, nice addition!

supervisor/utils/systemd_journal.py
```diff
@@ -0,0 +1,114 @@
...
+from datetime import UTC, datetime
+from functools import wraps
+
+from aiohttp import ClientResponse
+
+from supervisor.host.const import LogFormatter
+
+
+class MalformedBinaryEntry(Exception):
+    """Raised when binary entry in the journal isn't followed by a newline."""
```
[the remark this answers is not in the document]
sairon: Sure. Shouldn't it also be a subclass of HassioError then, so it's handled more gracefully in the API?

supervisor/api/host.py
```diff
@@ -175,26 +186,31 @@ async def advanced_logs(
 
         if ACCEPT in request.headers and request.headers[ACCEPT] not in [
             CONTENT_TYPE_TEXT,
+            CONTENT_TYPE_X_LOG,
             "*/*",
         ]:
             raise APIError(
-                "Invalid content type requested. Only text/plain supported for now."
+                "Invalid content type requested. Only text/plain and text/x-log "
```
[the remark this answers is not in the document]
sairon: I can't think of any other right now, and adding it will likely mean some larger changes around the place, so I vote for keeping it like this.

supervisor/utils/systemd_journal.py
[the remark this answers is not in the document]
mdegat01: Oh yes definitely. I didn't notice that. Please make it a subclass of `HassioError` otherwise we'll treat it as an unhandled internal error everywhere.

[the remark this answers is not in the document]
sairon: Done, I've also adjusted the name and added a test for it.

PR #4968, opened 2024-03-19
Unsupported if wrong image used on virtualization

mdegat01:
## Proposed change

Home Assistant OS has a specific image for use on virtualized systems called `ova`. It is possible to install the `generic-x86-64` and people do but then they report issues when expected options like guest tools are missing. This lets them know they have installed the wrong image and the repair will direct them to the correct one for their environment.

## Type of change

- [ ] Dependency upgrade

## Additional information

     - Translation strings for repair: https://github.com/home-assistant/core/pull/113882
     - Unsupported reason page: https://github.com/home-assistant/home-assistant.io/pull/31971
     - API Doc: https://github.com/home-assistant/developers.home-assistant/pull/2117
- Link to cli pull request:

## Checklist


mdegat01:
@agners do you want to write the unsupported page for this one? We'll need a new page at `https://www.home-assistant.io/more-info/virtualized_image` (assuming we're ok with that identifier for this issue)

supervisor/resolution/evaluations/virtualization_image.py
```diff
@@ -0,0 +1,36 @@
...
+    @property
+    def states(self) -> list[CoreState]:
+        """Return a list of valid states when this evaluation can run."""
+        return [CoreState.STARTUP]
+
+    async def evaluate(self):
+        """Run evaluation."""
+        if not self.sys_os.available:
+            return False
+        return self.sys_dbus.systemd.virtualization and self.sys_os.board != "ova"
```
sairon, line 36: I _think_ it's valid for `generic-aarch64` to run in a VM too.

supervisor/resolution/evaluations/virtualization_image.py
```diff
@@ -0,0 +1,39 @@
...
+    @property
+    def states(self) -> list[CoreState]:
+        """Return a list of valid states when this evaluation can run."""
+        return [CoreState.STARTUP]
+
+    async def evaluate(self):
+        """Run evaluation."""
+        if not self.sys_os.available:
+            return False
+        return self.sys_host.info.virtualization and self.sys_os.board not in {
```
agners, line 36: I was thinking to use the Chassis property and check for `vm` here. But checking virtualization is most likely largely equivalent. It seems that if Chassis can't be read through DMI BIOS interface, it falls back to the virtualization detection (see https://github.com/systemd/systemd/blob/v252/src/hostname/hostnamed.c#L283-L286).

I guess with this we also mark container as unsupported, which most likely make sense. I don't think it is reasonable or possible to run full HAOS in a container :sweat_smile:

supervisor/core.py
```diff
@@ -194,6 +194,9 @@ async def start(self):
         """Start Supervisor orchestration."""
         self.state = CoreState.STARTUP
 
+        # Evaluate the system
+        await self.sys_resolution.evaluate.evaluate_system()
```
agners, line 198: Hm, is this additional evaluation really necessary? It seems to me that the one at the end of setup just above should be enough? Or maybe we can drop that one instead?

supervisor/resolution/evaluations/virtualization_image.py
mdegat01: So is this good then or should I change to chassis? Or check both?

agners: Either way should work, I think it is fine to use the virtualization property.

agners (approved):

PR #4972, opened 2024-03-20
Use Systemd Journal API for all logs endpoints in API

### LINKED ISSUE #3309: RFC: Extend Logging to allow Selecting/Limiting entries and Follow mode
## Context

Currently log endpoints return a fixed amount of logs (or all available logs in the kernel buffer in the case of the host logs). Sometimes more logs should be returned, or users want to browse older logs. Also it often convenient to observe logs while they are being generated. Linux tools typically have a follow mode (`-f`). This RFC aims to specify how these features can be exposed through he API.

### Selecting/Limiting entries

To select certain entries or limit the amount of entries some type of selection should be possible. This is also often called paging, where a start number and item count is passed (e.g. via query parameters `start=20&count=50`).

Entries in the context of logs today are typically log lines.

In RESTful API's often query parameters are used for pagination. There is also the option of using the [Range header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Range). Since with RESTful API's one tries to use HTTP features as much as possible, it seems a very good fit for this use case:

_Header_

* `Range`: `entries=:num_skip:num_entries`: Allows to select a number of log entries. Negative numbers for `num_skip` allow to select from the end of the log.

If the `Range` header is not present, the same behavior as implemented today would apply (100 log entries).

Note: This idea is taken directly from [systemd-journal-gatewayd's Header support](https://www.freedesktop.org/software/systemd/man/systemd-journal-gatewayd.service.html#Range%20header). However, it is generic enough that it can be implemented for other log sources easily.

## Follow mode

Continuous logging can be archived by keeping the HTTP connection open and pass new logs as they come in from the log source. This can be implemented using [aiohttp's Streaming API](https://docs.aiohttp.org/en/stable/streams.html).

To keep backwards compatibility, the following/streaming mode should be explicitly enabled. I'd suggest

_URI query string_

- `follow`: If specified the API doesn't close the HTTP connection and continuous to pass new log entries to the client.

When the query string is set, the API endpoint continuously return logs as they come in. With HTTP 1.1 `Transfer-Encoding` `Chunked` is being used.

sairon:
## Proposed change

Replace all logs endpoints using container logs with wrapped advanced_logs function, adding possibility to get logs from previous boots and following the logs. Supervisor logs are an excetion where Docker logs are still used - in case an exception is raised while accessing the Systemd logs, they're used as fallback - otherwise we wouldn't have an easy way to see what went wrong.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR fixes or closes issue: fixes #3309
- Link to cli pull request: https://github.com/home-assistant/cli/pull/471

## Checklist

- [x] Documentation added/updated for [developers.home-assistant.io][docs-repository]: https://github.com/home-assistant/developers.home-assistant/pull/2124

### DIFF
--- a/supervisor/api/__init__.py
+++ b/supervisor/api/__init__.py
@@ -7,6 +7,7 @@ from typing import Any
 from aiohttp import web
 from aiohttp_fast_url_dispatcher import FastUrlDispatcher, attach_fast_url_dispatcher
 
+from ..addons.addon import Addon
 from ..const import AddonState
 from ..coresys import CoreSys, CoreSysAttributes
 from ..exceptions import APIAddonNotInstalled
@@ -71,8 +72,14 @@ class RestAPI(CoreSysAttributes):
         self._runner: web.AppRunner = web.AppRunner(self.webapp, shutdown_timeout=5)
         self._site: web.TCPSite | None = None
 
+        # share single host API handler for reuse in logging endpoints
+        self._api_host: APIHost | None = None
+
     async def load(self) -> None:
         """Register REST API Calls."""
+        self._api_host = APIHost()
+        self._api_host.coresys = self.coresys
+
         self._register_addons()
         self._register_audio()
         self._register_auth()
@@ -102,10 +109,41 @@ class RestAPI(CoreSysAttributes):
 
         await self.start()
 
+    def _register_advanced_logs(self, path: str, syslog_identifier: str):
+        """Register logs endpoint for a given path, returning logs for single syslog identifier."""
+
+        self.webapp.add_routes(
+            [
+                web.get(
+                    f"{path}/logs",
+                    partial(self._api_host.advanced_logs, identifier=syslog_identifier),
+                ),
+                web.get(
+                    f"{path}/logs/follow",
+                    partial(
+                        self._api_host.advanced_logs,
+                        identifier=syslog_identifier,
+                        follow=True,
+                    ),
+                ),
+                web.get(
+                    f"{path}/logs/boots/{{bootid}}",
+                    partial(self._api_host.advanced_logs, identifier=syslog_identifier),
+                ),
+                web.get(
+                    f"{path}/logs/boots/{{bootid}}/follow",
+                    partial(
+                        self._api_host.advanced_logs,
+                        identifier=syslog_identifier,
+                        follow=True,
+                    ),
+                ),
+            ]
+        )
+
     def _register_host(self) -> None:
         """Register hostcontrol functions."""
-        api_host = APIHost()
-        api_host.coresys = self.coresys
+        api_host = self._api_host
 
         self.webapp.add_routes(
             [
@@ -261,11 +299,11 @@ class RestAPI(CoreSysAttributes):
             [
                 web.get("/multicast/info", api_multicast.info),
                 web.get("/multicast/stats", api_multicast.stats),
-                web.get("/multicast/logs", api_multicast.logs),
                 web.post("/multicast/update", api_multicast.update),
                 web.post("/multicast/restart", api_multicast.restart),
             ]
         )
+        self._register_advanced_logs("/multicast", "hassio_multicast")
 
     def _register_hardware(self) -> None:
         """Register hardware functions."""
@@ -352,12 +390,39 @@ class RestAPI(CoreSysAttributes):
                 web.get("/supervisor/ping", api_supervisor.ping),
                 web.get("/supervisor/info", api_supervisor.info),
                 web.get("/supervisor/stats", api_supervisor.stats),
-                web.get("/supervisor/logs", api_supervisor.logs),
                 web.post("/supervisor/update", api_supervisor.update),
                 web.post("/supervisor/reload", api_supervisor.reload),
                 web.post("/supervisor/restart", api_supervisor.restart),
                 web.post("/supervisor/options", api_supervisor.options),
                 web.post("/supervisor/repair", api_supervisor.repair),
+            ]
+        )
+
+        async def get_supervisor_logs(*args, **kwargs):
+            try:
+                kwargs["identifier"] = "hassio_supervisor"
+                return await self._api_host.advanced_logs(*args, **kwargs)
+            except Exception:  # pylint: disable=broad-exception-caught
+                # Supervisor logs are critical, so catch everything, log the exception
+                # and try to return Docker container logs as the fallback
+                _LOGGER.exception(
+                    "Failed to get supervisor logs using advanced_logs API"
+                )
+                kwargs.pop("identifier")
+                return await api_supervisor.logs(*args, **kwargs)
+
+        self.webapp.add_routes(
+            [
+                web.get("/supervisor/logs", get_supervisor_logs),
+                web.get(
+                    "/supervisor/logs/follow",
+                    partial(get_supervisor_logs, follow=True),
+                ),
+                web.get("/supervisor/logs/boots/{bootid}", get_supervisor_logs),
+                web.get(
+                    "/supervisor/logs/boots/{bootid}/follow",
+                    partial(get_supervisor_logs, follow=True),
+                ),
             ]
         )
 
@@ -369,7 +434,6 @@ class RestAPI(CoreSysAttributes):
         self.webapp.add_routes(
             [
                 web.get("/core/info", api_hass.info),
-                web.get("/core/logs", api_hass.logs),
                 web.get("/core/stats", api_hass.stats),
                 web.post("/core/options", api_hass.options),
                 web.post("/core/update", api_hass.update),
@@ -381,11 +445,12 @@ class RestAPI(CoreSysAttributes):
             ]
         )
 
+        self._register_advanced_logs("/core", "homeassistant")
+
         # Reroute from legacy
         self.webapp.add_routes(
             [
                 web.get("/homeassistant/info", api_hass.info),
-                web.get("/homeassistant/logs", api_hass.logs),
                 web.get("/homeassistant/stats", api_hass.stats),
                 web.post("/homeassistant/options", api_hass.options),
                 web.post("/homeassistant/restart", api_hass.restart),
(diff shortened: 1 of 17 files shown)


mdegat01 (changes requested):
Mostly minor comments although there is an issue with the addon logs one that requires some tweaks, see the comment below.

supervisor/api/__init__.py
```diff
@@ -443,13 +510,32 @@ def _register_addons(self) -> None:
...
                 web.post("/addons/{addon}/stdin", api_addons.stdin),
                 web.post("/addons/{addon}/security", api_addons.security),
                 web.get("/addons/{addon}/stats", api_addons.stats),
             ]
         )
 
+        async def get_addon_logs(request, *args, **kwargs):
+            addon: Addon = api_addons.get_addon_for_request(request)
+            kwargs["identifier"] = f"addon_{addon.slug}"
+            return await self._api_host.advanced_logs(request, *args, **kwargs)
```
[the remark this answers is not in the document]
sairon: > Or move all the logic in self._api_host.advanced_logs into another method without the annotation and have both places call that and then add the annotation to this method.

This sounds like the best option to me - splitting the `advanced_logs` into a method fetching the logs and and API endpoint handler. However, it creates an inconsistency between the add-on logs endpoints and the other ones, because with `api_process_raw(CONTENT_TYPE_BINARY)` you always get a response with this mimetype, instead of the `text/plain` that is returned normally (and it's also a bit weird you get a different type than what you specified in the `Accept` header IMO). Moreover, "plain" usage of `advanced_logs` annotated with `api_process` returns the content type specified within the function, but errors are returned as `application/json` (this was always the case, not introduced by my PR).

And one more thing - while `api_process_raw` handles the errors sensibly, it does not return bad request - it always returns response with HTTP 200. Also this was always there, and I'm not sure if it should be changed within the scope of this PR, or even changed at all, because I might not see the big picture there.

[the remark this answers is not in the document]
mdegat01: > However, it creates an inconsistency between the add-on logs endpoints and the other ones, because with `api_process_raw(CONTENT_TYPE_BINARY)` you always get a response with this mimetype, instead of the `text/plain` that is returned normally (and it's also a bit weird you get a different type than what you specified in the Accept header IMO). 

Oh that's fair. I'm not sure we currently have a way in this code to react to the `Accept` header but that does make sense. I was mostly concerned about the consistent error handling to be honest.

Perhaps in that case it would be best to make the `web.Response` object in the method and return that? And then make a new annotation that exclusively handles errors. It assumes any non-exception return is already a `web.Response` object and just returns it as is.

> Moreover, "plain" usage of advanced_logs annotated with api_process returns the content type specified within the function, but errors are returned as application/json (this was always the case, not introduced by my PR).

Oh I didn't realize that. I thought the code in there seemed to attempt to encode its return to binary so it was always that, I guess I didn't look at it closely enough. I'm open to suggestions here, just want to make sure we aren't bubbling up exceptions past our code and instead returning error responses.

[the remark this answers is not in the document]
sairon: Since I didn't want to break things and `api_proces_raw` doesn't behave in the way I'd like (e.g. it always returns HTTP 200 and doesn't handle all `HassioError` types), I created a new decorator for API functions that's something between `api_process` and `api_process_raw`. I also used it for other endpoints to address the consistency problems I mention above.

sairon:
This PR should also address this older RFC: https://github.com/home-assistant/supervisor/issues/3309

supervisor/api/utils.py
```diff
@@ -101,13 +101,15 @@ async def wrap_api(api, *args, **kwargs):
...
                 msg_type = content
-            except (APIError, APIForbidden) as err:
+            except (APIError, APIForbidden, HassioError) as err:
                 msg_data = str(err).encode()
-                msg_type = CONTENT_TYPE_BINARY
+                msg_type = error_type or CONTENT_TYPE_BINARY
             except HassioError:
                 msg_data = b""
-                msg_type = CONTENT_TYPE_BINARY
+                msg_type = error_type or CONTENT_TYPE_BINARY
```
mdegat01, line 112:
```suggestion
            except HassioError as err:
                msg_data = str(err).encode()
                msg_type = CONTENT_TYPE_BINARY
                msg_type = error_type or CONTENT_TYPE_BINARY
```
Tbh, I'm not entirely sure how this code passed the linter since `HassioError` is in there twice in two different `except` clauses. But `APIForbidden` extends from `APIError` which extends from `HassioError`. So if you have `HassioError` in the first one then the others are just redundant. And the second except for just `HassioError` will never be called.

Although really I think the code here should be this:
```py
            status = 200
            try:
                msg_data = await method(api, *args, **kwargs)
                if isinstance(msg_data, (web.Response, web.StreamResponse)):
                    return msg_data
                msg_type = content
            except APIError as err:
                msg_data = str(err).encode()
                msg_type = error_type or CONTENT_TYPE_BINARY
                status = err.status
            except HassioError:
                msg_data = str(err).encode()
                msg_type = error_type or CONTENT_TYPE_BINARY
                status = 400

            return web.Response(body=msg_data, content_type=msg_type, status=status)
```
Because right now we don't reply with 403s when an `APIForbidden` is raised. But I don't think you're actually using this annotation at all anymore are you? Perhaps we should just leave all this for another PR.

supervisor/api/utils.py
```diff
@@ -116,6 +118,28 @@ async def wrap_api(api, *args, **kwargs):
...
+    """Ensure errors are handled and returned with specified content_type."""
+
+    def decorator(method):
+        async def wrapper(api, *args, **kwargs):
+            status = 200
+            try:
+                response = await method(api, *args, **kwargs)
+            except HassioError as err:
+                response = str(err)
+                status = 400
```
mdegat01, line 131:
```suggestion
            except APIError as err:
                response = str(err)
                status = err.status
            except HassioError as err:
                response = str(err)
                status = 400
```
This way if a method raises an `APIForbidden` (or if we make other HTTP Status codes into exceptions in the future) it will use the correct status code.

supervisor/api/utils.py
```diff
@@ -116,6 +118,28 @@ async def wrap_api(api, *args, **kwargs):
     return wrap_method
 
 
+def api_process_custom(content_type):
```
mdegat01, line 121: This annotation looks kind of redundant `api_process_raw` to be honest. It's set up exactly the same way except this one only actually supports content type text since it always returns errors as plain text, regardless of the specified content type. Whereas `api_process_raw` only actually supports content type binary because it always encodes errors as binary, regardless of the specified content type.

What if we just did this instead?
```py
def api_process_raw(content, *, error_type=None):
    """Wrap content_type into function."""

    def wrap_method(method):
        """Wrap function with raw output to rest api."""
        async def wrap_api(api, *args, **kwargs):
            """Return api information."""
            try:
                msg_data = await method(api, *args, **kwargs)
            except HassioError as err:
                return api_return_error(err, error_type=error_type or CONTENT_TYPE_BINARY)

            if isinstance(msg_data, (web.Response, web.StreamResponse)):
                return msg_data
            return web.Response(body=msg_data, content_type=content)

def api_return_error(
    error: Exception | None = None, message: str | None = None, error_type: str | None = None
) -> web.Response:
    """Return an API error message."""
    if error and not message:
        message = get_message_from_exception_chain(error)
        if check_exception_chain(error, DockerAPIError):
            message = format_message(message)
    if not message:
        message = "Unknown error, see supervisor"
    
    status = 400
    if (is_api_error := isinstance(error, APIError)):
        status = error.status

    match error_type:
        case CONTENT_TYPE_TEXT:
            return web.Response(body=message, content_type=error_type, status=status)
        case CONTENT_TYPE_BINARY:
            return web.Response(body=message.encode(), content_type=error_type, status=status)
        case _:
            result = {
                JSON_RESULT: RESULT_ERROR,
                JSON_MESSAGE: message,
            }
            if is_api_error and error.job_id:
                result[JSON_JOB_ID] = error.job_id

    return web.json_response(
        result,
        status=status,
        dumps=json_dumps,
    )
```
This has a couple benefits:
1. `api_process_raw` works and we don't need to create a new `api_process_custom`
2. `api_process_raw` actually supports content type text and binary and isn't just secretly binary only because of how it does error
3. Shared error handling in all our annotations. Including how we make the message from the exception (since that is currently quite different in `api_process_raw` and `api_process_custom` which doesn't seem right)

mdegat01 (changes requested):

supervisor/api/utils.py
mdegat01: Ok actually start from the last comment in these 3. Since if you make that change, these other two are resolved. Sorry I realized after the other two that there might be a better way to do this then a new annotation.

sairon: TBH this was actually committed accidentally - it just slipped in because I have a habit of running `git add -u ; pre-commit ; git diff` to re-format the code and forgot to revert these changes when I implemented `api_process_custom` :facepalm:

supervisor/api/utils.py
sairon: This is definitely better - and something I also wanted to do, I was just hesitant to make larger adjustments to it - for example I didn't know why it had `msg_data` set to empty string for `HassioError` and returned message text for `APIError, APIForbidden`. So I agree, let's go with your suggestion, thanks!

(I just needed to do adjust the match cases to dotted attributes because the suggested syntax is invalid, and reformatted the code.)

mdegat01 (approved):
Looks great 👍

sairon:
> This is changing the content type from binary to text. Are the current users (probably only frontend?) able to handle this?

Yes, frontend doesn't really care about the content type. Host logs used text in the previous version as well, now it's consistent for all `.../logs` endpoints.

> I guess the reverse can be true to. But yeah, since we try journald logs first, it will just not be noticeable in this case.

I expect the journald erros to be rather persistent (until a reboot, etc.) than intermittent. And Docker logs are used only as the fallback, so if it also fails, you're done :)

PR #4991, opened 2024-03-29
Allow adoption of existing data disk

mdegat01:
## Proposed change

When a user plugs in an existing data disk from another system we encourage them to rename it otherwise it causes a race condition at startup. However users often expect another option - to use it as their data disk. For example if they just purchased a new system they might expect to move the external data disk from one to the other and have it keep working.

This adds that option. It largely works as is. However at startup we now check all the images to see if we have the right ones for the current machine/architecture. Because if you move a data disk from one system to another we may need to get all new images depending on what changed.

PR in core to add translation strings and support for this new repair: https://github.com/home-assistant/core/pull/114891

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/addons/addon.py
+++ b/supervisor/addons/addon.py
@@ -198,6 +198,12 @@ class Addon(AddonModel):
         with suppress(DockerError):
             await self.instance.attach(version=self.version)
 
+        with suppress(DockerError):
+            await self.instance.check_image(
+                self.version, self._image(self.data), self.arch
+            )
+            self.persist[ATTR_IMAGE] = self.instance.image
+
     @property
     def ip_address(self) -> IPv4Address:
         """Return IP of add-on instance."""
--- a/supervisor/addons/addon.py
+++ b/supervisor/addons/addon.py
@@ -195,21 +195,19 @@ class Addon(AddonModel):
         )
 
         await self._check_ingress_port()
+        default_image = self._image(self.data)
         try:
             await self.instance.attach(version=self.version)
 
             # Ensure we are using correct image for this system
-            image = self._image(self.data)
-            await self.instance.check_image(self.version, image, self.arch)
+            await self.instance.check_image(self.version, default_image, self.arch)
         except DockerError:
-            _LOGGER.info(
-                "No %s addon Docker image %s found", self.slug, self.instance.image
-            )
+            _LOGGER.info("No %s addon Docker image %s found", self.slug, self.image)
             with suppress(AddonsError):
-                await self.instance.install(self.version, image, arch=self.arch)
-        else:
-            self.persist[ATTR_IMAGE] = image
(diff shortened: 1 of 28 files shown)


supervisor/addons/addon.py
```diff
@@ -198,6 +198,11 @@ async def load(self) -> None:
         with suppress(DockerError):
             await self.instance.attach(version=self.version)
 
+        with suppress(DockerError):
+            image = self._image(self.data)
+            await self.instance.check_image(self.version, image, self.arch)
+            self.persist[ATTR_IMAGE] = image
```
pvizeli, line 204: Doing this kind of action at the `load` is very early in the boot state. If there is no internet (which we not know as it's to early in the boot state) it can fail - today we still boot. We should make sure that we still boot without internet. Maybe it could make sense to to fix that after we loaded the NM and know if we can do this kind of operation or even as fixup after the system is booted and we have all system component loaded and working

mdegat01: All the load methods are called during setup stage, not boot. Like that particular one is called as part of [this](https://github.com/home-assistant/supervisor/blob/a894c4589eed3c3ac03370d9d74284c1a6f4086b/supervisor/core.py#L149). I also checked before adding it and plugins, home assistant and addons are all loaded after dbus and host so that check should be functional by then.

Still that is why I added the suppress there just like we were doing before, to prevent failing startup over this.

supervisor/plugins/base.py
```diff
@@ -160,6 +165,8 @@ async def load(self) -> None:
             await self.instance.attach(
```
agners, line 165: This is attaching to a running container. It seems `DockerError` is raised if that fails. So, when we adopt a new disk, I expect this to raise `DockerError`. This in turn calls `install()` :thinking: 

How is `check_image` getting called here when we actually need a new image? :thinking:

agners: Hm, in the add-on case, we suppress `DockerError`, so I see how this works in that case. But in general, it seems to me we want to check the image before attach no? Or maybe as part of attach?

agners (commented):
This lead to a bit more changes than I expected :cry: But there are also some nice cleanups (like de-duplicating the plug-ins `install()` method).

There are quite some special cases, like locally built add-on and add-on which use a architecture not equals to Supervisor's architecture (a supported/fallback arch). From what I can tell, this code seems to handle them correctly.

supervisor/plugins/base.py
mdegat01: That attaches to a container if it exists. If it does not exist then it just looks for the expected image and stores its metadata instead. It only raises an exception if both the container and the expected image do not exist

agners (approved):

Issue #5021, opened 2024-04-17
Unable to view supervisor (or addon) logs when running in devcontainer


lildude:
### Describe the issue you are experiencing

Attempting to view any of the logs within the devcontainer fails with `Failed to get supervisor logs, No systemd-journal-gatewayd Unix socket available`:

![Failed to get supervisor logs, No systemd-journal-gatewayd Unix socket available](https://github.com/home-assistant/supervisor/assets/627280/5740ea3d-1a65-4473-8bda-a9b34704ff85)

I first noticed this when my addon devcontainer suddenly started reporting this error and then discovered opening this repo's devcontainer experiences the same issue.

After a lot of digging and debugging, https://github.com/home-assistant/supervisor/pull/4963 appears to be the PR that introduced the changes that "broke" the ability to view the logs when running in the devcontainer. Checking out the commit before, 0e0fadd7, and following the repro steps "fixes" the issue.

Forcing the addon container to use supervisor version `2024.03.1.dev1803` in `/etc/supervisor_scripts/common` within the addon devcontainer resolves the issue too.

### What type of installation are you running?

Home Assistant Supervised

### Which operating system are you running on?

Debian

### Steps to reproduce the issue

1. Clone https://github.com/home-assistant/supervisor to your system
2. Open in VSCode
3. Accept the prompt to re-open in a container
4. Start the supervisor by running the "Run Supervisor" task
5. Open 0.0.0.0:9123 in your browser and configure as normal
6. Go to Settings -> System -> Logs -> Supervisor
7. 💥 

### Anything in the Supervisor logs that might be useful for us?

```
2024-04-17 11:18:25.410 WARNING (MainThread) [supervisor.resolution.evaluations.base] Systemd journal is not working correctly or inaccessible (more-info: https://www.home-assistant.io/more-info/unsupported/systemd_journal)
2024-04-17 11:18:25.411 WARNING (MainThread) [supervisor.resolution.evaluations.base] NetworkManager is not correctly configured (more-info: https://www.home-assistant.io/more-info/unsupported/network_manager)
2024-04-17 11:18:25.412 WARNING (MainThread) [supervisor.resolution.evaluations.base] Docker cgroup version 2 is not supported! {'1'} (more-info: https://www.home-assistant.io/more-info/unsupported/cgroup_version)
[...]

PR #5022, opened 2024-04-18
Fix Supervisor logs fallback

sairon:
## Proposed change

Supervisor logs fallback in get_supervisor_logs didn't work properly because the exception was caught in api_process_raw instead. This was not discovered in tests because the side effect raised OSError, which isn't handled there.

To address that, I split the advanced_logs to two functions, one being a wrapped API handler, one being plain function returning response without any additional error handling. The tests now check for both cases of errors (HassioError and random generic Python error).

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR is related to issue: #5021
- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/api/__init__.py
+++ b/supervisor/api/__init__.py
@@ -401,7 +401,7 @@ class RestAPI(CoreSysAttributes):
 
         async def get_supervisor_logs(*args, **kwargs):
             try:
-                return await self._api_host.advanced_logs(
+                return await self._api_host.advanced_logs_handler(
                     *args, identifier="hassio_supervisor", **kwargs
                 )
             except Exception as err:  # pylint: disable=broad-exception-caught
--- a/supervisor/api/host.py
+++ b/supervisor/api/host.py
@@ -1,4 +1,5 @@
 """Init file for Supervisor host RESTful API."""
+
 import asyncio
 from contextlib import suppress
 import logging
@@ -163,8 +164,7 @@ class APIHost(CoreSysAttributes):
                 raise APIError() from err
         return possible_offset
 
-    @api_process_raw(CONTENT_TYPE_TEXT, error_type=CONTENT_TYPE_TEXT)
-    async def advanced_logs(
+    async def advanced_logs_handler(
         self, request: web.Request, identifier: str | None = None, follow: bool = False
     ) -> web.StreamResponse:
         """Return systemd-journald logs."""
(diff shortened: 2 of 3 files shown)


mdegat01 (changes requested):
We should avoid letting unhandled errors like `OSError` bubble up if we don't have to. Rather then changing then changing the API methods like this, couldn't we just add `OSError` to the `except` line here?
https://github.com/home-assistant/supervisor/blob/b09230a884d93eff0f8baf7db60f39d061af2710/supervisor/host/logs.py#L161-L164

Or if you don't think that should raise `HostServiceError` then catch it separately and raise a more appropriate exception instead that extends from `HassioError`. And take whatever other actions you think are required first (log, etc.)

sairon:
> We should avoid letting unhandled errors like `OSError` bubble up if we don't have to.

It wouldn't actually bubble up in the test case - it will be caught by the "too broad" exception cause.

> Rather then changing then changing the API methods like this, couldn't we just add `OSError` to the `except` line here?

The `OSError` is actually one of the random (yet possible) exception classes, so catching that is not the solution.

> Or if you don't think that should raise `HostServiceError` then catch it separately and raise a more appropriate exception instead that extends from `HassioError`. And take whatever other actions you think are required first (log, etc.)

To sum it up - the broad exception handler is there to make sure that `get_supervisor_logs` always returns some output if possible - because without Supervisor logs, you're totally in the dark. This is really just to handle rare edge cases - and if those happen, you should still see the error in the output through the `_LOGGER.exception` called there. We **do not** want this to be handled in any of the upper layers because it is there just as a fallback. If something is really screwed up, you'll get another stack trace within the exception handler.

What I realize now, knowing the Supervisor codebase a bit more, I think that a call to `capture_exception` in the generic `Exception` handler is what is _really_ necessary, so we learn about those those rather rare cases that are caught in Sentry. Am I right?

mdegat01 (approved):

supervisor/api/__init__.py
```diff
@@ -401,7 +401,7 @@ def _register_supervisor(self) -> None:
 
         async def get_supervisor_logs(*args, **kwargs):
             try:
-                return await self._api_host.advanced_logs(
+                return await self._api_host.advanced_logs_handler(
                     *args, identifier="hassio_supervisor", **kwargs
                 )
             except Exception as err:  # pylint: disable=broad-exception-caught
```
agners, line 407: Now that we have a known use case (as this PR got triggered by devcontainer usage, see #5021), can we handle this specific and make it not report to Sentry?

I guess the exact error here might depond on weather systemd-journal-gatewayd is available? E.g. would the exception be different if we were to revert https://github.com/home-assistant/devcontainer/pull/6?

agners: Let's get this handled in a follow up PR. This PR does what it advertises: Fix the fallback as it is intended currently.

agners (approved):

PR #5023, opened 2024-04-18
Use signals to recognize new disks immediately

mdegat01:
## Proposed change

Currently when a user connects a new data disk they have to wait for the host reload scheduled task to run before they can see it as a data disk target or learn about errors related to its label. This PR processes signals from UDisks2 to learn immediately when disk devices are connected or disconnected and make them available right away.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist


tests/dbus_service_mocks/udisks2.py
```diff
@@ -0,0 +1,41 @@
...
+
+def setup(object_path: str | None = None) -> DBusServiceMock:
+    """Create dbus mock object."""
+    return UDisks2()
+
+
+class UDisks2(DBusServiceMock):
+    """UDisks2 base object mock.
+
+    gdbus introspect --system --dest org.freedesktop.UDisks2 --object-path /org/freedesktop/UDisks2/Manager
```
sairon, line 19: Tiny nit, just mocks a different path:
```suggestion
    gdbus introspect --system --dest org.freedesktop.UDisks2 --object-path /org/freedesktop/UDisks2
```

agners (approved):

supervisor/dbus/udisks2/__init__.py
```diff
@@ -170,4 +230,5 @@ def shutdown(self) -> None:
             block_device.shutdown()
         for drive in self.drives:
             drive.shutdown()
+        self.udisks2_object_manager.shutdown()
```
sairon, line 233: How fast is disconnecting the previous objects from the bus? Can we assume `on_interfaces_added` won't be called before the manager is shut down?

agners: From debug logs when attaching a disk, it looks like all is called in the same Thread and hence event loop, so it *probably* doesn't matter in what order these operations are executed. But logically it makes more sense to shutdown the object manager first, which stops the signal handling.

I've changed the order and tested shutdown, still works without errors.

sairon (approved):

PR #5084, opened 2024-05-16
Create issue for detached addons

mdegat01:
## Proposed change

Create an issue if we detect a detached addon so we can make this into a repair and inform users they aren't receiving updates for these addons.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist


mdegat01:
@agners I choose not to include the repo URL in stored addon data. Detached addons where the repo was removed generally shouldn't happen other then current users hit with bugs in the past (and that won't help them).

Each addon includes a URL linking to where to find information for that particular addon. So I'm going to link to that in the repair and tell people to find more info there. That is available even for existing users in this situation.

supervisor/resolution/checks/detached_addon.py
```diff
@@ -0,0 +1,46 @@
...
+
+    @property
+    def context(self) -> ContextType:
+        """Return a ContextType enum."""
+        return ContextType.ADDON
+
+    @property
+    def states(self) -> list[CoreState]:
+        """Return a list of valid states when this check can run."""
+        return [CoreState.RUNNING, CoreState.SETUP]
```
agners, line 46: Doesn't `RUNNING` makes this a periodical check?

I think once at startup should be enough.

pvizeli, line 46:
```suggestion
        return [CoreState.STARTUP, CoreState.RUNNING]
```
At this point, addons are loaded, SETUP can be to early I feel. I guess we need RUNNING as well detect detach add-ons after store updates

mdegat01: @pvizeli we can do either. Currently the healthcheck during `SETUP` is at the very end after everything else in that step so all addon information is loaded:
https://github.com/home-assistant/supervisor/blob/67ec89c9567016182d77d22968cb9632820f091f/supervisor/core.py#L191

I just removed `RUNNING` from there as per @agners request although I can add it back if we want. Once at startup of supervisor seemed like enough, this should be uncommon other then possibly the source code of a local add-on being removed so I don't think we need to catch this immediately.

pvizeli (approved):
Okay, let's test that

PR #5133, opened 2024-06-12
No update available if update cannot be installed on system

mdegat01:
## Proposed change

If there is an update for the addon but it cannot be installed because the system does not meet the requirements for it, API should not specify that an update is available.

Note: This PR will have an immediate impact on the addons info panel in frontend as it looks to this field to determine when to show UI about the update. It will not affect core currently since core update entities are true any time latest version != current version. They have no current support for the concept that an update is out there but it can't be installed here.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Enhanced update logic to check for the latest version before updating add-ons.

- **Tests**
  - Added a new test to validate update availability based on version compatibility.

### DIFF
--- a/supervisor/addons/addon.py
+++ b/supervisor/addons/addon.py
@@ -285,9 +285,13 @@ class Addon(AddonModel):
     @property
     def need_update(self) -> bool:
         """Return True if an update is available."""
-        if self.is_detached:
+        if self.is_detached or self.version == self.latest_version:
             return False
-        return self.version != self.latest_version
+
+        with suppress(AddonsNotSupportedError):
+            self._validate_availability(self.data_store)
+            return True
+        return False
 
     @property
     def dns(self) -> list[str]:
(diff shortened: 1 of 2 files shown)


mdegat01:
Ok so I just checked. After this PR if there is an update but it cannot be installed on the system then the frontend says nothing about it in the addon info panel. No update panel with the changelog or any info about there being a newer version, just doesn't acknowledge it.

That's functional but it would be ideal if the frontend acknowledged that there was an update and  said its not available on this system. Although it would need more info if we wanted to tell the user why there is no update (home assistant version, platform/machine not supported, etc.)

agners:
> Note: This PR will have an immediate impact on the addons info panel in frontend as it looks to this field to determine when to show UI about the update. It will not affect core currently since core update entities are true any time latest version != current version. They have no current support for the concept that an update is out there but it can't be installed here.

In discussions, we kinda agreed that hiding it completely is not good UX either. Ideally, we'd have a flag to tell the UI that the update is there, but can't be applied. But this is missing for update entities in general.

This wasn't really a user driven effort, and it doesn't seem a very important/urgent issue. So we decided to stay with status quo.

PR #5205, opened 2024-07-24
Restrict stopping core during migrations with force option

mdegat01:
## Proposed change

We should avoid stopping core while it is doing an offline DB migration to prevent corruption. The following APIs now check for an offline DB migration and return an error if one is in progress with a `force` option to do it anyway:
- `/homeassistant/rebuild`
- `/homeassistant/restart`
- `/homeassistant/stop`
- `/host/reboot`
- `/host/shutdown`

In addition `/homeassistant/update` is blocked during an offline DB migration but does not provide a `force` option, users should wait to do that update.

One notable omission here is backup restore. Backup restore will generally restore the DB as well so waiting for the migration to complete first seems wasteful in time and resources. We do have the `exclude_db` option now so there is some risk but seeing as that is an advanced option it seemed ok to leave it. I can add the restriction there too though if we prefer.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to documentation pull request: https://github.com/home-assistant/developers.home-assistant/pull/2260
- Link to cli pull request: https://github.com/home-assistant/cli/pull/502

## Checklist

## Summary by CodeRabbit

- **New Features**
	- Introduced `ATTR_FORCE` to enhance control over shutdown and restart operations.
	- Added new methods to check for ongoing database migrations before executing critical operations.
	- New exception class for managing migration-related service availability issues.

- **Bug Fixes**
	- Improved API error handling by ensuring operations are blocked during database migrations unless forced.

- **Tests**
	- Expanded test cases to validate behavior during data migrations, confirming that critical operations can be blocked or forced appropriately.


supervisor/api/homeassistant.py
```diff
@@ -66,13 +67,31 @@
...
+        vol.Optional(ATTR_FORCE, default=False): vol.Boolean(),
     }
 )
 
 
 class APIHomeAssistant(CoreSysAttributes):
     """Handle RESTful API for Home Assistant functions."""
 
+    async def _check_offline_migration(self, force: bool = False) -> None:
+        """Check and raise if there's an offline DB migration in progress."""
```
agners, line 85: Should we maybe also prevent restart during an online migration? :thinking: 

@emontnemery, opinions?

agners (commented):
Just to be clear, it is generally safe to restart  right? I assume the migration is a database transaction, so it completes or gets rolled back?

But yeah, typically you don't want to interrupt the process unnecessarily, so preventing restart during this time is certainly a good idea :+1:

supervisor/api/homeassistant.py
emontnemery: Yes, we should. Core already rejects the service call for shutting down during online migration.

agners: Let's do this in a follow up PR. The new feature is offline migration, and this prevents at least restart in that case.

agners (approved):

PR #5272, opened 2024-08-26
Add `manual_only` option to addon boot config

mdegat01:
## Proposed change

Some addons like the silabs flasher should never be set to auto-start under any circumstance. Add a `manual_only` option to `boot` which allows addon developers to prevent this when the addon requires it.

As this is the first PR that also requires client library changes to support it, updated the Pull Request template to mention that alongside CLI and API Doc.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to documentation pull request: https://github.com/home-assistant/developers.home-assistant/pull/2286
- Link to cli pull request:
- Link to client library pull request: https://github.com/home-assistant-libs/python-supervisor-client/pull/2

## Checklist

## Summary by CodeRabbit

- **New Features**
	- Introduced a new configuration entity for addon boot management, allowing for enhanced control over boot modes.
	- Added a new property to streamline retrieval of boot configurations.
	- Implemented validation to prevent changes to boot options under specific conditions.
	- Expanded pull request template to improve contributor clarity and checklist comprehensiveness.

- **Bug Fixes**
	- Improved handling of boot configurations to ensure expected behavior during updates.

- **Tests**
	- Enhanced testing framework with new tests to validate boot configuration logic and API responses related to boot mode changes. 

- **Documentation**
	- Updated documentation to reflect changes in boot configuration handling and properties.


supervisor/const.py
```diff
@@ -382,12 +382,27 @@
 ROLE_ALL = [ROLE_DEFAULT, ROLE_HOMEASSISTANT, ROLE_BACKUP, ROLE_MANAGER, ROLE_ADMIN]
 
 
+class AddonBootConfig(StrEnum):
+    """Boot mode config for the add-on."""
+
+    AUTO = "auto"
+    MANUAL = "manual"
+    MANUAL_FORCED = "manual_forced"
```
agners, line 390: I find `manual_forced` rather confusing. The `boot` option says weather an add-on should be started on boot, does manual force mean that it is manually forced to always boot? :thinking: 

I found simply `disabled` better, but maybe that is just a me :sweat_smile: 

Maybe the option `boot` should actually be more verbose, like `start_on_boot`. Then `disabled` definitely would be clearer.

Maybe `disallowed`?

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -38,6 +38,7 @@
 - This PR is related to issue:
 - Link to documentation pull request:
 - Link to cli pull request:
+- Link to client library pull request:
```
agners, line 41: Yeah this makes sense, and ideally we also add a pull request template to the client library, which links back.

This is not really related to this PR, so it should be a separate.

supervisor/const.py
```diff
@@ -382,12 +382,27 @@
...
     """Boot mode for the add-on."""
 
     AUTO = "auto"
     MANUAL = "manual"
 
+    @classmethod
+    def _missing_(cls, value: str) -> Self | None:
+        """Convert 'forced' config values to their counterpart."""
+        if value == AddonBootConfig.MANUAL_FORCED:
+            return AddonBoot.MANUAL
```
agners, line 403: Why is this needed exactly? The config will use `AddonBootConfig` which know the new value, and the API we'll not change no? :thinking:

agners (changes requested):

supervisor/const.py
mdegat01: Because the `boot` option is currently set to fallback on the config value if the user has not overwritten it by changing the setting. And the config value now can have a new option which does not make sense for `boot`. Because an addon is either booted automatically at startup or manually on demand, this third option is not really a new boot option it simply prevents user changes to the boot option.

So my two options were to add an if statement here to convert `manual_forced` into `manual`:
https://github.com/home-assistant/supervisor/blob/24050f594cc09ac8cc23fb05910da151166aa732/supervisor/addons/model.py#L159-L161

Or add the if statement there in the enum itself. The latter option seemed cleaner to me, especially if we do decide to add an `auto_forced` in the future.

supervisor/const.py
agners: Let's go with `manual_only` as discussed in our meeting.

agners (approved):

PR #5276, opened 2024-08-27
Persist currently selected slot on successful boot

sairon:
## Proposed change

When boot slot is selected manually in GRUB, the system boots into this slot and marks it as good. However, the boot order is not changed, so in the next boot (after an explicit or unexpected reboot) HAOS returns to the version in the other slot. This might be confusing because if the system has been running for some time, the user can forget they have changed the boot slot to fix issue they had.

This gets more confusing if the "other" boot slot is selected manually three times in a row. Let's say we have ORDER="A B". This means that every time GRUB starts, it wants to boot slot A. If the slot B is selected instead, only A_TRY is incremented, system boots into slot B and marks slot B as good (B_OK=1, B_TRY=0). On another boot, this repeats, yet A_TRY is incremented again. Until it reaches 3, the slot A would be always chosen automatically, only after that it would boot to slot B, presuming slot A is dead. The ORDER variable will be still unchanged though.

This commit only makes sure that when the system is marked as healthy, the slot is both marked as good AND active, updating the ORDER variable as well. Because the X_TRY counter is incremented by GRUB, if we want the other slot not to be marked as bad, we need to adjust the logic in OS's grub.cfg as well, because Supervisor can't know whether it's apppropriate to change other slot's state or not.

I also took the courtesy to adjust the logging a bit, to include the stack trace in the error log if marking the slot fails somehow.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

## Summary by CodeRabbit

- **New Features**
	- Enhanced the `mark_healthy` functionality to allow marking multiple states for the booted partition, improving operational flexibility.
	
- **Bug Fixes**
	- Improved error handling by using more appropriate logging for exceptions, enhancing debugging capabilities.

- **Improvements**
	- Expanded logging for successful operations to include detailed feedback on state changes, providing better visibility into the system's status.

### DIFF
--- a/supervisor/os/manager.py
+++ b/supervisor/os/manager.py
@@ -366,11 +366,19 @@ class OSManager(CoreSysAttributes):
     async def mark_healthy(self) -> None:
         """Set booted partition as good for rauc."""
         try:
-            response = await self.sys_dbus.rauc.mark(RaucState.GOOD, "booted")
+            responses = [
+                await self.sys_dbus.rauc.mark(RaucState.ACTIVE, "booted"),
+                await self.sys_dbus.rauc.mark(RaucState.GOOD, "booted"),
+            ]
         except DBusError:
-            _LOGGER.error("Can't mark booted partition as healthy!")
+            _LOGGER.exception("Can't mark booted partition as healthy!")
         else:
-            _LOGGER.info("Rauc: %s - %s", self.sys_dbus.rauc.boot_slot, response[1])
+            _LOGGER.info(
+                "Rauc: slot %s - %s, %s",
+                self.sys_dbus.rauc.boot_slot,
+                responses[0][1],
+                responses[1][1],
+            )
             await self.reload()
 
     @Job(


supervisor/os/manager.py
```diff
@@ -366,11 +366,19 @@ async def update(self, version: AwesomeVersion | None = None) -> None:
     async def mark_healthy(self) -> None:
         """Set booted partition as good for rauc."""
         try:
-            response = await self.sys_dbus.rauc.mark(RaucState.GOOD, "booted")
+            responses = [
+                await self.sys_dbus.rauc.mark(RaucState.ACTIVE, "booted"),
+                await self.sys_dbus.rauc.mark(RaucState.GOOD, "booted"),
+            ]
```
mdegat01, line 372: If order of operations doesn't matter we should use `asyncio.gather`:
```py
responses = await asyncio.gather(
    self.sys_dbus.rauc.mark(RaucState.ACTIVE, "booted"),
    self.sys_dbus.rauc.mark(RaucState.GOOD, "booted"),
)
```
However does order of operations matter? If we had a situation where the first succeeded and the other failed is there one we would prefer to be the one that succeeded? Also would we want to specifically log which failed? Because we aren't doing that right now.

mdegat01 (changes requested):

supervisor/os/manager.py
sairon: I had the same thought and my conclusion is that the order shouldn't matter - as it's not possible to do it atomically, marking it as active first sounds a bit more logical (so you don't end up with "good" but inactive slot) but in the end I don't expect only one of the operation would fail, unless there was some race condition. However, if we do as you suggest and one of the operations really fails for some reason, the other one still might be performed. Wouldn't it be better to do it sequentially as in the original code to get at least some consistency?

About logging, just in case one of the operation fails, the cause should be traceable from the logged exception, shouldn't it?

mdegat01: Yea it sounds like there's one we would prefer be the one that succeeded in the case of a failure on the second so I agree then, I would leave the code as is and do it sequentially.

> About logging, just in case one of the operation fails, the cause should be traceable from the logged exception, shouldn't it?

Ah right, its `_LOGGER.exception` now instead of `error`. Ok makes sense then.

agners (approved):
LGTM!

PR #5283, opened 2024-08-30
Make network API replace IP/WiFi settings

### LINKED ISSUE #4442: Supervisord crashes when a second network interface has no gateway
### The problem

The problem is rather simple: Supervisord assumes that every network interface will have a set of an IP+CIDR and a gateway address. This is a flawed assumption, as with multiple network cards realistically only the primary one should have a default route gateway.

Attempting to save a config like following, with the `enp6s18` being set to DHCP:
image

Results in an error:
Screenshot 2023-02-05 at 11 34 46

### What version of Home Assistant Core has the issue?

2023.2.2

### What was the last working version of Home Assistant Core?

N/A

Home Assistant OS

### Integration causing the issue

Supervisord itself

### Link to integration documentation on our website

_No response_

### Diagnostics information

_No response_

### Example YAML snippet

_No response_

### Anything in the logs that might be useful for us?

I isolated the problem to this exception, produced in the `supervisord` container:

image

### Additional information

Putting a gateway on that second card seems to solve the issue. However, I'm not sure if it will not cause problems down the line as generally I don't want a gateway on that card.

agners:
## Proposed change
This changes the network API behavior to replace IP configuration and WiFi settings instead of updating them.

It also adds support to set DNS (via `nameservers` property) when method `auto` is used. This allows to set a DNS server in case DHCP does not provide one, or overrides the 

Update by replacing the whole IP configuration has advantages:
- It allows to specify no gateway if not required (fixes #4442)
- It allows to clear particular aspect of the IP config by omitting the property, which feels rather natural

This also works out quite well with how the frontend behaves, it essentially keeps the behavior as it is today:
- Setting to "auto" only provides the `method` property. So this clears lingering `nameservers`.
- Setting to "static" always sets all fields (`method`, `addresses`, `gateway` and `nameservers`), so no change in behavior.

However, this does change the CLI behavior: Missing options now really lead to clearing that particular field. However, this seems the more expected behavior, e.g. the following sequence:

```
ha network update enp1s0 --ipv4-method static --ipv4-address 192.168.87.123/24 --ipv4-nameserver 192.168.87.2
ha network update enp1s0 --ipv4-method static --ipv4-address 192.168.87.123/24
```

With the patch style API, this would not have cleared the DNS. Clearing was only possible by switching to method `auto` and then back to `static`.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist


agners:
An alternative to this would be to allow `null` for `gateway` (and potentially also `nameservers` as an alternative to an empty list) to clear these settings. This essentially would keep the patch style API.

However, the support for `nameservers` in `auto` mode would be problematic with the current frontend: The frontend code only sets the `method` property. Hence when switching from `static` the `nameservers` would get inherited. This might not what the user intended.

Also the CLI would need a way to set no nameserver (currently that is not possible since the `--ipvX-nameserver` argument is additive only). The CLI I don't see that critical, we can implement the replace type behavior for the CLI client side (which probably would make sense).

Another option here would be to make the complete API replace style. For the frontend this would work without a problem as all aspects (`ipv4`, `ipv6`, `wifi` and `enabled` are set always.

supervisor/api/network.py
```diff
@@ -209,20 +209,17 @@ async def interface_update(self, request: web.Request) -> None:
         for key, config in body.items():
             if key == ATTR_IPV4:
                 interface.ipv4setting = replace(
-                    interface.ipv4setting
-                    or IpSetting(InterfaceMethod.STATIC, [], None, []),
+                    IpSetting(InterfaceMethod.STATIC, [], None, []),
                     **config,
                 )
```
mdegat01, line 214:
```suggestion
                interface.ipv4setting = IpSetting(**config)
```
There's no reason to do the `replace`as we aren't patch updating an existing object anymore, we're making a new object and using it. We should update the schema to includle `InterfaceMethod.STATIC, [], None, []` as the default values of their respective fields instead so we get those when the client doesn't provide their own.

Alternatively we can add a `from_dict` class method to those dataclasses and use that here, passing it `config`.

supervisor/api/network.py
```diff
@@ -209,20 +209,17 @@ async def interface_update(self, request: web.Request) -> None:
...
             elif key == ATTR_WIFI:
                 interface.wifi = replace(
-                    interface.wifi
-                    or WifiConfig(
+                    WifiConfig(
                         WifiMode.INFRASTRUCTURE, "", AuthMethod.OPEN, None, None
                     ),
                     **config,
```
mdegat01, line 225: I forget, do we have UI around this or is purely set via CLI? If we have UI then I'd just want to check that all the fields you can set are actually exposed in the UI before okaying this. Unfortunately I'm realizing I don't actually have a wifi-based test device at the moment, I'll have to look into that

mdegat01 (changes requested):
I think this makes sense. Minor style change to the existing code.

One thing I was thinking about though, what if we mapped the existing code to `PATCH /network/interface/{interface}/update`? Like update `POST /network/interface/{interface}/update` exactly as you're showing but if we receive a `PATCH` request to that endpoint it does what it does today?

Then what we can do is simply update the CLI to make a `PATCH` call instead of a `POST` call to keep it working as it does today. And if we want we can add a `--replace` parameter or something that makes it do the `POST` call and clear existing values.

That way we don't have to try and collect the existing values and add them to the postbody to maintain existing behavior in the CLI, Supervisor will handle that. The frontend will continue to use `POST` and behave the way we want it to.

supervisor/api/network.py
agners: We use also different standards in other places:

https://github.com/home-assistant/supervisor/blob/05e0c7c3ab949ec26c98f5dc98c4d8cf52ee3c0c/supervisor/host/configuration.py#L116

I'd prefer to keep this explicitly for the API.

agners: Ah I see you meant to add it to the schema validation. I've implemented that just now. It required vol.Maybe(), but in my mind this is a correct use-case for null values.

PR #5287, opened 2024-09-03
Add "lines" and "verbose" query parameters for advanced logs

sairon:
## Proposed change

Since headers are clumsy considering the Core proxy between the frontend and Supervisor, add a way to adjust number of lines and verbose log format using query parameters as well. If both query parameters and headers are supplied, prefer the former, as it's more prominent when reading through the request logs.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

- [ ] Documentation added/updated for [developers.home-assistant.io][docs-repository] TBD

## Summary by CodeRabbit

- **New Features**
	- Enhanced logging functionality with support for "verbose" and "lines" query parameters, allowing users to customize log retrieval.
  
- **Tests**
	- Introduced a new test to validate the advanced logging API's behavior with query parameters, ensuring reliable functionality.

### DIFF
--- a/supervisor/api/host.py
+++ b/supervisor/api/host.py
@@ -222,10 +222,14 @@ class APIHost(CoreSysAttributes):
                 "supported for now."
             )
 
-        if request.headers[ACCEPT] == CONTENT_TYPE_X_LOG:
+        if "verbose" in request.query or request.headers[ACCEPT] == CONTENT_TYPE_X_LOG:
             log_formatter = LogFormatter.VERBOSE
 
-        if RANGE in request.headers:
+        if "lines" in request.query:
+            lines = request.query.get("lines", DEFAULT_RANGE)
+            # entries=cursor[[:num_skip]:num_entries]
+            range_header = f"entries=:-{lines}:"
+        elif RANGE in request.headers:
             range_header = request.headers.get(RANGE)
         else:
             range_header = f"entries=:-{DEFAULT_RANGE}:"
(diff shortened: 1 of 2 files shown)


supervisor/api/host.py
```diff
@@ -222,10 +222,14 @@ async def advanced_logs_handler(
                 "supported for now."
             )
 
-        if request.headers[ACCEPT] == CONTENT_TYPE_X_LOG:
+        if "verbose" in request.query or request.headers[ACCEPT] == CONTENT_TYPE_X_LOG:
```
agners, line 225: Hm, is `verbose` the right word here? :thinking: 

These are the same logs, just with more metadata no? Maybe we should have named the LogFormatter already different? :thinking: 

From what I can tell we add timestamps currently, maybe just `timestamp` or `timestamped`? :thinking:

sairon: I named the fields same way as we call them in the CLI. Also, [it's not just timestamp](https://github.com/home-assistant/supervisor/blob/6c1203e4bfcdd49e87398088577fc99850fa01b0/supervisor/utils/systemd_journal.py#L37C12-L59), so that would be a wrong name for the format - it's also host name, journal identifier (which - currently - only varies in host logs) and PID of the process.

agners: Hm, ok in that case let's leave as is :+1:

agners (approved):

PR #5438, opened 2024-11-20
Add support for cloud backups in Core

mdegat01:
## Proposed change

This is an initial PR on what cloud backup support would look like in Supervisor after discussion with @emontnemery . Noted the backup and restore processes and the significant changes below for review.

The way a cloud backup will work from Supervisor's perspective is as follows:

1. Core tell Supervisor to make a backup
2. If the backup will permanently reside in `/backup` or a backup mount, that will be the `location`. If not (it is only a cloud backup) then core will use a new `location` called `.cloud_backup`
3. Supervisor makes the backup at the specified location
4. Supervisor notifies core whenever a new backup is done with a WS Message that includes the slug, job ID and path to the backup file that Core understands
5. Core grabs file and uploads it to backup sync agents as necessary. This may include copying it to other backup mounts via the `hassio` backup sync agent in core
6. When done, if the file is in `/cloud_backup` then core deletes it. Otherwise it leaves it

The way a cloud restore will work from Supervisor's perspective is as follows:

1. Core downloads the file into `/cloud_backup`
2. Core calls `/backups/reload` providing the location `.cloud_backup` and the filename so Supervisor picks it up
3. Core tells Supervisor to restore this new backup by slug
4. When done, Supervisor expects core to delete it or move it as necessary

Significant changes to make this function:

1. `/backup` is now mapped into core's container read-write
2. All backup mounts are mapped as subfolders of `/backup` like share and media mounts. This will also affect all addons which map `backup`
3. A new folder at `/mnt/data/supervisor/core/backup` is mapped to Core's container at `/cloud_backup`. This is the `.cloud_backup` location
4. Supervisor searches the `.cloud_backup` location for backups on reload. However this location is hidden from the info APIs in `/backups` as only Core can see it and Supervisor expects backups in it to be transient. Additionally API requests which seek to make backups in this location or interact with backups in this location will be rejected with a 403 Forbidden if not from Core

Smaller detail - backups may exist in multiple locations Supervisor sees. This has always been true. Prior to this PR if we found two backups with the same slug during reload the last one was all we remembered. Now the first one is remembered and we keep track of all the locations we found it to return that info from the API.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:
    - https://github.com/home-assistant-libs/python-supervisor-client/pull/31 

## Checklist

## Release Notes

- **New Features**
  - Introduced new backup location management with enhanced validation for cloud backups.
  - Added constants for various paths, improving clarity in configuration.
  - New properties for core backup paths in the configuration settings.

- **Bug Fixes**
  - Corrected constant naming errors to ensure consistency across the application.
  - Enhanced error handling for backup operations and mount management.

- **Tests**
  - Expanded test coverage for backup functionalities, including reload and access control scenarios.
  - Updated tests to reflect new response structures and error handling improvements.

- **Documentation**
  - Updated documentation to reflect new properties and constants related to backup management.


mdegat01:
There is a lot of test case work to do still that I'm working on. I put this up so people could provide feedback on the approach as its a good amount of change to support this initiative.

supervisor/backups/backup.py
```diff
@@ -178,12 +182,44 @@ def docker(self, value: dict[str, Any]) -> None:
...
+        return self.locations[0]
+
+    @property
+    def all_locations(self) -> set[str | None]:
+        """Return all locations this backup was found in."""
+        return self._locations.keys()
+
+    @property
+    def locations(self) -> list[str | None]:
+        """Return locations this backup was found in except cloud backup (unless that's the only one)."""
```
agners, line 197: Seems a bit weird to me that we only show the cloud location if it is the only one, isn't that a bit inconsistent?

supervisor/docker/homeassistant.py
```diff
@@ -109,23 +115,36 @@ def mounts(self) -> list[Mount]:
...
                         source=self.sys_config.path_extern_media.as_posix(),
-                        target="/media",
+                        target=PATH_MEDIA.as_posix(),
                         read_only=False,
                         propagation=PropagationMode.RSLAVE.value,
                     ),
+                    Mount(
+                        type=MountType.BIND,
+                        source=self.sys_config.path_extern_backup.as_posix(),
+                        target=PATH_BACKUP.as_posix(),
```
agners, line 138: Why is the bind mount for regular backup needed? And do we handle backups on external network storage as well?

supervisor/mounts/manager.py
```diff
@@ -141,6 +141,15 @@ async def load(self) -> None:
                 ]
             )
 
+        # Bind all backup mounts to directories in backup
```
agners, line 144: Ah I see. I am a bit worried, so now we need to maintain another location where network storage are bind mounted :cold_sweat: 

Hm, why didn't we do that in first place?

agners:
> 1. Core tell Supervisor to make a backup
> ..
> 4\. Supervisor notifies core whenever a new backup is done with a WS Message that includes the slug, job ID and path to the backup file that Core understands

Why is the WS message even needed? Since Core triggers the backup, can't it simply wait for the command to complete and then take the backup? :thinking:

supervisor/backups/backup.py
mdegat01: I just didn't want this list to be empty ever. If a backup exists exclusively in `.cloud_backup` no one will be aware of it except core. They're the only ones who will know its slug to ask more info about it and figured it was fine to put that location in the list in this case. In all other cases I would prefer not to show `.cloud_backup` as it will confuse all other clients who have no idea what that location is.

supervisor/docker/homeassistant.py
mdegat01: Because core will only use the location `.cloud_backup` for cloud-only backups. If the user wants it stored locally or in a network mount core will use that location instead. And we'll provide them with a path to the file within `/backup` to use for further processing.

I want it like this because of network mounts. I want to avoid writing backups to the local disk any time we can. So if the user wants a backup stored on their backup mount and google drive only then Supervisor will write it directly to the backup mount and core will read it out of there to upload it to google drive. No unnecessary writes on the local disk.

supervisor/mounts/manager.py
mdegat01: Mostly just because we didn't need to. There wasn't any reason to expose backup mounts to others because supervisor was doing all the operations on them. I'm not exactly sure how addons like Google Drive sync, etc. will handle it though, it depends on how they process the `/backup` folder.

The only other options I could think of was bind mounting the entire `/mounts` folder to supervisor into core's container. This would work but it seemed worse for a few reasons:

1. All the share and media mounts are available in multiple places in core's container now
2. A user could now muck with this container in an automation/script and really mess up supervisor (like if they deleted the folder)
3. I don't know if we'll have other types of mounts in the future we want to keep more hidden

Exposing addons to backup network mounts felt less risky. I couldn't think of any reason we didn't do it to begin with other then "we didn't have to".

supervisor/backups/backup.py
agners: > I just didn't want this list to be empty ever.  [...] In all other cases I would prefer not to show .cloud_backup as it will confuse all other clients who have no idea what that location is.

Hm, ok other clients might not know about cloud backup, but maybe they should, so nothing funky happens accidentally (e.g. if something alters the backup file? Then maybe it is better if it has the chance to know?)

I mean, I get your arguments. But I think rather than adding this complexity, and potentially hiding something from other users which might be(come) important, why not just leave it here. If it turns out to be a problem, we can remove it then. :man_shrugging:

supervisor/mounts/manager.py
agners: > I'm not exactly sure how addons like Google Drive sync, etc. will handle it though, it depends on how they process the `/backup` folder.

Hm, I see, that could actually have potential implications, e.g. if they browse often etc. it will lead to network traffic etc.

Instead of going to a separate mount, could we optionally mount without mount propagation maybe? :thinking:

agners (approved):
Generally LGTM, we can still tweak things as need until release :+1:

supervisor/mounts/manager.py
mdegat01: Just for anyone else interested - tried this, did not work as we hoped. When propagation is set to `rprivate` the addons can still see everything in the network mounts. Its just changes to backup mounts require a restart of the addon. So it really just makes it inconsistent, doesn't help.

PR #5447, opened 2024-11-22
Trigger HA USB rescan on device removal

puddly:
## Proposed change

Currently, `usb/scan` is called only when new hardware is added. I am interested in Core having the ability to detect when hardware is removed (https://github.com/home-assistant/core/pull/131224). To do this, Core needs to be notified when Supervisor detects a USB device removal.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

## Summary by CodeRabbit

- **New Features**
	- Enhanced hardware event handling by adding support for device removal events in the Home Assistant module.

### DIFF
--- a/supervisor/homeassistant/module.py
+++ b/supervisor/homeassistant/module.py
@@ -305,6 +305,9 @@ class HomeAssistant(FileConfiguration, CoreSysAttributes):
 
         # Register for events
         self.sys_bus.register_event(BusEvent.HARDWARE_NEW_DEVICE, self._hardware_events)
+        self.sys_bus.register_event(
+            BusEvent.HARDWARE_REMOVE_DEVICE, self._hardware_events
+        )
 
     def write_pulse(self):
         """Write asound config to file and return True on success."""


mdegat01 (approved):
Seems reasonable to me. I notice this is in draft but it seems ready? Is there something else you want to add @puddly ?

mdegat01:
I guess one thing that does come to mind for me, is a usb scan expensive? We don't technically have to do that right? The alternative here is to tell core exactly which device is removed and have core remove that from its cache instead of doing a full scan and rebuilding the cache from scratch. Unless there's other info that needs to be collected which Supervisor doesn't have?

puddly:
The Core side confusingly named, this really only scans serial ports and does so via `from serial.tools.list_ports import comports`, which just globs some paths in `/dev/` and reads some files in `/sys`: https://github.com/pyserial/pyserial/blob/7aeea35429d15f3eefed10bbb659674638903e3a/serial/tools/list_ports_linux.py#L91-L106.

On a Yellow with a USB stick plugged in:

```python
In [1]: from serial.tools.list_ports import comports

In [2]: %timeit comports()
6.09 ms ± 98.5 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

puddly:
I've tested this out with Core and it works just fine. Unfortunately, Core has a *60 second* cooldown on serial port rescans so we'll need some tweaking on the Core side to get this to work.

Going forward, do you think a dedicated API to forward added and removed USB and serial devices would be better?

agners (approved):

PR #5521, opened 2025-01-02
Check if backup is still available when accessing

agners:
## Proposed change
When accessing a backup (e.g. to restore, download or delete it) check if the locations of the backup file are still available. If not, force a backup reload instead.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/api/backups.py
+++ b/supervisor/api/backups.py
@@ -143,11 +143,26 @@ SCHEMA_REMOVE = vol.Schema(
 class APIBackups(CoreSysAttributes):
     """Handle RESTful API for backups functions."""
 
-    def _extract_slug(self, request):
+    async def _extract_slug(self, request):
         """Return backup, throw an exception if it doesn't exist."""
         backup = self.sys_backups.get(request.match_info.get("slug"))
         if not backup:
             raise APINotFound("Backup does not exist")
+
+        # Reload in case locations of this backup are no longer in sync
+        # This usually means the user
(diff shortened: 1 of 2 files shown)


supervisor/api/backups.py
```diff
@@ -143,11 +143,26 @@ def _ensure_list(item: Any) -> list:
 class APIBackups(CoreSysAttributes):
     """Handle RESTful API for backups functions."""
 
-    def _extract_slug(self, request):
+    async def _extract_slug(self, request):
```
emontnemery, line 146: I think this method should be renamed since it does (and already did before this PR) more than just get the slug from the request

emontnemery (approved):
Looks good to me, except for the method name. That could be changed in a follow-up though.

supervisor/api/backups.py
mdegat01: Tbh I would actually prefer we don't rename it. Or at the very least, please keep `_extract` at the beginning. We can change it to `_extract_backup_by_slug` if we prefer. But if you look at all the other files under `api`, every one that takes an id of something in the resource path has a matching `_extract_<x>` method that extracts that id and returns the resource or raises. I'd like to keep that consistency to make these easy to find.

agners:
This global refresh is not ideal in any case. We've decided to handle missing backups on Core side, and on Supervisor side just update the cache for the backup in question specifically.

PR #5548, opened 2025-01-14
Add env on core restart due to restore

mdegat01:
## Proposed change

Add `SUPERVISOR_RESTORE_JOB_ID` env to home assistant container when it restarts due to a restore. This will be there even if the restore failed as long as Home Assistant was actually stopped as part of the restore.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Added ability to remove Home Assistant container during backup restoration and shutdown processes.
  - Introduced environment variable tracking for restore job IDs.

- **Bug Fixes**
  - Refined Home Assistant restoration and container management logic.
  - Updated job stage handling during backup restoration.

- **Refactor**
  - Modified job identifier management from UUID to string representation.
  - Enhanced container lifecycle control mechanisms.

- **Tests**
  - Added new test coverage for Home Assistant restore process and environment variable handling.
  - Streamlined assertions related to the restoration process in existing tests.

### DIFF
--- a/supervisor/homeassistant/core.py
+++ b/supervisor/homeassistant/core.py
@@ -316,6 +316,18 @@ class HomeAssistantCore(JobGroup):
             self.sys_resolution.create_issue(IssueType.UPDATE_FAILED, ContextType.CORE)
             raise HomeAssistantUpdateError()
 
+    def _is_restore(self) -> str | None:
+        """Return id of current restore job if a restore job is in progress."""
+        job = self.sys_jobs.current
+        while job.parent_id:
+            job = self.sys_jobs.get_job(job.parent_id)
+            if job.name in {
+                "backup_manager_full_restore",
+                "backup_manager_partial_restore",
+            }:
+                return job.uuid
+        return None
+
     @Job(
         name="home_assistant_core_start",
         limit=JobExecutionLimit.GROUP_ONCE,
@@ -345,7 +357,7 @@ class HomeAssistantCore(JobGroup):
             self.sys_homeassistant.write_pulse()
 
             try:
-                await self.instance.run()
+                await self.instance.run(restore_job_id=self._is_restore())
             except DockerError as err:
                 raise HomeAssistantError() from err
 
@@ -356,10 +368,10 @@ class HomeAssistantCore(JobGroup):
         limit=JobExecutionLimit.GROUP_ONCE,
         on_condition=HomeAssistantJobError,
     )
-    async def stop(self) -> None:
+    async def stop(self, *, remove_container: bool = False) -> None:
         """Stop Home Assistant Docker."""
         try:
-            return await self.instance.stop(remove_container=False)
+            return await self.instance.stop(remove_container=remove_container)
         except DockerError as err:
             raise HomeAssistantError() from err
 
--- a/supervisor/homeassistant/core.py
+++ b/supervisor/homeassistant/core.py
@@ -316,18 +316,6 @@ class HomeAssistantCore(JobGroup):
             self.sys_resolution.create_issue(IssueType.UPDATE_FAILED, ContextType.CORE)
             raise HomeAssistantUpdateError()
 
-    def _is_restore(self) -> str | None:
-        """Return id of current restore job if a restore job is in progress."""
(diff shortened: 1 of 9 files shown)


supervisor/homeassistant/core.py
```diff
@@ -345,7 +357,7 @@ async def start(self) -> None:
             self.sys_homeassistant.write_pulse()
 
             try:
-                await self.instance.run()
+                await self.instance.run(restore_job_id=self._is_restore())
```
agners, line 360: Could we maybe move this further up in the call chain, e.g. pass it through from `update()` and `start()`, and actually determine the job id in `supervisor/backups/backup.py:restore_homeassistant()`? This would isolate this backup related logic a bit better, with the downside of some additional argument. But IMHO, this would be nicer overall.

mdegat01: > Could we maybe move this further up in the call chain, e.g. pass it through from `update()` and `start()`, and actually determine the job id in `supervisor/backups/backup.py:restore_homeassistant()`?

Actually, this code path does not start the home assistant container during a restore. Its confusing I know because it includes calls to `start`. But if you notice at the top of `update` it checks if the container is running and that will always be false during a restore, because the first thing `restore_homeassistant` does is stop the container before calling that, right here: https://github.com/home-assistant/supervisor/blob/60c9b81dabc02d71b8bab1da3bd41f4321a6cb8b/supervisor/backups/backup.py#L775

So the place to add that parameter would be here:
https://github.com/home-assistant/supervisor/blob/60c9b81dabc02d71b8bab1da3bd41f4321a6cb8b/supervisor/backups/manager.py#L646-L651

I considered that. The reason I didn't is because of this:
1. I felt like I should also add it to `update` and `start` as you described "just in case". I am not totally sure that there aren't edge cases that cause Home Assistant to get started during update and not wait until the end and I would always want that restore ID there
2. When I was writing it I thought `_block_till_run` included rollback-on-crash logic and I would want to add it there too (https://github.com/home-assistant/supervisor/blob/60c9b81dabc02d71b8bab1da3bd41f4321a6cb8b/supervisor/homeassistant/core.py#L482-L526). I see now that it doesn't so this reason is invalid
3. IF the API responsivity check [here](https://github.com/home-assistant/supervisor/blob/60c9b81dabc02d71b8bab1da3bd41f4321a6cb8b/supervisor/backups/manager.py#L653-L658) fails perhaps we should do a rollback. And then we would want it included there

In the end I felt it was cleaner if instead of tracking down where all core can be started during a restore I made it so any time core was started during a restore the env was there. Plus jobs in general are mostly outside the execution logic in context vars so handling it like this felt fitting with how jobs work, picking up the job from context rather then an argument.

agners: > Actually, this code path does not start the home assistant container during a restore. Its confusing I know because it includes calls to `start`. But if you notice at the top of `update` it checks if the container is running and that will always be false during a restore, because the first thing `restore_homeassistant` does is stop the container before calling that, right here:

Ok I see.

> So the place to add that parameter would be here:
> 
> https://github.com/home-assistant/supervisor/blob/60c9b81dabc02d71b8bab1da3bd41f4321a6cb8b/supervisor/backups/manager.py#L646-L651
> [...]
> In the end I felt it was cleaner if instead of tracking down where all core can be started during a restore I made it so any time core was started during a restore the env was there.

In the end, this is a bit question of style. Personally I prefer such things to be explicit, even if it means that we miss a particular call chain today. It makes the code flow more obvious and easier to track down issues later on, IMHO. But I see, depending on how retries/rollback would work, this can actually become quite a chain of calls, which then might not be all that nice.

Anyhow, I am fine using job context to figure out if this is a Core start related to restore.

However, I'd suggest to move `_is_restore` into `supervisor/backups/manager.py`. It does add circular calls, but we already call into manager.py from core.py for the backup on update :shrug: 

Or at the very least make the job names `backup_manager_full_restore` and `backup_manager_partial_restore` a constant which we use in manager.py and core.py, that way the coupling is a bit more explicit.

agners (approved):

PR #5569, opened 2025-01-22
Backup protected status can vary per location

mdegat01:
## Proposed change

Backups can now be easily encrypted and decrypted while streaming and core does this regularly. Therefore two backups that are otherwise identical except one is encrypted and one is not need to be considered the same. And Supervisor should tell clients the protected status of backups across locations.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

## Summary by CodeRabbit

- **New Features**
	- Enhanced backup location management with detailed protection status.
	- Added ability to track protected backup locations.

- **Improvements**
	- Improved backup information retrieval.
	- More robust handling of backup file paths and encryption status.
	- Centralized password validation for backup locations.

- **Bug Fixes**
	- Updated backup location data structure for more accurate tracking.

- **Tests**
	- Added new test cases for backup encryption and protection scenarios.
	- Enhanced existing tests to validate new backup location structure.

These changes provide a more comprehensive approach to managing backup locations, ensuring detailed information about backup protection and file paths.


agners:
The `tests/backups/test_manager.py::test_backup_remove_error` test failure actually uncovers a bug in that test in general: The backup hasn't been in the `test` location at all, this test should have failed even before this change :see_no_evil: .

I've updated the test to actually copy the backup to that mount location as well, and fixed the test accordingly.

supervisor/backups/backup.py
```diff
@@ -252,7 +258,27 @@ def data(self) -> dict[str, Any]:
 
     def __eq__(self, other: Any) -> bool:
         """Return true if backups have same metadata."""
-        return isinstance(other, Backup) and self._data == other._data
+        if not isinstance(other, Backup):
+            return False
+
+        # Compare all fields except ones about protection. Current encryption status does not affect equality
+        keys = self._data.keys() | other._data.keys()
+        for k in keys - {ATTR_PROTECTED, ATTR_CRYPTO}:
```
agners, line 266: Tests with an unencrypted/encrypted backup also made the `registries` setting to light up:

```
2025-01-23 11:56:19.845 DEBUG (MainThread) [supervisor.backups.backup] Backup 5ba8f1d7 and 5ba8f1d7 not equal because docker field has different value: {'registries': {'hub.docker.com': {'username': 'agners', 'password': '/aDMOAXyCd5fG1ZR2j4nHA=='}}} and {'registries': {'hub.docker.com': {'username': 'agners', 'password': 'gaga'}}}
2025-01-23 11:56:19.845 ERROR (MainThread) [supervisor.backups.manager] Ignoring backup 5ba8f1d7 in backup_test due to: Backup in backup_test and None both have slug 5ba8f1d7 but are not the same!
```

Technically we could dive into the field and only ignore the password (since this is the only part which is encrypted/unencrypted). However, the idea is to move the settings into a Supervisor specific tarball anyways, so I'd say we just ignore the whole attribute for now:

```suggestion
        for k in keys - {ATTR_PROTECTED, ATTR_CRYPTO, ATTR_DOCKER}:
```

agners (changes requested):

agners (approved):

supervisor/backups/backup.py
```diff
@@ -263,7 +263,7 @@ def __eq__(self, other: Any) -> bool:
 
         # Compare all fields except ones about protection. Current encryption status does not affect equality
         keys = self._data.keys() | other._data.keys()
-        for k in keys - {ATTR_PROTECTED, ATTR_CRYPTO}:
+        for k in keys - {ATTR_PROTECTED, ATTR_CRYPTO, ATTR_DOCKER}:
```
emontnemery, line 266: Do we have tests to ensure these attributes are ignored?

tests/backups/test_manager.py
```diff
@@ -1734,12 +1734,20 @@ async def test_backup_remove_error(
     healthy_expected: bool,
 ):
     """Test removing a backup error."""
```
emontnemery, line 1736: Why did this test start failing?

tests/backups/test_manager.py
```diff
@@ -1734,12 +1734,20 @@ async def test_backup_remove_error(
...
-    copy(get_fixture_path("backup_example.tar"), coresys.config.path_backup)
-    await coresys.backups.reload(location=None, filename="backup_example.tar")
+    backup_file = get_fixture_path("backup_example.tar")
+    if location_name is None:
+        copy(backup_file, coresys.config.path_backup)
+        location = None
+    else:
+        (mount_dir := coresys.config.path_mounts / location_name).mkdir()
+        copy(backup_file, mount_dir)
+        location = coresys.mounts.get(location_name)
```
emontnemery, line 1744: Instead of the if-statement, consider adding additional parametrization variable, for the subdirectory and maybe the key to get the location from `coresys.mounts.get`

supervisor/backups/backup.py
mdegat01: There's a test to confirm an encrypted backup and unencrypted backup can be consolidated, which means those fields were ignored. I didn't have `docker` in there before so my backup fixtures didn't have that, I'll have to check on that.

Technically for docker we should only be ignoring it if one backup is encrypted and one isn't, otherwise it should be the same. I don't know if we want to deal with that for a field to be deprecated though

mdegat01:
> The `tests/backups/test_manager.py::test_backup_remove_error` test failure actually uncovers a bug in that test in general: The backup hasn't been in the `test` location at all, this test should have failed even before this change 🙈 .
> 
> I've updated the test to actually copy the backup to that mount location as well, and fixed the test accordingly.

Wait what was the failure? I didn't think the file not being in that location would cause a failure, I'm pretty sure other tests do that too?

supervisor/backups/backup.py
agners: > I don't know if we want to deal with that for a field to be deprecated though

Yeah let's not bother about that.

But adding `docker` attributes and make sure they get ignored in the test make sense :+1: .

agners: Ok, I've updated the test fixture files to include a docker repository setting. The test fails without the change above, but with the change merged succeeds :+1: .

tests/backups/test_manager.py
agners: Essentially, because one of the backup didn't exist before. But previously the test just "created" the location, and now because `all_locations` it is a dictionary in a dictionary that lead to an error.

A bit more detail in the comment below at https://github.com/home-assistant/supervisor/pull/5569#issuecomment-2609783007.

tests/backups/test_manager.py
agners: Unfortunately how to get the location is different for None and mount backups. I've reused existing code now, so that makes the if statement unnecessary.

PR #5590, opened 2025-01-30
Handle non-existing backup file

agners:
## Proposed change
Currently handling if the backup tar-file of a previously scanned tar file vanished is rather inconsistent. Most API's return with 400 error code, but error handling varies from simple error message in logs (e.g. when deleting) up to printing a stack trace (e.g. restoring with password).

This PR consistently uses error 404 when the backup file vanished.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

## Summary by CodeRabbit

- **New Features**
	- Introduced a new `BackupFileNotFoundError` exception to improve error handling for backup operations.

- **Bug Fixes**
	- Enhanced error reporting for scenarios involving missing backup files.
	- Improved API response for backup-related file not found errors.

- **Tests**
	- Added new test cases to verify error handling for backup file removal and restore operations.

These updates enhance the robustness of backup functionalities by providing clearer error messages and improved handling of missing backup files.

### DIFF
--- a/supervisor/api/backups.py
+++ b/supervisor/api/backups.py
@@ -49,7 +49,7 @@ from ..const import (
     CoreState,
 )
 from ..coresys import CoreSysAttributes
-from ..exceptions import APIError, APIForbidden, APINotFound, BackupFileNotFoundError
+from ..exceptions import APIError, APIForbidden, APINotFound
 from ..jobs import JobSchedulerOptions
 from ..mounts.const import MountUsage
 from ..resolution.const import UnhealthyReason
@@ -321,8 +321,6 @@ class APIBackups(CoreSysAttributes):
             if event_task in pending:
                 event_task.cancel()
             return (backup_task, job.uuid)
-        except BackupFileNotFoundError as err:
-            raise APINotFound(str(err)) from err
         finally:
             self.sys_bus.remove_listener(listener)
 
@@ -408,11 +406,8 @@ class APIBackups(CoreSysAttributes):
             self.sys_backups.do_restore_full, backup, **body
         )
 
-        try:
-            if background and not restore_task.done() or await restore_task:
-                return {ATTR_JOB_ID: job_id}
-        except BackupFileNotFoundError as err:
-            raise APINotFound(str(err)) from err
+        if background and not restore_task.done() or await restore_task:
+            return {ATTR_JOB_ID: job_id}
         raise APIError(
             f"An error occurred during restore of {backup.slug}, check job '{job_id}' or supervisor logs for details",
             job_id=job_id,
@@ -430,11 +425,9 @@ class APIBackups(CoreSysAttributes):
         restore_task, job_id = await self._background_backup_task(
             self.sys_backups.do_restore_partial, backup, **body
         )
-        try:
-            if background and not restore_task.done() or await restore_task:
-                return {ATTR_JOB_ID: job_id}
-        except BackupFileNotFoundError as err:
-            raise APINotFound(str(err)) from err
+
+        if background and not restore_task.done() or await restore_task:
+            return {ATTR_JOB_ID: job_id}
         raise APIError(
             f"An error occurred during restore of {backup.slug}, check job '{job_id}' or supervisor logs for details",
             job_id=job_id,
@@ -464,10 +457,7 @@ class APIBackups(CoreSysAttributes):
         else:
             self._validate_cloud_backup_location(request, backup.location)
 
-        try:
-            self.sys_backups.remove(backup, locations=locations)
-        except BackupFileNotFoundError as err:
-            raise APINotFound(str(err)) from err
+        self.sys_backups.remove(backup, locations=locations)
 
     @api_process
     async def download(self, request: web.Request):
--- a/supervisor/api/backups.py
+++ b/supervisor/api/backups.py
@@ -412,7 +412,7 @@ class APIBackups(CoreSysAttributes):
             if background and not restore_task.done() or await restore_task:
                 return {ATTR_JOB_ID: job_id}
         except BackupFileNotFoundError as err:
-            raise APIError(str(err)) from err
+            raise APINotFound(str(err)) from err
         raise APIError(
             f"An error occurred during restore of {backup.slug}, check job '{job_id}' or supervisor logs for details",
             job_id=job_id,
--- a/supervisor/api/backups.py
+++ b/supervisor/api/backups.py
@@ -49,7 +49,7 @@ from ..const import (
     CoreState,
 )
 from ..coresys import CoreSysAttributes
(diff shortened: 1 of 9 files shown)


supervisor/api/backups.py
```diff
@@ -406,8 +408,11 @@ async def restore_full(self, request: web.Request):
             self.sys_backups.do_restore_full, backup, **body
         )
 
-        if background and not restore_task.done() or await restore_task:
-            return {ATTR_JOB_ID: job_id}
+        try:
+            if background and not restore_task.done() or await restore_task:
+                return {ATTR_JOB_ID: job_id}
+        except BackupFileNotFoundError as err:
+            raise APIError(str(err)) from err
```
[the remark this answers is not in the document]
agners: Right, fixed!

supervisor/backups/manager.py
```diff
@@ -292,24 +293,27 @@ def remove(
...
+                ) from err
             except OSError as err:
                 if err.errno == errno.EBADMSG and location in {
                     None,
                     LOCATION_CLOUD_BACKUP,
                 }:
                     self.sys_resolution.unhealthy = UnhealthyReason.OSERROR_BAD_MESSAGE
-                _LOGGER.error("Can't remove backup %s: %s", backup.slug, err)
-                return False
+                raise err
```
mdegat01, line 311: Don't raise `OSError`. This will kick out of all sorts of things and be treated as an unhandled system error. It will also change the status code from `400` to `500` I believe, though the job might turn it into a generic `JobException` first.

Returning `False` here actually does result in an "Unknown error, check logs" type error today (although not after this PR since it seems like you removed that). If we want to raise instead that's fine but we should raise an exception that extends from `HassioError` to denote that its been handled at this point.

supervisor/exceptions.py
```diff
@@ -659,6 +659,10 @@ class BackupJobError(BackupError, JobException):
     """Raise on Backup job error."""
 
 
+class BackupFileNotFoundError(BackupError):
```
mdegat01, line 662:
```suggestion
class BackupFileNotFoundError(BackupError, APINotFound):
```
Then you can remove all the `try...except` logic added in `api/backups.py` converting `BackupFileNotFoundError` into `APINotFound`, that exception will just result in a 404 as is.

mdegat01 (changes requested):
Main change is that we should only raise exceptions that extend from `HassioError` once we catch a system one like `OSError` in an except block to denote to the upstream that its been handled. Also I think extending from `APINotFound` is a bit cleaner.

supervisor/exceptions.py
agners: Hm, that is a pattern we haven't used so far. Isn't that violating a bit separation of concerns here, as in the backup errors should not be dealing with API things? :thinking: Sure, this explicit catches add code, but it does seem cleaner to me.

supervisor/backups/manager.py
agners: > Returning `False` here actually does result in an "Unknown error, check logs" type error today (although not after this PR since it seems like you removed that). If we want to raise instead that's fine but we should raise an exception that extends from `HassioError` to denote that its been handled at this point.

The only reason delete didn't succeed was when there is a OSError. I think using return value is the wrong approach here. Delete should always work, we don't have a "internal" reason we reject a delete. All reasons are essentially unexpected, which should be handled with exceptions. So I think returning exception here is correct/better, IMHO.

But yes, I agree we should not let OSError bubble up to API if we don't have to.

I've converted the error to a BackupError, which extends from `HassioError`. That is in line with other places in the same file, and will lead to "Unknown error, check logs" in those cases.

supervisor/exceptions.py
mdegat01: I don't really think it violates that personally. Things can have multiple inheritance in python and these are all basically just mixins. If an exception is expected to bubble up to the API and should always result in a particular status code it seems better to define it this way.

If anything we might want to consider auditing our other exception in a similar fashion. The only reason there isn't more examples of this right now is because Supervisor returned only 3 status codes:
- `200` for success
- `400` for handled error with a likely descriptive message (which meant the error extended from `HassioError` internally)
- `500` for unhandled error with an "unknown error, check logs" response (whcih meant the error did not extend from `HassioError` internally)

This isn't the situation anymore, Supervisor tries to use the status code to convey info about what went wrong. So more exceptions should be changed to translate to the appropriate code.

agners: Ok, but if we go down that road, then we should be consequent and mark the Supervisor Exception with the appropriate API error everywhere. E.g.

> * `400` for handled error with a likely descriptive message (which meant the error extended from `HassioError` internally)

Instead of special casing `HassioError` in API, we should make `HassioError` extend from `APIError`.

Or we go the other way around: Add the special exception in `api_process` (e.g. make `BackupFileNotFoundError` an `APINotFound` there).

I think I would prefer the latter.

agners: Well, it wouldn't really hurt if `HassioError` extends from `APIError` for these general cases.

But yeah, sort of that is my point: Any exception we define is potentially only used within Supervisor. Starting to mark some "random" exception how they should translate into API errors seems weird to me. What if we use and handle `BackupFileNotFoundError` somewhere not API related?

What do you think about handling `BackupFileNotFoundError` in `api_process` instead? :thinking:

mdegat01: Sure that's fine. it just might turn into a long list. But that's fine, future problems, we can always refactor if we don't like it

tests/api/test_backups.py
```diff
@@ -503,6 +503,14 @@ async def test_restore_immediate_errors(
     assert resp.status == 400
     assert "No Home Assistant" in (await resp.json())["message"]
 
+    # mock_partial_backup._locations[None]["path"].unlink()
+    resp = await api_client.post(
+        f"/backups/{mock_partial_backup.slug}/restore/partial",
+        json={"background": True},
```
mdegat01, line 509: This should be invalid. The client library won't let you make a request that asks supervisor to do a partial restore of nothing at all. I guess the schema needs some fixing 🙈

supervisor/exceptions.py
mdegat01: i realized this proposal was setting us up for a lot of `isinstance` checks to see what the type of error is. Those have notoriously poor performance so I refactored it to remove the need for them entirely from `api_return_error` instead of adding more. Let me know what you think.

agners: Uh nice, yeah I like that! :+1:

tests/api/test_backups.py
agners: Yeah probably make sense. Thanks for adjusting the test, seems sensible now :+1: .

PR #5593, opened 2025-01-31
Avoid backup size difference by sorting JSON keys

agners:
## Proposed change
It seems that the order of keys in the `homeassistant.json` file leads to a different homeassistant.tar.gz size, which in turn leads to a different overall backup size. It seems that running the full test suite compared to run tests individually lead to a different order of keys in the JSON file.

This PR sorts the keys when dumping the JSON. With that change, the backup size is consistent.

Note that in theory, the size of the gzip still could change since there is random data in the json file. It may compress to different sizes. Currently that doesn't seem to be a problem, but in case it is, we either need to get rid of the random data or just drop the size check from pytest.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

## Summary by CodeRabbit

- **Chores**
	- Updated JSON serialization to include key sorting
	- Added a comment in backup test to clarify potential size variation considerations

### DIFF
--- a/tests/api/test_backups.py
+++ b/tests/api/test_backups.py
@@ -1036,8 +1036,9 @@ async def test_protected_backup(
     assert body["data"]["backups"][0]["locations"] == [None]
     assert body["data"]["backups"][0]["protected"] is True
 
-    # NOTE: Maybe it is not safe to compare size here, as random data in the
-    # backup might change the size (due to gzip).
+    # NOTE: It might not be safe to check size here, as random data in the
+    # backup (e.g. isntance UUID, backup slug) might change the size (due
+    # to gzip).
     assert body["data"]["backups"][0]["location_attributes"] == {
         ".local": {
             "protected": True,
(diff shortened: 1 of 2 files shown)


tests/api/test_backups.py
```diff
@@ -1035,6 +1035,9 @@ async def test_protected_backup(
     assert body["data"]["backups"][0]["location"] is None
     assert body["data"]["backups"][0]["locations"] == [None]
     assert body["data"]["backups"][0]["protected"] is True
+
+    # NOTE: Maybe it is not safe to compare size here, as random data in the
+    # backup might change the size (due to gzip).
```
emontnemery, line 1040: You could mention the slug/backup id, which is derived from the date, as well as the date itself will differ

emontnemery (approved):

tests/api/test_backups.py
agners: Improved the comment.

sairon (approved):
This will eventually change structure of all JSONs we write, but I guess they are not supposed to be human-readable, so it won't hurt. It should be noted though that [docs](https://github.com/ijl/orjson?tab=readme-ov-file#opt_sort_keys) say this about the `OPT_SORT_KEYS`:

> It has a substantial performance penalty and is not recommended in general.

I don't know if there's a place that writes JSONs frequently. If not, it shouldn't be an issue and I'm fine with that.

agners:
Superseeded by #5594.

PR #5599, opened 2025-02-03
Handle non-existing file in Backup password check too

agners:
## Proposed change
Make sure we handle a non-existing backup file also when validating the password.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
	- Enhanced error handling for backup operations to present clearer messages when backups are missing or cannot be read.
	- Improved the backup password validation process by enabling a smoother recovery path when specific errors occur.
	- Streamlined logic for validating backup locations and passwords to ensure accuracy based on specified locations.

- **New Features**
	- Added a new test function to validate the password for backups, covering various scenarios and edge cases.

- **Tests**
	- Updated test functions to allow dynamic selection of backup files, improving flexibility and robustness of test scenarios.
	- Refined handling of backup instance properties during tests, enhancing clarity and maintainability.


supervisor/backups/manager.py
```diff
@@ -727,10 +727,14 @@ async def _validate_location_password(
...
-                raise BackupInvalidError(
-                    f"Invalid password for backup {backup.slug}", _LOGGER.error
-                )
+            try:
+                if not await backup.validate_password():
+                    raise BackupInvalidError(
+                        f"Invalid password for backup {backup.slug}", _LOGGER.error
+                    )
+            except BackupFileNotFoundError:
+                self.sys_create_task(self.sys_backups.reload(location))
```
mdegat01, line 736:
```suggestion
                self.sys_create_task(self.reload(location))
```
`sys_backups` is how other objects get to backup manager but that's where we are already. Although we could just do this call in the other `except` before raising `BackupFileNotFoundError` and drop this `try` `except` if we prefer, we don't have to schedule the `reload` from here specifically.

mdegat01 (changes requested):
Should probably have a test case as well. Should just be able to add a parameter set of a partial restore with an invalid password to here:
https://github.com/home-assistant/supervisor/blob/4c04f364a3bd265842bbc765fed7a7c2e7e00fa2/tests/api/test_backups.py#L1122-L1163

supervisor/backups/manager.py
agners: We don't have the location in the other try except...

mdegat01: oh, good point  👍

mdegat01:
@agners in retrospect, our conversation above should've raised warning flags to me. The `validate_password` method needs to know the location, otherwise it cannot check the password correctly. Since we now support having the same backup in mulitiple locations with different encryption options, we cannot validate the password without knowing the location. We were only validating the password using the default location even if that's not where we were restoring from.

I adjusted the code to fix this bug, removed the unnecessary `try` `catch` and added a test case to keep this all working going forward.

mdegat01 (approved):

agners:
> @agners in retrospect, our conversation above should've raised warning flags to me. The `validate_password` method needs to know the location, otherwise it cannot check the password correctly. Since we now support having the same backup in mulitiple locations with different encryption options, we cannot validate the password without knowing the location. We were only validating the password using the default location even if that's not where we were restoring from.

Uh interesting, that could explain some of the issues we have been seeing that the password was not considered valid in some cases maybe :thinking: 


> I adjusted the code to fix this bug, removed the unnecessary `try` `catch` and added a test case to keep this all working going forward.

Cool, thanks! There was some more fallout in not properly mocked test cases (the test cases mocked the `protected` property directly, instead of the new ultimate source of truth in `all_locations`. Tests should be fine now.

supervisor/backups/backup.py
```diff
@@ -433,7 +433,13 @@ def _validate_file() -> bool:
                     _LOGGER.exception("Unexpected error validating password")
                     return True
 
-        return await self.sys_run_in_executor(_validate_file)
+        try:
+            await self.sys_run_in_executor(_validate_file)
```
agners: The return statement here was actually missing :see_no_evil: 

I've added a test to cover `validate_file`, with that this should no longer happen.

PR #5603, opened 2025-02-05
Connectivity for version fetch

mdegat01:
## Proposed change

Delay initial fetch of version information if there is no network connectivity on the host until either it has been re-established or the connectivity check has been disabled. This way users that turn on their systems before plugging in the ethernet cable or setting up wifi don't have to reboot or force a reload via the CLI to get started.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist


supervisor/updater.py
```diff
@@ -180,6 +197,28 @@ def auto_update(self, value: bool) -> None:
...
+            return
+
+        connectivity_check: bool | None = changed.get(DBUS_ATTR_CONNECTION_ENABLED)
+        connectivity: bool | None = changed.get(DBUS_ATTR_CONNECTIVITY)
+
+        # If there's no connectivity checks, stop waiting for connection
+        # Else when host says we're online, attempt to fetch version data and disable listener
+        if (
+            not connectivity_check
+            or connectivity == ConnectivityState.CONNECTIVITY_FULL
```
agners, line 214: So I am a bit worried that this breaks in some cases where the NetworkManager connectivity check fails but somehow Supervisor is still able to connect to the internet (e.g. differences in DNS fallback logic on host side, which is used by the Host connectivity check).

That said, I think we should resolve those issues, and rely fully rely on NetworkManager's judgement.

tests/test_updater.py
```diff
@@ -71,3 +76,39 @@ async def test_os_update_path(coresys: CoreSys, version: str, expected: str):
...
         assert coresys.updater.version_hassos == AwesomeVersion(expected)
+
+
+@pytest.mark.parametrize(
+    ("connectivity", "check_enabled"),
+    [
+        (ConnectivityState.CONNECTIVITY_FULL, True),
+        (ConnectivityState.CONNECTIVITY_NONE, False),
+    ],
+)
```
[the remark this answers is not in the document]
mdegat01: These aren't valid. The first one is impossible cause the connectivity check is disabled. The second is the initial state that triggers the scenario under test

pvizeli (approved):
nice

supervisor/jobs/__init__.py
```diff
@@ -213,9 +213,9 @@ def _notify_on_job_change(
 
         if attribute.name == "done":
             if value is False:
-                self.sys_bus.fire_event(BusEvent.SUPERVISOR_JOB_START, job.uuid)
+                self.sys_bus.fire_event(BusEvent.SUPERVISOR_JOB_START, job)
             if value is True:
-                self.sys_bus.fire_event(BusEvent.SUPERVISOR_JOB_END, job.uuid)
+                self.sys_bus.fire_event(BusEvent.SUPERVISOR_JOB_END, job)
```
mdegat01, line 218: I realize this feels unrelated. As it turns out, trying to use these in test made me realize they are entirely useless. The listener only receives a `uuid` and it can't actually count on the job still being around to look up more info about it, so there's really no way to reliably know what job started/ended. Passing the actual job in the event makes these useful (and it was needed in the test)

agners (approved):

PR #5649, opened 2025-02-19
Remove I/O in event loop for add-on backup and restore

agners:
## Proposed change
Remove I/O in event loop for add-on backup and restore operations. On backup, this moves the add-on shutdown before metadata is stored in the backup, which slightly lenghens the time the add-on is actually stopped.

However, the biggest contributor here is likely adding the image itself if it is a local backup. However, since that is the minority of cases, I've opted for simplicity over optimizing for this case.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

## Summary by CodeRabbit

- **Refactor**
  - Reorganized backup and restoration workflows for add-ons, ensuring more robust data validation and streamlined processing.
  - Simplified Docker image export and profile backup operations by shifting from asynchronous to direct synchronous execution for improved reliability.

- **Tests**
  - Updated test cases to align with the new synchronous operations, ensuring consistent error handling and stability.

### DIFF
--- a/supervisor/addons/addon.py
+++ b/supervisor/addons/addon.py
@@ -1238,46 +1238,45 @@ class Addon(AddonModel):
         Returns a Task that completes when addon has state 'started' (see start)
         for cold backup. Else nothing is returned.
         """
-        wait_for_start: Awaitable[None] | None = None
 
-        with TemporaryDirectory(dir=self.sys_config.path_tmp) as temp:
-            temp_path = Path(temp)
+        def _addon_backup(
+            store_image: bool,
+            metadata: dict[str, Any],
+            apparmor_profile: str | None,
+            addon_config_used: bool,
+        ):
+            """Start the backup process."""
+            with TemporaryDirectory(dir=self.sys_config.path_tmp) as temp:
+                temp_path = Path(temp)
 
-            # store local image
-            if self.need_build:
+                # store local image
+                if store_image:
+                    try:
+                        self.instance.export_image(temp_path.joinpath("image.tar"))
+                    except DockerError as err:
+                        raise AddonsError() from err
+
+                # Store local configs/state
                 try:
-                    await self.instance.export_image(temp_path.joinpath("image.tar"))
-                except DockerError as err:
-                    raise AddonsError() from err
-
-            data = {
-                ATTR_USER: self.persist,
-                ATTR_SYSTEM: self.data,
-                ATTR_VERSION: self.version,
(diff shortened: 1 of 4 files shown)


supervisor/addons/addon.py
```diff
@@ -1303,19 +1302,37 @@ def _write_tarfile():
...
+        was_running = await self.begin_backup()
+        try:
+            _LOGGER.info("Building backup for add-on %s", self.slug)
+            await self.sys_run_in_executor(
+                _addon_backup,
+                self.need_build,
+                data,
+                apparmor_profile,
+                self.addon_config_used,
+            )
```
mdegat01, line 1326: This is just personal preference but for inline function definitions like this I find it more confusing to pass everything in as parameters. Since every input arg corresponds to exactly one value in the outer function but you have to scroll up and down to figure out what that is. Whereas if we just closed around `self` then those at least would be the same in the inner and the outer one.

Just a nitpick note though, not a big deal either way.

mdegat01 (approved):
LGTM 👍

supervisor/addons/addon.py
agners: Yeah it is closure vs. explicit parameter passing. Closures are a bit tricky as when they bind the variable is slightly unpredictable, e.g. this example:

```python
async def fun():
    test = {"count": 0}  # Mutable dictionary

    def _inner(test):
        print("Do something")
        print("Do something")
        print(f"Inner sees: {test['count']}")  # Uses passed value

    loop = asyncio.get_running_loop()

    future = loop.run_in_executor(None, _inner, test)  # Pass a deep copy

    test = {"count": 42}  # Mutable dictionary
    # Modify test *after* scheduling _inner

    await future  # Wait for _inner to execute

asyncio.run(fun())
```

When passing, you are sure that you'll see 0. However, binding you might get 0 or 42, depending on when the closure gets executed (in my testing I had to delete/add teh two "Do something" prints to adjust the timings, but I got 0 or 42).

But I do see you argument about readability. On the other hand, the variable names also document somewhat what is meant. Ideally we'd use explicit keyword arguments, but that doesn't work with `run_in_executor`.

agners: Actually, we could do this with partial, it would become this:

```python
            await self.sys_run_in_executor(
                partial(
                    _addon_backup,
                    store_image=self.need_build,
                    metadata=data,
                    apparmor_profile=apparmor_profile,
                    addon_config_used=self.addon_config_used,
                )
            )
```

PR #5652, opened 2025-02-20
FileConfiguration uses executor for I/O

mdegat01:
## Proposed change

More I/O out of the event loop. Started as a search for `glob` but ran into `FileConfiguration` and that is the focus of this PR. Didn't really have much of a choice but to do them all in one PR due to the dependencies between the methods and all the dependencies on this utility class. Should cover it though

Small breaking change added - the code deprecated for years at this point in `migrate_system_env` has finally been removed. As the listed deprecation window has long passed on everything in there (and it has been many more years since Supervisor stored config that way) it seems safe to remove at this point.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

## Summary by CodeRabbit

- **New Features**
	- Introduced asynchronous configuration loading across key system components, enhancing responsiveness and performance.
- **Refactor**
	- Converted numerous synchronous operations to asynchronous ones to enable non-blocking execution across add-ons, plugins, and core services.
- **Tests**
	- Updated the test suite to support asynchronous execution, ensuring reliable validation of the new asynchronous behavior.

### DIFF
--- a/supervisor/__main__.py
+++ b/supervisor/__main__.py
@@ -55,7 +55,7 @@ if __name__ == "__main__":
     loop.run_until_complete(coresys.core.connect())
 
     bootstrap.supervisor_debugger(coresys)
-    bootstrap.migrate_system_env(coresys)
+    loop.run_until_complete(bootstrap.migrate_system_env(coresys))
 
     # Signal health startup for container
     run_os_startup_check_cleanup()
--- a/supervisor/__main__.py
+++ b/supervisor/__main__.py
@@ -55,7 +55,6 @@ if __name__ == "__main__":
     loop.run_until_complete(coresys.core.connect())
 
     bootstrap.supervisor_debugger(coresys)
-    loop.run_until_complete(bootstrap.migrate_system_env(coresys))
 
     # Signal health startup for container
     run_os_startup_check_cleanup()
--- a/supervisor/bootstrap.py
+++ b/supervisor/bootstrap.py
@@ -55,35 +55,35 @@ _LOGGER: logging.Logger = logging.getLogger(__name__)
 
 async def initialize_coresys() -> CoreSys:
     """Initialize supervisor coresys/objects."""
-    coresys = CoreSys()
+    coresys = await CoreSys().load_config()
 
     # Initialize core objects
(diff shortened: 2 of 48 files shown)


supervisor/bootstrap.py
```diff
@@ -256,8 +256,8 @@ def migrate_system_env(coresys: CoreSys) -> None:
             coresys.config.addons_repositories
         )
         coresys.config._data[ATTR_ADDONS_CUSTOM_LIST] = []
-        coresys.store.save_data()
-        coresys.config.save_data()
+        await coresys.store.save_data()
+        await coresys.config.save_data()
```
agners, line 260: Have you considered leaving `read_data`/`save_data` a regular function with blocking I/O, and use additional helper functions (like `load_config` you've already introduced) instead? I think this would allow to group I/O into a single executor call maybe. :thinking:

mdegat01: So the tricky part here is even `save_data` can actually modify `self._data`. That's effectively a global in all these classes so modifying it doesn't seem like something we should do in the executor.

I think bootstrap is the main place we would want to leverage this anyway, basically anywhere else is just reading or writing data from one object. I had actually planned to come back to bootstrap at some point already. I'd like to do some of these in parallel with `asyncio.gather` if possible but didn't want to do that in this PR since there could be dependencies between them. I can take a look at how to group some of these then by refactoring into helpers.

supervisor/__main__.py
```diff
@@ -55,7 +55,7 @@ def run_os_startup_check_cleanup() -> None:
     loop.run_until_complete(coresys.core.connect())
 
     bootstrap.supervisor_debugger(coresys)
-    bootstrap.migrate_system_env(coresys)
+    loop.run_until_complete(bootstrap.migrate_system_env(coresys))
```
agners, line 58: This feels a bit hacky.

That said, there is more I/O being done in `migrate_system_env`. But `migrate_system_env` seems to be old code anyways, even the comment says we can remove it in 2022. So let's just drop that function entirely, so we don't have to touch this.

agners (approved):
LGTM!

PR #5673, opened 2025-02-25
Temporary directory to executor

mdegat01:
## Proposed change

Move all remaining usage of `TemporaryDirectory` into executor for creation and breakdown of folder.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Introduced asynchronous configuration loading and non-blocking operations for core services and add-on management, ensuring critical processes complete reliably.
  
- **Refactor**
  - Transitioned numerous system operations to asynchronous execution to enhance system responsiveness, error handling, and resource management.
  
- **Tests**
  - Updated test implementations to align with asynchronous workflows, ensuring consistent and reliable verification of system functionality.

### DIFF
--- a/supervisor/addons/addon.py
+++ b/supervisor/addons/addon.py
@@ -982,7 +982,7 @@ class Addon(AddonModel):
         def install_update_profile() -> Path:
             nonlocal tmp_folder
             tmp_folder = TemporaryDirectory(dir=self.sys_config.path_tmp)
-            profile_file = Path(tmp_folder, "apparmor.txt")
+            profile_file = Path(tmp_folder.name, "apparmor.txt")
             adjust_profile(self.slug, self.path_apparmor, profile_file)
             return profile_file
 
--- a/supervisor/addons/addon.py
+++ b/supervisor/addons/addon.py
@@ -977,11 +977,21 @@ class Addon(AddonModel):
             return
 
         # Need install/update
-        with TemporaryDirectory(dir=self.sys_config.path_tmp) as tmp_folder:
-            profile_file = Path(tmp_folder, "apparmor.txt")
+        tmp_folder: TemporaryDirectory | None = None
 
+        def install_update_profile() -> Path:
+            nonlocal tmp_folder
+            tmp_folder = TemporaryDirectory(dir=self.sys_config.path_tmp)
+            profile_file = Path(tmp_folder, "apparmor.txt")
             adjust_profile(self.slug, self.path_apparmor, profile_file)
+            return profile_file
+
+        try:
+            profile_file = await self.sys_run_in_executor(install_update_profile)
             await self.sys_host.apparmor.load_profile(self.slug, profile_file)
+        finally:
+            if tmp_folder:
+                await self.sys_run_in_executor(tmp_folder.cleanup)
 
     async def uninstall_apparmor(self) -> None:
         """Remove AppArmor profile for Add-on."""
--- a/supervisor/api/backups.py
+++ b/supervisor/api/backups.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import asyncio
 from collections.abc import Callable
 import errno
+from io import IOBase
 import logging
 from pathlib import Path
 import re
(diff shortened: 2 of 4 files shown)


supervisor/addons/addon.py
```diff
@@ -977,11 +977,21 @@ async def install_apparmor(self) -> None:
...
 
         # Need install/update
-        with TemporaryDirectory(dir=self.sys_config.path_tmp) as tmp_folder:
-            profile_file = Path(tmp_folder, "apparmor.txt")
+        tmp_folder: TemporaryDirectory | None = None
 
+        def install_update_profile() -> Path:
+            nonlocal tmp_folder
+            tmp_folder = TemporaryDirectory(dir=self.sys_config.path_tmp)
+            profile_file = Path(tmp_folder, "apparmor.txt")
```
agners, line 985: I think this should be:
```suggestion
            profile_file = Path(tmp_folder.name, "apparmor.txt")
```

(since that is what the TemporaryDirectory context manager in `__enter__` returns).

supervisor/addons/addon.py
```diff
@@ -977,11 +977,21 @@ async def install_apparmor(self) -> None:
...
             adjust_profile(self.slug, self.path_apparmor, profile_file)
+            return profile_file
+
+        try:
+            profile_file = await self.sys_run_in_executor(install_update_profile)
             await self.sys_host.apparmor.load_profile(self.slug, profile_file)
+        finally:
+            if tmp_folder:
+                await self.sys_run_in_executor(tmp_folder.cleanup)
```
agners, line 994: Maybe we can avoid using the nonlocal variable like so? :thinking: 

```suggestion
        def install_update_profile() -> Path:
            tmp_folder = TemporaryDirectory(dir=self.sys_config.path_tmp)
            try:
                profile_file = Path(tmp_folder.name, "apparmor.txt")
                adjust_profile(self.slug, self.path_apparmor, profile_file)
                return tmp_folder, profile_file
            except:
                tmp_folder.cleanup()

        tmp_folder, profile_file = await self.sys_run_in_executor(install_update_profile)
        try:
            await self.sys_host.apparmor.load_profile(self.slug, profile_file)
        finally:
            await self.sys_run_in_executor(tmp_folder.cleanup)
```

supervisor/api/backups.py
```diff
@@ -518,29 +519,28 @@ async def upload(self, request: web.Request):
             except vol.Invalid as ex:
                 raise APIError(humanize_error(filename, ex)) from None
 
-        with TemporaryDirectory(dir=tmp_path.as_posix()) as temp_dir:
-            tar_file = Path(temp_dir, "backup.tar")
+        temp_dir: TemporaryDirectory | None = None
+        backup_file_stream: IOBase | None = None
+
+        def open_backup_file() -> Path:
+            nonlocal temp_dir, backup_file_stream
```
agners, line 526: Similar to above, we could make the error handling part of `open_backup_file`. The advantage is that the "consumer" of the function doesn't need to care. As in, if the function returns, then the consumer is responsible for cleanup. If not, we can consider things to be clean.

It is a bit a style question, either case should technically be fine.

mdegat01: Actually its not just style here. I originally wrote it that way because I also prefer to use returns. The reason I switched is because if the file open fails in this function then it raises and there is no return. This leaves us unable to clean up the temporary directory we created. The alternative seemed to be creating the temporary directory and opening the file in two separate executor tasks but I picked less clean code over worse performance as seemingly the lesser of two evils.

I suppose I could've used a return for `backup_file_stream` rather then a `nonlocal` like this. But I guess once I realized I had to use `nonlocal` for `temp_dir` I decided to not mix and match output methods.

agners: After discussion out-of-band, we agreed that the two should be technically equivalent. Since functions are inline and not meant to be used separately, making them taking care of cleanup themselfs is not really that big of an advantage. Having cleanup always in the main function might also be seen as improving overall readability. In any case, I am fine either way. So let's consider this closed.

agners (approved):

PR #5688, opened 2025-02-28
Move `read_text` to executor

mdegat01:
## Proposed change

Begin moving calls to `read_text` to executor. This PR was intended to cover all calls to `read_text` but then I realized our sentry method `capture_exception` was doing a call to `read_text`. That required a lot of changes to move all calls to that to the executor so I stopped there and will continue in a follow-up PR.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

## Summary by CodeRabbit

- **Refactor**
  - Transitioned numerous operations—including error handling, file access, and system queries—to an asynchronous model for improved responsiveness and smoother operations.
- **Tests**
  - Updated testing routines to accommodate asynchronous behavior, ensuring reliable validations.


supervisor/api/store.py
```diff
@@ -68,6 +69,15 @@
...
 
 
+def _read_static_file(path: Path) -> Any:
+    """Read in a static file asset for API output.
+
+    Must be run in executor.
+    """
+    with path.open("r") as asset:
+        return asset.read()
+
```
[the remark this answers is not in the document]
mdegat01: Huh yea, good catch coderabbit 👍

supervisor/addons/addon.py
```diff
@@ -1530,7 +1530,7 @@ async def _restart_after_problem(self, state: ContainerState):
                 except AddonsError as err:
                     attempts = attempts + 1
                     _LOGGER.error("Watchdog restart of addon %s failed!", self.name)
-                    capture_exception(err)
+                    await self.sys_run_in_executor(capture_exception, err)
```
agners, line 1533: I was actually wondering that, is `capture_exception` really blocking I/O? I guess it is, but it nowhere really says, from what I can tell? :thinking: 

In any case, I am a bit worried that we don't get some very fatal issues anymore with this e.g. what if the error leads to a segfault right afterwards? In that case the Sentry wouldn't make it out.

Now I am not sure, maybe that is a bit scripted, and given that people who opt-in to sentry might actually block unnecessarily in their production system, we should probably really do this in an executor.

But I'd like to have this centralized. Can you create a `async_capture_exception` which calls the executor? This would allow us to "opt-out", e.g. for a dev instance or something.

mdegat01: > I was actually wondering that, is capture_exception really blocking I/O? I guess it is, but it nowhere really says, from what I can tell? 🤔

It is. Though not necessarily because of Sentry. I also am not sure if the library i blocking I/O or not. But it doesn't seem to support async and makes HTTP calls so I think it is?

But the reason I know its I/O is because of us. We add read the remaining free space from the disk every time and attach that info as metadata in the call here:
https://github.com/home-assistant/supervisor/blob/696dcf6149840f1217ab4c04406280d42219eee9/supervisor/misc/filter.py#L83

That was how it ended up in part of this PR. I found the `read_text` call in the `free_space` property and backtracked to where it was used.

supervisor/utils/log_format.py
```diff
@@ -12,14 +12,14 @@
...
+async def format_message(message: str) -> str:
+    """Return a formatted message if it's known."""
     try:
         match = RE_BIND_FAILED.match(message)
         if match:
             return f"Port '{match.group(1)}' is already in use by something else on the host."
     except TypeError as err:
         _LOGGER.error("The type of message is not a string - %s", err)
-        capture_exception(err)
+        await async_capture_exception(err)
```
agners, line 23: Hm, it seems rather annoying that we have to convert so much to co-routines just because of that odd chance that the type here is wrong. In sentry there is not a single entry from this point.

And actually, checking code I think it is not even possible to get this ValueError :thinking: 

I'd prefer if we just drop this capture. We could convert it into a `_LOGGER.critical` (which ends up in Sentry, not sure how asyncio friendly this is exactly :wink: ), and/or don't handle the exception at all, and let it "bubble up".

agners (approved):
LGTM!

PR #5696, opened 2025-02-28
Load resolution evaluation, check and fixups early

agners:
## Proposed change
Before #5652, these modules were loaded in the constructor, hence early in `initialize_coresys()`. Moving them late actually exposed an issue where NetworkManager connectivity setter couldn't get the `connectivity_check` evaluation, leading to an exception early in bootstrap.

Technically, it might be safe to load the resolution modules only in `Core.connect()`, however then we'd have to load them separately for pytest. Let's go conservative and load them the same place where they got loaded before #5652.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

## Summary by CodeRabbit

- **Refactor**
  - Enhanced the system’s startup process to ensure all resolution-related modules are consistently loaded before subsequent operations, improving overall efficiency and maintainability.
- **Tests**
  - Adjusted test cases to align with the revamped module-loading process, ensuring reliable and consistent behavior during system startup.

### DIFF
--- a/supervisor/resolution/module.py
+++ b/supervisor/resolution/module.py
@@ -46,6 +46,12 @@ class ResolutionManager(FileConfiguration, CoreSysAttributes):
         self._unsupported: list[UnsupportedReason] = []
         self._unhealthy: list[UnhealthyReason] = []
 
+    async def load_modules(self):
+        """Load resolution evaluation, check and fixup modules."""
+        await self._evaluate.load_modules()
+        await self._check.load_modules()
+        await self._fixup.load_modules()
+
     @property
     def data(self) -> dict[str, Any]:
         """Return data."""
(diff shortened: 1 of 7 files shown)


supervisor/resolution/module.py
```diff
@@ -46,6 +46,12 @@ def __init__(self, coresys: CoreSys):
         self._unsupported: list[UnsupportedReason] = []
         self._unhealthy: list[UnhealthyReason] = []
 
+    async def load_modules(self):
+        """Load resolution evaluation, check and fixup modules."""
+        await self._evaluate.load_modules()
+        await self._check.load_modules()
+        await self._fixup.load_modules()
```
mdegat01, line 53: I think we should change each of these 3 `load_modules` modules to not be `async` and run all of them in the executor from here. Rather then having each one do its own loading in a separate executor task.

I know since each of them is setting class data this is normally something we would avoid doing in the executor. But I believe it should be safe in this case given the point at which this code is called. There is nothing else running that could be looking at resolution center data until this completes since we're still in bootstrap.

If you want to be extra safe can make those methods raise if called a second time to ensure they cannot be used outside of bootstrap accidentally and create race conditions.

mdegat01 (approved):

PR #5697, opened 2025-02-28
Capture warnings and report to sentry

agners:
## Proposed change
By default, warnings are simply printed to stderr. This makes them easy to miss in the log. Capture warnings and user Python logger to log them with warning level.

Also, if the message is an instance of Exception (which it typically is), report the warning to Sentry. This is e.g. useful for asyncio RuntimeWarning warnings "coroutine was never awaited".

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Enhanced the application's error monitoring system. Warnings and exceptions are now tracked more effectively, helping improve overall stability and facilitating faster issue resolution.

### DIFF
--- a/supervisor/bootstrap.py
+++ b/supervisor/bootstrap.py
@@ -4,8 +4,10 @@
 import logging
 import os
 import signal
+import warnings
 
 from colorlog import ColoredFormatter
+from sentry_sdk import capture_exception
 
 from .addons.manager import AddonManager
 from .api import RestAPI
@@ -222,6 +224,13 @@ def initialize_system(coresys: CoreSys) -> None:
         config.path_addon_configs.mkdir()
 
 
+def warning_handler(message, category, filename, lineno, file=None, line=None):
+    """Warning handler which logs warnings using the logging module."""
+    _LOGGER.warning("%s:%s: %s: %s", filename, lineno, category.__name__, message)
+    if isinstance(message, Exception):
+        capture_exception(message)
+
+
 def initialize_logging() -> None:
     """Initialize the logging."""
     logging.basicConfig(level=logging.INFO)
@@ -248,6 +257,7 @@ def initialize_logging() -> None:
             },
         )
     )
+    warnings.showwarning = warning_handler
 
 
 def check_environment() -> None:


supervisor/bootstrap.py
```diff
@@ -221,6 +223,13 @@
         )
         config.path_addon_configs.mkdir()
 
+def warning_handler(message, category, filename, lineno, file=None, line=None):
+    """Custom warning handler that logs warnings using the logging module."""
+    _LOGGER.warning(f"{filename}:{lineno}: {category.__name__}: {message}")
+    if isinstance(message, Exception):
+        capture_exception(message)
+        raise Exception()
+
```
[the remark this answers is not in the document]
agners: Whops yeah that was test code. Fixed now.

mdegat01:
I get what you're trying to do here but I'm not sure its a good idea to merge this before even attempting to tackle the problem of reducing/eliminating dangling tasks that we have right now. This may create a lot of noise in sentry an not end up being super helpful. I think maybe we should attempt to tackle the problem first and then put something like this in to help us catch and clean up any we missed.

supervisor/bootstrap.py
```diff
@@ -222,6 +224,13 @@ def initialize_system(coresys: CoreSys) -> None:
         config.path_addon_configs.mkdir()
 
 
+def warning_handler(message, category, filename, lineno, file=None, line=None):
+    """Warning handler which logs warnings using the logging module."""
+    _LOGGER.warning("%s:%s: %s: %s", filename, lineno, category.__name__, message)
+    if isinstance(message, Exception):
+        capture_exception(message)
```
mdegat01, line 231: I get what you're trying to do here but I'm not sure its a good idea to merge this before even attempting to tackle the problem of reducing/eliminating dangling tasks that we have right now. This may create a lot of noise in sentry an not end up being super helpful. I think maybe we should attempt to tackle the problem first and then put something like this in to help us catch and clean up any we missed.

agners: What I am after here are RuntimeWarning warnings "coroutine was never awaited". From what I can tell, those are not the same as not awaited tasks. I don't think not awaited task cause any runtime warnings. Afaik, they only show up when they raise an exception currently.

mdegat01 (approved):

Issue #5606, opened 2025-02-05
SUPERVISOR - Failed to get logs using advanced API (NEW - NOT THE SAME)


RafAustralia:
### Describe the issue you are experiencing

Hi all 

I have read through both previous issues and cases which simply refer to each other and are now closed and limited...

This issue has not been solved from research I have done.

I was told this issue is a similar issue to what happens when a docker is installed, but now I have this issue with bare metal HAOS install.

### What type of installation are you running?

Home Assistant OS

### Which operating system are you running on?

Home Assistant Operating System

### Steps to reproduce the issue

1. At random times supervisor logs stop and supervisor crashes all together
2.We have rebuilt new NVME 2 Drive and new HAOS and reinstated the backup - same or very similar issue occurs.
3. When supervisor crashes - system cannot be accessed, nor rebooted nor diagnosed in any other way but via ssh
4. Happens now around once weekly...
5. Might be related to Websockets, might be related to supervisors connectivity all together.
6. It appears there is an issue with obtaining advanced logs when call is being issued to get these... and timouts occur perhaps...
...

### Anything in the Supervisor logs that might be useful for us?

```txt
copy of the error, since rebuild last night:

2025-02-05 21:52:41.298 INFO (MainThread) [supervisor.resolution.fixup] Starting system autofix at state running
2025-02-05 21:52:41.298 INFO (MainThread) [supervisor.resolution.fixup] System autofix complete
2025-02-05 22:06:30.107 ERROR (MainThread) [supervisor.api] Failed to get supervisor logs using advanced_logs API
Traceback (most recent call last):
  File "/usr/src/supervisor/supervisor/api/host.py", line 262, in advanced_logs_handler
    await response.prepare(request)
  File "/usr/local/lib/python3.13/site-packages/aiohttp/web_response.py", line 453, in prepare
    return await self._start(request)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.13/site-packages/aiohttp/web_response.py", line 461, in _start
    await self._write_headers()
  File "/usr/local/lib/python3.13/site-packages/aiohttp/web_response.py", line 538, in _write_headers
    await writer.write_headers(status_line, self._headers)
  File "/usr/local/lib/python3.13/site-packages/aiohttp/http_writer.py", line 152, in write_headers
    self._write(buf)
    ~~~~~~~~~~~^^^^^
[...]

PR #5713, opened 2025-03-03
Avoid stack traces for known advanced logs errors

agners:
## Proposed change
Known advanced errors, like connection reset, should not lead to stack traces in the Supervisor logs. Let those errors bubble up to the API error handling.

Fixes: #5606

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Enhanced error handling for asynchronous log retrieval operations. Specific errors, such as connectivity issues, are now managed and propagated more transparently. This update improves overall system reliability by ensuring that any errors encountered do not lead to unexpected behavior, offering a more consistent and clear experience when accessing detailed logs.

### DIFF
--- a/supervisor/api/__init__.py
+++ b/supervisor/api/__init__.py
@@ -9,7 +9,7 @@ from aiohttp import web
 
 from ..const import AddonState
 from ..coresys import CoreSys, CoreSysAttributes
-from ..exceptions import APIAddonNotInstalled, HostNotSupportedError
+from ..exceptions import APIAddonNotInstalled, APIError, HostNotSupportedError
 from ..utils.sentry import async_capture_exception
 from .addons import APIAddons
 from .audio import APIAudio
@@ -403,6 +403,10 @@ class RestAPI(CoreSysAttributes):
                 return await self._api_host.advanced_logs_handler(
                     *args, identifier="hassio_supervisor", **kwargs
                 )
+            except APIError as err:
+                # Advanced logs are generally available, but we did encounter an error
+                # (e.g. connection reset). Those errors should bubble up to the API.
+                raise err
             except Exception as err:  # pylint: disable=broad-exception-caught
                 # Supervisor logs are critical, so catch everything, log the exception
                 # and try to return Docker container logs as the fallback


sairon (commented):
I think the proper fix for the linked issue should be #5715 (but note that both PRs fix only the first error mentioned there, not https://github.com/home-assistant/supervisor/issues/5606#issuecomment-2649270104).

I am unsure if we should silence the APIErrors though, do you have other examples where it makes sense?

sairon:
> I wonder though, should we maybe try .. catch them and print a debug log? You might want to be aware during development that these connections are being dropped/resetted. 🤔

Yeah, maybe, that wouldn't hurt. I'll update the other PR accordingly.

> To be clear, it won't be entirely silenced: This will trigger `api_process` default API error handling. e.g. in the case of a connection reset on systemd-journal-gatewayd side, `Connection reset when trying to fetch data from systemd-journald.` would be returned with an HTTP 400 error. That should then show up to the user as an error.

Right, I was a bit unclear by saying "silenced". It's been a while but IIRC the goal was to catch all those unexpected errors and report them to Sentry, because they should not occur in normal operation. And once we learn about the possible failures, add special handling only to those that we really don't care about (or better said - it's sufficient for them to be just logged).

agners:
> Right, I was a bit unclear by saying "silenced". It's been a while but IIRC the goal was to catch all those unexpected errors and report them to Sentry, because they should not occur in normal operation. 

Agreed, and that is what Exception handling does here.

That said, this actually should no longer be necessary since #5681: With that change, by default all unhandled exception will be reported to Sentry (if optted-in, that is). So if Sentry is the only argument, we could also remove this. That said, this does annotate the error a bit (`Failed to get supervisor logs using advanced_logs API`). If we let it bubble up to aiohttp, it will just say `Error handling request from 172.30.32.1`...

> And once we learn about the possible failures, add special handling only to those that we really don't care about (or better said - it's sufficient for them to be just logged).

Yes, and I am considering `ConnectionResetError` as one of this known issues, we handle and re-raise with `APIError`.

Although, i wonder if that exception even can get raised systemd-journal-gatewayd side (so in `journal_logs_reader`).  The `ClientResponse` `read()` [supposedly](https://docs.aiohttp.org/en/stable/client_reference.html#aiohttp.ClientResponse.read) only raises `aiohttp.ClientResponseError`.

But then again, in actual testing, by `sudo systemctl restart systemd-journal-gatewayd` while following logs, it's actually `ClientPayloadError` errors which get raised. Which brings us back to...

> (but note that both PRs fix only the first error mentioned there, not [#5606 (comment)](https://github.com/home-assistant/supervisor/issues/5606#issuecomment-2649270104)).

...and I think we should try to handle this more gracefully.

Now I think my fix is indeed wrong: APIError is not handled in this case, since we dont use the `@api_process` annotation. In fact, we probably can't use regular error handling here, as we are already streaming the response. It is sort of too late for regular http error handling.

In any case, agreed, this here is the wrong fix. So closing.

We probably could try to reconnect "behind the scenes". But not sure if it's worth the effort. Maybe just logging as a warning that a client error while talking to systemd-joournal-gatewayd happened, and move on.

PR #5720, opened 2025-03-03
Move write of core state to executor

mdegat01:
## Proposed change

Move the write of core state to executor

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist


supervisor/core.py
```diff
@@ -57,32 +57,38 @@ def healthy(self) -> bool:
         """Return true if the installation is healthy."""
         return len(self.sys_resolution.unhealthy) == 0
 
-    def _write_run_state(self, new_state: CoreState):
+    async def _write_run_state(self):
```
agners, line 60: If a single field is read accessed only, I usually prefer passing it as a parameter over accessing the local state. This would allow to make it a class method, but that is probably not really an advantage. It's more that it feels more encapsulated, avoids any potential concurrency issue (especially since it will be accessed in an executor), and generally is more explicit in what `_write_run_state` will end up writing.

Not sure, but I would guess that `self._state` is slightly more expensive as it does some lookup on the object instance.

mdegat01: It's not accessing the value in the executor. It passes that value into `write_text` as a parameter and that's the only thing run in the executor. I can change it if we want but it felt odd to me like that. It kind of implied that you could write a state that was different then supervisor's current state which we wouldn't really want.

As for performance impact, I really doubt the difference is meaningful. Instance variables are still fast. And this also changes very infrequently. Its basically a couple times during start up, a couple times during shutdown and then twice per backup.

agners: > It's not accessing the value in the executor. It passes that value into `write_text` as a parameter and that's the only thing run in the executor.

Right! So there is no asyncio concurrency issue, and since we run this always in the event loop, there should be really no concurrency issues, ever. But passing state as an argument would not even need proof of that, it becomes inherently clear, which is kinda nice, IMHO.

> I can change it if we want but it felt odd to me like that. It kind of implied that you could write a state that was different then supervisor's current state which we wouldn't really want.

Yeah that's why it is private, it can only be used from within, and we control explicitly when state gets written. :shrug:  But yeah, you could see this as an argument to read back current state from the instance variable.

> As for performance impact, I really doubt the difference is meaningful. Instance variables are still fast. And this also changes very infrequently. Its basically a couple times during start up, a couple times during shutdown and then twice per backup/restore.

Sure, this is probably not relevant here, especially since this is doing I/O which is sooo much more costly in comparison.

Anyhow, if you prefer looking up the current state in that method, fine by me.

agners (approved):
LGTM, thanks.

PR #5731, opened 2025-03-05
Add blockbuster library and find I/O from unit tests

mdegat01:
## Proposed change

The intent behind this PR was to add the [Blockbuster](https://pypi.org/project/blockbuster/) and enable it for all unit tests. That would help us find remaining blocking I/O in the event loop and then keep it out going forward.

Unfortunately it found far too much. So this PR is more of a checkpoint on that path before it gets too big. It adds the library and what will be our `autouse` fixture and fixes a good chunk of the I/O found from unit tests.

Note that the library is intentionally added to `requirements.txt` rather then `requirements_test.txt`. This is because eventually the plan is to add a supervisor option to enable and disable it on live systems. This will allow us to turn it on in dev systems to further find and eliminate blocking I/O beyond what is currently in our test coverage.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

## Summary by CodeRabbit

- **New Features**
	- Integrated a new dependency to enhance functionality.
	- Added a new asynchronous test for setting the supervisor's timezone via the API.
	- Introduced a new test file to validate the functionality of the panel API.
- **Refactor**
	- Improved asynchronous processing across various service tasks, including API calls, backup handling, log management, and timezone configuration for better responsiveness.
	- Modified several methods to utilize asynchronous calls for improved performance.
- **Tests**
	- Updated test procedures to align with asynchronous changes, ensuring improved stability and reliability.
	- Enhanced tests to utilize asynchronous execution for various functions, improving responsiveness and performance.


supervisor/addons/addon.py
```diff
@@ -1456,10 +1462,16 @@ def _restore_data():
...
+                def check_data_config() -> tuple[bool, bool]:
+                    return (self.path_data.is_dir(), self.path_config.is_dir())
+
                 _LOGGER.info("Restoring data and config for addon %s", self.slug)
-                if self.path_data.is_dir():
+                data_is_dir, config_is_dir = await self.sys_run_in_executor(
+                    check_data_config
+                )
+                if data_is_dir:
                     await remove_data(self.path_data)
```
agners, line 1473: From what I can tell, `remove_data` should be fine if data dir doesn't exist. So we could run this unconditionally :thinking:

supervisor/addons/build.py
```diff
@@ -103,25 +96,40 @@ def additional_labels(self) -> dict[str, str]:
         """Return additional Docker labels."""
         return self._data[ATTR_LABELS]
 
-    @property
-    def is_valid(self) -> bool:
+    def dockerfile(self) -> Path:
```
agners, line 99: I'd prefer if we name such property replacement with a verb, e.g.

```suggestion
    def get_dockerfile(self) -> Path:
```

This also has the advantage if we were to use them as a property still, not end up with unnoticed surprises.

supervisor/api/__init__.py
```diff
@@ -751,7 +751,10 @@ def _register_store(self) -> None:
         )
 
     def _register_panel(self) -> None:
-        """Register panel for Home Assistant."""
+        """Register panel for Home Assistant.
+
+        Must be run in executor.
```
agners, line 756: Why? `joinpath` should be fine no? :thinking:

supervisor/api/utils.py
```diff
@@ -174,12 +175,21 @@ def api_return_ok(data: dict[str, Any] | None = None) -> web.Response:
...
+    use_executor: bool = False,
 ) -> dict[str, Any]:
     """Validate request data with schema."""
     data: dict[str, Any] = await request.json(loads=json_loads)
     try:
-        data_validated = schema(data)
+        if use_executor:
+            data_validated = await asyncio.get_running_loop().run_in_executor(
+                None, schema, data
+            )
```
agners, line 190: This also surprises me a bit. Why/in what case does schema validation do I/O?

supervisor/host/logs.py
```diff
@@ -51,13 +52,19 @@ def __init__(self, coresys: CoreSys):
...
+        ) or await self.sys_run_in_executor(SYSTEMD_JOURNAL_GATEWAYD_SOCKET.is_socket)
+        return self
 
     @property
     def available(self) -> bool:
         """Check if systemd-journal-gatwayd is available."""
-        if os.environ.get("SUPERVISOR_SYSTEMD_JOURNAL_GATEWAYD_URL"):
-            return True
-        return SYSTEMD_JOURNAL_GATEWAYD_SOCKET.is_socket()
+        return self._available
```
agners, line 67: Is there a specific reason why we keep this as property? I think I'd prefer this as a is/check function (`is_systemd_journal_gatewayd_available()` or `check_...`).

agners (changes requested):

supervisor/addons/addon.py
mdegat01: Ok I'll check that. I was trying to make minimal logic changes to this one to keep bugs from getting in so I mostly just replaced as is.

supervisor/addons/build.py
mdegat01: Yea that's fair. I'll have to do that to a bunch of these, ill scan it 👍

supervisor/api/utils.py
mdegat01: There's only one schema that needs this right now. This one calls `validate_timezone`:
https://github.com/home-assistant/supervisor/blob/e787c17edc94f8ad35d8d9ab9c918f61f44b41fb/supervisor/api/supervisor.py#L63
Which calls `get_time_zone`:
https://github.com/home-assistant/supervisor/blob/1fb4d1cc11a52af0f29636c0008f94487fec0c45/supervisor/utils/validate.py#L22
Which must be run in the executor because `ZoneInfo` does blocking I/O:
https://github.com/home-assistant/supervisor/blob/1fb4d1cc11a52af0f29636c0008f94487fec0c45/supervisor/utils/dt.py#L71-L76

mdegat01: So I realized there wasn't actually a test case for this. I found this one in reverse. Blockbuster identified `get_time_zone` as doing blocking I/O so I traced back all the places it was used. I added a test locally I'll add with this PR to confirm this:
```py
async def test_api_supervisor_options_timezone(
    api_client: TestClient, coresys: CoreSys
):
    """Test setting supervisor timezone via API."""
    assert coresys.timezone == "Etc/UTC"

    resp = await api_client.post(
        "/supervisor/options", json={"timezone": "Europe/Zurich"}
    )
    assert resp.status == 200

    assert coresys.timezone == "Europe/Zurich"
```
And then when I let supervisor try to run that schema in the event loop here's the stack trace from the error Blockbuster raises:
```
File "/home/vscode/.local/ha-venv/lib/python3.13/site-packages/aiohttp/web_protocol.py", line 480, in _handle_request
  File "/home/vscode/.local/ha-venv/lib/python3.13/site-packages/aiohttp/web_app.py", line 569, in _handle
    return await handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/vscode/.local/ha-venv/lib/python3.13/site-packages/aiohttp/web_middlewares.py", line 117, in impl
    return await handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/workspaces/supervisor/tests/conftest.py", line 506, in _security_middleware
    return await handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/workspaces/supervisor/supervisor/api/utils.py", line 65, in wrap_api
    answer = await method(api, *args, **kwargs)
  File "/workspaces/supervisor/supervisor/api/supervisor.py", line 128, in options
    body = await api_validate(SCHEMA_OPTIONS, request)  # , use_executor=True)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspaces/supervisor/supervisor/api/utils.py", line 192, in api_validate
    data_validated = schema(data)
  File "/home/vscode/.local/ha-venv/lib/python3.13/site-packages/voluptuous/schema_builder.py", line 205, in __call__
    return self._compiled([], data)
           ~~~~~~~~~~~~~~^^^^^^^^^^
  File "/home/vscode/.local/ha-venv/lib/python3.13/site-packages/voluptuous/schema_builder.py", line 549, in validate_dict
    return base_validate(path, data.items(), out)
  File "/home/vscode/.local/ha-venv/lib/python3.13/site-packages/voluptuous/schema_builder.py", line 330, in validate_mapping
    cval = cvalue(key_path, value)
  File "/home/vscode/.local/ha-venv/lib/python3.13/site-packages/voluptuous/schema_builder.py", line 779, in validate_callable
    return schema(data)
  File "/workspaces/supervisor/supervisor/utils/validate.py", line 25, in validate_timezone
    if get_time_zone(timezone) is not None:
       ~~~~~~~~~~~~~^^^^^^^^^^
  File "/workspaces/supervisor/supervisor/utils/dt.py", line 77, in get_time_zone
    return zoneinfo.ZoneInfo(time_zone_str)
           ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.13/zoneinfo/_tzpath.py", line 76, in find_tzfile
    if os.path.isfile(filepath):
       ~~~~~~~~~~~~~~^^^^^^^^^^
  File "<frozen genericpath>", line 39, in isfile
```

supervisor/addons/build.py
mdegat01: Oh actually this is the only one. I have a follow-up branch already finishing this effort, I think I'm mixing up the two. I'll make sure to correct any instances of this in the follow-up branch.

supervisor/api/utils.py
agners: Hm, I see. Yeah it does seem weird to me to do the whole validation in a separate thread. It think we should make sure that none of our `voluptuous` checks do I/O, so we can always validate in the main event loop.

For this particular case, we can move the validation into `async def options()` directly. So what I would suggest to do is make timezone in `SCHEMA_SUPERVISOR_CONFIG`  just a string, and then validate "manually" in `options()` if a timezone is supplied (maybe we can even make it smart, and check if the timezone changed. Because validation is only needed if it actually changed).

supervisor/api/__init__.py
agners: Hm, I see, surprising, but ok. It would mean we need to call from a executor. Can we consider aiohttp a thread safe library? :thinking:

mdegat01: Ok after discussion I think I figured out how to separate making the static asset and modifying the webapp object so we aren't modifying the common object in the executor. I also added a testcase for frontend assets so I can confirm it works before and after my changes here.

supervisor/api/__init__.py
```diff
@@ -107,6 +118,17 @@ async def load(self) -> None:
...
+
+            def process_configs() -> list[web.StaticResource]:
+                return [
+                    web.StaticResource(config.prefix, config.path)
+                    for config in static_resource_configs
+                ]
+
+            for resource in await self.sys_run_in_executor(process_configs):
+                self.webapp.router.register_resource(resource)
+
```
bdraco, line 131: It would be a bit nicer to group the executor jobs into a single call like we do in core since the cost of creating the future and waiting for it etc is not small

mdegat01: Isn't that what I did here? I'm making all the static resources in one executor call and then registering all those afterwards

bdraco: Ignore me. I misread the code.

supervisor/api/utils.py
bdraco: We have https://github.com/bluetooth-devices/aiozoneinfo for this in core.. but probably won't help this case

mdegat01: Sure, validation split and it only runs on change.

agners (approved):
LGTM, thanks!

PR #5759, opened 2025-03-17
Enable mypy part 1 (addons and api)

mdegat01:
## Proposed change

On a quest to enable mypy in CI to help us catch more issues before they become bugs. As we've had it off for so long there are quite a few issues so this will take several PRs. This PR covers the `addons` and `api` modules.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

## Summary by CodeRabbit

- **Refactor**
  - Consolidated error handling and state management to improve stability.
  - Streamlined API interfaces with clearer type annotations and default handling.
  - Updated methods for adding unsupported reasons, suggestions, issues, and unhealthy reasons to use dedicated methods instead of direct assignments.

- **Chore**
  - Enhanced default parameter usage for more robust request processing.
  - Updated internal tests to align with improved error tracking and resolution methods.
  - Modified test cases to utilize method calls for adding suggestions and issues, enhancing encapsulation.

### DIFF
--- a/supervisor/api/audio.py
+++ b/supervisor/api/audio.py
@@ -124,7 +124,7 @@ class APIAudio(CoreSysAttributes):
     @api_process
     async def set_volume(self, request: web.Request) -> None:
         """Set audio volume on stream."""
-        source: StreamType = StreamType(request.match_info.get("source", ""))
+        source: StreamType = StreamType(request.match_info["source"])
         application: bool = request.path.endswith("application")
         body = await api_validate(SCHEMA_VOLUME, request)
 
@@ -137,7 +137,7 @@ class APIAudio(CoreSysAttributes):
     @api_process
     async def set_mute(self, request: web.Request) -> None:
         """Mute audio volume on stream."""
-        source: StreamType = StreamType(request.match_info.get("source", ""))
+        source: StreamType = StreamType(request.match_info["source"])
         application: bool = request.path.endswith("application")
         body = await api_validate(SCHEMA_MUTE, request)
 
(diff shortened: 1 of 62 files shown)


agners (approved):

supervisor/api/audio.py
```diff
@@ -137,7 +137,7 @@ async def set_volume(self, request: web.Request) -> None:
     @api_process
     async def set_mute(self, request: web.Request) -> None:
         """Mute audio volume on stream."""
-        source: StreamType = StreamType(request.match_info.get("source"))
+        source: StreamType = StreamType(request.match_info.get("source", ""))
```
agners, line 140: As discussed, this really should always be present. For those cases let's use the dictionary syntax:
```suggestion
        source: StreamType = StreamType(request.match_info["source"])
```

supervisor/api/supervisor.py
```diff
@@ -230,19 +230,12 @@ async def update(self, request: web.Request) -> None:
...
-                        self.sys_updater.reload(),
-                        self.sys_homeassistant.secrets.reload(),
-                        self.sys_resolution.evaluate.evaluate_system(),
-                    ]
-                ]
-            )
+        return asyncio.gather(
+            asyncio.shield(self.sys_updater.reload()),
+            asyncio.shield(self.sys_homeassistant.secrets.reload()),
+            asyncio.shield(self.sys_resolution.evaluate.evaluate_system()),
```
sairon, line 238: This change causes a regression in CLI, causing `ha su reload` to fail:
```
# ha su reload
json: cannot unmarshal array into Go struct field Response.data of type map[string]interface {}
# ha su reload --raw-json
{"result":"ok","data":[null,null,null]}# 
```

Previously it returned just `None`, so an empty `data` object. The changed type hint hides this, return value is now `Awaitable[tuple[None, None, None]]`.

agners: Hm, I see wait previously conflated the results essentially. Simply using wait probably doesn't work either.

Maybe simplest way out is to just make this a coroutine which returns None :shrug: 

```python
    async def reload(self, request: web.Request) -> None:
        """Reload add-ons, configuration, etc."""
        asyncio.gather(
            asyncio.shield(self.sys_updater.reload()),
            asyncio.shield(self.sys_homeassistant.secrets.reload()),
            asyncio.shield(self.sys_resolution.evaluate.evaluate_system()),
        )
        return None
```

PR #5770, opened 2025-03-20
Add API for swap configuration

sairon:
## Proposed change

Add HTTP API for swap size and swappiness to /os/config/swap. Individual options can be set in JSON and are calling the DBus API added in OS Agent 1.7.x, available since OS 15.0. Check for presence of OS of the required version and return 404 if the criteria are not met.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR is related to issue: https://github.com/home-assistant/os-agent/pull/222
- Link to documentation pull request: https://github.com/home-assistant/developers.home-assistant/pull/2611
- Link to cli pull request: TBD
- Link to client library pull request: TBD

## Checklist


supervisor/api/__init__.py
```diff
@@ -258,6 +258,14 @@ def _register_os(self) -> None:
             ]
         )
 
+        # Config options
+        self.webapp.add_routes(
+            [
+                web.get("/os/config/swap", api_os.config_swap_info),
+                web.post("/os/config/swap", api_os.config_swap_options),
```
mdegat01, line 265: Obviously more of a nitpick but we have `/os/config/sync` up above, think we should just add these two to the list up here:
https://github.com/home-assistant/supervisor/blob/c2455f892619facbda38a679d1638b7bbe341408/supervisor/api/__init__.py#L242

supervisor/api/os.py
```diff
@@ -212,3 +222,41 @@ async def boards_other_info(self, request: web.Request) -> dict[str, Any]:
             )
 
         return {}
+
+    @api_process
+    async def config_swap_info(self, request: web.Request) -> dict[str, Any]:
+        """Get swap settings."""
+        if not self.coresys.os.available or self.coresys.os.version < "15.0":
+            raise HTTPNotFound()
```
mdegat01, line 230: Looking at the status codes, this is the only one that makes sense. That being said, we can provide a better explanation then this so I think we should. Something like telling them this API requires the system be HAOS with version >= 15.0 . And perhaps log this as well.

supervisor/api/os.py
```diff
@@ -65,6 +68,13 @@
         vol.Optional(ATTR_SYSTEM_HEALTH_LED): vol.Boolean(),
     }
 )
+
+SCHEMA_SWAP_OPTIONS = vol.Schema(
+    {
+        vol.Optional(ATTR_SWAP_SIZE): vol.Coerce(str),
+        vol.Optional(ATTR_SWAPPINESS): vol.Coerce(int),
```
mdegat01, line 75: This seems very open-ended, is it actually this open-ended? Can we provide any further validation on the string? Maybe `vol.Match([1-9][0-9]M)`? I'm guessing based on the test cases. Although actually if `M` is the only allowed unit we should probably just require a positive integer instead and add the `M` for them.

Similar question on swappiness. Is there a valid range of values? If its just a positive number then we can do `vol.All(int, vol.Range(min=0))` or `min=1` if 0 is invalid.

Additionally we don't want coerce for `str` and `int`. Those data types can be expressed in JSON so we should require that the JSON has a `str` or an `int`, not something that can be coerced into a `str` or `int`

mdegat01 (changes requested):

mdegat01:
@sairon btw this outside diff range comment by coderabbit [here](https://github.com/home-assistant/supervisor/pull/5770#pullrequestreview-2703088127) is right:

> **Actionable comments posted: 3**
> 🔭 Outside diff range comments (1)
> 
> > supervisor/dbus/agent/__init__.py (1)> `97-99`: _⚠️ Potential issue_
> > > **Missing swap in the 'all' property.**
> > > The `all` property returns a list of all managed DBUS interfaces, but it doesn't include the newly added `self._swap`. This could cause issues with methods that rely on this list being complete, such as the `connect` and `shutdown` methods.
> > > ```diff
> > > @property
> > > def all(self) -> list[DBusInterface]:
> > >     """Return all managed dbus interfaces."""
> > > -    return [self.apparmor, self.board, self.cgroup, self.datadisk, self.system]
> > > +    return [self.apparmor, self.board, self.cgroup, self.datadisk, self.system, self.swap]
> > > ```

`self.swap` needs to be added to the list returned by `all` in that object or else the dbus object won't be loaded at startup here:
https://github.com/home-assistant/supervisor/blob/c2455f892619facbda38a679d1638b7bbe341408/supervisor/dbus/agent/__init__.py#L112-L114

supervisor/api/os.py
sairon: Thanks, I'll improve the validations. Swappiness can be only 0-200, that's what Linux accepts. For swap size it's actually quite permissive: https://github.com/home-assistant/operating-system/blob/15.0/buildroot-external/rootfs-overlay/usr/libexec/haos-swapfile#L4-L14

I'll add a regexp that matches what's defined there.

sairon:
Thanks for the comments, I've addressed them and rebased the PR.

About the "Missing swap in the 'all' property" - couldn't be that also tested in unit tests?

supervisor/api/os.py
```diff
@@ -212,3 +224,45 @@ async def boards_other_info(self, request: web.Request) -> dict[str, Any]:
...
 
         return {}
+
+    @api_process
+    async def config_swap_info(self, request: web.Request) -> dict[str, Any]:
+        """Get swap settings."""
+        if not self.coresys.os.available or self.coresys.os.version < "15.0":
+            raise APINotFound(
+                "Home Assistant OS 15.0 or newer required for swap settings"
+            )
```
agners, line 234: Shouldn't this be a server error? E.g. 501Not Implemented? :thinking: 

Or maybe 405 Method Not Allowed, but then we'd need to respond with what is allowed :thinking: 

We didn't use such error codes in the past, but for new APIs we are free to use whatever is appropriate for RESTful APIs. If we go for 501, we should implement a new API exception class.

sairon: 405 means only the method is wrong, different one should work. This is not the case. 

5xx codes are usually indicators of "something bad happening". Also, MDN docs say about [501](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/501):
> Servers are required to support [GET](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Methods/GET) and [HEAD](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Methods/HEAD), and therefore must not return 501 in response to requests with these methods.

I think that 404 used for cases like this is a common pattern in RESTful APIs and personally think it's still the best choice. In the end the body of the error message will be presented to the API consumer and different causes for the API not present don't require different handling, so I don't see much value in making it different.

mdegat01: Yea when I went through the list I came to the same conclusion.

- 501 Not Implemented sounds correct but its really about HTTP methods and that note on GET and HEAD rules it out for us
- 424 Failed Dependency sounded almost correct too but its really about a specific circumstance in WebDAV
- Generic 500 didn't feel correct because this isn't an unknown error. And neither did 400 because its not really a client error either

It would be nice if there was some kind of "update required" status code but I guess that'd be pretty unusual. Normally clients don't have any say over the version of the server. 404 with an explanation seemed like the best fit to me as well.

supervisor/api/os.py
```diff
@@ -65,6 +68,15 @@
         vol.Optional(ATTR_SYSTEM_HEALTH_LED): vol.Boolean(),
     }
 )
+
+RE_SWAP_SIZE = re.compile(r"^\d+([KMG](i?B)?|B)?$", re.IGNORECASE)
```
mdegat01, line 72: So I definitely like this validation better. But I guess one thing for you to consider here is do we need all these options? Just because the OS supports it doesn't necessarily mean we have to provide the full set to our clients. If we think they're unlikely to need anything besides MB and GB we can restrict it to those options for simplicity for example.

I'll leave it up to you. We can always add more options later as people need them but once added we can't take options back without a breaking change.

mdegat01 (approved):

supervisor/api/os.py
sairon: > Just because the OS supports it doesn't necessarily mean we have to provide the full set to our clients.

OTOH if you can set it manually in the `/etc/defaults/` file and its accepted, not accepting back the value that will be shown in GET isn't good either. I prefer to keep support for kilos, as it's what `/proc/meminfo` shows (as `kB`, even though it's technically `KiB`), that's mainly why I implemented this laxly.

supervisor/api/os.py
agners: Hm, ok I see, 501 seems really not appropriate as this is a `GET` request.

Probably 400 or 404 are really the best options here, I am fine either way.

agners (approved):

PR #5805, opened 2025-04-03
Fix mypy issues in docker, hardware and homeassistant modules

mdegat01:
## Proposed change

As it says on the tin.

Note that one particular mypy issue in the `homeassistant` module is separated into #5806 for reasons that should make sense from its description.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

## Summary by CodeRabbit

- **Refactor**
	- Streamlined image handling and validation processes across various components for improved clarity and encapsulation.
	- Enhanced the structure of the DockerInterface and its error handling mechanisms.
	- Updated return types in several methods to allow for `None`, improving error handling capabilities.
- **Bug Fixes**
	- Improved validation and error handling to prevent issues during system updates, integration, and network operations.
	- Added checks to ensure robust error management in various methods.
- **Tests**
	- Updated test setups and dependency injection to boost reliability and consistency across system validations.

These refinements will contribute to a more robust and dependable experience during updates and overall operations.

### DIFF
--- a/supervisor/hardware/manager.py
+++ b/supervisor/hardware/manager.py
@@ -1,8 +1,9 @@
 """Hardware Manager of Supervisor."""
 
+from __future__ import annotations
+
 import logging
 from pathlib import Path
-from typing import Self
 
 import pyudev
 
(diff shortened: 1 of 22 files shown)


supervisor/hardware/manager.py
```diff
@@ -62,9 +62,16 @@ def __init__(self, coresys: CoreSys):
...
+        self._monitor = HwMonitor(self.coresys, self.udev)
         return self
 
+    @property
+    def udev(self) -> pyudev.Context:
+        """Return Udev context instance."""
+        if not self._udev:
+            raise RuntimeError("Udev context not initialized!")
+        return self._udev
+
```
agners, line 74: Maybe we could use a factory class method instead, e.g.:

```python
class HardwareManager(CoreSysAttributes):

    def __init__(self, coresys: CoreSys, udev: pyudev.Context, monitor: HwMonitor):
        """Initialize Hardware Monitor object."""
        self._devices: dict[str, Device] = {}
        self._udev: pyudev.Context = udev
        self._monitor: HwMonitor = monitor
        self._helper: HwHelper = HwHelper(coresys)
        self._policy: HwPolicy = HwPolicy(coresys)
        self._disk: HwDisk = HwDisk(coresys)

    @classmethod
    async def create(cls: HardwareManager, coresys: CoreSys) -> HardwareManager:
        """Complete initialization of a HardwareManager object within event loop."""
        udev = await coresys.run_in_executor(pyudev.Context)
        monitor = HwMonitor(coresys, udev)
        return cls(coresys, udev, monitor)
```

Now this introduces a somewhat new pattern here, but I think it is better then the current solution as it ensures that the variables are not None from the beginning of the objects lifetime and makes the Runtime exception unnecessary.

supervisor/homeassistant/module.py
```diff
@@ -184,14 +184,13 @@ def default_image(self) -> str:
...
-    def image(self) -> str:
+    def image(self) -> str | None:
         """Return image name of the Home Assistant container."""
         if self._data.get(ATTR_IMAGE):
             return self._data[ATTR_IMAGE]
         return self.default_image
 
-    @image.setter
-    def image(self, value: str | None) -> None:
+    def set_image(self, value: str | None) -> None:
```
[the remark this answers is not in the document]
mdegat01: Oh, I actually mean to do this. Yes, we can and should change the property to this:
```py
    @property
    def image(self) -> str | None:
```
Because it always returns `str`.

But the setter must accept `None`. Because its the way to return to the default image if a user has overridden it temporarily via the API. And mypy will not allow the property to have a return type which is different from the type of the argument on the setter.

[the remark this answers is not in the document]
agners: I guess you meant to this?
```py
    @property
    def image(self) -> str:
```

> But the setter must accept `None`. Because its the way to return to the default image if a user has overridden it temporarily via the API. And mypy will not allow the property to have a return type which is different from the type of the argument on the setter.

Ah I see, yeah then having a separate `set_image` seems fine here :+1:

supervisor/hardware/manager.py
mdegat01: Yea I like that, good call

supervisor/hardware/manager.py
```diff
@@ -48,28 +49,30 @@
...
-        return self
+    @classmethod
+    async def create(cls: type[HardwareManager], coresys: CoreSys) -> HardwareManager:
+        """Complete initialization of a HardwareManager object within event loop."""
+        return cls(coresys, await coresys.run_in_executor(pyudev.Context))
+
+    @property
+    def udev(self) -> pyudev.Context:
+        """Return Udev context instance."""
+        return self._udev
```
agners, line 72: At this point, I think we can just not have a property here or do you have a good reason to keep the property?

agners (approved):
LGTM, thanks!

PR #5833, opened 2025-04-23
Add dedicated version update refresh for main components

agners:
## Proposed change
Currently we have the /refresh_updates endpoint which updates the main component versions (Core, OS, Supervisor, Plug-ins) and the add-on store at the same time. This leads to much more updates then necessary.

To allow more fine grained update refresh control introduce a new endpoint /refresh_update which only refreshes the main component versions.

The /store/reload endpoint already allows to update the add-on store only.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to documentation pull request: https://github.com/home-assistant/developers.home-assistant/pull/2651
- Link to cli pull request:
- Link to client library pull request: https://github.com/home-assistant-libs/python-supervisor-client/pull/90

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Introduced a new API endpoint to refresh updater information independently from other components.  
- **Tests**
  - Added tests to verify the new updater refresh endpoint functions correctly.

### DIFF
--- a/supervisor/api/__init__.py
+++ b/supervisor/api/__init__.py
@@ -345,6 +345,9 @@ class RestAPI(CoreSysAttributes):
         api_root.coresys = self.coresys
 
         self.webapp.add_routes([web.get("/info", api_root.info)])
+        self.webapp.add_routes([web.post("/reload_updates", api_root.reload_updates)])
+
+        # Deprecated
         self.webapp.add_routes([web.post("/refresh_updates", api_root.refresh_updates)])
         self.webapp.add_routes(
             [web.get("/available_updates", api_root.available_updates)]
--- a/supervisor/api/__init__.py
+++ b/supervisor/api/__init__.py
@@ -347,7 +347,7 @@ class RestAPI(CoreSysAttributes):
         self.webapp.add_routes([web.get("/info", api_root.info)])
         self.webapp.add_routes([web.post("/reload_updates", api_root.reload_updates)])
 
-        # Deprecated
+        # Discouraged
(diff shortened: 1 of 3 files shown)


supervisor/api/__init__.py
```diff
@@ -345,6 +345,9 @@ def _register_root(self) -> None:
         api_root.coresys = self.coresys
 
         self.webapp.add_routes([web.get("/info", api_root.info)])
+        self.webapp.add_routes([web.post("/reload_updates", api_root.reload_updates)])
+
+        # Deprecated
```
sairon, line 350: Should we handle it as an actual deprecation with a deprecation schedule at some point or is it rather "discouraged" now?

sairon (approved):
Looks good. But let's not forgot about the client library and documentation update (along with the deprecation notice probably).

supervisor/api/__init__.py
agners: Yeah good point, ideally we should have a deprecation schedule. Today both can be found, notes about deprecated features without date and things with a clear removal date (but which we didn't follow up on yet).

At least `/refresh_updates` API is currently actively used by Core. As per when we actually can remove those depends how long back we want to support Core versions (which is a discussion we also recently started).

And since this API is still used by Core, it also depends until when/if we get the new update refresh handling done.

I was considering not adding any note here, but I prefer to point out that users should rather use other API endpoints (like the new `/reload_updates` and `/store/reload`) instead.

So yeah, discouraged is probably the right approach for now :+1: .

agners:
```suggestion
        # Discouraged
```

PR #5332, opened 2024-10-08
Recreate aiohttp session on connectivity check

agners:
## Proposed change
It seems to actually get a proper connectivity check run, we need a new vanilla ClientSession() object.

## Type of change

- [ ] Dependency upgrade
- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (which adds functionality to the supervisor)
- [ ] Breaking change (fix/feature causing existing functionality to break)
- [ ] Code quality improvements to existing code or addition of tests

## Additional information

- This PR fixes or closes issue: fixes #
- This PR is related to issue:
- Link to documentation pull request:
- Link to cli pull request:
- Link to client library pull request:

## Checklist

- [ ] The code change is tested and works locally.
- [ ] Local tests pass. **Your PR cannot be merged unless tests pass**
- [ ] There is no commented out code in this PR.
- [ ] I have followed the [development checklist][dev-checklist]
- [ ] The code has been formatted using Ruff (`ruff format supervisor tests`)
- [ ] Tests have been added to verify that the new code works.

If API endpoints or add-on configuration are added/changed:

- [ ] Documentation added/updated for [developers.home-assistant.io][docs-repository]
- [ ] [CLI][cli-repository] updated (if necessary)
- [ ] [Client library][client-library-repository] updated (if necessary)

[dev-checklist]: https://developers.home-assistant.io/docs/en/development_checklist.html
[docs-repository]: https://github.com/home-assistant/developers.home-assistant
[cli-repository]: https://github.com/home-assistant/cli
[client-library-repository]: https://github.com/home-assistant-libs/python-supervisor-client/

PR #5852, opened 2025-04-28
Use separate ClientSession for connectivity check

agners:
## Proposed change
Instead of using the default aiohttp session, lets create a dedicated ClientSession for the connectivity check. This allows us to reconfigure the session for the connectivity check without affecting the global default session.

This is another attempt to fix #5332, which got revert in #5851.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability of connectivity checks to reduce false negatives by using a dedicated connection session and enhanced error handling.

- **Chores**
  - Optimized internal handling of network connections for connectivity checks to improve stability.

### DIFF
--- a/supervisor/supervisor.py
+++ b/supervisor/supervisor.py
@@ -8,6 +8,7 @@ from ipaddress import IPv4Address
 import logging
 from pathlib import Path
 from tempfile import TemporaryDirectory
+from typing import Final
 
 import aiohttp
 from aiohttp.client_exceptions import ClientError
@@ -38,6 +39,10 @@ from .resolution.const import ContextType, IssueType, UnhealthyReason
 from .utils.codenotary import calc_checksum
 from .utils.sentry import async_capture_exception
 
+_CONNECTIVITY_CHECK_TIMEOUT: Final[aiohttp.ClientTimeout] = aiohttp.ClientTimeout(
+    total=10
+)
+
 _LOGGER: logging.Logger = logging.getLogger(__name__)
 
 
@@ -57,6 +62,14 @@ class Supervisor(CoreSysAttributes):
         self.coresys: CoreSys = coresys
         self.instance: DockerSupervisor = DockerSupervisor(coresys)
         self._connectivity: bool = True
+        self._connectivity_client: aiohttp.ClientSession = (
+            self._create_connectivity_client()
+        )
+
+    def _create_connectivity_client(self) -> aiohttp.ClientSession:
+        """Create a new connectivity client."""
+        connector = aiohttp.TCPConnector(use_dns_cache=False, force_close=True)
+        return aiohttp.ClientSession(connector=connector)
 
     async def load(self) -> None:
         """Prepare Supervisor object."""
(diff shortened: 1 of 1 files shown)


supervisor/supervisor.py
```diff
@@ -57,6 +62,14 @@ def __init__(self, coresys: CoreSys):
         self.coresys: CoreSys = coresys
         self.instance: DockerSupervisor = DockerSupervisor(coresys)
         self._connectivity: bool = True
+        self._connectivity_client: aiohttp.ClientSession = (
+            self._create_connectivity_client()
+        )
+
+    def _create_connectivity_client(self) -> aiohttp.ClientSession:
+        """Create a new connectivity client."""
+        connector = aiohttp.TCPConnector(use_dns_cache=False, force_close=True)
```
agners, line 71: The idea behind force_close is to avoid that the connectivity check reuses an existing connection. Right now, force_close wouldn't be necessary since default keep-alive is 15s, and our connectivity check is scheduled no more often than every 30s.

I was hoping that `use_dns_cache` avoids the necessity to recreate the session. But it seems without recreating the session `aiohttp` continous to raise `TimeoutError` even after the DNS issue has been cleared (as described in #5332).

agners:
As a reproducer, this slightly enhanced script previously posted in #5332 can be used:

```python
import asyncio
import logging

import aiohttp


async def fetch(client):
    timeout = aiohttp.ClientTimeout(total=2)
    async with client.head(
        "https://checkonline.home-assistant.io/online.txt", timeout=timeout
    ) as resp:
        assert resp.status == 200


async def main():
    session1 = aiohttp.ClientSession()
    conn = aiohttp.TCPConnector(use_dns_cache=False, force_close=True)
    # async with aiohttp.ClientSession() as client:
    async with aiohttp.ClientSession(connector=conn) as client:
        while True:
            logging.info("Start fetch")
            try:
                await fetch(client)
                logging.info("Fetch finished without Exception.")
            except Exception as e:
                # logging.exception(e)
                logging.error("Error is %r", e)
            # Test must wait for longer than 15 seconds, default keep-alive
            # or use force_close=True
            await asyncio.sleep(20)


logging.basicConfig(
    format="%(asctime)s %(levelname)s %(message)s",
    level=logging.INFO,
    datefmt="%Y-%m-%d %H:%M:%S",
)
asyncio.run(main())

```

My test then involves following steps:
1. Use my router as only DNS server, block access to it via drop firewall rules.
2. Start the script above. This leads to `TimeoutError`, as expected.
3. Add a working DNS server (e.g. 8.8.8.8) to my host OS (e.g. via Supervisor network config)
4. Observe continuous `TimeoutError`

I would expect that  DNS is resolved again, especially when `use_dns_cache=False` is used. But it seems something caches the (DNS) results still, and leads to continuous `TimeoutError`. When I start another instance of the above script it can fetch from the server immediately.

agners:
I've tried to strace the python process in that case, and it seems that DNS requests get made to an existing DNS server. Looking at open sockets reveals that the process is still polling from the initial DNS server:

```
sudo lsof -n -i -a -p 53577
COMMAND   PID USER   FD   TYPE DEVICE SIZE/OFF NODE NAME
python3 53577  sag    6u  IPv4 677586      0t0  UDP 192.168.87.123:46819->192.168.87.1:domain
```

agners:
This wouldn't really be useful as other users of the `ClientSession` would still suffer from having degraded connectivity (since it seems a issue of the `ClientSession` instance).

Closing in favor of #5862.

PR #5862, opened 2025-05-01
Recreate aiohttp ClientSession after DNS plug-in load

### LINKED ISSUE #5857: Supervisor using DNS settings from host OS /etc/resolv.conf

While researching for alternative fixes for #5332, I realized (in https://github.com/home-assistant/supervisor/pull/5852#issuecomment-2838249710) that aiodns (used by aiohttp if available) loads `/etc/resolv.conf` once on startup, and remains using that configuration throughout the `ClientSession` lifetime. Since we intend to only create a single ClientSession to leverage connection pooling etc., this means that Supervisor remains using the initial DNS configuration throughout its lifetime (especially since recreating the session also proved to be problematic, see #5851).

Reading `/etc/resolv.conf` right before creating the session shows that it has the following content at that point:

```
# Generated by Docker Engine.
# This file can be edited; Docker Engine will not make further changes once it
# has been modified.

nameserver 192.168.87.1
search lan

# Based on host file: '/run/systemd/resolve/resolv.conf' (legacy)
# Overrides: []
```

Shortly after Supervisor connects the container to the hassio network:
```
2025-04-29 13:48:36.309 INFO (MainThread) [supervisor.docker.supervisor] Connecting Supervisor to hassio-network
```
This  seems to trigger a `/etc/resolv.conf` change to:

```
# Generated by Docker Engine.
# This file can be edited; Docker Engine will not make further changes once it
# has been modified.

nameserver 127.0.0.11
search lan
options ndots:0

# Based on host file: '/run/systemd/resolve/resolv.conf' (internal resolver)
# ExtServers: [host(192.168.87.1)]
# Overrides: []
# Option ndots from: internal
```

And finally, Supervisor itself updates `/etc/resolv.conf`:
```
2025-04-29 13:49:17.446 INFO (MainThread) [supervisor.plugins.dns] Updated /etc/resolv.conf
```

Which ends up being this:
```
search local.hass.io

nameserver 172.30.32.3

nameserver 127.0.0.11

options ndots:0
```

However, since the client session gets created early on, all this subsequent changes don't apply. This is especially also problematic when changing the DNS settings via network configuration dialog: The Supervisor doesn't seem to learn about this change until next restart. This is what has been observed in #5332.

Home Assistant Supervised

Debian 12

1. Start Supervisor
2. Change DNS
3. Observe that aiohttp doesn't make use of the new DNS

```txt
Relevant parts above.
```

### System information

Dev environment

### Supervisor diagnostics

_No response_

### Additional information

_No response_

agners:
## Proposed change
Create a temporary ClientSession early in case we need to load version information from the internet. This doesn't use the final DNS setup and hence might fail to load in certain situations since we don't have the fallback mechanims in place yet. But if the DNS container image is present, we'll continue the setup and load the DNS plug-in. We then can recreate the ClientSession such that it uses the DNS plug-in.

This works around an issue with aiodns, which today doesn't reload `resolv.conf` automatically when it changes. This lead to Supervisor using the initial `resolv.conf` as created by Docker. It meant that we did not use the DNS plug-in (and its fallback capabilities) in Supervisor. Also it meant that changes to the DNS setup at runtime did not propagate to the aiohttp ClientSession (as observed in #5332).

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR fixes or closes issue: fixes #5857
- Link to cli pull request:

## Checklist

## Summary by CodeRabbit

- **Refactor**
  - Improved initialization and management of network sessions for better reliability and error handling.
  - Enhanced DNS plugin setup to ensure correct DNS server usage during network operations.
  - Centralized valid API states for consistent system validation.

- **Tests**
  - Updated test suites to use improved network session and connectivity mocks, increasing test reliability and clarity.
  - Added new fixtures and mock classes to streamline HTTP response handling in tests.
  - Revised test function signatures to support new mocking strategies and dependencies.
  - Simplified HTTP call mocking by directly mocking session methods in supervisor tests.

### DIFF
--- a/supervisor/coresys.py
+++ b/supervisor/coresys.py
@@ -21,6 +21,7 @@ from .const import (
     ENV_SUPERVISOR_MACHINE,
     MACHINE_ID,
     SERVER_SOFTWARE,
+    VALID_API_STATES,
 )
 
 if TYPE_CHECKING:
@@ -111,6 +112,12 @@ class CoreSys:
 
     async def init_websession(self) -> None:
         """Initialize global aiohttp ClientSession."""
+        if self.core.state in VALID_API_STATES:
+            # Make sure we don't reinitialize the session if the API is running (see #5851)
+            raise RuntimeError(
+                "Initialize ClientSession is not safe when API is running"
+            )
+
         if self._websession:
             await self._websession.close()
 
--- a/supervisor/coresys.py
+++ b/supervisor/coresys.py
@@ -68,7 +68,6 @@ class CoreSys:
 
         # External objects
         self._loop: asyncio.BaseEventLoop = asyncio.get_running_loop()
-        self._websession: aiohttp.ClientSession = aiohttp.ClientSession()
 
         # Global objects
         self._config: CoreConfig = CoreConfig()
@@ -100,11 +99,7 @@ class CoreSys:
         self._security: Security | None = None
         self._bus: Bus | None = None
         self._mounts: MountManager | None = None
-
-        # Set default header for aiohttp
-        self._websession._default_headers = MappingProxyType(
-            {aiohttp.hdrs.USER_AGENT: SERVER_SOFTWARE}
-        )
+        self._websession: aiohttp.ClientSession | None = None
 
         # Task factory attributes
         self._set_task_context: list[Callable[[Context], Context]] = []
@@ -113,6 +108,17 @@ class CoreSys:
         """Load config in executor."""
         await self.config.read_data()
         return self
+
+    async def init_websession(self) -> None:
+        """Initialize global aiohttp ClientSession."""
+        if self._websession:
+            await self._websession.close()
+
+        session = aiohttp.ClientSession(
+            headers=MappingProxyType({aiohttp.hdrs.USER_AGENT: SERVER_SOFTWARE})
+        )
+
+        self._websession = session
 
     async def init_machine(self):
         """Initialize machine information."""
@@ -165,6 +171,8 @@ class CoreSys:
     @property
     def websession(self) -> aiohttp.ClientSession:
         """Return websession object."""
+        if self._websession is None:
+            raise RuntimeError("WebSession not setup yet")
         return self._websession
 
     @property
--- a/supervisor/coresys.py
+++ b/supervisor/coresys.py
@@ -114,8 +114,18 @@ class CoreSys:
         if self._websession:
             await self._websession.close()
 
+        resolver = aiohttp.AsyncResolver()
+
+        # pylint: disable=protected-access
+        _LOGGER.debug(
+            "Initialize ClientSession with AsyncResolver. Using nameservers %s",
+            resolver._resolver.nameservers,
+        )
+        connector = aiohttp.TCPConnector(loop=self.loop, resolver=resolver)
+
         session = aiohttp.ClientSession(
-            headers=MappingProxyType({aiohttp.hdrs.USER_AGENT: SERVER_SOFTWARE})
+            headers=MappingProxyType({aiohttp.hdrs.USER_AGENT: SERVER_SOFTWARE}),
+            connector=connector,
         )
 
         self._websession = session
--- a/supervisor/coresys.py
+++ b/supervisor/coresys.py
@@ -115,7 +115,7 @@ class CoreSys:
         if self.core.state in VALID_API_STATES:
             # Make sure we don't reinitialize the session if the API is running (see #5851)
             raise RuntimeError(
-                "Initialize ClientSession is not safe when API is running"
+                "Initializing ClientSession is not safe when API is running"
             )
 
         if self._websession:
@@ -125,7 +125,7 @@ class CoreSys:
 
         # pylint: disable=protected-access
         _LOGGER.debug(
-            "Initialize ClientSession with AsyncResolver. Using nameservers %s",
+            "Initializing ClientSession with AsyncResolver. Using nameservers %s",
             resolver._resolver.nameservers,
         )
         connector = aiohttp.TCPConnector(loop=self.loop, resolver=resolver)
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -230,6 +230,9 @@ filterwarnings = [
     "ignore:pkg_resources is deprecated as an API:DeprecationWarning:dirhash",
     "ignore::pytest.PytestUnraisableExceptionWarning",
 ]
+markers = [
+    "no_mock_init_websession: disable the autouse mock of init_websession for this test",
+]
 
 [tool.ruff]
 lint.select = [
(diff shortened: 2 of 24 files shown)


agners:
> Hopefully #5864 will take care of this 🤞

Unfortunately, it seems not :cry: I've removed the re-init and instead read the nameservers currently reported by the `DNSResolver` instance of the aiohttp `AsyncResolver`. However, even after inserting a `asyncio.sleep(10)`, the nameserver still seems to remain the same name server (127.0.0.11 in this case), while `/etc/resolv.conf` has the DNS plug-in IP listed first (`172.30.32.3`)  :thinking: .

I've double checked, the relevant Python package versions look good:
```
f2baee4a1873:/# pip list | grep -e pycares -e aiodns
aiodns                     3.3.0
pycares                    4.8.0
```

bdraco:
> However, even after inserting a `asyncio.sleep(10)`, the nameserver still seems to remain the same name server (127.0.0.11 in this case),

Can you share how you are testing this?

agners:
Ok, I've dig into what is going on here, and it seems c-ares rely on inotify to monitor for changes:

https://github.com/c-ares/c-ares/blob/v1.34.5/src/lib/event/ares_event_configchg.c#L157-L162

So we monitor `/etc` for changes. However, crucially this doesn't work for two rather common situations:

1. systemd-resolved or NetworkManager might use symlinks to store `resolv.conf` in a location more suitable where changes are made often. E.g. systemd-resolved symlink /etc/resolv.conf to `/run/systemd/resolve/stub-resolv.conf`. With that, even when opening `/etc/resolv.conf`, there is no "change" in `/etc` directly, since the file modified is located elsewhere.
2. Containers usually have /etc part of the overlayfs file system. Since `/etc` is not on the overlayed file system, changes will not get monitored.

So unfortunately, the current implementation is not really reliable in detecting `/etc/resolv.conf` changes :cry: 

Ideally we just resolve through a statically configured resolver which is reliably gets DNS changes through other mechanisms. The plug-in DNS actually does that already, its just that Supervisor is a special situation since it sets up the plug-in first.

Ideally we'd make the local resolver available on OS level, such that it is there from the very beginning.

bdraco:
> So we monitor `/etc` for changes. However, crucially this doesn't work for two rather common situations:
> 
> 1. systemd-resolved or NetworkManager might use symlinks to store `resolv.conf` in a location more suitable where changes are made often. E.g. systemd-resolved symlink /etc/resolv.conf to `/run/systemd/resolve/stub-resolv.conf`. With that, even when opening `/etc/resolv.conf`, there is no "change" in `/etc` directly, since the file modified is located elsewhere.
> 2. Containers usually have /etc part of the overlayfs file system. Since `/etc` is not on the overlayed file system, changes will not get monitored.

Is this something we can get `c-ares` to solve upstream?

agners:
> > So we monitor `/etc` for changes. However, crucially this doesn't work for two rather common situations:
> > 
> > 1. systemd-resolved or NetworkManager might use symlinks to store `resolv.conf` in a location more suitable where changes are made often. E.g. systemd-resolved symlink /etc/resolv.conf to `/run/systemd/resolve/stub-resolv.conf`. With that, even when opening `/etc/resolv.conf`, there is no "change" in `/etc` directly, since the file modified is located elsewhere.
> > 2. Containers usually have /etc part of the overlayfs file system. Since `/etc` is not on the overlayed file system, changes will not get monitored.
> 
> Is this something we can get `c-ares` to solve upstream?

For 1, I guess we could follow the symlink and monitor that parent directory.

For 2, which is actual the problem here, we could monitor `/etc/resolv.conf` directly. From what I understand, Docker creates `/etc/resolv.conf` in the overlayed file system, so monitoring the file itself should trigger when it gets updated. But this relies on current Docker behavior.

agners:
Since this works right now, and is really the only re-initialization we need (DNS changes afterwards are handled by the resoolver running in the DNS plug-in, the `resolv.conf` of Supervisor stays static), let's move forward with this PR. It also includes some nice tests cleanup etc. Once `resolv.conf` changes correctly get picked up, we can simply drop the ClientSession re-initialization after writing the `resolv.conf`.

supervisor/coresys.py
```diff
@@ -114,6 +110,33 @@ async def load_config(self) -> Self:
         await self.config.read_data()
         return self
 
+    async def init_websession(self) -> None:
+        """Initialize global aiohttp ClientSession."""
+        if self.core.state in VALID_API_STATES:
+            # Make sure we don't reinitialize the session if the API is running (see #5851)
+            raise RuntimeError(
+                "Initialize ClientSession is not safe when API is running"
```
sairon, line 118:
```suggestion
                "Initializing ClientSession is not safe when API is running"
```

supervisor/coresys.py
```diff
@@ -114,6 +110,33 @@ async def load_config(self) -> Self:
...
+            )
+
+        if self._websession:
+            await self._websession.close()
+
+        resolver = aiohttp.AsyncResolver()
+
+        # pylint: disable=protected-access
+        _LOGGER.debug(
+            "Initialize ClientSession with AsyncResolver. Using nameservers %s",
```
sairon, line 128:
```suggestion
            "Initializing ClientSession with AsyncResolver. Using nameservers %s",
```
(not as bad as the previous one, but still)

sairon (approved):
Looks good, if the tests somehow catch how things are initialized, should be fine.

FWIW, I was thinking about another approach at this - and that's using short-lived `ClientSession` where relevant. So subclass `ClientSession`, move the code that's currently in `init_websession` to the class initializer, and use it as a context manager instead of `sys_websession` where it makes sense (namely for requests going out of the docker networks). The `websession` property then could initialize the session (along with an `INFO` logging entry maybe) instead of raising an error. Anyway with 9d869e1aefc680b9dde07d76c9fca7189a1bcc25 implemented we might be fine with the single one without the risk of races. Probably just a food for thought now.

agners:
> FWIW, I was thinking about another approach at this - and that's using short-lived `ClientSession` where relevant. So subclass `ClientSession`, move the code that's currently in `init_websession` to the class initializer, and use it as a context manager instead of `sys_websession` where it makes sense (namely for requests going out of the docker networks). The `websession` property then could initialize the session (along with an `INFO` logging entry maybe) instead of raising an error. Anyway with [9d869e1](https://github.com/home-assistant/supervisor/commit/9d869e1aefc680b9dde07d76c9fca7189a1bcc25) implemented we might be fine with the single one without the risk of races. Probably just a food for thought now.

Yeah I assume that as long as there are no calls through the API, all code using the web session (during startup) is sequential, and hence we should be free to safely initialize. I guess we'll see if that is indeed the case on Sentry rather quickly.

Creating `ClientSession` for callees and let them use it as a context on their end would make sure the lifetime is controlled on caller side, which would solve the problem. But it also would not allow to share the session which brings features such as connection pooling etc.

PR #5879, opened 2025-05-12
Add opt-in IPv6 for containers

davidrapan:
## Proposed change

Hi,
I access my HA instances through nginx and purely using IPv6. But since it is exposed like this to the outside, I have to use the Ban IP feature, which currently doesn't work because nginx communicates with the HA instance using IPv4 and the forward header doesn't work ~~because it refuses to mix these families~~ for some reason (therefore it's banning docker IP of the container). :)

I have this well tested for many months already and there are no side effects to this configuration.

Step I used to reconfigure existing `hassio` instance (it requires to run the image directly):
```
for ID in $(docker ps --format '{{ .ID }}'); do docker stop $ID; done
for ID in $(docker ps -a --format '{{ .ID }}'); do docker rm $ID; done
docker network rm hassio
docker network create hassio -d bridge --ipv6 --subnet fd0c::/64 --subnet 172.30.32.0/23 --ip-range 172.30.33.0/24 --gateway 172.30.32.1 --opt com.docker.network.bridge.name=hassio
reboot
```

After restart, all containers will also receive an IPv6 address that they can use for outgoing connections but the most important thing is that Ban IP now sees the real IPv6 of the connected client.

Link to my FR: https://community.home-assistant.io/t/docker-hassio-network-with-ipv6-enabled-by-default/760739 I made a while ago _but didn't had the chance to look into supervisor until now._

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to documentation pull request: https://github.com/home-assistant/developers.home-assistant/pull/2677
- Link to cli pull request: https://github.com/home-assistant/cli/pull/567

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Added support for enabling or disabling IPv6 for Docker networks via the API and configuration.
  - Docker network settings now include an option to toggle IPv6 support.
  - Asynchronous initialization of Docker network with IPv6-aware migration and container attachment improvements.
  - Introduced Docker container naming constants for Supervisor and Observer components.

- **Bug Fixes**
  - Standardized naming and usage of network mask constants for improved clarity and consistency.
  - Enhanced error handling in Docker network operations.

- **Documentation**
  - Updated docstrings and comments to consistently use correct "IPv4" and "IPv6" capitalization.

- **Tests**
  - Introduced new tests to verify Docker network IPv6 enablement and network recreation logic.
  - Updated existing tests to reflect changes in constants and improve documentation clarity.


agners (commented):
> I access my HA instances through nginx and purely using IPv6. 

By nginx you mean the official [NGINX proxy add-on](https://github.com/home-assistant/addons/tree/master/nginx_proxy)?

> But since it is exposed like this to the outside, I have to use the Ban IP feature, which currently doesn't work because nginx communicates with the HA instance using IPv4 and the Forward header doesn't work because it refuses to mix these families (which is then banning docker IP of the container). :)

Thanks for explaining your specific use case here! :pray: 

I was wondering if X-Forwarded-For indeed doesn't allow mixing IP addresses, but that doesn't seem the problem per-se, even the [X-Forwarded-For examples at mdn](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-Forwarded-For) show headers with mixed address families. I also don't think this is a NGINX problem. This is purely a Docker problem: We expose the outside IP address using Docker, which seemingly makes it look like originating from the Docker network's hassio host IP (`172.30.32.1`). Presumably this is just because Docker uses `docker-proxy` by default like so:

```
/usr/bin/docker-proxy -proto tcp -host-ip :: -host-port 1443 -container-ip 172.30.33.6 -container-port 443 -use-listen-fd
```

Side quest: Interestingly enough there is also an IPv4 docker-proxy running, so why does it work for IPv4 then? Shouldn't that also lead to a masqueraded IPv4 source IP? It turns out that besides the proxy, there is also a iptables rule:

```
iptables -t nat -L DOCKER -n -v
...
    8   480 DNAT       tcp  --  !hassio *       0.0.0.0/0            0.0.0.0/0            tcp dpt:1443 to:172.30.33.6:443

``` 
So it seem the IPv4 proxy is only a fallback :thinking: :shrug: 

Anyhow, I can see why using IPv6 can help here, even in the NAT case: This probably makes Docker use the same DNAT rules for IPv6 as well, hence the source IP is no longer masqueraded.

That said, I am a bit worried what it does for users without IPv6 connectivity: IP client software running in add-on who try to connect to hosts with IPv6 addresses (where the nameserver resolved AAAA) might try connecting to those hosts using IPv6. Maybe this is not a (big) problem because the add-on use ULA address range and clients will ultimately fallback to IPv4. But there is some risk of things not working as expected, I think. Ideally we make that feature an opt-in first maybe? :thinking:

davidrapan:
> By nginx you mean the official [NGINX proxy add-on](https://github.com/home-assistant/addons/tree/master/nginx_proxy)?

Yes.

> I was wondering if X-Forwarded-For indeed doesn't allow mixing IP addresses..

The thing is, when I initially investigated the problem, I even logged the packet headers and they just weren't there, so something is going wrong. I just thought let's try give the containers IPv6 and let's see.. and it worked so my conclusion was nginx is not handing over v6 (but who knows if nginx even know the source IP, proly not) to a server on v4 (in other words, out-of-spec behavior no matter who's to blame here).

So when is the whole traffic after the changes on v6 it very likely hides some problem somewhere (docker-proxy is a strong candidate) but I wasn't able to find out more.

> Anyhow, I can see why using IPv6 can help here, even in the NAT case: This probably makes Docker use the same DNAT rules for IPv6 as well, hence the source IP is no longer masqueraded.

Yes, it's easy to see when observing the traffic there.

> That said, I am a bit worried what it does for users without IPv6 connectivity: IP client software running in add-on who try to connect to hosts with IPv6 addresses (where the nameserver resolved AAAA) might try connecting to those hosts using IPv6. Maybe this is not a (big) problem because the add-on use ULA address range and clients will ultimately fallback to IPv4. But there is some risk of things not working as expected, I think. Ideally we make that feature an opt-in first maybe? 🤔

This can't be a problem practically, because containers don't get a default gateway to v6. So there are no attempts to initiate any v6 connections, except when they need to respond, etc. But I thought about it and tried `hassio` configured like this in a v4-only network and didn't experience any issues whatsoever.

Hmmm, it would be ideal but I guess it would need `hassio` network bridge re-initialization (like I did manually), right?

agners:
> This can't be a problem practically, because containers don't get a default gateway to v6. So there are no attempts to initiate any v6 connections, except when they need to respond, etc. But I thought about it and tried `hassio` configured like this in a v4-only network and didn't experience any issues whatsoever.

Hm, yeah I guess no gateway will make any connection attempt fail immediately, and likely cause to try out the next address (family). But the exact behavior is always a bit tricky, depends on how well the client code is written.

> Hmmm, it would be ideal but I guess it would need `hassio` network bridge re-initialization (like I did manually), right?

Yeah the current code essentially doesn't migrate existing installations. Which maybe is actually not a bad thing, just to see how things work first... :thinking: 

But ultimately we should migrate existing installations.

supervisor/docker/network.py
```diff
@@ -42,53 +47,59 @@ def containers(self) -> list[str]:
...
-        return DOCKER_NETWORK_MASK[6]
+        return DOCKER_IPV4_NETWORK_MASK[6]
 
     def _get_network(self) -> docker.models.networks.Network:
         """Get supervisor network."""
         try:
-            return self.docker.networks.get(DOCKER_NETWORK)
+            if (network := self.docker.networks.get(DOCKER_NETWORK))["EnableIPv6"]:
+                return network
+            network.remove()
```
davidrapan, line 82: What I'm not sure about is that when we are here during the initialization flow (after reboot) are any containers already connected to the network?

agners: No that is early during boostrap, DockerAPI `post_init`. So it should be safe to migrate here. :+1:

davidrapan: Great then. I've added test for the migration (finally works but I had to slightly alter `_get_network` function because I simply wasn't able to patch functions used inside, no chance, hah)

davidrapan:
Hi, @agners, so apparently containers are now getting a default gateway v6? (I could swear that when I did the first tests, they didn't have it):
```
➜  ~ docker exec -i addon_core_nginx_proxy ip -6 route 
fd0c::/64 dev eth0  metric 256 
fe80::/64 dev eth0  metric 256 
default via fd0c::1 dev eth0  metric 1024 
multicast ff00::/8 dev eth0  metric 256 
```
v6 ping:
```
➜  ~ docker exec -i addon_core_nginx_proxy ping -6 google.com
PING google.com (2a00:1450:4014:80a::200e): 56 data bytes
64 bytes from 2a00:1450:4014:80a::200e: seq=0 ttl=116 time=34.882 ms
```
However, the preferable route is still through v4:
```
➜  ~ docker exec -i addon_core_nginx_proxy ping google.com 
PING google.com (142.251.36.110): 56 data bytes
64 bytes from 142.251.36.110: seq=0 ttl=116 time=32.297 ms
```

davidrapan:
> I definitely need to do some testing with different scenarios of broken IPv6 connectivity.

I did some myself today and so far so good.

But anyway, according to [RFC 6724](https://datatracker.ietf.org/doc/html/rfc6724) IPv4 has preference over IPv6 ULA by default:
```
Prefix        Precedence Label
::1/128               50     0
::/0                  40     1
::ffff:0:0/96         35     4
2002::/16             30     2
2001::/32              5     5
fc00::/7               3    13
::/96                  1     3
fec0::/10              1    11
3ffe::/16              1    12
```
v4 (::ffff:0:0/96) has precedence 35 while v6 ULA (fc00::/7) only 3, "we say that address A has higher precedence than address B, meaning that our algorithm will prefer to sort destination address A before destination address B".

BTW, all checks are now "completable". 😉

agners:
First of all, thanks a lot for your work and the additional research! :star_struck: 

> But anyway, according to [RFC 6724](https://datatracker.ietf.org/doc/html/rfc6724) IPv4 has preference over IPv6 ULA in `sourcefilter` by default:
> 
> ```
> Prefix        Precedence Label
> ::1/128               50     0
> ::/0                  40     1
> ::ffff:0:0/96         35     4
> 2002::/16             30     2
> 2001::/32              5     5
> fc00::/7               3    13
> ::/96                  1     3
> fec0::/10              1    11
> 3ffe::/16              1    12
> ```

Interesting, where is that list from/what command prints that?

> 
> v4 (::ffff:0:0/96) has precedence 35 while v6 ULA (fc00::/7) only 3, "..we say that address A has higher precedence than address B, meaning that our algorithm will prefer to sort destination address A before destination address B." And it's the same for source too.

Thanks, yeah that make sense.

> Hi @agners, finally, I also found confirmation that [musl (which is used in Alpine Linux) follows RFC 6724 since version 1.1.3](https://git.musl-libc.org/cgit/musl/commit/?h=v1.1.3)

Hm, this applies for `getaddrinfo`. Sometimes user space does their own thing, e.g. aiodns uses c-ares, and Home Assistant Core uses aiohttp and aiodns for most of the work. That said, I just checked this particular case, it seems that [c-ares also implements sorting according to RFC 6724](https://c-ares.org/docs/ares_getaddrinfo.html), so it should behave the same way. Still, i fear that some user space does apps do things wrong. I guess we can consider this a bug of those apps, but it will appear as a breaking change for users potentially.

I wonder if under certain circumstances (server down) the error will change since with an IPv6 address available we'll attempt connecting to the server using the ULA IPv6 if IPv4 fails, so maybe a socket connection attempt ends with a routing error instead of not responding or similar. This is probably not a big deal, but just things which could be influenced.

Unfortunately, the update concept of Supervisor doesn't allow downgrade. So once this is merged and shipped, the only thing to opt out of it is if we would revert it and ship a new release.

For the reasons above, I'd like to make this an option, which we first set the default to false (make it opt-in), and then then rollout to everyone by changing the default to true. We have the mechanism to do this already, e.g. the `diagnostics` option works exactly like that: If not set by user, the state is None, where we fall back to default (which is false in that case). These options can be set with `ha supervisor options`, e.g. for this case `ha supervisor options --bridge-ipv6=true`). From my point of view it's fine to apply the change only on restart (e.g. require a `ha supervisor restart` after changing the option).

With that, we can get this merged soon and roll out to interested users to test soon. Later, when we enable it by default, we can also easily test potential regressions (e.g. if people report connectivity issue, we can ask them to temporarily disable the feature to see if that was the source of the problem).

I think code wise it won't be all that complex. What is missing then is that we also need to add the revert logic in `_get_network`, essentially compare to the current setting and if it deviates, recreate the bridge.

davidrapan:
> Interesting, where is that list from/what command prints that?

It is from RFC and standard path to this configuration is [/etc/gai.conf](https://man7.org/linux/man-pages/man5/gai.conf.5.html). However in case of musl or kernel address selection it's hardcoded.

It is from the RFC and the standard path to this configuration is [/etc/gai.conf](https://man7.org/linux/man-pages/man5/gai.conf.5.html). However, in the case of musl it is hardcoded.

> Hm, this applies for getaddrinfo. Sometimes user space does their own thing, e.g. aiodns uses c-ares, and Home Assistant Core uses aiohttp and aiodns for most of the work. That said, I just checked this particular case, it seems that [c-ares also implements sorting according to RFC 6724](https://c-ares.org/docs/ares_getaddrinfo.html), so it should behave the same way.

And **kernel address selection**.

Of course, the application is free to `bind()` to any address it wants.

And I totally agree w/ you. I'm very much in favor of the users being able to choose exactly the configuration they want. Forcing it both ways would be too restrictive. I'll try adding a command to the `ha options` namespace. 😉

davidrapan:
I can't find tests for configuration items, are they missing?

davidrapan:
Okay, it should be done now. Supervisor is running and now I'm just waiting for the tests to complete.

davidrapan:
I'm now getting
```
Load key ".ssh/id_rsa": error in libcrypto
```
what's going on?

supervisor/const.py
```diff
@@ -124,6 +126,7 @@
 ATTR_BOARD = "board"
 ATTR_BOOT = "boot"
 ATTR_BRANCH = "branch"
+ATTR_BRIDGE_IPV6 = "bridge_ipv6"
```
agners, line 129: Somehow that name doesn't convince me anymore :see_no_evil:  It affects more than add-ons, also plug-ins. Maybe `use_ipv6_on_bridge`, or `bridge_use_ipv6`? Do you have a good idea?

davidrapan: I'd go with enable_ipv6 (same name as docker uses for clarity) prefixed by docker, bridge or both.

supervisor/docker/manager.py
```diff
@@ -93,6 +94,11 @@ def __init__(self):
         """Initialize the JSON configuration."""
         super().__init__(FILE_HASSIO_DOCKER, SCHEMA_DOCKER_CONFIG)
 
+    @property
+    def bridge_ipv6(self) -> bool:
+        """Return configuration for bridge ipv6."""
+        return self._data.get(ATTR_BRIDGE_IPV6, False)
+
```
agners, line 101: Hm, you've added the property here too, which actually is a good point: This should be underneath the `ha docker` command. I guess it would mean that we need to introduce `ha docker options` command :thinking:  But it shouldn't be that big of a deal, we have that underneath other commands already :shrug:

supervisor/const.py
agners: I think when we have it underneath the `ha docker` command, we can just remove the prefix. Just `enable_ipv6` is fine by me in that case.

supervisor/docker/manager.py
davidrapan: You mean cli right?

supervisor/const.py
davidrapan: You mean directly on the `ha docker` or `ha docker network --enable-ipv6=True`?

supervisor/docker/network.py
```diff
@@ -42,55 +68,60 @@ def containers(self) -> list[str]:
...
+
+    def _get_network(self, enable_ipv6: bool = False) -> docker.models.networks.Network:
         """Get supervisor network."""
         try:
-            return self.docker.networks.get(DOCKER_NETWORK)
+            if (network := self._get(DOCKER_NETWORK)).attrs.get(
+                DOCKER_ENABLEIPV6
+            ) == enable_ipv6:
+                return network
+            network.remove()
```
agners, line 113: Aactually, this is a problem when Supervisor just restarts: In that case we attach to the existing container. Also, we have the observer container by default set to auto-start, in case Supervisor fails to start. The observer hence is also attached to the network.

And removing the network seems not to be possible when there is a container attached to it :see_no_evil: 

```
  File "/usr/local/lib/python3.13/site-packages/docker/api/client.py", line 275, in _raise_for_status
    response.raise_for_status()
    ~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/usr/local/lib/python3.13/site-packages/requests/models.py", line 1024, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 403 Client Error: Forbidden for url: http+docker://localhost/v1.49/networks/d1a2bf591e0f57ebdccabadde552423fa988e588da1626f3f06c6ffaea9cf47c

The above exception was the direct cause of the following exception:

  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/usr/src/supervisor/supervisor/__main__.py", line 55, in <module>
    coresys = loop.run_until_complete(bootstrap.initialize_coresys())
  File "/usr/local/lib/python3.13/asyncio/base_events.py", line 725, in run_until_complete
    return future.result()
           ~~~~~~~~~~~~~^^
  File "/usr/src/supervisor/supervisor/bootstrap.py", line 58, in initialize_coresys
    coresys.docker = await DockerAPI(coresys).post_init()
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/supervisor/supervisor/docker/manager.py", line 136, in post_init
    self._network = DockerNetwork(self._docker, self.config.bridge_ipv6)
                    ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/supervisor/supervisor/docker/network.py", line 51, in __init__
    self._network: docker.models.networks.Network = self._get_network(enable_ipv6)
                                                    ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^
  File "/usr/src/supervisor/supervisor/docker/network.py", line 113, in _get_network
    network.remove()
    ~~~~~~~~~~~~~~^^
  File "/usr/local/lib/python3.13/site-packages/docker/models/networks.py", line 91, in remove
    return self.client.api.remove_network(self.id)
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^
  File "/usr/local/lib/python3.13/site-packages/docker/utils/decorators.py", line 19, in wrapped
    return f(self, resource_id, *args, **kwargs)
  File "/usr/local/lib/python3.13/site-packages/docker/api/network.py", line 186, in remove_network
    self._raise_for_status(res)
    ~~~~~~~~~~~~~~~~~~~~~~^^^^^
  File "/usr/local/lib/python3.13/site-packages/docker/api/client.py", line 277, in _raise_for_status
    raise create_api_error_from_http_exception(e) from e
          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^
  File "/usr/local/lib/python3.13/site-packages/docker/errors.py", line 39, in create_api_error_from_http_exception
    raise cls(e, response=response, explanation=explanation) from e
docker.errors.APIError: 403 Client Error for http+docker://localhost/v1.49/networks/d1a2bf591e0f57ebdccabadde552423fa988e588da1626f3f06c6ffaea9cf47c: Forbidden ("error while removing network: network hassio has active endpoints (addon_core_ssh, hassio_observer, hassio_cli, hassio_dns, hassio_audio, addon_d4456232_puppet, addon_core_whisper, addon_core_configurator)")
```

Not sure how to deal with this, probably the easiest is to check if only the observer container runs, and if so just shut it down. It would mean that we can only migrate on a full system restart, but I think that restriction is fine.

Maybe a full shutdown is not needed, but a simple detaching/reattaching is enough :thinking: 

I guess at that point we could just loop through all containers and do that, but I'd say that would most likely break a bunch of user space applications of running add-ons :cold_face:

davidrapan: > I guess at that point we could just loop through all containers and do that, but I'd say that would most likely break a bunch of user space applications of running add-ons 🥶

This. :-D

Exactly my current approach..

I mean, this is w/ supervisor already running?

davidrapan:
I cannot build w/ `cryptography==45.0.2` for some reason.

supervisor/docker/manager.py
agners: Yeah, `ha docker options` on CLI. I probably should rather mention the endpoint, in this case `/docker/options`.

agners:
> I cannot build w/ `cryptography==45.0.2` for some reason.

In a quick test it works here using devcontainer :thinking: 

With the cryptography bump PR merged wheels for supported architecture should be on our wheels server.

davidrapan:
I've changed v6 prefix to `fd0c:ac1e:2100::/48`, d0c as Docker and `ac1e:2100` is hexadecimal `172.30.33.0`.

supervisor/docker/network.py
```diff
@@ -42,55 +68,65 @@ def containers(self) -> list[str]:
...
+                try:
+                    network.remove()
+                    _LOGGER.info(
+                        "Migrating Supervisor network to %s",
+                        "IPv4/IPv6 Dual-Stack" if enable_ipv6 else "IPv4-Only",
+                    )
+                except docker.errors.APIError:
+                    _LOGGER.info(
+                        "Supervisor restart is required to apply network changes"
+                    )
```
agners, line 121: So a Supervisor restart is not enough, since we leave containers running in that case. At the very least it needs a system restart, so:

```suggestion
                    _LOGGER.info(
                        "Migrating Supervisor network to %s",
                        "IPv4/IPv6 Dual-Stack" if enable_ipv6 else "IPv4-Only",
                    )
                    network.remove()
                except docker.errors.APIError:
                    _LOGGER.warning(
                        "System restart is required to apply Supervisor network changes"
                    )
```

But I think even that is not enough: The Observer runs with restart policy always, which makes it run on startup, probably even before Supervisor :cry: 

I think we need to really add some special code here: Check if Observer plug-in is running, if so either remove and reattach to the hassio network, or simply stop it and rely on later code to start it again.

davidrapan: So we need to disconnect containers from the network. Something like:
```
                except docker.errors.APIError:
                    if (containers := self.docker.networks.containers):
                        for container in containers:
                            self.docker.networks.disconnect(container)
```
?

davidrapan: Something like this?

supervisor/config.py
```diff
@@ -118,6 +119,16 @@ def version(self, value: AwesomeVersion) -> None:
...
+    @property
+    def enable_ipv6(self) -> bool | None:
+        """Return IPv6 configuration for docker network."""
+        return self._data.get(ATTR_ENABLE_IPV6)
+
+    @enable_ipv6.setter
+    def enable_ipv6(self, value: bool) -> None:
+        """Set IPv6 configuration for docker network."""
+        self._data[ATTR_ENABLE_IPV6] = value
+
```
agners, line 131: Since we expose the Docker IPv6 config in the `/docker/options` endpoint I'd also store the config in the Docker configuration file (`DockerConfig` class, extend `SCHEMA_DOCKER_CONFIG`).

supervisor/docker/network.py
```diff
@@ -42,55 +68,74 @@ def containers(self) -> list[str]:
...
+                    ):
+                        for container in containers:
+                            network.disconnect(container, force=True)
+                        network.remove()
+                    else:
+                        _LOGGER.warning(
+                            "System appears to be running, "
+                            "not applying Supervisor network change. "
+                            "Reboot your system to apply the change."
+                        )
```
agners, line 129: We need to return the found network interface in this case, otherwise the code attempts to create a network below.

```suggestion
                        )
                        return network
```

supervisor/docker/network.py
```diff
@@ -42,55 +68,74 @@ def containers(self) -> list[str]:
...
+            if network := self._get(DOCKER_NETWORK):
+                if network.attrs.get(DOCKER_ENABLEIPV6) == enable_ipv6:
+                    return network
+                _LOGGER.info(
+                    "Migrating Supervisor network to %s",
+                    "IPv4/IPv6 Dual-Stack" if enable_ipv6 else "IPv4-Only",
+                )
+                if containers := network.containers:
+                    if all(
+                        container.name in ("hassio_supervisor", "hassio_observer")
```
agners, line 118: There is a constant for the observer `OBSERVER_DOCKER_NAME`. I don't think there is one for Supervisor, but I guess we should introduce one.

```suggestion
                        container.name in ("hassio_supervisor", OBSERVER_DOCKER_NAME)
```

We must reattach the Observer after recreating the network. The Supervisor gets attached later anyways, but probably easier to just reattach both.

supervisor/docker/network.py
```diff
@@ -42,55 +68,74 @@ def containers(self) -> list[str]:
...
+    def _get(self, network_id: str) -> docker.models.networks.Network:
+        """Get a network by ID."""
+        return self.docker.networks.get(network_id)
 
-    def _get_network(self) -> docker.models.networks.Network:
+    def _create(self, **kwargs: Any) -> docker.models.networks.Network:
+        """Create a new network."""
+        return self.docker.networks.create(**kwargs)
+
+    def _get_network(self, enable_ipv6: bool = False) -> docker.models.networks.Network:
```
agners, line 106: This now all happens in the constructor, which is not ideal.

Let's introduce a `async def post_init` and call it from `DockerAPI`'s `post_init`. With that we can also run the API calls in a executor, since those are blocking calls.

agners (changes requested):
I've tested the detach part, it seems to work nicely. By default on startup it seems that Supervisor is not attached (the OS does not attach the container by default), but the Observer is running and attached. So with the detach logic we indeed can migrate the network on startup. :+1:

supervisor/api/docker.py
```diff
@@ -64,18 +98,3 @@ async def remove_registry(self, request: web.Request):
...
-        for hostname, registry in self.sys_docker.config.registries.items():
-            data_registries[hostname] = {
-                ATTR_USERNAME: registry[ATTR_USERNAME],
-            }
-        return {
-            ATTR_VERSION: self.sys_docker.info.version,
-            ATTR_STORAGE: self.sys_docker.info.storage,
-            ATTR_LOGGING: self.sys_docker.info.logging,
-            ATTR_REGISTRIES: data_registries,
-        }
```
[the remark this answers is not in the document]
davidrapan: Hah yeah, I have those tendencies that compels me to alphabetically (or how things are used/presented) order stuff.. 😆

supervisor/docker/network.py
davidrapan: I was looking for constants, but I started (and ended up with the supervisor) so I thought the observer proly didn't have any either. 😆

supervisor/docker/network.py
davidrapan: Oh yeah, sure thing, I forgot that one!

supervisor/docker/network.py
davidrapan: I'm not sure where to find the aliases and IPs for reattaching back... 🤔

supervisor/config.py
agners: I meant to really move these two properties in `DockerConfig`. The property would then be accessible using `self.sys_docker.enable_ipv6` (instead of `self.sys_config.enable_ipv6`).

supervisor/api/docker.py
```diff
@@ -30,10 +31,43 @@
...
+        body = await api_validate(SCHEMA_OPTIONS, request)
+
+        if ATTR_ENABLE_IPV6 in body:
+            self.sys_config.enable_ipv6 = body[ATTR_ENABLE_IPV6]
+
+        # Save changes before processing addons in case of errors
+        await self.sys_updater.save_data()
+        await self.sys_config.save_data()
+
+        await self.sys_resolution.evaluate.evaluate_system()
```
agners, line 69: We don't need all the logic from the Supervisor option call, as we don't change properties which might be relevant for updater/resolution system.

So this should boil down to:

```suggestion
            self.sys_docker.config.enable_ipv6 = body[ATTR_ENABLE_IPV6]

        await self.sys_docker.config.save_data()
```

agners (changes requested):
The migration logic worked just fine here :+1: .

I think with the config location adjusted, this is ready to be merged!

davidrapan:
Awesome! Thank you. 🎉

PR #5882, opened 2025-05-13
Fallback to threaded resolver in case AsyncResolver fails

agners:
## Proposed change
In case the c-ares based AsyncResolver fails to initialize, let's fallback to the threaded resolver. This would have helped for aiodns 3.3.0 issue when clients were unable to allocate more inotify watches.

This is fixed in aiodns 3.4.0, but we should still fallback to the threaded resolver as a precautionary measure.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability of network connectivity by adding a fallback mechanism for DNS resolution, ensuring continued operation if the primary resolver fails. Error occurrences are now properly logged.

### DIFF
--- a/supervisor/coresys.py
+++ b/supervisor/coresys.py
@@ -13,6 +13,7 @@ from types import MappingProxyType
 from typing import TYPE_CHECKING, Any, Self, TypeVar
 
 import aiohttp
+from pycares import AresError
 
 from .config import CoreConfig
 from .const import (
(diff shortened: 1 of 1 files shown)


supervisor/coresys.py
```diff
@@ -121,13 +122,17 @@ async def init_websession(self) -> None:
...
-        resolver = aiohttp.AsyncResolver()
+        try:
+            resolver = aiohttp.AsyncResolver(loop=self.loop)
+            # pylint: disable=protected-access
+            _LOGGER.debug(
+                "Initializing ClientSession with AsyncResolver. Using nameservers %s",
+                resolver._resolver.nameservers,
+            )
+        except AresError as err:
+            _LOGGER.exception("Unable to initialize async DNS resolver: %s", err)
```
sairon, line 133: We don't want to capture that, right?

sairon (approved):
LGTM, just one question.

supervisor/coresys.py
sairon: Or, with the aiodns updated to 3.4.0, we could, eventually removing it later? So we'd learn about other ways it can fail.

agners: The aiodns 3.4.0 update got already merged, so initializing should no longer fail in that particular case. But just in case there is some other situation where this happens, I'd like to add this fallback. But I'd also like to know about it.

sairon: But that was my point - would you learn about it with the current code? I thought only level `CRITICAL` gets logged to Sentry if it's handled (because we configure `LoggingIntegration(level=logging.INFO, event_level=logging.CRITICAL)`). `_LOGGER.exception` only creates `ERROR` log entry, which is lower priority.

agners: Uh right, I somehow thought that `exception` makes it to Sentry. I'll make a follow up PR!

PR #5892, opened 2025-05-15
Add IPv6 address generation mode & privacy extensions

davidrapan:
## Proposed change

Currently default `stable-privacy` is probably sufficient for some users, but for the rest of us, there should be an option to easily change this configuration therefore I'm introducing the changes in this PR.

Added are two new **opt-in** configuration items for IPv6:
- `addr_gen_mode`
  - with options: `eui64`, `stable-privacy`, `default-or-eui64` and `default`
- `ip6_privacy`
  - with three options: `enabled`, `enabled` but prefer public, `disabled` and `default`

More 'bout the options [here](https://www.networkmanager.dev/docs/api/1.50.2/settings-ipv6.html)

Configuration of
```
addr_gen_mode: eui64
ip6_privacy: enabled
```
combines the best of both "worlds" because Privacy Extensions also adds an additional temporary address that is used for all ongoing connections, just like in the "stable-privacy" generation mode, but still has a fixed and deterministic address for all incoming connections.

_I've been using this configuration myself since day 1 and it should also be the default in my opinion, but I understand that some may have a different one._

Edit: I just found out that there have been some attempts in the past to bring this configuration to HA:
* https://github.com/home-assistant/operating-system/issues/1687
  * https://github.com/home-assistant/supervisor/pull/4510
  * https://github.com/home-assistant/frontend/pull/17722

So hopefully this one will be successful, hah 😆 But I also think it's not necessary to expose this configuration in the frontend.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to documentation pull request: https://github.com/home-assistant/developers.home-assistant/pull/2670
- Link to cli pull request: https://github.com/home-assistant/cli/pull/566

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Added support for configuring IPv6 address generation mode and privacy settings through the network API.
  - IPv6 settings now include new parameters for address generation mode and privacy in configuration and display.

- **Bug Fixes**
  - Improved validation and serialization of IPv6 configuration options in network interface settings.

- **Tests**
  - Updated tests to verify the presence and correctness of new IPv6 configuration options.


davidrapan:
> How is address selection properly implement/what APIs allow to get the correct IP?

It is in RFC 6724 and is simply enforced by a temporary over public rule in **kernel address selection** or left to the application to bind() to a specific source address (binding to a public one with privacy extensions enabled would violate the RFC).

> Do you know how that plays out with e.g. mDNS specifically? How well is address selection in mDNS responder and similar implementations which use the IP address implemented?

Both are announced and the source address is a link-local.

> It uses ifaddr to get a list of interfaces and IP addresses from what I remember, but I don't think it does any address selection.

This is correct behavior. In cases like this, the kernel address selection is in charge.

I'm not sure about OpenThread Border as I've never used it, but I think it will be a similar case to mDNS. Service-specific announcements have very little to do with source address selection. It's all about where the packets are coming from, not what their content is. :) And that should be at least configurable? I guess.

BUT, that's why everything presented here is opt-in for those who know what's going on. HA never initiated any connections outside using anything other than a temporary address.

If you want me to do any specific tests, let me know. :)

agners:
> > Do you know how that plays out with e.g. mDNS specifically? How well is address selection in mDNS responder and similar implementations which use the IP address implemented?
> 
> Both are announced and the source address is a link-local.
> 

I am not really worried what the source address of the multicast packet is, it's largely irrelevant for the receiving system.

What is relevant is the payload of mDNS packets, which contains `A` and `AAAA` information of the host. What I wonder is if mDNS responder correctly handle such rotating addresses, or rather, how they should handle them correctly. From what I've seen in Python Zeroconf code, by default it announces all addresses returned by ifaddr of a particular interface. What happens if that address lifetime expires before the receiving system is using that address? Probably the mDNS announcement should align the record lifetime with the address. Or probably the address deprecation feature is useful for this, so mDNS should only announce non-deprecated address, and the deprecation period should at least be as long as the announced record cache TTL value. 

This really deviates a bit from that PR, but I try to understand the ramification of enabled privacy extensions for Home Assistant as a whole, and tap your expertise on this topic  :sweat_smile: 

> > It uses ifaddr to get a list of interfaces and IP addresses from what I remember, but I don't think it does any address selection.
> 
> This is correct behavior. In cases like this, the kernel address selection is in charge.

I am worried about the remote system here. From what I understand, there is no way for the remote system to tell which address is which.

> BUT, that's why everything presented here is opt-in for those who know what's going on. HA never initiated any connections outside using anything other than a temporary address.

Ok, yeah I didn't look at the code just yet, but I was assuming so. Can you clarify that in the PR description as well, so that it will be clear in the future what to expect from this change?

Also, as you wrote in the description, not adding this to the frontend make sense, at least for now.

davidrapan:
> This really deviates a bit from that PR, but I try to understand the ramification of enabled privacy extensions for Home Assistant as a whole, and tap your expertise on this topic

* `ip addr: valid_lft 271sec, preferred_lft 121sec`
* mDNS TTL: 120 seconds

> I am worried about the remote system here. From what I understand, there is no way for the remote system to tell which address is which.

In the case of mDNS? It doesn't matter.

> Ok, yeah I didn't look at the code just yet, but I was assuming so. Can you clarify that in the PR description as well, so that it will be clear in the future what to expect from this change?

Sure.

> Also, as you wrote in the description, not adding this to the frontend make sense, at least for now.

👍

davidrapan:
And I do rely heavily on the mDNS. I'm connecting almost everything within my network using `.local`. Using 3 VLANs, HA is the only one (not counting network elements) connected to all of them and router is running custom build of avahi mDNS repeater with hardcoded mac filter for HA :-D ...

agners:
> 
>     * `ip addr: valid_lft 271sec, preferred_lft 121sec`
> 
>     * mDNS TTL: 120 seconds
> 

Hm, ok I see https://datatracker.ietf.org/doc/html/rfc6762#section-10 says 120 seconds for the host records (including A and AAAA). It seems that non-host records have much higher values by default, it was probably what I had in mind. Python Zeroconf seems to also use the RFC values, so we should be fine.

> 
> > I am worried about the remote system here. From what I understand, there is no way for the remote system to tell which address is which.
> 
> In the case of mDNS? It doesn't matter.
> 

Ok, yeah with the above it doesn't matter because the TTL is lower than the lifetime of the IP address. As long as the client doesn't store/use the address longer than TTL (which probably should be seen as a bug), we should be fine.

Related: Do you happen to know how an open TCP connection behaves in face of this? :thinking:

davidrapan:
This temporary address is not generated that often (RFC says the default TEMP_PREFERRED_LIFETIME is 1 day). + I've never seen it used as a destination address, so I'd have to do some synthetic testing, but my best guess is a simple dc. :)

supervisor/dbus/network/setting/generate.py
```diff
@@ -93,10 +100,21 @@ def _get_ipv4_connection_settings(ipv4setting) -> dict:
...
+            if ipv6setting.addr_gen_mode == InterfaceAddrGenMode.EUI64:
+                ipv6[CONF_ATTR_IPV6_ADDR_GEN_MODE] = Variant("i", 0)
+            else:
+                ipv6[CONF_ATTR_IPV6_ADDR_GEN_MODE] = Variant("i", 1)
+            if ipv6setting.ip6_privacy == InterfaceIp6Privacy.DISABLED:
+                ipv6[CONF_ATTR_IPV6_PRIVACY] = Variant("i", 0)
+            elif ipv6setting.ip6_privacy == InterfaceIp6Privacy.ENABLED_PREFER_PUBLIC:
+                ipv6[CONF_ATTR_IPV6_PRIVACY] = Variant("i", 1)
+            else:
+                ipv6[CONF_ATTR_IPV6_PRIVACY] = Variant("i", 2)
```
agners, line 117: Hm, we decided to use `link-local` IPv6 mode when the user selects disabled, do we want these settings maybe also apply then? :thinking:

supervisor/dbus/const.py
```diff
@@ -210,6 +210,21 @@ class InterfaceMethod(StrEnum):
     LINK_LOCAL = "link-local"
 
 
+class InterfaceAddrGenMode(IntEnum):
+    """Interface addr_gen_mode."""
+
+    EUI64 = 0
+    STABLE_PRIVACY = 1
```
agners, line 217: Modern NetworkManager seem to know two more modes, default-or-eui64 and default.

See: https://www.networkmanager.dev/docs/api/1.50.2/settings-ipv6.html

In my test instance the existing profiles seem to have 3.

supervisor/api/network.py
```diff
@@ -212,25 +232,31 @@ async def interface_update(self, request: web.Request) -> None:
...
-                interface.ipv6setting = IpSetting(
-                    config.get(ATTR_METHOD, InterfaceMethod.STATIC),
-                    config.get(ATTR_ADDRESS, []),
-                    config.get(ATTR_GATEWAY),
-                    config.get(ATTR_NAMESERVERS, []),
+                interface.ipv6setting = Ip6Setting(
+                    method=config.get(ATTR_METHOD, InterfaceMethod.STATIC),
+                    addr_gen_mode=config.get(
+                        ATTR_ADDR_GEN_MODE, InterfaceAddrGenMode.STABLE_PRIVACY
+                    ),
```
agners, line 245: I think it would be better if we use the new default value mentioned before here.

supervisor/dbus/network/setting/generate.py
```diff
@@ -93,10 +100,21 @@ def _get_ipv4_connection_settings(ipv4setting) -> dict:
...
-def _get_ipv6_connection_settings(ipv6setting) -> dict:
+def _get_ipv6_connection_settings(ipv6setting: Ip6Setting | None) -> dict:
     ipv6 = {}
     if not ipv6setting or ipv6setting.method == InterfaceMethod.AUTO:
         ipv6[CONF_ATTR_IPV6_METHOD] = Variant("s", "auto")
+        if ipv6setting:
+            if ipv6setting.addr_gen_mode == InterfaceAddrGenMode.EUI64:
+                ipv6[CONF_ATTR_IPV6_ADDR_GEN_MODE] = Variant("i", 0)
+            else:
+                ipv6[CONF_ATTR_IPV6_ADDR_GEN_MODE] = Variant("i", 1)
```
agners, line 111: And then handle the new options here too.

Now there is a problem with backwards compatibility: In general we try to support old versions of NetworkManager. So we should be careful to only write values which are valid. Since this is a new feature, I don't care much if the new setting `default-or-eui64` is used. But I think it isn't safe to use `default` (3) here since it will break saving network settings if someone has a older OS version still :cry: 

What we could do is just not set the attribute, I think that makes NetworkManager to use the default, so essentially just handle settings 0 through 2, and don't set anything if 3 :thinking:

supervisor/dbus/const.py
```diff
@@ -210,6 +210,21 @@ class InterfaceMethod(StrEnum):
...
+
+
+class InterfaceIp6Privacy(IntEnum):
+    """Interface ip6_privacy."""
+
```
agners, line 222: There is also -1 for default.

agners (changes requested):

supervisor/dbus/const.py
davidrapan: Oh, of course! I just wanted to skip them (and forgot to up the counter accordingly xD) as I don't like these nondeterministic "states" but I'll add them all if you think we need them.

supervisor/api/network.py
davidrapan: I'm bit confused now, I thought stable-privacy should be default?

supervisor/dbus/network/setting/generate.py
davidrapan: `DEFAULT_OR_EUI64` it is then.

supervisor/api/network.py
agners: The above linked NetworkManager docs say:

> The special values "default" and "default-or-eui64" will fallback to the global connection default as documented in the NetworkManager.conf(5) manual

And in NetworkManager.conf(5) says:

> ipv6.ip6-privacy
> If ipv6.ip6-privacy is unset, use the content of "/proc/sys/net/ipv6/conf/default/use_tempaddr" as last fallback.

Which seems to be 0 on HAOS.

supervisor/dbus/network/setting/generate.py
agners: What do you mean? The absence of the property means default (3), from the linked page:

> Previously it used to default to "stable-privacy". On D-Bus, the absence of an addr-gen-mode setting equals "default". 

 :thinking:

supervisor/api/network.py
davidrapan: Yup that's sysctl. 👍

supervisor/dbus/network/setting/generate.py
davidrapan: That if-elif set will end w/ this value and DEFAULT won't be handled.

supervisor/host/configuration.py
```diff
@@ -46,6 +55,14 @@ class IpSetting:
     nameservers: list[IPv4Address | IPv6Address]
 
 
+@dataclass(slots=True)
+class Ip6Setting(IpSetting):
+    """Represent a user IPv6 setting."""
+
+    addr_gen_mode: InterfaceAddrGenMode = InterfaceAddrGenMode.DEFAULT_OR_EUI64
+    ip6_privacy: InterfaceIp6Privacy = InterfaceIp6Privacy.DISABLED
```
agners, line 63: With this, if IPv6 is not set we actually set explicit `addr-gen-mode` and `ip6-privacy` via D-Bus.

To make the system behave as it is currently, we should choose defaults which don't set those modes, so:

```suggestion
    addr_gen_mode: InterfaceAddrGenMode = InterfaceAddrGenMode.DEFAULT
    ip6_privacy: InterfaceIp6Privacy = InterfaceIp6Privacy.DEFAULT
```

With that change, the two modes don't get written via D-Bus, making sure that if things stay the way they are today if not explicitly set.

supervisor/api/network.py
```diff
@@ -212,25 +232,31 @@ async def interface_update(self, request: web.Request) -> None:
...
             elif key == ATTR_IPV6:
-                interface.ipv6setting = IpSetting(
-                    config.get(ATTR_METHOD, InterfaceMethod.STATIC),
-                    config.get(ATTR_ADDRESS, []),
-                    config.get(ATTR_GATEWAY),
-                    config.get(ATTR_NAMESERVERS, []),
+                interface.ipv6setting = Ip6Setting(
+                    method=config.get(ATTR_METHOD, InterfaceMethod.STATIC),
+                    addr_gen_mode=config.get(
+                        ATTR_ADDR_GEN_MODE, InterfaceAddrGenMode.DEFAULT_OR_EUI64
```
agners, line 244: Same here:

```suggestion
                        ATTR_ADDR_GEN_MODE, InterfaceAddrGenMode.DEFAULT
```

I am not sure if this is really relevant in practice as I think the real D-Bus API will always return the default if `addr-gen-mode` never got set. But this makes test work as we expect things.

supervisor/api/network.py
```diff
@@ -277,19 +303,25 @@ async def create_vlan(self, request: web.Request) -> None:
...
         if ATTR_IPV6 in body:
-            ipv6_setting = IpSetting(
-                body[ATTR_IPV6].get(ATTR_METHOD, InterfaceMethod.AUTO),
-                body[ATTR_IPV6].get(ATTR_ADDRESS, []),
-                body[ATTR_IPV6].get(ATTR_GATEWAY, None),
-                body[ATTR_IPV6].get(ATTR_NAMESERVERS, []),
+            ipv6_setting = Ip6Setting(
+                method=body[ATTR_IPV6].get(ATTR_METHOD, InterfaceMethod.AUTO),
+                addr_gen_mode=body[ATTR_IPV6].get(
+                    ATTR_ADDR_GEN_MODE, InterfaceAddrGenMode.DEFAULT_OR_EUI64
```
agners, line 317: and here..

```suggestion
                    ATTR_ADDR_GEN_MODE, InterfaceAddrGenMode.DEFAULT
```

Unless you have good reason to use a deviate from the NetworkManager default for 3 (see https://www.networkmanager.dev/docs/api/1.50.2/settings-ipv6.html) :thinking: .

agners (changes requested):
With the suggested changes applied the tests then also need some adjustments. Other than those defaults, LGTM!

supervisor/host/configuration.py
davidrapan: But didn't you said it's dangerous to set addr_gen_mode to 3 ? In terms of backward compatibility?

supervisor/api/network.py
davidrapan: Yeah I though that it could possibly make a mess on an older OS.

supervisor/host/configuration.py
davidrapan: Hm, true, just didn't think that this slight misalignment between the two configs is acceptable. But I guess it ONLY applies fot those older OS's so it's most likely fine. In terms of clarity. 😉

supervisor/dbus/network/setting/generate.py
```diff
@@ -93,10 +100,25 @@ def _get_ipv4_connection_settings(ipv4setting) -> dict:
...
     ipv6 = {}
     if not ipv6setting or ipv6setting.method == InterfaceMethod.AUTO:
         ipv6[CONF_ATTR_IPV6_METHOD] = Variant("s", "auto")
+        if ipv6setting:
+            if ipv6setting.addr_gen_mode == InterfaceAddrGenMode.EUI64:
+                ipv6[CONF_ATTR_IPV6_ADDR_GEN_MODE] = Variant("i", 0)
+            elif ipv6setting.addr_gen_mode == InterfaceAddrGenMode.STABLE_PRIVACY:
+                ipv6[CONF_ATTR_IPV6_ADDR_GEN_MODE] = Variant("i", 1)
+            elif ipv6setting.addr_gen_mode == InterfaceAddrGenMode.DEFAULT_OR_EUI64:
+                ipv6[CONF_ATTR_IPV6_ADDR_GEN_MODE] = Variant("i", 2)
```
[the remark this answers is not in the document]
davidrapan: Great idea! I like it.

agners:
There are some ruff issues still.

Ideally also rebase, I merged two PRs. I think then we are good!

davidrapan:
Anyway, builds still fails
```
Load key ".ssh/id_rsa": error in libcrypto
```
for some reason since the last bump of crypto. :/

When I revert the bump, everything works.

agners:
How do you build Supervisor?

agners:
> When I revert the bump, everything works.

Ah I see, you meant in CI. That is because the wheels bump gets triggered. I wonder though why that is the case, the necessary wheels should be present 🤔

agners:
Yeah our infrastructure builds wheels for musllinux automatically. However, that wheels build should only trigger if a PR  touches `requirements.txt` or `build.yaml`. Wheels build do fail for PRs from forks currently because CI can't reed the necessary keys to upload wheels to our servers (for security reasons).

Your PR should not touch those files, so no wheels build should be triggered. :thinking: 

It seems to work now, so I think it's fine.

davidrapan:
Yeah I think I get it:
```
if: needs.init.outputs.requirements == 'true'
```
this evaluates to True and should. 😉

Yes! It should be all good and polished now!

davidrapan:
I don't understand that AwesomeVersion thing. When it's the class it complains it's not valid... but it doesn't like string either...

agners:
> Edit: Ah, now it's a mocking issue, or even better it's a mocking issue all along. :-D

Yes, and we have fixture for that. Just updated the test, should be fine now.

davidrapan:
Great, thanks!

agners:
Added another test to validate `addr_gen_mode` is correctly set. And actually realized that 1.40.0 already knows value default value 3, so I've fixed that.

agners (approved):
LGTM, thanks!

davidrapan:
Awesome! 🎉

PR #5901, opened 2025-05-20
Avoid initializing Blockbuster on Supervisor info call

agners:
## Proposed change
Instead of creating an instance of Blockbuster to simply check if Bluckbuster is enabled, use a global variable to store the instance of Blockbuster and only initialize it when needed. This avoids unnecessary initialization of Blockbuster when it is not required.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

## Summary by CodeRabbit

- **Refactor**
  - Improved internal management of the BlockBuster feature with a new streamlined activation system. No changes to user-facing functionality.

### DIFF
--- a/supervisor/utils/blockbuster.py
+++ b/supervisor/utils/blockbuster.py
@@ -2,6 +2,40 @@
 
 import logging
 
+from blockbuster import BlockBuster as _BlockBuster
+
+_LOGGER: logging.Logger = logging.getLogger(__name__)
+
+
+class BlockBuster:
+    _instance: _BlockBuster | None = None
+
+    @classmethod
+    def is_enabled(cls):
+        """Return true if blockbuster detection is enabled."""
+        if cls._instance is None:
+            return False
+        for _, fn in cls._instance.functions.items():
+            return fn.activated
+        return False
+
+    @classmethod
+    def activate(cls):
+        """Activate blockbuster detection."""
+        _LOGGER.info("Activating BlockBuster blocking I/O detection")
+        if cls._instance is None:
+            cls._instance = _BlockBuster()
+        cls._instance.activate()
+
+    @classmethod
+    def deactivate(cls):
+        """Deactivate blockbuster detection."""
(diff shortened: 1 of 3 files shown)


supervisor/utils/blockbuster.py
```diff
@@ -1,24 +1,28 @@
 """Activate and deactivate blockbuster for finding blocking I/O."""
```
sairon, line 1: GH doesn't let me apply suggestion over the whole file, but this should be the whole module content:
```suggestion
"""Activate and deactivate blockbuster for finding blocking I/O."""

import logging

from blockbuster import BlockBuster as _BlockBuster

_LOGGER: logging.Logger = logging.getLogger(__name__)


class BlockBuster:
    _instance: _BlockBuster | None = None

    @classmethod
    def is_enabled(cls):
        """Return true if blockbuster detection is enabled."""
        if cls._instance is None:
            return False
        for _, fn in cls._instance.functions.items():
            return fn.activated
        return False

    @classmethod
    def activate(cls):
        """Activate blockbuster detection."""
        _LOGGER.info("Activating BlockBuster blocking I/O detection")
        if cls._instance is None:
            cls._instance = _BlockBuster()
        cls._instance.activate()

    @classmethod
    def deactivate(cls):
        """Deactivate blockbuster detection."""
        _LOGGER.info("Deactivating BlockBuster blocking I/O detection")
        if cls._instance:
            cls._instance.deactivate()
```

sairon (commented):
Instead of the ugly global and lint suppressions, what about turning it into a kind of singleton class and use class methods to access it? It would also require refactoring of the usages (unless we implement complementary methods that will use the class in this module, which I find a bit unnecessary).

supervisor/utils/blockbuster.py
agners: Yeah seems cleaner. I'll rename it to `BlockBusterManager` to avoid the naming issue.

sairon (approved):
This indeed looks cleaner, thanks!

PR #5906, opened 2025-05-22
Use s6-overlay read-only mode by default

agners:
## Proposed change
To avoid accidential writes to the Supervisor root filesystem, we might use the Docker read-only mode at one point. This is not yet the default, but using s6-overlay with the read-only flag seems not to have any downsides. So enable this by default.

To start Supervisor with read-only root file system teh following arguments have to be used: `--read-only --tmpfs /run:exec`.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

## Summary by CodeRabbit

- **Chores**
	- Updated container configuration to support a read-only root filesystem for enhanced security.


sairon (approved):
I didn't know what it does at first but reading the docs, it should be indeed harmless:

> S6_READ_ONLY_ROOT (default = 0): When running in a container whose root filesystem is read-only, set this env to 1 to inform init stage 2 that it should copy user-provided initialization scripts from /etc to /run/s6/etc before it attempts to change permissions, etc. See [Read-Only Root Filesystem](https://github.com/just-containers/s6-overlay/tree/d6f37bb2052cfc5705154315419dcacb3975f11e#read-only-root-filesystem) for more information.

agners:
I don't see anything in `/run/s6/etc` afterwards, so I guess we don't have a "user-provided initialization script" which needs that treatment, or the docs are outdated/things get copied somewhere else. Whatever it might be, it seems to not hurt, so :shrug: .

agners:
> Maybe it would have better to actually set this flag via the Docker run command once we actually set it read-only 🤔

On the other hand, s6 would still do that remount attempt I guess, leading to the AppArmor deny event. Maybe a `S6_RUN_IS_FINE_I_KNOW_WHAT_I_AM_DOING` would be better? :thinking: :sweat_smile:

PR #5942, opened 2025-06-12
Fix mypy issues in `misc`, `mounts` and `os` modules

mdegat01:
## Proposed change

As it says on the tin

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved error handling and validation in OS management and D-Bus connection features.
  - Enhanced type safety and consistency across mount and backup management.

- **Refactor**
  - Standardized dictionary key usage and type annotations for mounts.
  - Updated method signatures and return types for better clarity and reliability.
  - Made systemd unit state updates more explicit by passing unit objects.

- **New Features**
  - Added support for an optional mount version field.

- **Style**
  - Improved code clarity with explicit type casting and annotations.
  - Broadened exception handling to include all BaseException types.

### DIFF
--- a/supervisor/mounts/manager.py
+++ b/supervisor/mounts/manager.py
@@ -5,7 +5,7 @@ from collections.abc import Awaitable
 from dataclasses import dataclass
 import logging
 from pathlib import PurePath
-from typing import Self
+from typing import Self, cast
 
 from attr import evolve
 
@@ -56,7 +56,7 @@ class MountManager(FileConfiguration, CoreSysAttributes):
     async def load_config(self) -> Self:
         """Load config in executor."""
         await super().load_config()
-        self._mounts: dict[str, Mount] = {
+        self._mounts = {
             mount[ATTR_NAME]: Mount.from_dict(self.coresys, mount)
             for mount in self._data[ATTR_MOUNTS]
         }
@@ -177,7 +177,7 @@ class MountManager(FileConfiguration, CoreSysAttributes):
             if mounts[i].failed_issue in self.sys_resolution.issues:
                 continue
             if not isinstance(errors[i], MountError):
-                await async_capture_exception(errors[i])
+                await async_capture_exception(cast(Exception, errors[i]))
 
             self.sys_resolution.add_issue(
                 evolve(mounts[i].failed_issue),
@@ -219,7 +219,7 @@ class MountManager(FileConfiguration, CoreSysAttributes):
         conditions=[JobCondition.MOUNT_AVAILABLE],
         on_condition=MountJobError,
     )
-    async def remove_mount(self, name: str, *, retain_entry: bool = False) -> None:
+    async def remove_mount(self, name: str, *, retain_entry: bool = False) -> Mount:
         """Remove a mount."""
         # Add mount name to job
         self.sys_jobs.current.reference = name
--- a/supervisor/mounts/manager.py
+++ b/supervisor/mounts/manager.py
@@ -5,7 +5,7 @@ from collections.abc import Awaitable
 from dataclasses import dataclass
 import logging
 from pathlib import PurePath
-from typing import Self, cast
+from typing import Self
 
 from attr import evolve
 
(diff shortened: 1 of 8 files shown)


supervisor/os/manager.py
```diff
@@ -61,20 +62,22 @@ def from_dict(cls, data: SlotStatusDataType) -> "SlotStatus":
...
             status=data.get("status"),
-            activated_count=data.get("activated.count"),
+            activated_count=cast(int | None, data.get("activated.count")),
             activated_timestamp=datetime.fromisoformat(data["activated.timestamp"])
             if "activated.timestamp" in data
             else None,
-            boot_status=data.get("boot-status"),
+            boot_status=RaucState(data["boot-status"])
+            if "boot-status" in data
+            else None,
```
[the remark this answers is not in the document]
mdegat01: As these are all frozen dataclasses the intent was to avoid runtime conversion and have type-checker only hints

supervisor/os/data_disk.py
```diff
@@ -166,12 +167,16 @@ def available_disks(self) -> list[Disk]:
...
+            CheckMultipleDataDisks, self.sys_resolution.check.get("multiple_data_disks")
+        )
 
     @property
     def check_disabled_data_disk(self) -> CheckDisabledDataDisk:
         """Resolution center check for disabled data disk."""
-        return self.sys_resolution.check.get("disabled_data_disk")
+        return cast(
+            CheckDisabledDataDisk, self.sys_resolution.check.get("disabled_data_disk")
+        )
```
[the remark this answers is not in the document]
mdegat01: `None` is never returned by that get, it raises on miss

supervisor/mounts/manager.py
```diff
@@ -177,7 +177,7 @@ async def _mount_errors_to_issues(
             if mounts[i].failed_issue in self.sys_resolution.issues:
                 continue
             if not isinstance(errors[i], MountError):
-                await async_capture_exception(errors[i])
+                await async_capture_exception(cast(Exception, errors[i]))
```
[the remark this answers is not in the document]
mdegat01: Sentry's `capture_exception` actually accepts `BaseException`, corrected the typing of our wrapper 👍

supervisor/mounts/mount.py
```diff
@@ -196,8 +181,11 @@ async def load(self) -> None:
         if not await self.is_mounted():
             await self.reload()
 
-    async def _update_state(self) -> UnitActiveState | None:
+    async def _update_state(self) -> None:
         """Update mount unit state."""
+        if not self.unit:
```
agners: We should add a log here, debug level if expected, warning or critical level if this is unexpected.

supervisor/os/manager.py
```diff
@@ -93,17 +96,17 @@ def to_dict(self) -> SlotStatusDataType:
         if self.sha256 is not None:
             out["sha256"] = self.sha256
         if self.size is not None:
-            out["size"] = self.size
+            out["size"] = cast(c_uint64, self.size)
```
agners: Are you sure it wouldn't be better to actually runtime convert this values here? If we don't convert them, does D-Bus next maybe do those later on or how does that work? If it anyway needs to be converted at runtime to c_uint64 later on, I think it would be better to do it here.

agners (changes requested):

supervisor/os/manager.py
mdegat01: In all current use cases they already are. The only way we get an instance of this object is by running `from_dict` on an instance of this:
https://github.com/home-assistant/supervisor/blob/bdbd09733a7744094499d7509cc3ed6c69633287/supervisor/dbus/rauc.py#L28-L48

And then this one is a frozen dataclass so it cannot be modified after creation. So in reality they are always the correct type.

Now if we ever had a reason to make an instance of this dataclass some other way then yea, that would be a concern. Currently it would only be overhead. Is it worth it "just in case"? I don't know when we would ever make an instance of this manually 🤔

supervisor/mounts/mount.py
mdegat01: I can add a debug log if we want. This is really just type check stuff as it can't actually occur. All the places that call this private method first attempt to update the unit info and fail out if they can't.

supervisor/os/manager.py
mdegat01: This method isn't used. Removing it instead to simplify things

agners (approved):

PR #5946, opened 2025-06-12
Fix mypy issues in `plugins` and `resolution`

mdegat01:
## Proposed change

As it says on the tin. Only two modules because `resolution` had a few more then expected.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Added a dedicated method for retrieving locally installed add-ons.
  - Introduced an abstract method for stopping system plugins, requiring implementation in subclasses.

- **Bug Fixes**
  - Improved handling of missing or invalid references in various operations to prevent errors.
  - Enhanced checks for disk and add-on status to avoid processing when data is incomplete or invalid.
  - Refined access to plugin tokens for stricter type enforcement.

- **Refactor**
  - Updated several internal methods to use the new local add-on retrieval approach.
  - Improved encapsulation and error handling for DNS plugin templates.
  - Clarified type annotations and return types for evaluation and resolution methods.
  - Streamlined iteration and error handling in DNS server checks.

- **Tests**
  - Added and updated fixtures to ensure plugin tokens are available during test execution.
  - Adjusted and removed certain test assertions to align with updated error handling and logging behavior.

- **Documentation**
  - Updated type annotations and method signatures for better clarity and consistency.

### DIFF
--- a/supervisor/resolution/evaluations/lxc.py
+++ b/supervisor/resolution/evaluations/lxc.py
@@ -35,7 +35,7 @@ class EvaluateLxc(EvaluateBase):
     async def evaluate(self) -> bool:
         """Run evaluation."""
 
-        def check_lxc():
+        def check_lxc() -> bool:
             with suppress(OSError):
                 if "container=lxc" in Path("/proc/1/environ").read_text(
                     encoding="utf-8"
--- a/supervisor/resolution/evaluations/systemd.py
+++ b/supervisor/resolution/evaluations/systemd.py
@@ -6,6 +6,14 @@ from ...host.const import HostFeature
 from ..const import UnsupportedReason
 from .base import EvaluateBase
 
+FEATURES_REQUIRED: tuple[HostFeature, ...] = (
+    HostFeature.HOSTNAME,
+    HostFeature.SERVICES,
+    HostFeature.SHUTDOWN,
+    HostFeature.REBOOT,
+    HostFeature.TIMEDATE,
+)
+
 
 def setup(coresys: CoreSys) -> EvaluateBase:
     """Initialize evaluation-setup function."""
@@ -33,12 +41,5 @@ class EvaluateSystemd(EvaluateBase):
     async def evaluate(self) -> bool:
         """Run evaluation."""
         return any(
-            feature not in self.sys_host.features
-            for feature in (
-                HostFeature.HOSTNAME,
-                HostFeature.SERVICES,
-                HostFeature.SHUTDOWN,
-                HostFeature.REBOOT,
-                HostFeature.TIMEDATE,
-            )
+            feature not in self.sys_host.features for feature in FEATURES_REQUIRED
         )


mdegat01:
@agners as an added bonus, this PR also includes fixing issues in `security` and `services` (because there are none  💯 )

supervisor/resolution/evaluations/dbus.py
```diff
@@ -29,6 +29,6 @@ def states(self) -> list[CoreState]:
         """Return a list of valid states when this evaluation can run."""
         return [CoreState.INITIALIZE]
 
-    async def evaluate(self) -> None:
+    async def evaluate(self) -> bool:
         """Run evaluation."""
-        return not SOCKET_DBUS.exists()
+        return not await self.sys_run_in_executor(SOCKET_DBUS.exists)
```
mdegat01, line 34: This is technically unrelated to the PR but it's so small I put it in. Looks like I/O in the event loop we missed. Or are paths in `/run` different and this was intentionally omitted?

agners: > Or are paths in `/run` different and this was intentionally omitted?

They are probably on a tmpfs in ram, so most likely won't actually block on IO. But it could lead to memory reclaim operations or similar. I don't think it should be considered non-blocking, so running in an executor is the right thing to do :+1:

supervisor/plugins/observer.py
```diff
@@ -90,6 +91,10 @@ async def start(self) -> None:
             _LOGGER.error("Can't start observer plugin")
             raise ObserverError() from err
 
+    async def stop(self) -> None:
```
agners, line 94: Is this new functionality? :thinking:

supervisor/addons/manager.py
```diff
@@ -67,6 +67,10 @@ def get(self, addon_slug: str, local_only: bool = False) -> AnyAddon | None:
             return self.store.get(addon_slug)
         return None
 
+    def get_local_only(self, addon_slug: str) -> Addon | None:
+        """Return an installed add-on from slug."""
```
agners, line 71: Maybe we should replace "local" with "installed" at one point, since we also have add-ons which are in a local "store". But in any case, such a change should not be part of this PR.

supervisor/plugins/observer.py
```diff
@@ -90,6 +91,10 @@ async def start(self) -> None:
...
+        """Raise. Supervisor should not stop observer."""
+        raise RuntimeError("Stopping observer without a restart is not supported!")
+
```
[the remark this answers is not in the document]
mdegat01: This is intended to be a dramatic failure. It should never be possible for code to call this as it is not implemented for this plugin.

supervisor/plugins/observer.py
mdegat01: Not exactly. It was in every plugin except observer because Supervisor cannot stop observer, it can only restart it. But we were using it like it was in the `BasePlugin` class elsewhere. So I refactored to put it in `BasePlugin` and made the observer override raise a `RuntimeError` saying it was not implemented in that particular plugin.

supervisor/addons/manager.py
mdegat01: yea I actually wrote `get_installed_only` and then changed it back because it seemed like that was addressing a different problem.

supervisor/plugins/observer.py
mdegat01: https://github.com/home-assistant/supervisor/blob/9ef2352d1217cda1a8c20cfbbac7a789308325a9/supervisor/plugins/manager.py#L125-L135

Here is where we were using it like it was in `BasePlugin` btw

agners: Ok, yeah make sense then!

supervisor/addons/manager.py
agners: Yeah if we rename, we should also change local to installed in other places. Not part of this PR.

supervisor/plugins/observer.py
```diff
@@ -60,7 +61,7 @@ def latest_version(self) -> AwesomeVersion | None:
     @property
     def supervisor_token(self) -> str:
         """Return an access token for the Observer API."""
-        return self._data.get(ATTR_ACCESS_TOKEN)
+        return cast(str, self._data[ATTR_ACCESS_TOKEN])
```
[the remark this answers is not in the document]
mdegat01: This is true. But the existing code also made this same assumption and would simply fail much more silently if we got `None` back. I decided to prefer a louder, clearer failure if this case actually occurs so we can fix it. I don't think it does though because we have no issue reports suggesting it (and it would certainly have failed).

PR #5957, opened 2025-06-17
Fix mypy issues in `store`, `utils` and all source files not in a module

mdegat01:
## Proposed change

Last one! Fixes all remaining mypy issues in all source files excluding tests. After merge we can turn on mypy in CI for source excluding tests in a follow-up. 

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist


supervisor/utils/systemd_journal.py
```diff
@@ -107,17 +107,17 @@ async def journal_logs_reader(
                 # followed by a newline as separator to the next field.
                 if not data.endswith(b"\n"):
                     raise MalformedBinaryEntryError(
-                        f"Failed parsing binary entry {data}"
+                        f"Failed parsing binary entry {data.decode('utf-8')}"
```
agners: Hm, since this is binary, maybe we should not try to decode but print as hex? Or at least be forgiving in decoding the data:

```suggestion
                        f"Failed parsing binary entry {data.decode('utf-8', errors="replace")}"
```

@sairon any thoughts on this? Are binary data often strings? :thinking:

supervisor/store/repository.py
```diff
@@ -139,7 +140,7 @@ async def update(self) -> bool:
 
     async def remove(self) -> None:
         """Remove add-on repository."""
-        if self.type != StoreType.GIT:
+        if not self.git or self.type == StoreType.CORE:
             raise StoreError("Can't remove built-in repositories!", _LOGGER.error)
 
-        await self.git.remove()
+        await cast(GitRepoCustom, self.git).remove()
```
agners, line 146: We probably should get away with `StoreType` and instead use actual `Repository` classes handling the individual repository types (with a base class for the shared logic).

supervisor/arch.py
```diff
@@ -40,7 +40,7 @@ def default(self) -> str:
     @property
     def supervisor(self) -> str:
         """Return supervisor arch."""
-        return self.sys_supervisor.arch
+        return self.sys_supervisor.arch or self._default_arch
```
agners, line 43: Hm, `_default_arch` is not initialized in `__init__` so technically it could be None as well :thinking: 

What is weird here is `_default_arch` is potentially initialized from `supervisor` property, in case `detect_cpu()` doesn't find an arch. So this ends up circular, potentially :see_no_evil: 

We could break this by making detect CPU fall back differently, e.g.:

```python
    def detect_cpu(self) -> str:
        """Return the arch type of local CPU."""
        cpu = platform.machine()
        for check, value in MAP_CPU.items():
            if cpu.startswith(check):
                return value
        if self.sys_supervisor.arch:
            _LOGGER.warning("Unknown CPU architecture %s, falling back to Supervisor architecture.", cpu)
            return self.sys_supervisor.arch
        _LOGGER.warning("Unknown CPU architecture %s, assuming CPU architecture equals Supervisor architecture.", cpu)
        return cpu
```

supervisor/utils/systemd_journal.py
mdegat01: I mostly just figured if the decode didn't work it would be unintelligible and useless as a log anyway. But I do like the forgiving decode

supervisor/store/repository.py
mdegat01: We have this already. `GitRepo` is base and subclasses are `GitRepoHassIO` and `GitRepoCustom`. The problem is `GitRepoHassIO` does not support `remove`.

This is the same situation as `stop` in `BasePlugin` in the last PR. All subclasses except for 1 support the method so the implementation path we choose was only implement it in the subclasses that supported it so the one that didn't simply didn't have that method. But that meant anytime you're acting on a variable which could hold any plugin you can't call `stop`.

For plugins I choose to make `stop` an `abstractmethod` in `BasePlugin` and have the override in observer simply raise `RuntimeError`. This was because there were 6 subclasses (with potential for more in the future) so it seemed easier and more maintainable that way there. As there were only two subclasses this time and not much possibility for more in future, a one-off `cast` seemed easier and more maintainable here.

I'm fine with doing the abstract and raise method if we want though. But I think you had planned a rework of all this anyway so might end up being wasted effort.

supervisor/arch.py
mdegat01: It can't be `None`. If we attempt to read it before it has been set it will raise an error for uninitialized variable but it will not be `None`. We declare it but do not set it in `__init__` in this case. It's unusual since we do `str | None = None` almost everywhere else but rather convenient for typing purposes this time.

I like your alternative fallback though. You're right, I accidentally made a loop 🙈

supervisor/store/repository.py
agners: Yeah let's tackle this with the git store repository handling rework.

PR #5974, opened 2025-06-25
Use Docker BuildKit to build addons

felipecrs:
## Proposed change

Use Docker BuildKit to build addons by delegating the build to a temporary container.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Closes https://github.com/orgs/home-assistant/discussions/77

Tests made manually:

- [x] Build add-ons normally
- [x] When squash is set, ignore it and log a warning
- [x] Failed build should display logs in UI

https://github.com/user-attachments/assets/f5289d52-fa93-46ca-83ee-67a375247021

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Improved add-on build process to use a dedicated Docker CLI container, enhancing reliability and error reporting.
  * Added support for explicit platform and Dockerfile selection during add-on builds.
  * Enhanced detection of known container images for better evaluation and troubleshooting.

* **Bug Fixes**
  * Improved error handling and logging during add-on builds.

* **Tests**
  * Updated and expanded tests to reflect changes in the build process and command execution.
  * Added utility functions to assist with test assertions.
  * Enhanced test fixtures for more accurate version handling.

### DIFF
--- a/supervisor/addons/build.py
+++ b/supervisor/addons/build.py
@@ -130,6 +130,7 @@ class AddonBuild(FileConfiguration, CoreSysAttributes):
 
         build_cmd = [
             "docker",
+            "buildx",
             "build",
             ".",
             "--tag",


felipecrs:
@coderabbitai this nitpick you mentioned is not correct. Ruff formatter actually prefers the other way.

felipecrs:
@agners I'm done with the changes. This PR is ready for review if you are ok with the approach.

I can also write a Home Assistant developer blog post for announcing this as well.

felipecrs:
I also attached a testing demonstration to the PR description.

mdegat01:
Can we add the image here:
https://github.com/home-assistant/supervisor/blob/8378d434d489d6960c186dbb0df1ba702548e131/supervisor/resolution/evaluations/container.py#L59-L66

That way if the check happens to run while a build is in progress and sees the container it won't mark the system as unsupported. Looks good though, really like the idea.

agners:
> Additionally, while I was on it, I resolved a few other nitpicks:
> 
> 1. `docker.run_command` only works with `Detach=True`, but was not including it in the default `container.run` args (i.e. was broken by default)
> 2. `docker_home_assistant_execute_command` was executing using `homeassistant:latest` image because `version` arg was actually not being honored by `docker.run_command`.
> 3. All `container.remove()` calls now include `v=True`. This will also delete whatever anonymous (and ephemeral) volumes that were created by that container like how `docker run --rm` does.
>    
>    * This avoids potential leftovers in the system depending on whether the image being executed had `VOLUME` instructions in its Dockerfile.

Nice finds :star_struck: ! Can you make this drive-by changes separate PR(s)?

We'll likely have to iterate and test this PR a bit before it can go in. But these changes are rather small and can go in quickly. Also, it is better if these fixes are separate commits just in case we run into problems (bisect)/or have to revert.

felipecrs:
Absolutely, I'll work on all the requests. Thank you both.

supervisor/addons/build.py
```diff
@@ -121,39 +121,65 @@ def build_is_valid() -> bool:
...
+            **self.additional_args,
+        }
 
-        return args
+        for key, value in build_args.items():
+            build_cmd.extend(["--build-arg", f"{key}={value}"])
+
+        # The build path must be the addon path in the host filesystem
+        build_path = str(self.addon.path_location).removeprefix("/data/")
+        build_path = f"/mnt/supervisor/{build_path}"
```
felipecrs: This section can break if these paths ever change.

This is my main concern with this PR.

I am working on a better solution.

supervisor/addons/build.py
```diff
@@ -121,39 +121,65 @@ def build_is_valid() -> bool:
...
+
+        return {
+            "command": build_cmd,
+            "volumes": {
+                "/var/run/docker.sock": {"bind": "/var/run/docker.sock", "mode": "rw"},
```
felipecrs: I should investigate if `/var/run/docker.sock` in host can be automatically inferred with `docker-py` for better portability. CLI reference:

```console
$ docker context inspect | jq -r '.[0].Endpoints.docker.Host' | sed "s^unix://^^"
/var/run/docker.sock
```

felipecrs: I did come up with a solution:

```python
def _get_host_socket_path(self) -> str:
        """Get the Docker socket path on the host.

        This is equivalent to:
        $ docker context inspect | jq -r '.[0].Endpoints.docker.Host' | sed "s^unix://^^"

        Returns:
            str: The socket path on the host filesystem

        """
        try:
            context = docker.ContextAPI.inspect_context()

            if not isinstance(context, dict):
                raise TypeError(
                    f"Expected context to be a dictionary, got: {type(context).__name__}"
                )

            endpoints = context.get("Endpoints", {})
            docker_endpoint = endpoints.get("docker", {})
            docker_host = docker_endpoint.get("Host", "")

            # Remove the unix:// prefix
            if docker_host.startswith("unix://"):
                return docker_host[7:]

            raise ValueError(
                "Docker host does not start with 'unix://', cannot determine socket path"
            )

        except (docker.errors.DockerException, requests.RequestException) as err:
            raise DockerError("Failed to get Docker host socket path") from err
```

But I think it's not worth it. Chance of this changing are extremely low.

felipecrs: Found a const for it. All good.

supervisor/addons/build.py
felipecrs: Found consts for these dirs. All good.

felipecrs:
@mdegat01, @agners, my last remaining concern is if the Supervisor will delete dangling builder images at some point (for example older versions of `docker.io/library/docker` once HAOS gets updated with newer docker versions).

What do you think about it?

felipecrs:
@agners this PR is ready to be reviewed.

felipecrs:
I'm not sure I should adhere to @coderabbitai's comments.

supervisor/addons/build.py
```diff
@@ -121,39 +122,74 @@ def build_is_valid() -> bool:
...
-                "BUILD_FROM": self.base_image,
-                "BUILD_VERSION": version,
-                "BUILD_ARCH": self.sys_arch.default,
-                **self.additional_args,
-            },
+        # The addon path will be mounted from the host system
+        relative_path = self.addon.path_location.relative_to(
+            self.sys_config.path_supervisor
+        )
+        return str(self.sys_config.path_extern_supervisor / relative_path)
```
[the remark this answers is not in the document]
agners: Since add-ons are always a subdirectory of the Supervisor data path this is really not necessary. We actually would want to let such a exception bubble up so that we'd see it in Sentry.

agners: There is a helper to do that, search for `local_to_extern_path`.

supervisor/addons/build.py
```diff
@@ -121,39 +122,74 @@ def build_is_valid() -> bool:
...
+
+    def get_docker_args(
+        self, version: AwesomeVersion, image_tag: str
+    ) -> dict[str, Any]:
+        """Create a dict with Docker run args.
+
+        Must be run in executor.
```
agners: Do we do any I/O here? If no, then this does not need to run in an executor.

supervisor/docker/addon.py
```diff
@@ -673,10 +675,40 @@ async def _build(self, version: AwesomeVersion, image: str | None = None) -> Non
...
+            addon_image_tag = f"{image or self.addon.image}:{version!s}"
+
+            docker_version = self.sys_docker.info.version
+            builder_version_tag = f"{docker_version.major}.{docker_version.minor}.{docker_version.micro}-cli"
+
+            builder_name = f"addon_builder_{self.addon.slug}"
+
+            # Remove dangling builder container if it exists by any chance
+            with suppress(docker.errors.NotFound):
+                self.sys_docker.containers.get(builder_name).remove(force=True, v=True)
```
agners, line 694: In what case could this happen? Do we maybe need to try stop it in case it hanged and Supervisor got restarted? :thinking:

agners (commented):
This looks quite good. With the minor things addressed, I think we can land this and include it in the next Supervisor release.

supervisor/docker/addon.py
felipecrs: Let's say an add-on takes 1 minute to build.

If during this minute the host abruptly reboots, the add-on builder will remain in the system and would tamper with future addon builds because the container name is pinned.

This approach ensures the addon rebuild will not be blocked upon such disaster and also ensures previous dangling build containers will be removed at some point (when the addon gets rebuilt).

By the way, on such occasion, I believe the cleanup would even happen automatically upon reboot since Supervisor would still have the addon marked as "rebuild pending".

supervisor/addons/build.py
felipecrs: Awesome, this allowed me to simplify stuff.

felipecrs:
@agners all comments handled. Also, re-tested. All good from my side.

supervisor/docker/addon.py
felipecrs: I clarified this through a code comment.

tests/common.py
```diff
@@ -105,6 +105,20 @@ def reset_last_call(func, group: str | None = None) -> None:
...
+    Taken from https://stackoverflow.com/a/69175987/12156188.
+    """
+
+    for c in a:
+        if c in b:
+            b = b[b.index(c) :]
+        else:
+            return False
+    return True
+
```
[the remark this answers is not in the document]
felipecrs: @agners would you like me to handle this? My first preference would be to keep the function as similar as possible to its StackOverflow reference.

supervisor/addons/build.py
```diff
@@ -121,39 +122,63 @@ def build_is_valid() -> bool:
...
-            },
+    def get_docker_args(
+        self, version: AwesomeVersion, image_tag: str
+    ) -> dict[str, Any]:
+        """Create a dict with Docker run args."""
+        dockerfile_path = self.get_dockerfile().relative_to(self.addon.path_location)
+
+        build_cmd = [
+            "docker",
+            "build",
```
agners, line 134: It seems that simply using `docker build` without special environment etc. actually still uses legacy builder:

image

This seems to do the trick:

```suggestion
            "buildx",
            "build",
```

agners (changes requested):
I've tested a bunch of Core add-ons by removing the `image` tag and let it build using Supervisor itself. Some raised an exception because the args were not strings (e.g. the Let's Encrypt add-on). This is a not a new issue, we need to address that separatly. I guess we should make sure that all build args are actual strings, since that is what Docker passes. 

```
2025-07-02 16:49:14.507 CRITICAL (MainThread) [supervisor.utils.common] Can't parse /data/addons/local/letsencrypt/build.yaml: expected str for dictionary value @ data['args']['CERTBOT_DNS_DREAMHOST_VERSION']. Got 1.0
expected str for dictionary value @ data['args']['CERTBOT_DNS_DUCKDNS_VERSION']. Got 1.5
2025-07-02 16:49:14.507 WARNING (MainThread) [supervisor.utils.common] Resetting /data/addons/local/letsencrypt/build.yaml to default
```

But with that manually fixed, all add-ons got build flawlessly. Even parallel builds seem to work just fine. Nice :tada:

agners (approved):
LGTM, thanks!

felipecrs:
@agners thanks a lot for welcoming my contribution. :D

felipecrs:
> I've tested a bunch of Core add-ons by removing the `image` tag and let it build using Supervisor itself.

BTW it would be great to have some mechanism for rebuilding an addon even if `image` is set. That would be super useful for trying local addon changes without having to comment out the `image` line (which risks committing that by mistake).

Maybe always attempting to build when invoked by the `ha addons rebuild` command?

agners:
> BTW it would be great to have some mechanism for rebuilding an addon even if `image` is set. That would be super useful for trying local addon changes without having to comment out the `image` line (which risks committing that by mistake).

Hm, seems a bit an edge case, but yeah why not.

> 
> Maybe always attempting to build when invoked by the `ha addons rebuild` command?

Currently it is refusing to rebuild with:

```
# ha addons rebuild a0d7b954_tailscale
Processing... Done.

Error: Can't rebuild a image based add-on
```

It could become a bit confusing, especially if a Dockerfile doesn't create reproducible docker images.  I think I'd prefer to have this behind a `--force` flag, just to make it clear that this is not a normal user operation.

felipecrs:
I completely agree with `ha addons rebuild --force`.

agners:
Note to myself: To run buildx command the following command can be used:

```
docker run --rm --mount type=bind,src=/var/run/docker.sock,dst=/var/run/docker.sock  docker.io/library/docker:28.3.3 docker buildx du 
```

PR #5977, opened 2025-06-27
Remove anonymous volumes when removing containers

felipecrs:
## Proposed change

All `container.remove()` calls now include `v=True`. This will also delete whatever anonymous (and ephemeral) volumes that were created by that container like how `docker run --rm` does.

This avoids potential leftovers in the system depending on whether the image being executed had `VOLUME` instructions in its Dockerfile.

## Type of change

- [ ] Dependency upgrade
- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (which adds functionality to the supervisor)
- [ ] Breaking change (fix/feature causing existing functionality to break)
- [ ] Code quality improvements to existing code or addition of tests

## Additional information

Extracted from https://github.com/home-assistant/supervisor/pull/5974.

## Checklist

- [x] The code change is tested and works locally.
- [x] Local tests pass. **Your PR cannot be merged unless tests pass**
- [x] There is no commented out code in this PR.
- [x] I have followed the [development checklist][dev-checklist]
- [x] The code has been formatted using Ruff (`ruff format supervisor tests`)
- [x] Tests have been added to verify that the new code works.

If API endpoints or add-on configuration are added/changed:

- [ ] Documentation added/updated for [developers.home-assistant.io][docs-repository]
- [ ] [CLI][cli-repository] updated (if necessary)
- [ ] [Client library][client-library-repository] updated (if necessary)

[dev-checklist]: https://developers.home-assistant.io/docs/en/development_checklist.html
[docs-repository]: https://github.com/home-assistant/developers.home-assistant
[cli-repository]: https://github.com/home-assistant/cli
[client-library-repository]: https://github.com/home-assistant-libs/python-supervisor-client/

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved container removal to also delete associated volumes, ensuring more thorough cleanup when containers are removed.

* **Tests**
  * Updated test cases to verify that container removal now includes the deletion of associated volumes.
  * Added new asynchronous tests for Docker command execution covering success, defaults, exceptions, cleanup, and output handling.

PR #5979, opened 2025-06-27
Fix `docker.run_command()` needing `detach` but not enforcing it

felipecrs:
## Proposed change

`docker.run_command` only works with `Detach=True`, but was not including it in the default `container.run` args (i.e. was broken by default).

This fixes it, and also remove the redundant arguments from `docker.run_command` callers.

## Type of change

- [ ] Dependency upgrade

## Additional information

Extracted from https://github.com/home-assistant/supervisor/pull/5974.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of container command execution to ensure consistent detachment and output stream behavior.
  * Adjusted retrieval of command exit codes for more reliable status reporting.

### DIFF
--- a/supervisor/docker/manager.py
+++ b/supervisor/docker/manager.py
@@ -311,6 +311,7 @@ class DockerAPI:
             container = self.docker.containers.run(
                 f"{image}:{tag}",
                 command=command,
+                detach=True,
                 network=self.network.name,
                 use_config_proxy=False,
                 **kwargs,
@@ -329,7 +330,7 @@ class DockerAPI:
                 with suppress(docker_errors.DockerException, requests.RequestException):
                     container.remove(force=True, v=True)
 
-        return CommandReturn(result.get("StatusCode"), output)
+        return CommandReturn(result["StatusCode"], output)
 
     def repair(self) -> None:
         """Repair local docker overlayfs2 issues."""
--- a/supervisor/docker/homeassistant.py
+++ b/supervisor/docker/homeassistant.py
@@ -213,9 +213,6 @@ class DockerHomeAssistant(DockerInterface):
             privileged=True,
             init=True,
             entrypoint=[],
-            detach=True,
-            stdout=True,
-            stderr=True,
             mounts=[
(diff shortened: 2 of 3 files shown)


felipecrs:
I don't understand why codecov is failing. It is pointing itself that was a 0.00% coverage change with this PR.

felipecrs:
Waiting for https://github.com/home-assistant/supervisor/pull/5977 to be merged first, as it adds the test that will fix codecov for this PR.

supervisor/docker/manager.py
```diff
@@ -311,6 +311,7 @@ def run_command(
             container = self.docker.containers.run(
                 f"{image}:{tag}",
                 command=command,
+                detach=True,
```
agners, line 314: The change generally looks good, from what I understand this is required because we call `container.wait()` below.

That said, is there a good reason we go down that road? We could also run it attached and simply let the `container.run()` block? :thinking:

agners:
Since `execute_command()` in `supervisor/docker/homeassistant.py` is the only call site of this function, and it is passing the detach, this PR effectively fixes a "local" bug, which is hidden by correct call args. So for the end user it doesn't fix a Supervisor bug, which is why I labeled it `refactor`. @felipecrs I hope you agree with that assessment.

felipecrs:
Sure, I completely agree.

supervisor/docker/manager.py
felipecrs: > That said, is there a good reason we go down that road? We could also run it attached and simply let the `container.run()` block? 🤔

Indeed `container.run()` could be used and in fact I was using it in https://github.com/home-assistant/supervisor/pull/5974 before I found out this function.

However, my intention was to minimize code changes as much as possible.

If you don't mind, I would prefer merging this. Later we can consider a bigger refactor for leveraging `container.run()` with `Detach=False` and `rm=True`.

agners (approved):
LGTM, thanks!

PR #5990, opened 2025-07-03
Refactor add-on store git repository more

agners:
## Proposed change
Developed using Claude Code. It describes the overall refactoring as follows:

  🏗️ Architecture Improvements Achieved:

  1. Eliminated magic strings - No more scattered "core" and "local" literals
  2. Clear repository types - SpecialRepository vs DefaultRepository enums
  3. Simplified class hierarchy - Direct classes instead of complex inheritance
  4. Cleaner factory method - Straightforward type-based creation
  5. Better separation of concerns - Special vs URL-based repositories clearly distinguished

  🔧 Files Updated:

  - supervisor/store/types.py - New clean type definitions
  - supervisor/store/repository.py - Simplified repository classes
  - supervisor/store/validate.py - Updated validation logic
  - supervisor/store/__init__.py - Updated store manager
  - supervisor/store/const.py - Removed StoreType enum
  - supervisor/store/data.py - Updated error handling with constants
  - tests/store/test_custom_repository.py - Updated imports
  - tests/conftest.py - Fixed test fixtures

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist


supervisor/store/repository.py
```diff
@@ -100,29 +101,68 @@
...
-        """Assume built-in repositories are always valid."""
+        """Assume local repository is always valid."""
         return True
 
+    async def load(self) -> None:
+        """Load addon repository."""
+        self._latest_mtime, _ = await self.sys_run_in_executor(
+            get_latest_mtime, self.sys_config.path_addons_local
+        )
+
```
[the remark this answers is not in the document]
agners: We did not do error handling before, let's not do this as part of this refactoring.

supervisor/store/repository.py
```diff
@@ -100,29 +101,68 @@ async def reset(self) -> None:
...
+    async def update(self) -> bool:
+        """Update add-on repository.
+
+        Returns True if the repository was updated.
+        """
+        # Check local modifications
+        latest_mtime, modified_path = await self.sys_run_in_executor(
+            get_latest_mtime, self.sys_config.path_addons_local
+        )
```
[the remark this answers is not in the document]
agners: We did not do error handling before, let's not do this as part of this refactoring.

supervisor/store/repository.py
```diff
@@ -138,6 +178,25 @@ async def update(self) -> bool:
```
mdegat01: EDIT: Ok apparently highlighting the entire block of code including a part github decided to collapse doesn't work either, you just simply can't make a suggestion on collapsed code 😠 

https://github.com/home-assistant/supervisor/blob/2e28451bdb94186667c017c954d116ad5e9cb725/supervisor/store/repository.py#L176-L177

This doesn't do anything after these changes and should be removed. The validate in this class always returns `True`

supervisor/store/repository.py
```diff
@@ -162,72 +221,25 @@ def validate_file() -> bool:
...
-        """Initialize object."""
-        super().__init__(coresys, url)
-        self._slug = get_hash_from_repository(url)
-        self._type = StoreType.GIT
-        self._git = GitRepo(coresys, coresys.config.path_addons_git / self._slug, url)
+        return await self._git.pull()
 
     async def remove(self) -> None:
         """Remove add-on repository."""
         await self._git.remove()
```
mdegat01: This brings us back to a situation where its possible to remove built-in repositories, which should never happen. And it also returns us to validating all built-in repositories except core like before the last PR.

I'm not entirely sure why this approach is better. It seems to return us to core as a super special built-in repository and the others are just custom repositories that happen to be in the default list. Which isn't what we want I don't think. I was trying to get us to the point where there was a special Local repository, built-in repositories and custom repositories. And the only thing special about core was the slug that we're kind of stuck with for compatibility reasons.

supervisor/store/types.py
```diff
@@ -0,0 +1,26 @@
...
+
+    COMMUNITY_ADDONS = "https://github.com/hassio-addons/repository"
+    ESPHOME = "https://github.com/esphome/home-assistant-addon"
+    MUSIC_ASSISTANT = "https://github.com/music-assistant/home-assistant-addon"
+
+
+# All repositories that are considered "built-in" (special + default)
+ALL_BUILTIN_REPOSITORIES = {repo.value for repo in SpecialRepository} | {
+    repo.value for repo in DefaultRepository
+}
```
mdegat01: This kind of hammers in the idea that core is more special then the other built-ins. Why do we want that? In my mind built-ins are built-ins. The only special one is local because its not git-based. I wish we could eliminate the special `core` "URL" but that seems not worth the compatibility risks at this point. Core shouldn't be any different from community addons, esphome or music assistant in my mind though.

agners: > This kind of hammers in the idea that core is more special then the other built-ins. Why do we want that? 

Agreed, we don't really want that. But it's how it is currently :shrug: . Claude came up with this separation, and at least to how things work today it does make sense. `StoreType` currently still has `CORE`, so treats it special too. `repository_urls` returns all git repos but the core repo (even though documentation says "Return source URL for all git repositories"). :thinking: 

But yeah, we should not code that into logic, but rather have code reflect where we want to go (Core not being special). I'll rollback some of the changes and adjust accordingly.

supervisor/store/validate.py
```diff
@@ -75,12 +23,12 @@ def ensure_builtin_repositories(addon_repositories: list[str]) -> list[str]:
...
     stable. This can have side effects when comparing data later on.
     """
-    return list(set(addon_repositories) | BUILTIN_REPOSITORIES)
+    return list(set(addon_repositories) | ALL_BUILTIN_REPOSITORIES)
 
 
 def validate_repository(repository: str) -> str:
     """Validate a valid repository."""
-    if repository in [StoreType.CORE, StoreType.LOCAL]:
+    if repository in BuiltinRepository:
```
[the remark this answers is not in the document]
agners: But in a quick test it seems to work:

```python
>>> "local" in BuiltinRepository
True
```

supervisor/store/data.py
```diff
@@ -169,7 +168,7 @@ def _get_addons_list() -> list[Path]:
                 self.sys_resolution.add_unhealthy_reason(
                     UnhealthyReason.OSERROR_BAD_MESSAGE
                 )
-            elif path.stem != StoreType.LOCAL:
+            elif repository != REPOSITORY_LOCAL:
```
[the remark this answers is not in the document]
agners: That is not true, repository is the static string "local" in that case.

supervisor/store/__init__.py
```diff
@@ -236,38 +226,52 @@ async def update_repositories(
...
+            repositories_to_add = set(list_repositories) - current_repositories
+            target_repositories = current_repositories | repositories_to_add
 
         # Add new repositories
         add_errors = await asyncio.gather(
             *[
-                self._add_repository(url, persist=False, add_with_errors=True)
-                if add_with_errors
+                # Use _add_repository to avoid JobCondition.SUPERVISOR_UPDATED
+                # to preven proper loading of repositories on startup.
```
mdegat01:
```suggestion
                # to prevent proper loading of repositories on startup.
```
This is not important at all. I only put this comment because Claude wrote this PR so I'm confused how there could be a typo...

supervisor/store/__init__.py
```diff
@@ -236,38 +226,52 @@ async def update_repositories(
...
+        # Remove repositories
         remove_errors = await asyncio.gather(
             *[
-                self.remove_repository(self.get_from_url(url), persist=False)
-                for url in old_rep - new_rep - BUILTIN_REPOSITORIES
+                self.remove_repository(repository, persist=False)
+                for repository in repositories_to_remove
             ],
             return_exceptions=True,
         )
```
mdegat01:
```suggestion
        remove_errors: list[BaseException | None] = []
        if replace:
            # Determine repositories to remove
            repositories_to_remove: list[Repository] = [
                repository
                for repository in self.all
                if repository.source not in target_repositories
                and not repository.is_builtin
            ]

            # Remove repositories
            remove_errors = await asyncio.gather(
                *[
                    self.remove_repository(repository, persist=False)
                    for repository in repositories_to_remove
                ],
                return_exceptions=True,
            )
```
I'm not entirely sure why the code didn't have this conditional already because reading it it was simply looping over an empty list 100% of the time if `replace` was `False`. Now its looping over a non-empty list doing work but still never going to remove anything if `replace` is `False` so we should limit this code to run only when it might actually do something.

Also technically we don't need the `and not repository.is_builtin` because `target_repositories` already ensured that but I guess leaving it for clarity here is fine.

supervisor/store/types.py
```diff
@@ -0,0 +1,32 @@
+"""Repository type definitions for the store."""
```
mdegat01: This is a nitpick. But everywhere else in Supervisor these definitions would be in the corresponding `const.py` file of the module. We can leave it if we want but it is inconsistent with our existing patterns so it might confuse us later.

tests/conftest.py
```diff
@@ -595,18 +595,11 @@ def store_addon(coresys: CoreSys, tmp_path, repository):
...
 @pytest.fixture
-async def repository(coresys: CoreSys):
-    """Repository fixture."""
-    coresys.store._data[ATTR_REPOSITORIES].remove(
-        "https://github.com/hassio-addons/repository"
-    )
-    coresys.store._data[ATTR_REPOSITORIES].remove(
-        "https://github.com/esphome/home-assistant-addon"
-    )
+async def test_repository(coresys: CoreSys):
```
mdegat01, line 598: Love this. We should definitely use AI for all future name changes of test fixtures. It is such a pain normally because you can't track references with these so normal refactor tools don't work. They just end up getting left as originally defined forever because its too annoying by hand 🙈

tests/store/test_custom_repository.py
```diff
@@ -33,7 +42,7 @@ async def test_add_valid_repository(
     """Test add custom repository."""
     current = coresys.store.repository_urls
     with (
-        patch("supervisor.store.repository.RepositoryGit.load", return_value=None),
+        patch("supervisor.store.git.GitRepo.load", return_value=None),
```
mdegat01, line 45: Claude taking digs at me for mocking a bit too much 😆 It's right though, this is preferrable, not sure why I mocked the Repository load.

supervisor/store/types.py
agners: Ok yeah make sense, let's keep this pattern.

Also added this to the Copilot instructions in #6011.

supervisor/store/const.py
```diff
@@ -3,14 +3,39 @@
...
-    GIT = "git"
+    # Local repository (non-git, special handling)
+    LOCAL = REPOSITORY_LOCAL
+
+    # Git-based built-in repositories
+    CORE = REPOSITORY_CORE
+    COMMUNITY_ADDONS = "https://github.com/hassio-addons/repository"
+    ESPHOME = "https://github.com/esphome/home-assistant-addon"
+    MUSIC_ASSISTANT = "https://github.com/music-assistant/home-assistant-addon"
+
```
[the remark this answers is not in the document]
agners: It is the only place where this is hard coded, and its already in a const.py file. Adding a separate constant and reuse it in the Enum seems unnecessary to me.

supervisor/store/__init__.py
agners: Nice catch, yeah that make sense.

> Also technically we don't need the and not repository.is_builtin because target_repositories already ensured that but I guess leaving it for clarity here is fine.

Hm, good point. It also can be confusing for readers if such things are unnecessarily tested twice. I think I prefer to get rid of the `ensure_builtin_repositories` function. But let's separate this change in another PR.

PR #6006, opened 2025-07-09
Implement logging through Systemd native Journal protocol

sairon:
## Proposed change

Use JournaldLogHandler from logging-journald [1] to log events from Python logging to Systemd journal directly. This makes it possible to better filter the log messages based on priority and persist more useful metadata (like filename and line number) for the log events in the Journal. The Journal log priority is also distingushed based on different log priorities of logging events.

Another advantage is that stack traces are saved as a single log event rather than multiple separated lines printed to container stdout/stderr. The coloring isn't done using ANSI escape sequences from the ColoredFormatter anymore, resulting in plaintext MESSAGE fields (instead of byte arrays). Colors are added based on the priority metadata when the events are read from the Systemd Journal Gatewayd to maintain backwards compatibility.

To correctly match non-Python error logs, log entries must be written to the Journal with the same SYSLOG_IDENTIFIER field as the logs coming form the container directly (e.g. from s6, bashio, etc.). To obtain the Python logs through the docker logs command, container ID (obtained though a hacky method currently) is added to the log records as well - this is useful to keep the falllback method for getting the logs working.

[1] https://github.com/mosquito/logging-journald/
[2] https://systemd.io/JOURNAL_NATIVE_PROTOCOL/

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR is related to issue: https://github.com/home-assistant/operating-system/issues/4121
- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/host/sound.py
+++ b/supervisor/host/sound.py
@@ -339,7 +339,7 @@ class SoundControl(CoreSysAttributes):
                                     profile.name == card.profile_active.name,
                                 )
                                 for profile in card.profile_list
-                                if profile.available
+                                if profile.is_available
                             ],
                         )
                         for card in pulse.card_list()
--- a/supervisor/utils/logging.py
+++ b/supervisor/utils/logging.py
@@ -7,6 +7,8 @@ import logging.handlers
 import queue
 from typing import Any
 
+from logging_journald import Facility, JournaldLogHandler
+
 
 class AddonLoggerAdapter(logging.LoggerAdapter):
     """Logging Adapter which prepends log entries with add-on name."""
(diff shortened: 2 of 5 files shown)


supervisor/utils/systemd_journal.py
```diff
@@ -24,15 +38,34 @@ def decorator(func):
...
+        implicit_fields = ["__CURSOR", "SYSLOG_FACILITY", "PRIORITY"]
+        wrapper.required_fields = implicit_fields + required_fields
         return wrapper
 
     return decorator
 
 
+def _entry_is_haos_log(entry: dict[str, str]) -> bool:
+    """Check if the entry is a Home Assistant Operating System log entry."""
+    return entry.get("SYSLOG_FACILITY") == str(HAOSLogHandler.SYSLOG_FACILITY)
```
sairon, line 50: It appears that nothing else is using this facility in HAOS currently so it's fairly reliable indicator it's "our" logs.. `Facility.LOCAL7 == 23` is the default in `JournaldLogHandler` as well but I defined it on the `HAOSLogHandler` class to be explicit and stable.

supervisor/utils/logging.py
```diff
@@ -59,6 +61,52 @@ def close(self) -> None:
         self.listener = None
 
 
+class HAOSLogHandler(JournaldLogHandler):
```
sairon, line 64: If the Core should log through Systemd as well, I think it would be good to extract this to a shared Python package.

supervisor/utils/logging.py
```diff
@@ -59,6 +61,52 @@ def close(self) -> None:
...
+        # Currently we only have this hacky way of getting the container ID,
+        # we (probably) cannot get it without having some cgroup namespaces
+        # mounted in the container or passed it there using other means.
+        # Not obtaining it will only result in the logs not being available
+        # through `docker logs` command, so it is not a critical issue.
+        with open("/proc/self/mountinfo", mode="rb") as f:
+            for line in f:
+                if b"/docker/containers/" in line:
+                    container_id = line.split(b"/docker/containers/")[-1]
+                    return str(container_id.split(b"/")[0])
```
agners, line 86: We've discussed it elsewhere, but this seems a bit brittle.

Using the Docker `--cidfile` and bind mount that into the container seems a better approach. I think we need to store it somewhere persistent, as we don't recreate containers on restart. E.g. `/mnt/data/supervisor/cidfiles/${containername}`. For now, I'd do it just for Supervisor and Core, eventually we could add it to all the containers if there is a need. Probably best is to create this as a separate PR.

supervisor/utils/systemd_journal.py
agners: Hm, I see this is syslog backward compatibility, and the numbers are from [RFC5424](https://www.rfc-editor.org/rfc/rfc5424.html), interesting. 

It seems in my logs that quite some entries omit any `SYSLOG_FACILITY`. I wonder if we really want to use that as the means to identify our logs? :thinking: 

I see the library currently maps the standard logging priorities to the systemd-journald priority levels, maybe it would be better to store the Python logging level (which also can have custom levels) as a separate field (e.,g. `PYTHON_LEVEL`), and use that as a means to detect "our" logs? :thinking: I guess it would involve some changes to the library, but could be a good way to test how active that upstream project is.

supervisor/utils/logging.py
agners: Yeah make sense, it is fine by me to have this in Supervisor for the time being.

supervisor/host/sound.py
```diff
@@ -339,7 +339,7 @@ def _get_pulse_data() -> PulseData:
                                     profile.name == card.profile_active.name,
                                 )
                                 for profile in card.profile_list
-                                if profile.available
+                                if profile.is_available
```
sairon, line 342: Huh, this is probably side-effect of the IDE refactoring tool, shouldn't be here.

PR #6012, opened 2025-07-10
Drop ensure_builtin_repositories()

agners:
## Proposed change
With the new Repository classes we have the is_builtin property, so we can easily make sure that built-ins are not removed. The `ensure_builtin_repositories()` is really redundant logic at this point. This allows us to further cleanup the code by removing the `ensure_builtin_repositories()` function and the ALL_BUILTIN_REPOSITORIES constant.

This came up in discussions of PR https://github.com/home-assistant/supervisor/pull/5990#discussion_r2195466349.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

## Summary by CodeRabbit

* **Refactor**
  * Streamlined repository update logic for improved efficiency and maintainability.
  * Replaced static repository lists with dynamic enumeration-based lists for built-in repositories.
  * Ensured built-in repositories are always included during startup.
  * Changed repository collections from lists to sets for uniqueness and simplified handling.

* **Tests**
  * Updated tests to use the new enumeration-based approach for built-in repositories and set-based collections.

### DIFF
--- a/supervisor/store/__init__.py
+++ b/supervisor/store/__init__.py
@@ -21,7 +21,7 @@ from .addon import AddonStore
 from .const import FILE_HASSIO_STORE, BuiltinRepository
 from .data import StoreData
 from .repository import Repository
-from .validate import SCHEMA_STORE_FILE
+from .validate import DEFAULT_REPOSITORIES, SCHEMA_STORE_FILE
 
 _LOGGER: logging.Logger = logging.getLogger(__name__)
 
@@ -67,9 +67,9 @@ class StoreManager(CoreSysAttributes, FileConfiguration):
         # Make sure the built-in repositories are all present
         # This is especially important when adding new built-in repositories
         # to make sure existing installations have them.
-        all_repositories: set[str] = set(self._data.get(ATTR_REPOSITORIES, [])) | {
-            repo.value for repo in BuiltinRepository
-        }
+        all_repositories: set[str] = (
+            set(self._data.get(ATTR_REPOSITORIES, [])) | DEFAULT_REPOSITORIES
+        )
         await self.update_repositories(all_repositories, issue_on_error=True)
 
     @Job(
--- a/supervisor/store/__init__.py
+++ b/supervisor/store/__init__.py
@@ -21,7 +21,7 @@ from .addon import AddonStore
 from .const import FILE_HASSIO_STORE, BuiltinRepository
 from .data import StoreData
 from .repository import Repository
-from .validate import SCHEMA_STORE_FILE, ensure_builtin_repositories
+from .validate import SCHEMA_STORE_FILE
 
 _LOGGER: logging.Logger = logging.getLogger(__name__)
 
@@ -231,14 +231,8 @@ class StoreManager(CoreSysAttributes, FileConfiguration):
         """Update repositories by adding new ones and removing stale ones."""
         current_repositories = {repository.source for repository in self.all}
 
-        # Determine changes needed
-        if replace:
-            target_repositories = set(ensure_builtin_repositories(list_repositories))
-            repositories_to_add = target_repositories - current_repositories
-        else:
-            # When not replacing, just add the new repositories
-            repositories_to_add = set(list_repositories) - current_repositories
-            target_repositories = current_repositories | repositories_to_add
+        # Determine repositories to add
+        repositories_to_add = set(list_repositories) - current_repositories
 
         # Add new repositories
         add_errors = await asyncio.gather(
@@ -259,7 +253,7 @@ class StoreManager(CoreSysAttributes, FileConfiguration):
             repositories_to_remove: list[Repository] = [
                 repository
                 for repository in self.all
-                if repository.source not in target_repositories
+                if repository.source not in list_repositories
                 and not repository.is_builtin
             ]
 
--- a/supervisor/store/__init__.py
+++ b/supervisor/store/__init__.py
@@ -63,11 +63,14 @@ class StoreManager(CoreSysAttributes, FileConfiguration):
         return self.repositories[slug]
 
     async def load(self) -> None:
-        """Start up add-on management."""
-        # Init custom repositories and load add-ons
-        await self.update_repositories(
-            self._data[ATTR_REPOSITORIES], issue_on_error=True
-        )
+        """Start up add-on store management."""
+        # Make sure the built-in repositories are all present
+        # This is especially important when adding new built-in repositories
+        # to make sure existing installations have them.
+        all_repositories: set[str] = set(self._data.get(ATTR_REPOSITORIES, [])) | {
+            repo.value for repo in BuiltinRepository
+        }
+        await self.update_repositories(all_repositories, issue_on_error=True)
 
     @Job(
         name="store_manager_reload",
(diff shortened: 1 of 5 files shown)


supervisor/store/validate.py
```diff
@@ -44,11 +35,11 @@ def validate_repository(repository: str) -> str:
...
 repositories = vol.All([validate_repository], vol.Unique())
 
+DEFAULT_REPOSITORIES = [repo.value for repo in BuiltinRepository]
+
 SCHEMA_STORE_FILE = vol.Schema(
     {
-        vol.Optional(
-            ATTR_REPOSITORIES, default=list(ALL_BUILTIN_REPOSITORIES)
-        ): repositories,
+        vol.Optional(ATTR_REPOSITORIES, default=DEFAULT_REPOSITORIES): repositories,
```
mdegat01:
```suggestion
        vol.Optional(ATTR_REPOSITORIES, default=DEFAULT_REPOSITORIES.copy()): repositories,
```
The list is mutable so using a reference to it as the default value seems like a bad idea likely to lead to strange bugs.

supervisor/store/__init__.py
```diff
@@ -63,11 +63,14 @@ def get(self, slug: str) -> Repository:
...
-        await self.update_repositories(
-            self._data[ATTR_REPOSITORIES], issue_on_error=True
-        )
+        """Start up add-on store management."""
+        # Make sure the built-in repositories are all present
+        # This is especially important when adding new built-in repositories
+        # to make sure existing installations have them.
+        all_repositories: set[str] = set(self._data.get(ATTR_REPOSITORIES, [])) | {
+            repo.value for repo in BuiltinRepository
+        }
```
mdegat01: Since we should make a copy of `DEFAULT_REPOSITORIES` in order to use it in the validate, maybe better to just define it as a `set` to begin with? Then we can use it here as is and in the validate we can do `list(DEFAULT_REPOSITORIES)` (which also conveniently makes a copy).

mdegat01 (approved):
Makes sense. Small suggested tweak because I don't think we should use references to mutable constants as a default value in our validates.

supervisor/store/validate.py
agners: Uh yeah I kinda saw this as an actual performance improvement: Avoid creating another list object. I've asked Claude Code to test if a copy gets created, and it seems it does:

<details>

```py
#!/usr/bin/env python3
"""Test script to verify Voluptuous behavior with mutable default values."""

import voluptuous as vol

# Test 1: Basic mutable default behavior
print("=== Test 1: Basic mutable default behavior ===")

original_list = ["a", "b", "c"]
print(f"Original list before schema creation: {original_list}")

schema = vol.Schema({
    vol.Optional("items", default=original_list): [str]
})

print(f"Original list after schema creation: {original_list}")

# Test with empty input (should use default)
result1 = schema({})
print(f"First validation result: {result1}")
print(f"Original list after first validation: {original_list}")

# Modify the result to see if it affects the original
result1["items"].append("modified_in_result1")
print(f"Modified result1: {result1}")
print(f"Original list after modifying result1: {original_list}")

# Test with second validation
result2 = schema({})
print(f"Second validation result: {result2}")
print(f"Original list after second validation: {original_list}")

# Test 2: Modify original list after schema creation
print("\n=== Test 2: Modify original list after schema creation ===")

original_list2 = ["x", "y", "z"]
schema2 = vol.Schema({
    vol.Optional("items", default=original_list2): [str]
})

# Modify original list after schema creation
original_list2.append("added_after_schema")
print(f"Original list after modification: {original_list2}")

result3 = schema2({})
print(f"Validation result after original modification: {result3}")

# Test 3: Multiple schema instances with same default
print("\n=== Test 3: Multiple schema instances with same default ===")

shared_default = ["shared1", "shared2"]
schema_a = vol.Schema({
    vol.Optional("items", default=shared_default): [str]
})
schema_b = vol.Schema({
    vol.Optional("items", default=shared_default): [str]
})

result_a = schema_a({})
result_b = schema_b({})

print(f"Schema A result: {result_a}")
print(f"Schema B result: {result_b}")

# Modify result from schema A
result_a["items"].append("modified_via_a")
print(f"Modified schema A result: {result_a}")
print(f"Schema B result after A modification: {result_b}")

# Validate again with schema B
result_b2 = schema_b({})
print(f"Fresh schema B validation: {result_b2}")
print(f"Shared default after all operations: {shared_default}")
```
Restults in
```
=== Test 1: Basic mutable default behavior ===
Original list before schema creation: ['a', 'b', 'c']
Original list after schema creation: ['a', 'b', 'c']
First validation result: {'items': ['a', 'b', 'c']}
Original list after first validation: ['a', 'b', 'c']
Modified result1: {'items': ['a', 'b', 'c', 'modified_in_result1']}
Original list after modifying result1: ['a', 'b', 'c']
Second validation result: {'items': ['a', 'b', 'c']}
Original list after second validation: ['a', 'b', 'c']

=== Test 2: Modify original list after schema creation ===
Original list after modification: ['x', 'y', 'z', 'added_after_schema']
Validation result after original modification: {'items': ['x', 'y', 'z', 'added_after_schema']}

=== Test 3: Multiple schema instances with same default ===
Schema A result: {'items': ['shared1', 'shared2']}
Schema B result: {'items': ['shared1', 'shared2']}
Modified schema A result: {'items': ['shared1', 'shared2', 'modified_via_a']}
Schema B result after A modification: {'items': ['shared1', 'shared2']}
Fresh schema B validation: {'items': ['shared1', 'shared2']}
Shared default after all operations: ['shared1', 'shared2']
```

</details>

However, looking into Voluptuous code, it seems that this is more a side effect of the validation (which seems to run even when filling in the defaults). 

<details>

```
● Perfect! You were absolutely right to question my explanation. The default_factory function does just return the same reference - the copy protection happens during the validation process 
  itself.

  Here's what actually happens:

  The Real Copy Mechanism:

  1. Default Factory Returns Same Reference

  return lambda: value  # Returns the SAME list object

  2. Validation Process Creates Copies

  When validating a list, Voluptuous does this (_compile_sequence method):

  # Line 581: Create new output list
  out = []

  # Lines 585-592: Validate each element and append to NEW list
  for i, value in enumerate(data):
      cval = validate(index_path, value)
      out.append(cval)

  # Line 606: Return new list of same type
  return type(data)(out)

  3. Same for Dictionaries

  # Line 548: Create new output dictionary
  out = data.__class__()

  # Line 333: Add validated values to NEW dictionary
  out[new_key] = cval

  So You Were Right:

  - default_factory just creates a callable that returns the same object reference
  - The validation process itself creates copies of container types (lists, dicts, tuples)
  - This is why modifying validation results doesn't affect the original defaults

  The mutable default protection is a side effect of Voluptuous's design to create clean validation results, not an intentional feature to handle mutable defaults.

  Your intuition was spot-on - the factory function should return the same reference, and it does. The copying happens later in the validation pipeline.
```

</details>

Anyhow, I think what we can do here to avoid the copy in the typical case (where the repositories are loaded from disk) is to provide a factory lambda instead. I'll adjust accordingly.

supervisor/store/__init__.py
agners: Yeah seems sensible :+1:

mdegat01 (approved):
LGTM 👍

PR #6013, opened 2025-07-10
Split execution limit in concurrency and throttle parameters

agners:
## Proposed change
Currently the execution limit combines two orthogonal features: Limit concurrency and throttle execution. This change separates the two features, allowing for more flexible configuration of job execution.
    
Ultimately I want to get rid of the old limit parameter. But for ease of review and migration, I'd like to do this in two steps: First introduce the new parameters, and map the old limit parameters to the new parameters. Then, in a second step, remove the old limit parameter and migrate all users to the new concurrency and throttle parameters as needed.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist


supervisor/jobs/decorator.py
```diff
@@ -255,71 +321,11 @@ async def wrapper(
...
-                            else self.on_condition
-                        )
-                        raise on_condition(
-                            f"Rate limit exceeded, more than {self.throttle_max_calls} calls in {self.throttle_period(group_name)}",
-                        )
+                # Handle execution limits
+                await self._handle_concurrency_control(job_group, job)
+                if not await self._handle_throttling(group_name):
+                    self._release_concurrency_control(job_group)
+                    return  # Job was throttled, exit early
```
mdegat01: It looks like the deadlock is due to the fact that the group throttles raise when they get throttled, unlike the non-group throttles. Which means concurrency control is never released.

I'd suggest defining `_handle_concurrency_control` as an `asynccontextmanager` that uses a `try-finally` setup and calls `_release_concurrency_control` in the `finally`. Then you can wrap this entire block (including the job.start part) with this:
```py
async with self._handle_concurrency_control(job_group, job):
```
And drop the finally at the end since the concurrency control context manager just handles getting and releasing the lock in all cases.

Alternatively we could make the group throttles stop raising. I'm not sure why these two have inconsistent behavior 🙈

agners: > I'd suggest defining `_handle_concurrency_control` as an `asynccontextmanager` that uses a `try-finally` setup and calls `_release_concurrency_control` in the `finally`. Then you can wrap this entire block (including the job.start part) with this:
> 
> ```python
> async with self._handle_concurrency_control(job_group, job):
> ```

Uh yes, nice cleanup, I definitely implement that.

> It looks like the deadlock is due to the fact that the group throttles raise when they get throttled, unlike the non-group throttles. Which means concurrency control is never released.

I don't think group throttles raise, they simply return just like the regular throttle. The group throttling is handled in  `_handle_throttling()`, there is no raise statement there.

The problem is more delicate: We can't unlock group locks unless we start the job. See my comment in https://github.com/home-assistant/supervisor/pull/6013/commits/351de4d9a3655e3b3309f6d92168f7f38fcf3460. And because we don't start the job when throttled, we can't unlock. The problem seems to be in the Job group release implementation, it raises `supervisor.exceptions.JobException: Cannot release as caller does not own lock`, because `has_lock` evaluated to False (despite the fact I created the lock).

This is definitely fixable. But it made me realize that currently the group limits use the non-group concurrency checks. Probably not intentional though 🤷.

Note that this only affects `GROUP_THROTTLE_WAIT`. But with the orthogonal configuration, the user can use other combinations though...

I suggest the following approach:
1. Go ahead with the current implementation which aims to be a drop in replacement with the very same behavior.
2. Replace the decorators with the new settings
3. Fix Job group locks so they can be used with throttle limits.

agners: Hm, actually, while trying to fix job group locks (with https://github.com/home-assistant/supervisor/commit/c71daefd7f524d5ae335a5f83456d888fc3d7126), I realized that any throttle limits won't work with group locks. I'll adjust the RuntimeError accordingly.

agners: Actually, I tried implementing the context manager with https://github.com/home-assistant/supervisor/commit/7f584e386d04e6178ece50b6de3e0f14bc3c88ea, and it turns out that this suffers from the same problem as the group locks releasing problem: With that context manager, the group lock gets released outside the `with job.start():` block, essentially outside of the code path where the job is running. This again causes the lock to not get released properly (e.g. it makes the  `pytest tests/jobs/test_job_decorator.py::test_job_group_wait` deadlock) :cry:

mdegat01 (approved):
This looks good. But I think we need some test cases no? Maybe just on the combinations that were never possible before, like `REJECT` + `THROTTLE` or `REJECT` + `RATE_LIMIT`

mdegat01 (approved):

supervisor/jobs/decorator.py
agners: Addressed with #6074.

Issue #6021, opened 2025-07-15
Mark system as unhealthy when multiple operating system instances are detected


agners:
### Description

Add an Supervisor unhealthy condition when multiple Home Assistant OS installations are detected.

Multiple Home Assistant OS installations on a single system can be problematic because the update system uses partition labels (or for devices which still use MBR style partition labels, its based on partition entry UUID, which is made up from MBR disk signature plus partition number). 

In both cases, when two installations are present, the update system might update one of the two installations. However, the boot loader might choose the other installation. This leads to systems booting into an old OS version, or worse, in case the boot flow changed considerably the system might not boot anymore (e.g. when the boot partition uses incompatible device trees, like during update from HAOS 15.2 to 16.0 on Raspberry Pi based devices as well as Yellow, see https://github.com/home-assistant/operating-system/issues/4150).

### Additional context

- Related issue: https://github.com/home-assistant/operating-system/issues/4150

PR #6024, opened 2025-07-15
Mark system as unhealthy if multiple OS installations are found

agners:
## Proposed change
The HAOS update system relies on partition label and partition UUIDs to find the correct partition to update. If for some reason the user has multiple HAOS installations on a system, we might update the "wrong" partition.

This adds a check for dupplicate parittions, and  marks the system as unhealthy in case multiple partitions with the same partition label or UUID is found.

Implements: #6021

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to documentation pull request: https://github.com/home-assistant/home-assistant.io/pull/40022
- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/resolution/checks/duplicate_os_installation.py
+++ b/supervisor/resolution/checks/duplicate_os_installation.py
@@ -72,7 +72,7 @@ class CheckDuplicateOSInstallation(CheckBase):
                     identifier,
                     spec_type,
                     len(resolved),
-                    ", ".join(resolved),
+                    ", ".join(str(device.device) for device in resolved),
                 )
                 self.sys_resolution.add_unhealthy_reason(
                     UnhealthyReason.DUPLICATE_OS_INSTALLATION
--- a/supervisor/resolution/checks/duplicate_os_installation.py
+++ b/supervisor/resolution/checks/duplicate_os_installation.py
@@ -10,13 +10,22 @@ from .base import CheckBase
 
 _LOGGER: logging.Logger = logging.getLogger(__name__)
 
-# Partition labels to check for duplicates
+# Partition labels to check for duplicates (GPT-based installations)
 HAOS_PARTITIONS = [
     "hassos-boot",
     "hassos-kernel0",
     "hassos-kernel1",
     "hassos-system0",
     "hassos-system1",
+]
+
+# Partition UUIDs to check for duplicates (MBR-based installations)
+HAOS_PARTITION_UUIDS = [
+    "48617373-01",  # hassos-boot
+    "48617373-05",  # hassos-kernel0
+    "48617373-06",  # hassos-system0
+    "48617373-07",  # hassos-kernel1
+    "48617373-08",  # hassos-system1
 ]
 
 
@@ -30,6 +39,7 @@ class CheckDuplicateOSInstallation(CheckBase):
 
     async def run_check(self) -> None:
         """Run check if not affected by issue."""
+        # Check GPT-based installations (partition labels)
         for partition_label in HAOS_PARTITIONS:
             try:
                 resolved = await self.sys_dbus.udisks2.resolve_device(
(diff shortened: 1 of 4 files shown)


agners:
Example output from a Yellow with two installations:
```
2025-07-15 16:10:04.754 INFO (MainThread) [supervisor.resolution.checks.base] Run check for duplicate_os_installation/system
2025-07-15 16:10:04.859 WARNING (MainThread) [supervisor.resolution.checks.duplicate_os_installation] Found duplicate OS installation: hassos-boot partition exists on 2 devices
2025-07-15 16:10:04.859 INFO (MainThread) [supervisor.resolution.module] Create new issue duplicate_os_installation - system / None
```

sairon:
PR for the page reachable through the "Learn more" link:
- https://github.com/home-assistant/home-assistant.io/pull/40022

agners:
Updated message with some more info:
```
2025-07-15 17:43:57.635 INFO (MainThread) [supervisor.resolution.checks.base] Run check for duplicate_os_installation/system
2025-07-15 17:43:57.734 WARNING (MainThread) [supervisor.resolution.checks.duplicate_os_installation] Found duplicate OS installation: hassos-boot partition exists on 2 devices (/dev/mmcblk0p1, /dev/nvme0n1p1)
2025-07-15 17:43:57.735 INFO (MainThread) [supervisor.resolution.module] Create new issue duplicate_os_installation - system / None

```

supervisor/resolution/checks/duplicate_os_installation.py
```diff
@@ -0,0 +1,108 @@
...
+]
+
+# Partition UUIDs to check for duplicates (MBR-based installations)
+HAOS_PARTITION_UUIDS = [
+    "48617373-01",  # hassos-boot
+    "48617373-05",  # hassos-kernel0
+    "48617373-06",  # hassos-system0
+    "48617373-07",  # hassos-kernel1
+    "48617373-08",  # hassos-system1
+]
```
mdegat01, line 29: Is this likely to change in the future or not really? Just wondering if it should be loaded from an API or put in the version repository instead of here.

supervisor/resolution/checks/duplicate_os_installation.py
```diff
@@ -0,0 +1,108 @@
...
+                    len(resolved),
+                    ", ".join(str(device.device) for device in resolved),
+                )
+                self.sys_resolution.add_unhealthy_reason(
+                    UnhealthyReason.DUPLICATE_OS_INSTALLATION
+                )
+                self.sys_resolution.create_issue(
+                    IssueType.DUPLICATE_OS_INSTALLATION,
+                    ContextType.SYSTEM,
+                )
```
mdegat01, line 83: Can we help them fix this with a suggestion here? Maybe a suggestion for each installation with a fixup that removes that one when applied? Else we'll need some very technical doc telling users what to do to fix this. Though I guess anyone that got themselves in this situation is likely already technical since I assume it involved some amount of mucking about in the host system shell?

tests/resolution/check/test_check_duplicate_os_installation.py
```diff
@@ -0,0 +1,201 @@
...
+from supervisor.resolution.const import ContextType, IssueType, UnhealthyReason
+
+
+async def test_base(coresys: CoreSys):
+    """Test check basics."""
+    duplicate_os_installation = CheckDuplicateOSInstallation(coresys)
+    assert duplicate_os_installation.slug == "duplicate_os_installation"
+    assert duplicate_os_installation.enabled
+
+
```
mdegat01, line 23:
```suggestion

@pytest.mark.usefixtures("os_available")
```
Think this should get your test passing, I assume right now its kicking out on the "is OS" check. You'll have to add this to the rest too.

mdegat01 (approved):
I'm fine with the code and the suggestion is really just to fix the broken tests. Let me know what you think about the two comments though.

supervisor/resolution/checks/duplicate_os_installation.py
agners: This are essentially as static as the partition names above.

Most likely any added board will use the GPT partition table and hence the labels.

If anything, we'll move away from this label tied approach entirely, determine the partitions through other means, so a duplicate installation won't be a problem anymore. Until then most likely these just stay the same.

That is at least my take on this, @sairon any thoughts?

supervisor/resolution/checks/duplicate_os_installation.py
agners: I guess a possible fix would be renaming the partition labels (resp. change the MBR disk-signature) so we won't pick those up. But determining which one of the two is the right is essentially impossible. The problem is, we can't really tell what boot/kernel partition the bootloader used to boot the system, so we can't suggest a "correct" fix.

Depending on the SoC, we might actually have the information from the SoC boot ROM through some vendor specific format, which we'd need to translate in U-Boot, and then again pass to the OS.

But at that point, we better alter the boot flow to not rely on labels.

For now, the user will see an unsupported message, with the more info containing information how to resolve the problem, see https://github.com/home-assistant/home-assistant.io/pull/40022.

mdegat01: Yea I wasn't suggesting we try to guess the right one. I was thinking multiple suggestions, one for each partition with the duplicate name. They apply the one that renames the bad partition and the issue goes awa. But I suppose if we can't figure out the right one we can't provide them enough information to make that decision anyway.

agners: > But I suppose if we can't figure out the right one we can't provide them enough information to make that decision anyway.

Yeah that is pretty much the problem here. And I guess those who get which one is the right, figure that out on their own.

The more info page advices to create a full backup and download, then clear all and install the OS again. This is the safest way to go about it.

PR #6041, opened 2025-07-24
Add unsupported reason os_version and evaluation

mdegat01:
## Proposed change

Only the last 4 versions of OS are supported. Add an evaluation that marks system as unsupported if running a version of OS older then that.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to documentation pull request: https://github.com/home-assistant/home-assistant.io/pull/40132
- Link to cli pull request:
- Link to client library pull request: https://github.com/home-assistant-libs/python-supervisor-client/pull/140
- Link to Core PR for repair support: https://github.com/home-assistant/core/pull/149837

## Checklist

### DIFF
--- a/supervisor/resolution/const.py
+++ b/supervisor/resolution/const.py
@@ -49,6 +49,7 @@ class UnsupportedReason(StrEnum):
     NETWORK_MANAGER = "network_manager"
     OS = "os"
     OS_AGENT = "os_agent"
+    OS_VERSION = "os_version"
     PRIVILEGED = "privileged"
     RESTART_POLICY = "restart_policy"
     SOFTWARE = "software"
@@ -58,7 +59,6 @@ class UnsupportedReason(StrEnum):
     SYSTEMD_JOURNAL = "systemd_journal"
     SYSTEMD_RESOLVED = "systemd_resolved"
     VIRTUALIZATION_IMAGE = "virtualization_image"
-    OS_VERSION = "os_version"
 
 
 class UnhealthyReason(StrEnum):
--- a/supervisor/resolution/const.py
+++ b/supervisor/resolution/const.py
@@ -58,6 +58,7 @@ class UnsupportedReason(StrEnum):
     SYSTEMD_JOURNAL = "systemd_journal"
     SYSTEMD_RESOLVED = "systemd_resolved"
     VIRTUALIZATION_IMAGE = "virtualization_image"
+    OS_VERSION = "os_version"
 
 
 class UnhealthyReason(StrEnum):
--- a/supervisor/resolution/evaluations/os_version.py
+++ b/supervisor/resolution/evaluations/os_version.py
@@ -1,0 +1,51 @@
+"""Evaluation class for OS version."""
+
+from awesomeversion import AwesomeVersion, AwesomeVersionException
+
+from ...const import CoreState
+from ...coresys import CoreSys
+from ..const import UnsupportedReason
+from .base import EvaluateBase
+
+
+def setup(coresys: CoreSys) -> EvaluateBase:
+    """Initialize evaluation-setup function."""
+    return EvaluateOSVersion(coresys)
+
+
+class EvaluateOSVersion(EvaluateBase):
+    """Evaluate the OS version."""
+
+    @property
+    def reason(self) -> UnsupportedReason:
+        """Return a UnsupportedReason enum."""
+        return UnsupportedReason.OS_VERSION
+
+    @property
+    def on_failure(self) -> str:
+        """Return a string that is printed when self.evaluate is True."""
+        return f"OS version '{self.sys_os.version}' is more than 4 versions behind the latest '{self.sys_os.latest_version}'!"
+
+    @property
+    def states(self) -> list[CoreState]:
+        """Return a list of valid states when this evaluation can run."""
+        # Technically there's no reason to run this after STARTUP as update requires
+        # a reboot. But if network is down we won't have latest version info then.
+        return [CoreState.RUNNING, CoreState.STARTUP]
+
+    async def evaluate(self) -> bool:
+        """Run evaluation."""
+        if (
+            not self.sys_os.available
+            or not (current := self.sys_os.version)
+            or not (latest := self.sys_os.latest_version)
+            or not latest.major
+        ):
+            return False
+
+        # If current is more than 4 major versions behind latest, mark as unsupported
+        last_supported_version = AwesomeVersion(f"{int(latest.major) - 4}.0")
+        try:
+            return current < last_supported_version
+        except AwesomeVersionException:
+            return False
--- a/supervisor/resolution/evaluations/os_version.py
+++ b/supervisor/resolution/evaluations/os_version.py
@@ -48,4 +48,4 @@ class EvaluateOSVersion(EvaluateBase):
         try:
             return current < last_supported_version
         except AwesomeVersionException:
-            return False
+            return True
--- a/tests/resolution/evaluation/test_evaluate_os_version.py
+++ b/tests/resolution/evaluation/test_evaluate_os_version.py
@@ -30,7 +30,7 @@ from supervisor.resolution.evaluations.os_version import EvaluateOSVersion
         (
             "9ccda431973acf17e4221850b08f3280b723df8d",
             "15.0",
-            False,
+            True,
         ),  # Dev setup running on a commit hash, check skipped
     ],
 )
--- a/tests/resolution/evaluation/test_evaluate_os_version.py
+++ b/tests/resolution/evaluation/test_evaluate_os_version.py
@@ -21,6 +21,10 @@ from supervisor.resolution.evaluations.os_version import EvaluateOSVersion
         ("10.5", "10.5", False),  # up to date, supported
         ("10.5", "10.6", False),  # same major, supported
         ("10.0", "13.3", False),  # 3 major behind, supported
+        ("10.2.dev20240321", "15.0", True),  # 5 major behind, dev version, unsupported
+        ("10.2.dev20240321", "13.0", False),  # 3 major behind, dev version, supported
+        ("10.2.rc2", "15.0", True),  # 5 major behind, rc version, unsupported
+        ("10.2.rc2", "13.0", False),  # 3 major behind, rc version, supported
         (None, "15.0", False),  # No current version info, check skipped
         ("2.0", None, False),  # No latest version info, check skipped
         (
--- a/tests/resolution/evaluation/test_evaluate_os_version.py
+++ b/tests/resolution/evaluation/test_evaluate_os_version.py
@@ -1,0 +1,79 @@
+"""Test OS Version evaluation."""
+
+from unittest.mock import PropertyMock, patch
+
+from awesomeversion import AwesomeVersion
+import pytest
+
+from supervisor.const import CoreState
+from supervisor.coresys import CoreSys
+from supervisor.os.manager import OSManager
+from supervisor.resolution.evaluations.os_version import EvaluateOSVersion
+
+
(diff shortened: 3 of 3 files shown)


supervisor/resolution/evaluations/os_version.py
```diff
@@ -0,0 +1,51 @@
...
+            or not latest.major
+        ):
+            return False
+
+        # If current is more than 4 major versions behind latest, mark as unsupported
+        last_supported_version = AwesomeVersion(f"{int(latest.major) - 4}.0")
+        try:
+            return current < last_supported_version
+        except AwesomeVersionException:
+            return False
```
agners: From what I can tell we should really be comparing compatible versions here :thinking: . Should we log or capture that exception maybe?

supervisor/resolution/const.py
```diff
@@ -58,6 +58,7 @@ class UnsupportedReason(StrEnum):
     SYSTEMD_JOURNAL = "systemd_journal"
     SYSTEMD_RESOLVED = "systemd_resolved"
     VIRTUALIZATION_IMAGE = "virtualization_image"
+    OS_VERSION = "os_version"
```
agners: All other entries seem to be ordered by alphabet, let's add this one in the right spot as well.

agners (commented):
This will raise a repair with the title "Unsupported system - os_version" and text "System is unsupported due to os_version. Use the link to learn more and how to fix this.". I guess the full string should be "Home Assistant OS version" :thinking: 

Ideally we get translation strings in for this release (similar to https://github.com/home-assistant/frontend/pull/26190).

So this will mark everything older than 12.0 as unsupported (all HAOS installations with major version < (16 - 4)). From analytics, this are about 3.5% of users, around 11k according to opt-in statistics. This is well within [ADR0015 supported version definition](https://github.com/home-assistant/architecture/blob/master/adr/0015-home-assistant-os.md#supported-operating-systems-and-versions), so from that perspective we should be fine. But we should have documentation up on why we do this before promoting a Supervisor version with this code to stable.

supervisor/resolution/evaluations/os_version.py
sairon: Agreed, everything that comes out of our CI should be either an `AwesomeVersion CalVer` or `AwesomeVersion PEP 440` which are comparable, so only if someone has a homebrew build with some wacky version string this should fail.

tests/resolution/evaluation/test_evaluate_os_version.py
```diff
@@ -0,0 +1,79 @@
...
+        ("10.5", "10.5", False),  # up to date, supported
+        ("10.5", "10.6", False),  # same major, supported
+        ("10.0", "13.3", False),  # 3 major behind, supported
+        (None, "15.0", False),  # No current version info, check skipped
+        ("2.0", None, False),  # No latest version info, check skipped
+        (
+            "9ccda431973acf17e4221850b08f3280b723df8d",
+            "15.0",
+            False,
+        ),  # Dev setup running on a commit hash, check skipped
```
sairon, line 34: Maybe I'd add cases for some valid pre-release version strings in the mix, like `12.2.dev20240321` or `10.0.rc2`.

sairon (commented):
Stefan said it all, I agree with the deprecation proposal as well.

tests/resolution/evaluation/test_evaluate_os_version.py
agners: Good point. FWIW, I don't think we have OS versions with git hashes, do we? :thinking:

sairon: No, local build will usually be something like `16.1.dev0` without any meta information.

sairon: Other than that, there are also branch dev builds which have Unix timestamp suffix instead of YYYYMMDD, e.g. `dev1744721960`.

supervisor/resolution/evaluations/os_version.py
mdegat01: I was thinking of dev builds actually. Do you all always pull a published version or do you sometimes build from a particular commit or branch to test something?

mdegat01: The problem with logging or capturing it btw is this will run every time the healthcheck runs. So if it somehow fails we'll get a recurring event from that system or a recurring log for the local system. As you said there's no good reason for it to happen besides someone hacking around locally anyway so squashing it seemed fine. I can capture/log if we prefer but it'll be recurring if we do.

supervisor/resolution/const.py
mdegat01: I had claude start this PR and write the first pass. I guess it didn't pick up on the alphabetical order bit 😆

tests/resolution/evaluation/test_evaluate_os_version.py
mdegat01: Ok I can drop the commit hash test then. I was just trying to think of something that might occur that tested the version exception. Figured building from a commit or branch was something done once in a while but maybe not.

IF there's a valid test which tests that exception handling let me know. IF not I can drop it. I can also drop the exception handling if it should never come up. I'd probably prefer not to log or capture because of the recurring point I mentioned above but if that's what you both want I'll add it. Realistically that is more about OS then Supervisor and what we'd expect in the wild for OS.

mdegat01:
> This will raise a repair with the title "Unsupported system - os_version" and text "System is unsupported due to os_version. Use the link to learn more and how to fix this.". I guess the full string should be "Home Assistant OS version"

Good point. I'll put up a PR for that. I wonder if that should be a checklist item in the PR template? I guess its not that common but changes in `resolution` require it a lot 🤔 

But @agners your link confuses me. I didn't realize that was in the frontend codebase. Up until now I had been putting translation strings for these here:
https://github.com/home-assistant/core/blob/bddd4d621a98f5ba1f0c0e46b01d148e645b504c/homeassistant/components/hassio/strings.json#L118-L229

Do these not work? Or is there some reason we have to put this information in frontend as well? I'd prefer them to all be in the hassio component if possible to keep them in one place with the supervisor code in core...

supervisor/resolution/evaluations/os_version.py
agners: I am more worried that we hide an issue with this: E.g. what if the check gets run early, where we have no (valid) version available :thinking: . I guess chances of that happening are low so :shrug: .

tests/resolution/evaluation/test_evaluate_os_version.py
agners: > Figured building from a commit or branch was something done once in a while but maybe not.

That does happen, but builds from commit/branches still do not end up with version numbers which include commit hashes, at least the way the OS build scripts work currently. 

If we keep the exception handling discussed above, I agree then we should also keep that test case. So removing the git hash from this list should really be dependent on how the discussion above concludes.

supervisor/resolution/evaluations/os_version.py
mdegat01: That situation isn't handled by that `try-except`. That's handled here:
https://github.com/home-assistant/supervisor/blob/81137ad0517eb5c78828b77024882a28a8184327/supervisor/resolution/evaluations/os_version.py#L38-L44

Other then network issue blocking our ability to see updates it should be safe though. We can control when these run based on core state as you can see here: 
https://github.com/home-assistant/supervisor/blob/81137ad0517eb5c78828b77024882a28a8184327/supervisor/resolution/evaluations/os_version.py#L29-L34

The `SETUP` healthcheck is after all of Supervisor's various modules have been initially loaded.

agners: Hm, I see yeah this should all be correctly handled then, nice :+1: 

But in a way, this makes this exception even more unexpected, no? Completely unexpected situation should not silently discarded, but we should be aware so we can handle it properly.

I get that you considered git hashes to be a valid use case, but that is not the case (other than someone modifies OS build scripts etc., but with modified source anything can happen, we should not code for that case). So IMHO we should leave this unhandled.

mdegat01: I mostly just don't like noisy background processes. And since this will run regularly in the background without user interaction I don't like the idea of regular noise from exceptions.

That being said, maybe this exception handling should just return `True` instead of `False`? The point is to identify when you are running with an unsupported OS version. Running on a version that we can't compare to the official OS version is definitely unsupported.

agners: Yeah, that works for me, lets do that.

```suggestion
            return True
```

agners (approved):
LGTM!

supervisor/resolution/evaluations/os_version.py
agners: Yeah that will make it more likely that we'd notice if any strange version numbers make it here. Let's return `True` here :+1:

agners:
> But @agners your link confuses me. I didn't realize that was in the frontend codebase. Up until now I had been putting translation strings for these here: https://github.com/home-assistant/core/blob/bddd4d621a98f5ba1f0c0e46b01d148e645b504c/homeassistant/components/hassio/strings.json#L118-L229
> 
> Do these not work? Or is there some reason we have to put this information in frontend as well? I'd prefer them to all be in the hassio component if possible to keep them in one place with the supervisor code in core...

Hm, I see, it appears we have translations on both sides currently.

It seems that for the repairs itself the hassio integration translations are used.

But there is one place where we still use the frontend translations, namely in the System info page, where we show a list of unsupported/unhealthy reasons:

image

tests/resolution/evaluation/test_evaluate_os_version.py
```diff
@@ -0,0 +1,83 @@
...
+        ("10.2.dev20240321", "15.0", True),  # 5 major behind, dev version, unsupported
+        ("10.2.dev20240321", "13.0", False),  # 3 major behind, dev version, supported
+        ("10.2.rc2", "15.0", True),  # 5 major behind, rc version, unsupported
+        ("10.2.rc2", "13.0", False),  # 3 major behind, rc version, supported
+        (None, "15.0", False),  # No current version info, check skipped
+        ("2.0", None, False),  # No latest version info, check skipped
+        (
+            "9ccda431973acf17e4221850b08f3280b723df8d",
+            "15.0",
+            False,
```
agners:
```suggestion
            True,
```

supervisor/resolution/evaluations/os_version.py
```diff
@@ -0,0 +1,51 @@
...
+        # Technically there's no reason to run this after STARTUP as update requires
+        # a reboot. But if network is down we won't have latest version info then.
+        return [CoreState.RUNNING, CoreState.STARTUP]
+
+    async def evaluate(self) -> bool:
+        """Run evaluation."""
+        if (
+            not self.sys_os.available
+            or not (current := self.sys_os.version)
+            or not (latest := self.sys_os.latest_version)
```
agners, line 41: I've downgraded to OS 11.5 (which should be the last unsupported version) but it turns out this evaluation is actually not working correctly: this property has the latest version which we are allowed to upgrade, according to the OS upgrade path.

https://github.com/home-assistant/supervisor/blob/8a1e6b089535ac4f9f97b95a827b9a1c948218a2/supervisor/updater.py#L276-L285

For all recent major OS versions we mandate to upgrade to OS 15.2, so that property is 15.2. But 15 - 4 = 11, so the evaluation returns False. This pretty much makes all OS versions "supported" currently :see_no_evil: 

I guess we have to store both, the absolutely latest version and the next version we are allowed to upgrade.

mdegat01: Shoot. Right, the upgrade path 🙈

PR #6046, opened 2025-07-28
Storage space usage API

MindFreeze:
## Proposed change

We need this for better space management. To show charts like this in the frontend
image

Not sure if this needs to be a new API or if we should shove it into mounts. Reports only the main disk right now but eventually we'll need the same data for USB mounts and maybe some network mounts.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/api/__init__.py
+++ b/supervisor/api/__init__.py
@@ -37,6 +37,7 @@ from .resolution import APIResoulution
 from .root import APIRoot
 from .security import APISecurity
 from .services import APIServices
+from .storage import APIStorage
 from .store import APIStore
 from .supervisor import APISupervisor
 from .utils import api_process, api_process_raw
@@ -115,6 +116,7 @@ class RestAPI(CoreSysAttributes):
         self._register_services()
         self._register_store()
         self._register_supervisor()
+        self._register_storage()
 
         if static_resource_configs:
 
@@ -793,6 +795,16 @@ class RestAPI(CoreSysAttributes):
                 web.get("/docker/registries", api_docker.registries),
                 web.post("/docker/registries", api_docker.create_registry),
                 web.delete("/docker/registries/{hostname}", api_docker.remove_registry),
+            ]
+        )
+
+    def _register_storage(self) -> None:
+        """Register storage usage endpoint."""
+        api_storage = APIStorage()
+        api_storage.coresys = self.coresys
+        self.webapp.add_routes(
+            [
+                web.get("/storage/usage", api_storage.usage),
             ]
         )
 
--- a/supervisor/api/storage.py
+++ b/supervisor/api/storage.py
@@ -1,0 +1,92 @@
+"""Init file for the storage API."""
+
+from pathlib import Path
+
+from aiohttp import web
+
+from ..coresys import CoreSysAttributes
+from .utils import api_process
+
+ATTR_MAX_DEPTH = "max_depth"
+
+
+def get_dir_structure_sizes(path: Path, max_depth: int = 1):
+    """Return a recursive dict of subdirectories and their sizes, only if size > 0.
+
+    Excludes external mounts and symlinks to avoid counting files on other filesystems
+    or following symlinks that could lead to infinite loops or incorrect sizes.
+    """
+
+    size = 0
+    if not path.exists():
+        return {"size": size}
+
+    children = {}
+    root_device = path.stat().st_dev
+
+    for child in path.iterdir():
+        if not child.is_dir():
+            size += child.stat(follow_symlinks=False).st_size
+            continue
+
+        # Skip symlinks to avoid infinite loops
+        if child.is_symlink():
+            continue
+
+        try:
+            # Skip if not on same device (external mount)
+            if child.stat().st_dev != root_device:
+                continue
+        except (OSError, FileNotFoundError):
+            continue
+
+        child_result = get_dir_structure_sizes(child, max_depth - 1)
+        if child_result["size"] > 0:
+            size += child_result["size"]
+            if max_depth > 1:
+                children[child.name] = child_result
+
+    if children:
+        return {"size": size, "children": children}
+
+    return {"size": size}
+
+
+class APIStorage(CoreSysAttributes):
+    """Handle RESTful API for storage usage."""
+
+    @api_process
+    async def usage(self, request: web.Request) -> dict:
+        """Return a breakdown of storage usage for the system."""
+
+        max_depth = request.query.get(ATTR_MAX_DEPTH, 1)
+        try:
+            max_depth = int(max_depth)
+        except ValueError:
+            max_depth = 1
+
+        root_device = self.sys_config.path_homeassistant.stat().st_dev
+
+        async def dir_info(path):
+            return await self.sys_run_in_executor(
+                get_dir_structure_sizes, path, max_depth
+            )
+
+        addons = await dir_info(self.sys_config.path_addons_data)
+        media = await dir_info(self.sys_config.path_media)
+        share = await dir_info(self.sys_config.path_share)
+        backup = await dir_info(self.sys_config.path_backup)
+        tmp = await dir_info(self.sys_config.path_tmp)
+        config = await dir_info(self.sys_config.path_homeassistant)
+
+        return {
+            # Separate by device to allow for future expansion. This may be changed to partition in the future.
+            root_device: {
+                "addons": addons,
+                "media": media,
+                "share": share,
+                "backup": backup,
+                "tmp": tmp,
+                "config": config,
+            }
+        }


supervisor/api/host.py
```diff
@@ -289,3 +290,36 @@ async def advanced_logs(
...
+            )
+
+        return {
+            "size": used,
+            "children": {
+                "addons": await dir_info(self.sys_config.path_addons_data),
+                "media": await dir_info(self.sys_config.path_media),
+                "share": await dir_info(self.sys_config.path_share),
+                "backup": await dir_info(self.sys_config.path_backup),
+                "tmp": await dir_info(self.sys_config.path_tmp),
```
mdegat01: This folder is not user accessible. It's only used for temporary files within supervisor processes, no other container has access to it. So I'm not sure it belongs in this list? It feels like it belongs in a system bucket if we do that like your UI shows. Or if we trying to be very thorough even with system internal stuff there's likely a few more list.

For example: `path_addons_git` and `path_addons_core` for addon source files (can get pretty large with images and media and such). There's also `path_addons_local` though that one actually is user facing (it's mounted to `/addons` in other containers).

`path_core_backup` is another fully internal folder that could have large data. That is basically a place that only Home Assistant can put backups. It uses it when asked to make backups that are only intended for upload to a cloud backup location and not stored locally.

supervisor/api/host.py
```diff
@@ -289,3 +290,36 @@ async def advanced_logs(
...
+                "config": await dir_info(self.sys_config.path_homeassistant),
```
mdegat01: This should be called `homeassistant` not `config`. `/config` is legacy and in the process of being deprecated because it created a lot of confusion and made people think all config (including that of addons) should be in there. We don't want to bring that back.

On that note, we should also probably include `path_addon_configs` in here or include its size in `addons`. That is the folder where addons should put any config/files that they want users to be able to see and interact with outside the addon.

supervisor/api/host.py
```diff
@@ -289,3 +290,36 @@ async def advanced_logs(
...
+            },
```
mdegat01, line 320: `ssl` is missing from this list. I assume that's an oversight or did you think of that as part of the "system" bucket?

supervisor/api/host.py
```diff
@@ -289,3 +290,36 @@ async def advanced_logs(
...
+
+        async def dir_info(path):
+            return await self.sys_run_in_executor(
+                disk.get_dir_structure_sizes, path, max_depth
+            )
+
+        return {
+            "size": used,
+            "children": {
+                "addons": await dir_info(self.sys_config.path_addons_data),
```
mdegat01: As nice as the code looks this way, I think we may want to consider making one single function that we call in the executor that gets and returns all this information. Doing it this way will result in a lot more overhead scheduling tasks in the executor thread. Maybe something like this:

```py
def get_dir_sizes(disk: HwDisk, request: dict[str, Path], max_depth: int = 1) -> dict[str, Any]:
    """Accepts a dictionary of `name: Path` and returns a dictionary with `name: <size>`.

    Must be run in executor.
    """
    return {
        name: disk.get_dir_structure_sizes(path, max_depth)
        for name, path in request.items()
    }
```

supervisor/hardware/disk.py
```diff
@@ -53,17 +54,65 @@ def get_disk_total_space(self, path: str | Path) -> float:
...
+
+            # Skip symlinks to avoid infinite loops
+            if child.is_symlink():
+                continue
+
+            try:
+                # Skip if not on same device (external mount)
+                if child.stat().st_dev != root_device:
+                    continue
+            except (OSError, FileNotFoundError):
```
mdegat01: As a general practice within Supervisor, when accessing the filesystem if you encounter an `OSError` with `errno.EBADMSG` then we need to stop whatever we're doing and mark the system as unhealthy. This type of `OSError` is raised specifically when the disk is finished. It has reached the end of its lifecycle and cannot be written to, read from or both anymore. Usually happens with SDCards.

Take a look at this for an example of how we do that:
https://github.com/home-assistant/supervisor/blob/a004830131e673382fbb31205d6c65fc9428f34e/supervisor/supervisor.py#L180-L187

Can either raise or stop the loop at that point, whichever you prefer. If you receive that `errno` though then nothing further will work here and we shouldn't try.

mdegat01:
On the labels I added:
- Missing documentation - Supervisor's API is documented [here](https://developers.home-assistant.io/docs/api/supervisor/endpoints). So we'll need to update that for API changes
- Need client library - We'll need to add support for this to [our client library](https://github.com/home-assistant-libs/python-supervisor-client/) as Home Assistant is trying to use that for all supervisor access going forward
- Need cli - Less important but presumably it would be nice to access this info from the [cli](https://github.com/home-assistant/cli) in future

I can probably take care of these if you want, we just have to finalize the API first.

supervisor/api/host.py
MindFreeze: The buckets are very much open for discussion. These are just what came to mind initially

MindFreeze:
As we discussed, I updated the URL, tweaked the structure a bit and added "total_space" and "system". We can update the buckets more in the future but if the protocol is OK, I'll update the docs and lib.
IMO we don't need this in the cli as there you can just use `du` but we can add it if you want.

supervisor/api/__init__.py
```diff
@@ -198,6 +198,7 @@ def _register_host(self) -> None:
                 web.post("/host/reload", api_host.reload),
                 web.post("/host/options", api_host.options),
                 web.get("/host/services", api_host.services),
+                web.get("/host/disk/default/usage", api_host.disk_usage),
```
agners: For other API we use the plural, so `disks`.

We also discussed if we should fold the usage and other metrics (e.g. lifetime) into a single endpoint. Usually the frontend prefers if not too many calls need to be made. So this could just become `/host/disk/default`, but then also return the lifetime, and maybe even partition layout.

But then again, looking at the current implementation, we already have a query parameter. And I guess since getting exact usage can become an expensive operation, maybe it is better to have this as a separate endpoint.

@MindFreeze if you/the frontend team is fine to have it as a separate endpoint, I am fine too.

supervisor/api/host.py
```diff
@@ -289,3 +290,44 @@ async def advanced_logs(
...
+            },
+            max_depth,
+        )
+        return {
+            "total_space": total,
+            "used_space": used,
+            "children": {
+                **known_paths,
+                "system": {
+                    "used_space": used
```
agners: Hm, should we use that same structure for all types (above)?

supervisor/api/__init__.py
MindFreeze: This endpoint is going to be slow no matter what so frontend would not want to rely on it for other data that can be fetched instantly otherwise

supervisor/api/host.py
MindFreeze: What do you mean? It is the same recursive structure. Just `total_space` is only on the top level. `total_bytes` and `used_bytes` like you suggested would be fine as well

agners: Oh I see that is part of `get_dir_structure_sizes`. Nevermind!

agners:
I run this on my production instance with a depth of 2, this is the json I got back:

<details>

```json
{
  "result": "ok",
  "data": {
    "total_space": 984465879040,
    "used_space": 213384400896,
    "children": {
      "addons_data": {
        "used_space": 17918974193,
        "children": {
          "core_configurator": {
            "used_space": 173
          },
          "77f1785d_chip_tool": {
            "used_space": 2
          },
          "core_whisper": {
            "used_space": 3892688550
          },
          "a0d7b954_vscode": {
            "used_space": 176693578
          },
          "core_mosquitto": {
            "used_space": 845657
          },
          "982ee2c4_volvo2mqtt": {
            "used_space": 475
          },
          "core_ssh": {
            "used_space": 38000998
          },
          "a0d7b954_influxdb": {
            "used_space": 8218539131
          },
          "a0d7b954_grafana": {
            "used_space": 1215195
          },
          "a0d7b954_uptime-kuma": {
            "used_space": 671268550
          },
          "core_letsencrypt": {
            "used_space": 445994
          },
          "core_piper": {
            "used_space": 442690429
          },
          "a0d7b954_ssh": {
            "used_space": 6290
          },
          "core_git_pull": {
            "used_space": 2376
          },
          "a0d7b954_logviewer": {
            "used_space": 76
          },
          "core_mariadb": {
            "used_space": 4299634590
          },
          "core_silabs_flasher": {
            "used_space": 109
          },
          "core_speech-to-phrase": {
            "used_space": 132782908
          },
          "a0d7b954_spotify": {
            "used_space": 158
          },
          "a0d7b954_phpmyadmin": {
            "used_space": 70
          },
          "17fe96b6_puppet": {
            "used_space": 207
          },
          "a0d7b954_glances": {
            "used_space": 397
          },
          "core_matter_server": {
            "used_space": 44157845
          },
          "core_vlc": {
            "used_space": 67
          },
          "core_assist_microphone": {
            "used_space": 368
          }
        }
      },
      "addons_config": {
        "used_space": 20040109,
        "children": {
          "a0d7b954_emqx": {
            "used_space": 3355776
          },
          "core_matter_server": {
            "used_space": 16684333
          }
        }
      },
      "media": {
        "used_space": 619970,
        "children": {
          "frigate": {
            "used_space": 65536
          }
        }
      },
      "share": {
        "used_space": 18309324,
        "children": {
          "mosquitto": {
            "used_space": 116
          },
          "speech-to-phrase": {
            "used_space": 18309208
          }
        }
      },
      "backup": {
        "used_space": 174410946560
      },
      "ssl": {
        "used_space": 3267
      },
      "homeassistant": {
        "used_space": 134055761,
        "children": {
          "custom_components": {
            "used_space": 122467
          },
          "glances": {
            "used_space": 8981
          },
          "blueprints": {
            "used_space": 53063
          },
          "python-matter-server": {
            "used_space": 4351441
          },
          "ota": {
            "used_space": 240760
          },
          "iotawatt_ha": {
            "used_space": 75066
          },
          ".storage": {
            "used_space": 39609972
          },
          "custom_zha_quirks": {
            "used_space": 10573
          },
          "esphome": {
            "used_space": 2977631
          },
          ".git": {
            "used_space": 1527118
          },
          "automations": {
            "used_space": 26032
          },
          "frigate-hass-integration": {
            "used_space": 396310
          },
          "www": {
            "used_space": 371349
          },
          "hass-weather-srgssr": {
            "used_space": 521089
          },
          "tts": {
            "used_space": 10866909
          }
        }
      },
      "system": {
        "used_space": 20881451712
      }
    }
  }
}
```

</details>

This is from a rather simple installation on Home Assistant Green

<details>

```json
{
  "result": "ok",
  "data": {
    "total_space": 30016831488,
    "used_space": 5170909184,
    "children": {
      "addons_data": {
        "used_space": 5073,
        "children": {
          "core_ssh": {
            "used_space": 5073
          }
        }
      },
      "addons_config": {
        "used_space": 11067,
        "children": {
          "core_zwave_js": {
            "used_space": 11067
          }
        }
      },
      "media": {
        "used_space": 0
      },
      "share": {
        "used_space": 0
      },
      "backup": {
        "used_space": 11151360
      },
      "ssl": {
        "used_space": 0
      },
      "homeassistant": {
        "used_space": 2355866,
        "children": {
          ".storage": {
            "used_space": 66945
          },
          "blueprints": {
            "used_space": 5703
          }
        }
      },
      "system": {
        "used_space": 5157385818
      }
    }
  }
}
```

```
# du -sh *
6.2G    docker
484.4M  logs
16.0K   lost+found
20.0K   rngtest
61.4M   supervisor
1.3G    swapfile
```

</details>

What I generally notice is that quite some space is essentially unattributed and lands in system. I guess we could use the Docker API to get container image sizes. However, due to layer sharing using individual image sizes may add up to more space than actually occupied, which can lead to weird graphs in the end :see_no_evil:

agners:
When you look at the Home Assistant Green output, pretty much everything is accounted for System at this point.

```
# du -s -x *
3165276 docker
504204  logs
16      lost+found
20      rngtest
65036   supervisor
1325764 swapfile
```

Docker has the `/system/df` API (or `docker system df -v` command) which would allow for more detailed information. But as mentioned, the shared space makes things complicated to attribute :cry: 


```
REPOSITORY                                        TAG          IMAGE ID       CREATED        SIZE      SHARED SIZE   UNIQUE SIZE   CONTAINERS
agners/aarch64-hassio-supervisor                  latest       7e3dc674f09f   3 hours ago    490MB     53.77MB       436.2MB       1
ghcr.io/home-assistant/aarch64-hassio-cli         2025.08.1    b99ab04c1358   40 hours ago   62.7MB    53.77MB       8.972MB       1
ghcr.io/home-assistant/aarch64-hassio-audio       2025.08.0    de6ae9c3a553   3 days ago     96.2MB    53.77MB       42.39MB       1
ghcr.io/home-assistant/aarch64-hassio-dns         2025.08.0    31a532040358   3 days ago     77.5MB    53.77MB       23.73MB       1
ghcr.io/home-assistant/aarch64-hassio-multicast   2025.08.0    2afd5d4f11cb   3 days ago     53.9MB    53.77MB       170.7kB       1
ghcr.io/home-assistant/green-homeassistant        2025.8.0b2   fe0f4fde185d   10 days ago    2GB       0B            2GB           1
homeassistant/aarch64-addon-ssh                   9.18.0       54189453a186   7 weeks ago    132MB     0B            132.3MB       1
ghcr.io/home-assistant/aarch64-hassio-observer    2025.02.0    abf1bf606aca   5 months ago   9.53MB    0B            9.532MB       1
```

I am fine starting with the current rather simple and straight forward approach. It will help to have a rough overview where space goes, especially if users have local backups/media or add-ons with big user data (e.g. InfluxDB). But I think it will raise questions pretty quickly, e.g. especially on a newly installed system with a couple of Add-ons installed: Users will wonder why "Add-on" is still small, or "System" occupying so much space... Since the frontend for this API will be part of Core, we need to consider how we can evolve the API if needed.

MindFreeze:
> I run this on my production instance with a depth of 2, this is the json I got back:
> 
> This is from a rather simple installation on Home Assistant Green
> 
> What I generally notice is that quite some space is essentially unattributed and lands in system. I guess we could use the Docker API to get container image sizes. However, due to layer sharing using individual image sizes may add up to more space than actually occupied, which can lead to weird graphs in the end 🙈

Here are the charts from this json
image
image

MindFreeze:
Updated the url and structure as discussed. Also here is an example of another prod system
image

agners:
@MindFreeze one of the use case is certainly that users can learn which add-on uses lots of space. What is the current plan for the add-on names? The code currently returns directory names, which matches add-on slugs. is the idea that the frontend translates those or is the intention to do this in the backend later on?

MindFreeze:
Both are possible. With the current format, Supervisor can add the names as `label` or leave the labels empty and let frontend translate using the `id`. We can figure it out in the next PRs

MindFreeze:
The first iteration will only display the top level info anyway. Deeper visualizations will be done later

agners (approved):
LGTM, very nice addition, looking forward for this to land :star_struck:

PR #6074, opened 2025-08-06
Improve JobGroup locking with external ownership tracking

agners:
## Proposed change
Refactors JobGroup concurrency system to enable external lock management. This allows to release Job group locks outside of a running Job context, which allows the use of context manager in the decorator and execution limits combined with group concurrency parameters.

**JobGroup** (supervisor/jobs/job_group.py):
  - Add `release(job)` parameter to enable external lock release with ownership verification
  - Add `is_locked_by(job)` helper methods

**Job Decorator** (supervisor/jobs/decorator.py):
  - Add helper methods `_is_group_concurrency()` and `_is_group_throttle()` for cleaner logic separation
  - Remove all `cast(JobGroup, job_group)` calls by using proper type checking
  - Change from `asyncio.Semaphore()` to `asyncio.Lock()` for consistency with `JobGroup`
  - Enable throttling with group concurrency - remove artificial restriction since external lock release now works

**Tests** (tests/jobs/test_job_decorator.py):
  - Add test_group_concurrency_with_group_throttling() to verify the combination works correctly

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/jobs/decorator.py
+++ b/supervisor/jobs/decorator.py
@@ -447,8 +447,7 @@ class Job(CoreSysAttributes):
         """Release concurrency control locks."""
         if self._is_group_concurrency():
             # Group-level concurrency: delegate to job group
-            assert job_group is not None  # Should be validated during _post_init
-            job_group.release(job)
+            cast(JobGroup, job_group).release(job)
         elif self.concurrency in (JobConcurrency.REJECT, JobConcurrency.QUEUE):
             # Job-level concurrency: use semaphore
             if self.lock.locked():
@@ -460,9 +459,8 @@ class Job(CoreSysAttributes):
         """Handle concurrency control limits."""
         if self._is_group_concurrency():
             # Group-level concurrency: delegate to job group
-            assert job_group is not None  # Validated during _post_init
             try:
-                await job_group.acquire(
+                await cast(JobGroup, job_group).acquire(
                     job, wait=self.concurrency == JobConcurrency.GROUP_QUEUE
                 )
             except JobGroupExecutionLimitExceeded as err:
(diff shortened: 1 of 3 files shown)


supervisor/jobs/decorator.py
```diff
@@ -455,48 +441,56 @@ async def check_conditions(
...
+    ) -> None:
         """Release concurrency control locks."""
-        if self.concurrency == JobConcurrency.REJECT:
-            if self.lock.locked():
-                self.lock.release()
-        elif self.concurrency == JobConcurrency.QUEUE:
+        if self._is_group_concurrency():
+            # Group-level concurrency: delegate to job group
+            assert job_group is not None  # Should be validated during _post_init
+            job_group.release(job)
```
mdegat01: Is this actually better then cast? `cast` specifically has no runtime overhead and just informs the type checker. And raising an AssertionError here is just as catastrophic as a TypeError, its still bubbles all the way out as an unhandled exception.

mdegat01 (approved):
Looks good! Nice fix to the issue of lock management. I'm not sure I agree that using `assert` is better then simply using `cast` since it has a small amount of runtime overhead and the exception it raises if its condition is somehow false is just as catastrophic and confusing. But that's a nitpick, I'm fine with it if you like it better this way

supervisor/jobs/decorator.py
agners: Yeah good point, this does not really help. I hoped we could get rid of cast by separating `JobGroup` code paths from non-`JobGroup` code paths entirely, but this did lead to too much code duplication.

Went back to `cast`. The refactoring folded some calls into one already, so at least less casts then initially.

PR #6075, opened 2025-08-06
Avoid race condition on add-on installation

agners:
## Proposed change

The JobGroup concurrency decorator of the Addon class does not prevent concurrent installation since the AddonManager creates a separate object for each installation. Make sure only one add-on installation is running at a time by using the JobGroup decorator on the AddonManager install method instead.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/addons/manager.py
+++ b/supervisor/addons/manager.py
@@ -9,6 +9,8 @@ from typing import Self, Union
 
 from attr import evolve
 
+from supervisor.jobs.const import JobConcurrency
+
 from ..const import AddonBoot, AddonStartup, AddonState
 from ..coresys import CoreSys, CoreSysAttributes
 from ..exceptions import (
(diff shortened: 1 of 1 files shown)


mdegat01:
Does this happen with `update` and `uninstall` as well?  Maybe `uninstall` is ok but `update` is really no different from `install` from a docker perspective...

agners:
> Does this happen with `update` and `uninstall` as well? Maybe `uninstall` is ok but `update` is really no different from `install` from a docker perspective...

I don't think it is a problem for `update` or `uninstall` as in these cases we access the same object. In that case the `JobGroup` based concurrency control takes effect (`update` and `uninstall` have `JobConcurrency.GROUP_REJECT` in the `Addon` class). 

`install` is special because we *create* the Addon object (and hence the `JobGroup`) twice when running concurrently. It is really unique to `install` as this is the only place where we create `Addon` objects which is called from the user/API directly (`load` and `restore` create `Addon` object too, but I assume that in those cases concurrency is handled by the caller).

PR #6080, opened 2025-08-07
Handle git dir missing in load and pull

mdegat01:
## Proposed change

Improve handling for git information missing during startup and update of Addon stores. Currently a missing `.git` repo causes repeated failures that Supervisor does not fix. This ensures the repository is reset and put back on track.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/store/git.py
+++ b/supervisor/store/git.py
@@ -1,9 +1,11 @@
 """Init file for Supervisor add-on Git."""
 
 import asyncio
+from enum import StrEnum
 import errno
 import functools as ft
 import logging
+from os import listdir
 from pathlib import Path
 from tempfile import TemporaryDirectory
 
@@ -18,6 +20,32 @@ from ..utils import remove_folder
 from .validate import RE_REPOSITORY
 
 _LOGGER: logging.Logger = logging.getLogger(__name__)
+
+
+class _RepoDirectoryStatus(StrEnum):
+    """Basic directory status for repo."""
+
+    GOOD = "good"
+    EMPTY = "empty"
+    MISSING_GIT = "missing_git"
+
+
+def _check_repo_directory(path: Path) -> _RepoDirectoryStatus:
+    """Check repository directory.
+
+    Must be run in executor.
+    """
+    if not path.is_dir():
+        return _RepoDirectoryStatus.EMPTY
+
+    if (path / ".git").is_dir():
+        return _RepoDirectoryStatus.GOOD
+
+    return (
+        _RepoDirectoryStatus.MISSING_GIT
+        if listdir(path)
+        else _RepoDirectoryStatus.EMPTY
+    )
 
 
 class GitRepo(CoreSysAttributes):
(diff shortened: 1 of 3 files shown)


mdegat01:
Whoops. Tests are now actually deleting fixtures, that's not right

agners:
The error during startup in this situation is the following:

```
2025-08-11 11:15:48.276 ERROR (MainThread) [supervisor.store.git] Can't clone https://github.com/blakeblackshear/frigate-hass-addons.git repository: Cmd('git') failed due to: exit code(128)
  cmdline: git clone -v --recursive --depth=1 --shallow-submodules -- https://github.com/blakeblackshear/frigate-hass-addons.git /data/addons/git/95168ca3
  stderr: 'fatal: destination path '/data/addons/git/95168ca3' already exists and is not an empty directory.
'.
2025-08-11 11:15:48.276 ERROR (MainThread) [supervisor.store] Can't retrieve data from https://github.com/blakeblackshear/frigate-hass-addons.git due to
```

supervisor/store/git.py
```diff
@@ -20,6 +22,32 @@
...
+def _check_repo_directory(path: Path) -> _RepoDirectoryStatus:
+    """Check repository directory.
+
+    Must be run in executor.
+    """
+    if not path.is_dir():
+        return _RepoDirectoryStatus.EMPTY
+
+    if (path / ".git").is_dir():
+        return _RepoDirectoryStatus.GOOD
```
agners, line 42: `GOOD` could be a bit misleading here, in a way this does not guarantee that the git repository is actually in a working state. E.g. if `.git` directory is there but empty (`HEAD` etc. is missing), `git fsck` still complains with `fatal: not a git repository (or any of the parent directories): .git`. 

I've tested that situation with this change (`.git` exists, but `HEAD` missing), and it seems the patch goes down the "good" path, but then successfully resets the git repo:

```
2025-08-11 16:10:30.269 ERROR (MainThread) [supervisor.store.git] Can't load /data/addons/git/17fe96b6
2025-08-11 16:10:30.269 ERROR (MainThread) [supervisor.store] Can't load data from repository https://github.com/agners/balloob-home-assistant-addons due to 
2025-08-11 16:10:30.269 INFO (MainThread) [supervisor.resolution.module] Create new suggestion execute_reset - store / 17fe96b6
2025-08-11 16:10:30.270 INFO (MainThread) [supervisor.resolution.module] Create new issue fatal_error - store / 17fe96b6
...
2025-08-11 16:10:31.002 INFO (MainThread) [supervisor.resolution.fixups.store_execute_reset] Reset corrupt Store: 17fe96b6
```

In a way, I think simply remove the `.git` check again (essentially revert #5701), should be sufficient here, all other cases we just "assume" good, and rely on exceptions in case the git repo is corrupted.

If we want to determine if the git repository is actually good, we could use `git fsck`. But if exception handling of regular load works, I wonder if that would just make code unnecessary complicated :thinking: .

mdegat01: Is there something wrong with the check as written other then the name `GOOD`? I tried to describe that the enum was about the basic directory status. Like as in does it exist and does it have `.git`. I know that doesn't mean its good from a git perspective though, that's what I meant by "basic directory status". If you think calling it `GOOD` is confusing I can change that name though.

We already have a `git fsck` in there later so users will get that status. I'm fine with reverting all this if you want but it seems like a useful additional safety check. `.git` could go missing in a file corruption situation and then this handles it.

agners: My point is, that `.git` and `.git/HEAD` missing (and probably tons of other situation) are the same from a git perspective.

But we make a distinction here:
- `.git` missing is handled specifically via reset (which I think fails with an exception, since `.git` is missing... but it leads to a repair issue which then replaces)
- `.git/HEAD` missing isn't handled specifically, but runs into an exception too, which then leads to a repair issue down the road.

In the end, I think the current code over-complicates things unnecessary. The crucial fix here is that we _only_ try to clone when the  _root_ does *not* exists. As soon as it exists, we assume it is a git repo, and just attempt to use git commands. If that fails, then we resort to recreate it with the new logic.

So in a way, our error handling is good, I don't think we need to do something smart here.

agners:
Closing in favor of #6114.

PR #6094, opened 2025-08-12
Improve VLAN configuration

agners:
## Proposed change
The connection name for VLANs should include the parent interface name for better identification. This was originally the intention, but the interface object's name property was used which appears empty at that point.

Also only allow a single connection per interface and VLAN id. The regular network commands can be used to alter the configuration.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/dbus/network/setting/generate.py
+++ b/supervisor/dbus/network/setting/generate.py
@@ -176,9 +176,13 @@ def get_connection_from_interface(
 
     # Generate/Update ID/name
     if not name or not name.startswith("Supervisor"):
-        name = f"Supervisor {interface.name}"
         if interface.type == InterfaceType.VLAN:
-            name = f"{name}.{cast(VlanConfig, interface.vlan).id}"
+            # For VLANs, interface.name appears empty. Use the parent interface name
+            # since that is how we specify VLAN parent interface on creation
+            vlan_config = cast(VlanConfig, interface.vlan)
+            name = f"Supervisor {vlan_config.interface}.{vlan_config.id}"
+        else:
+            name = f"Supervisor {interface.name}"
 
     if interface.type == InterfaceType.ETHERNET:
         iftype = "802-3-ethernet"
--- a/supervisor/host/network.py
+++ b/supervisor/host/network.py
@@ -34,7 +34,7 @@ from ..jobs.const import JobCondition
 from ..jobs.decorator import Job
 from ..resolution.checks.network_interface_ipv4 import CheckNetworkInterfaceIPV4
 from .configuration import AccessPoint, Interface
-from .const import InterfaceMethod, WifiMode
+from .const import InterfaceMethod, InterfaceType, WifiMode
 
 _LOGGER: logging.Logger = logging.getLogger(__name__)
 
@@ -283,6 +283,15 @@ class NetworkManager(CoreSysAttributes):
 
         # Create new interface (like vlan)
         elif not inet:
+            # For VLAN interfaces, check if one already exists with same ID on same parent
+            if interface.type == InterfaceType.VLAN and interface.vlan:
+                vlan_interface_name = f"{interface.vlan.interface}.{interface.vlan.id}"
+                if vlan_interface_name in self.sys_dbus.network:
+                    raise HostNetworkError(
+                        f"VLAN {interface.vlan.id} already exists on interface {interface.vlan.interface}",
+                        _LOGGER.error,
+                    )
+
             settings = get_connection_from_interface(interface, self.sys_dbus.network)
 
             try:
--- a/supervisor/dbus/network/setting/__init__.py
+++ b/supervisor/dbus/network/setting/__init__.py
@@ -273,8 +273,8 @@ class NetworkSetting(DBusInterface):
         if CONF_ATTR_VLAN in data:
             if CONF_ATTR_VLAN_ID in data[CONF_ATTR_VLAN]:
                 self._vlan = VlanProperties(
-                    data[CONF_ATTR_VLAN][CONF_ATTR_VLAN_ID],
-                    data[CONF_ATTR_VLAN].get(CONF_ATTR_VLAN_PARENT),
+                    id=data[CONF_ATTR_VLAN][CONF_ATTR_VLAN_ID],
+                    parent=data[CONF_ATTR_VLAN].get(CONF_ATTR_VLAN_PARENT),
                 )
             else:
                 self._vlan = None


supervisor/host/network.py
```diff
@@ -283,6 +283,15 @@ async def apply_changes(
...
         # Create new interface (like vlan)
         elif not inet:
+            # For VLAN interfaces, check if one already exists with same ID on same parent
+            if interface.type == InterfaceType.VLAN and interface.vlan:
+                vlan_interface_name = f"{interface.vlan.interface}.{interface.vlan.id}"
+                if vlan_interface_name in self.sys_dbus.network:
+                    raise HostNetworkError(
+                        f"VLAN {interface.vlan.id} already exists on interface {interface.vlan.interface}",
+                        _LOGGER.error,
+                    )
```
mdegat01: Since `interface.name` is always empty for vlans, won't it go into this code path any time you are adding or updating a vlan interface? Currently I only see a test case for adding a new vlan in `tests/api/test_network.py`. I would recommend adding one for updating the details of an existing vlan to make sure this doesn't break that.

And if update does go down a separate code path somehow then a test case for the scenario you're trying to prevent (creating duplicate vlans) would be good. But I think the comment is misleading and the reason we don't have this check is to allow updates of existing vlan interfaces we've created.

supervisor/dbus/network/setting/generate.py
```diff
@@ -176,9 +176,13 @@ def get_connection_from_interface(
 
     # Generate/Update ID/name
     if not name or not name.startswith("Supervisor"):
-        name = f"Supervisor {interface.name}"
         if interface.type == InterfaceType.VLAN:
-            name = f"{name}.{cast(VlanConfig, interface.vlan).id}"
+            # For VLANs, interface.name appears empty. Use the parent interface name
+            # since that is how we specify VLAN parent interface on creation
+            vlan_config = cast(VlanConfig, interface.vlan)
+            name = f"Supervisor {vlan_config.interface}.{vlan_config.id}"
```
mdegat01: How will we handle backwards compatibility here? This is a sensible change but the result will be we abandon all existing vlan interfaces we've created right? Since when we generate the settings argument to pass to NM it will have a new name. So if someone goes to update an existing vlan interface they'll really just get a new one right? And the existing one will remain as well, unmodified.

We also don't have a `DELETE /network/interface/<interface>/vlan/<id>` API right now so there's no good way for them to clean this old vlan up without going to nmcli... 🙈

mdegat01 (changes requested):
It feels weird to do request changes without requesting specific code changes but I guess I do have feedback that needs addressing so I'll use it. I like the idea but I think there's a backwards compatibility issue here. And I think that code path in `apply_changes` is actually used for more then just create.

supervisor/host/network.py
agners: Updating a VLAN is handled by regular `ha network update` command, just using the interface with VLAN id (as in `<interface>.<vlanid>`). I think this code path is really only used to create VLAN, and only that.

supervisor/dbus/network/setting/generate.py
agners: > How will we handle backwards compatibility here? This is a sensible change but the result will be we abandon all existing vlan interfaces we've created right? Since when we generate the settings argument to pass to NM it will have a new name. So if someone goes to update an existing vlan interface they'll really just get a new one right? And the existing one will remain as well, unmodified.

I am not really sure how we identify the connection settings today for VLANs, I was assuming that this is based on actual device name and VLAN ID, not the given connection ID, but I might be wrong on that.

What I realized is that with this patch applied, when I update the settings, it generates a new config, and uses the interface UUID in the connection name :see_no_evil: 

```
$ nmcli con show
NAME                                                UUID                                  TYPE      DEVICE    
Supervisor enp1s0                                   1977b0fd-f9d0-43af-b98e-5f2ab6b1725d  ethernet  enp1s0    
lo                                                  de6ac105-924d-4574-8d0b-64ccf93ab654  loopback  lo        
Test2                                               ee736ea0-e2cc-4cc5-9c35-d6df94a56b47  ethernet  enp7s0    
Supervisor 1977b0fd-f9d0-43af-b98e-5f2ab6b1725d.10  93d8f537-4828-4cc7-be0f-0c5a588a84c0  vlan      enp1s0.10 
Supervisor 1977b0fd-f9d0-43af-b98e-5f2ab6b1725d.11  02120d96-4914-4ba6-a321-b4a82b0661be  vlan      enp1s0.11
```

So this definitely need some more thought.

> We also don't have a `DELETE /network/interface/<interface>/vlan/<id>` API right now so there's no good way for them to clean this old vlan up without going to nmcli... 🙈

Yes, that would be a nice addition as well :smile:

supervisor/host/network.py
mdegat01: Is `interface.name` filled in then on the update call? Because `ha network update` definitely calls this same `apply_changes` function. I thought you found that `interface.name` was always empty for vlan interfaces? Maybe I'm mixing up the two different `NetworkInterface` objects though, I always found it a bit confusing that we have one in `dbus` and one in `host`.

agners: > Is `interface.name` filled in then on the update call?

Yes, on the update call it is filled. It is only empty on the creation call (e.g. `ha network vlan enp1s0 10 --ipv4-method auto --ipv6-method disabled`). 

It seems that #4986 really didn't fix the problem properly, as on update the new code still causes the network configuration to be named `Supervisor enp1s0.10.10` :see_no_evil: 

Maybe it would be better to align the creation call with the update situation, where `interface.name` is set properly, and revert #4986 :thinking: .

agners: > Maybe it would be better to align the creation call with the update situation, where `interface.name` is set properly, and revert #4986 🤔 .

I've ended up implementing this.

supervisor/host/configuration.py
```diff
@@ -82,6 +82,11 @@ class VlanConfig:
     """Represent a vlan configuration."""
 
     id: int
+    # Note: On VLAN creation, parent is the interface name, but in the NetworkManager
+    # config the parent is set to the connection UUID in get_connection_from_interface().
+    # On network update (which we call in apply_changes() on VLAN creation), the
+    # parent is then set to that connection UUID in _map_nm_vlan(), hence we always
+    # operate with a connection UUID as interface!
```
agners, line 89: So yeah, this is already today the case, essentially since #4633.

```yaml
- connected: true
  enabled: true
  interface: enp1s0.10
  ipv4:
    address: []
    gateway: null
    method: auto
    nameservers: []
    ready: false
  ipv6:
    addr_gen_mode: default
    address:
    - fe80::a011:ec5f:ad9a:7ff6/64
    gateway: null
    ip6_privacy: default
    method: disabled
    nameservers: []
    ready: true
  mac: 52:54:00:2B:36:80
  primary: false
  type: vlan
  vlan:
    id: 10
    parent: 1977b0fd-f9d0-43af-b98e-5f2ab6b1725d
  wifi: null
```

We could argue this was a breaking change, as the output of network information changed. Maybe we should convert the connection UUID to the (current) parent interface name in the output :thinking: 

Since this is nothing new caused by this PR, I'd rather prefer to address any change in what parent represents in a separate PR.

supervisor/host/network.py
```diff
@@ -209,20 +211,54 @@ async def update(self, *, force_connectivity_check: bool = False):
...
+                f"Can't create new interface: {err}", _LOGGER.error
+            ) from err
+
+        await self.update(force_connectivity_check=True)
+
     async def apply_changes(
         self, interface: Interface, *, update_only: bool = False
     ) -> None:
         """Apply Interface changes to host."""
         inet: NetworkInterface | None = None
```
mdegat01:
```suggestion
```
I'm very confused why is not yelling about this since you dropped all the `None` checks in the if conditions below. But there's no reason to do this anymore. `self.sys_dbus.network.get` will always return a `NetworkInterface` object or raise, it will never return `None`. You can remove this declaration and the type can be correctly inferred below when the variable is set.

mdegat01 (approved):
I'm a little worried about the duplicate vlan entries but i guess if its not new behavior its ok. Small nitpick tweak (sort of, it seems like mypy should be unhappy right now and isn't which is odd 🤔 ). But in general LGTM 👍

supervisor/host/network.py
agners: With the new code mypy can infer that `inet` is not `None`. It also does the opposite, if you type it non-optional, but it detects that it can be `None`, it will raise. To some degree mypy just ignores typing and rather relies on type inference.

But yeah good catch, lets remove the line, it will make it clearer for readers as well :+1:

PR #6098, opened 2025-08-13
Stop refreshing the update information on outdated OS versions

agners:
## Proposed change
Add `JobCondition.OS_SUPPORTED` to the updater job to avoid refreshing update information when the OS version is unsupported.

This effectively freezes installations on unsupported OS versions and blocks Supervisor updates. Once deployed, this pretty much guarantees that any current Supervisor version will always run on at least the minimum supported OS version (except if the user manually intervenes, like deleting the updater file). This allows to drop compatibility code safely.

This requires to move the OS version check before Supervisor updater initialization to allow the `JobCondition.OS_SUPPORTED` to work correctly.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/updater.py
+++ b/supervisor/updater.py
@@ -69,6 +69,7 @@ class Updater(FileConfiguration, CoreSysAttributes):
         """Update internal data."""
         # If there's no connectivity, delay initial version fetch
         if not self.sys_supervisor.connectivity:
+            _LOGGER.debug("No Supervisor connectivity, delaying version fetch")
             if not self._connectivity_listener:
                 self._connectivity_listener = self.sys_bus.register_event(
                     BusEvent.SUPERVISOR_CONNECTIVITY_CHANGE, self._check_connectivity


agners:
With this change, on an unsupported Home Assistant OS, Supervisor stops updating the version information. The logs will have the following messages:

```
2025-08-13 11:30:43.039 INFO (MainThread) [__main__] Setting up Supervisor
...
2025-08-13 11:30:47.592 WARNING (MainThread) [supervisor.resolution.evaluations.base] OS version '11.5' is more than 4 versions behind the latest '15.2'! (more-info: https://www.home-assistant.io/more-info/unsupported/os_version)
...
2025-08-13 11:30:47.641 WARNING (MainThread) [supervisor.jobs] 'Updater.fetch_data' blocked from execution, unsupported OS version
...
2025-08-13 11:30:53.114 INFO (MainThread) [supervisor.resolution.check] Starting system checks with state setup
...
2025-08-13 11:30:53.127 INFO (MainThread) [supervisor.resolution.evaluate] System evaluation complete
...
2025-08-13 11:30:53.128 WARNING (MainThread) [supervisor.core] System running in a unsupported environment!
...
```

supervisor/os/manager.py
```diff
@@ -251,6 +251,10 @@ async def load(self) -> None:
         self._version = AwesomeVersion(cpe.get_version()[0])
         self._board = cpe.get_target_hardware()[0]
         self._os_name = cpe.get_product()[0]
+
+        # Evaluate OS version early to make OS_SUPPORTED Job condition work
+        await self.sys_resolution.evaluate.get("os_version")()
```
mdegat01: I'm not a fan of this. I get why you're doing it but besides creating this odd dependency on resolution center here, its actually running this before `ResolutionManager` loads. We get away with it today because `ResolutionManager` load doesn't really do much, just runs a healthcheck and schedules future ones. Any future change made assuming `ResolutionManager` was loaded before any of its functionality was used (a reasonable assumption) would cause this to break...

What if we made updater check if already has a version file and if so it delays the fetch? Either until the next time the task runs or schedule it to run when we transition to `CoreState.STARTUP`.

Seems like that would solve two problems. We could remove this special check and just let the normal `SETUP` healthcheck handle it. And if we don't have a version file then we fetch latest before checking if the OS is unsupported. Because having the latest version info is better then having nothing at all.

supervisor/resolution/evaluations/os_version.py
```diff
@@ -31,7 +31,7 @@ def states(self) -> list[CoreState]:
         """Return a list of valid states when this evaluation can run."""
         # Technically there's no reason to run this after STARTUP as update requires
         # a reboot. But if network is down we won't have latest version info then.
-        return [CoreState.RUNNING, CoreState.STARTUP]
+        return [CoreState.RUNNING, CoreState.STARTUP, CoreState.SETUP]
```
mdegat01:
```suggestion
        return [CoreState.RUNNING, CoreState.SETUP]
```
No point in listing `SETUP` and `STARTUP`. In each of those phases core stays in that state just long enough to run exactly one healthcheck and then moves on to `RUNNING`. The check in `STARTUP` would be redundant if we already ran it in `SETUP`

mdegat01 (changes requested):
There is something a bit odd about this. The primary difference between `Unsupported` and `Unhealthy` is that `Unhealthy` blocks updates. So why not just change `UnsupportedReason.OS_VERSION` to `UnhealthyVersion.OS_VERSION`?

This feels more right though. The system isn't actually unhealthy, its just moved to unsupported because its been too long since it was updated. Even if it is more work for us to make this one unsupported reason work like this 😆

supervisor/os/manager.py
agners: > What if we made updater check if already has a version file and if so it delays the fetch? Either until the next time the task runs or schedule it to run when we transition to `CoreState.STARTUP`.

Uh yes, I like that idea! We already delay in case there is no connectivity, so it is not uncommon that we have to delay (altough, I now wonder what happens if there is no version file and no connectivity, is that case handled somewhat gracefully today :thinking: ).

mdegat01: I think if we have no information on what the latest version is we assume whatever version of HAOS your on is supported right? So I guess the second problem wasn't actually an issue since the fetch wouldn't be blocked if you had no current version file. But it still lets us drop the unexpected dependency which seems like enough of a plus.

agners: With this, Supervisor now only updates when there are really no update information available, e.g.

```
2025-08-20 12:04:34.486 INFO (MainThread) [supervisor.updater] No OS update information found, force refreshing updater information
2025-08-20 12:04:34.486 INFO (MainThread) [supervisor.updater] Fetching update data from https://version.home-assistant.io/stable.json
```

Similar when plug-ins need update information.

supervisor/resolution/evaluations/os_version.py
mdegat01: This was marked resolved but it still shows all 3 core states. Did you mean to remove one or do you think it should be in all 3?

supervisor/updater.py
```diff
@@ -56,17 +56,26 @@ def __init__(self, coresys: CoreSys) -> None:
...
         """Update internal data."""
+        # If there's no connectivity, delay initial version fetch
+        if not self.sys_supervisor.connectivity:
+            if not self._connectivity_listener:
+                self._connectivity_listener = self.sys_bus.register_event(
+                    BusEvent.SUPERVISOR_CONNECTIVITY_CHANGE, self._check_connectivity
+                )
+            _LOGGER.info("No Supervisor connectivity, delaying version fetch")
+            return
+
```
mdegat01: So when this code was in `load` it would only delay the initial version fetch since `load` was only called once.

Now that this code is in `reload` it will happen repeatedly. We call `reload` on a schedule from a task. Any time that is called and the internet is down it will delay that until the internet comes back up. Whereas in the past it would've simply skipped that call to `reload` and tried again next time.

I think its ok like this but wanted to clarify the new behavior. Because this is a change in behavior, that comment is a little misleading. It's not just the initial fetch that can get delayed anymore, its all of them.

mdegat01 (approved):
I like this approach a lot better. Re-opened a comment from last time since it looks like we still have 3 `SETUP` and `STARTUP` listed but that's small. LGTM 👍

supervisor/resolution/evaluations/os_version.py
agners: Hm, I remember I applied it. It probably got lost in one of my rebase/force-pushes. Reapplied.

supervisor/updater.py
agners: Right, but since we have `JobCondition.SUPERVISOR_INTERNET` anyways, it won't get actually called. So this essentially gets rid of some warnings:

```
[supervisor.jobs] 'Updater.fetch_data' blocked from execution, no supervisor internet connection
```

Given that this was a warning before, maybe we should at least print a debug log message :thinking: 

From a behavior standpoint it should just lead to updates fetched as soon as we regain Internet connection. It should end up calling `fetch_data` not less then before.

PR #6102, opened 2025-08-16
Send progress updates during image pull for install/update

mdegat01:
## Proposed change

When pulling an image during an install or update of something, create a sub job per layer and use that to report progress on that layer. This way Home Assistant receives progress updates for an image pull and would be able to prevent that in the UI.

This isn't a perfect solution since we're still unable to report on the overall progress of the final image since we don't know how big each layer is until we get there. But it does provide a lot of progress information and could be used to provide users with some sense of progress during these tasks, unlike today.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/docker/interface.py
+++ b/supervisor/docker/interface.py
@@ -19,6 +19,7 @@ from docker.models.containers import Container
 from docker.models.images import Image
 import requests
 
+from ..bus import EventListener
 from ..const import (
     ATTR_PASSWORD,
     ATTR_REGISTRY,
@@ -39,13 +40,14 @@ from ..exceptions import (
     DockerRequestError,
     DockerTrustError,
 )
+from ..jobs import SupervisorJob
 from ..jobs.const import JOB_GROUP_DOCKER_INTERFACE, JobConcurrency
 from ..jobs.decorator import Job
 from ..jobs.job_group import JobGroup
 from ..resolution.const import ContextType, IssueType, SuggestionType
 from ..utils.sentry import async_capture_exception
-from .const import ContainerState, RestartPolicy
-from .manager import CommandReturn
+from .const import ContainerState, PullImageLayerStage, RestartPolicy
+from .manager import CommandReturn, PullLogEntry
 from .monitor import DockerContainerStateEvent
 from .stats import DockerStats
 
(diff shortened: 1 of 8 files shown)


supervisor/docker/interface.py
```diff
@@ -217,6 +219,117 @@ async def _docker_login(self, image: str) -> None:
...
+
+        # Pulling FS Layer is our marker for a layer that needs to be downloaded and extracted. Otherwise it already exists and we can ignore
+        job: SupervisorJob | None = None
+        if stage == PullImageLayerStage.PULLING_FS_LAYER:
+            job = self.sys_jobs.new_job(
+                name="Pulling image layer",
+                initial_stage=stage.status,
+                reference=reference.id,
+                parent_id=job_id,
+            )
```
agners, line 241: So this creates a new Job "out of the blue" for any pull operation. Will there be a way for the API consumer to match this jobs with whatever operation caused this? :thinking: 

E.g. if the user kicks off an add-on update or installation, it should have a way to extract progress information. Afaik, most of the operations which end up pulling a container image already have a job. Did you try attaching the progress to existing jobs :thinking:

mdegat01: Yes they can. Jobs have a hierarchy so the connection to a job that the caller can recognize is via the parent.

When listening on the WS connection every time a job is updated they will get a message with update details for that specific job. They'll have to kind of keep track of the hierarchy on their side and recognize that this one is a child of a job they are interested in.

When a client calls `GET /jobs/{uuid}` then they get the entire hierarchy for that job. So they can ask for info on the main Addon update job and see all these sub jobs all the way out to the individual docker layers.

That being said, I don't think addon install/update has a `background` option today like backup create/restore that quick returns a job ID and does the work asynchronously. I'll add that in a follow-up PR if we like this approach.

supervisor/docker/interface.py
```diff
@@ -217,6 +219,117 @@ async def _docker_login(self, image: str) -> None:
...
+            or not reference.status
+            or not (stage := PullImageLayerStage.from_status(reference.status))
+        ):
+            return
+
+        # Pulling FS Layer is our marker for a layer that needs to be downloaded and extracted. Otherwise it already exists and we can ignore
+        job: SupervisorJob | None = None
+        if stage == PullImageLayerStage.PULLING_FS_LAYER:
+            job = self.sys_jobs.new_job(
+                name="Pulling image layer",
```
agners:
```suggestion
                name="Pulling container image layer",
```

supervisor/docker/interface.py
agners: Actually, I see how this adds another Job "manually". So far we only create Jobs through the decorator. In a way, I feel this is a powerful mechanism to avoid this manual creation. The function where we track progress is already a Job named `docker_interface_install`, which IMHO fits perfectly the use case. Can't we just update that Jobs progress information?

In a quick test, something along these lines seem to work here:

```py
            job = self.sys_jobs.current

            async def process_pull_image_log(reference: PullLogEntry) -> None:
                self._process_pull_image_log(job, reference)

            listener = self.sys_bus.register_event(
                BusEvent.DOCKER_IMAGE_PULL_UPDATE, process_pull_image_log
            )
```

agners (changes requested):

supervisor/docker/interface.py
mdegat01: So yes, we can do this. But if we do then the job is constantly flipping between what progress its reporting. Because the layers are all pulled in parallel so the updates are all intermixed with each other.

And its no longer reporting on the progress of the task as a whole at all. I haven't yet figured out a way to update the progress of the main task based on the progress of the layers yet but I was hoping to. That's the most important progress bar to the user.

In addition with the current layout a client can call `GET /jobs/{uuid}` on the `docker_interface_install` job and see everything. They see all the `child_jobs` for each sub layer and the progress on each. If we make this change then `GET /jobs/{uuid}` is a useless API for this job. It tells you the progress of whatever image layer happened to provide the last update at that moment in time with no greater context at all. In practice that would mean if core's WS connection had a disconnect during this for whatever reason they would be 100% lost. There would be no way for them to determine overall current progress of the job and the layers and recover.

I'm just not sure what we gain here. The client has no issue locating all these sub jobs today. They hold on to the ID of the `docker_interface_install` job and look for any jobs with that as its `parent_id`. But we do lose a fair amount as only a client actively listening on the WS connection 100% of the time would have any clue what was going on. Everyone else would just see a random snapshot of the progress of one layer when they looked at the `docker_interface_install` job.

supervisor/docker/interface.py
```diff
@@ -217,6 +219,117 @@ async def _docker_login(self, image: str) -> None:
...
+                extra={
+                    "current": reference.progress_detail.current,
+                    "total": reference.progress_detail.total,
+                },
+            )
+        else:
+            job.update(progress=progress, stage=stage.status)
+
+        if stage == PullImageLayerStage.PULL_COMPLETE:
+            job.done = True
```
agners: Can we maybe make this part of the update call above? This would avoid one more message with essentially the same content, just one with `done=False` and the other with `done=True`:

```
2025-08-21 11:01:32.121 DEBUG (MainThread) [supervisor.homeassistant.websocket] Sending: {'type': <WSType.SUPERVISOR_EVENT: 'supervisor/event'>, 'data': {'event': <WSEvent.JOB: 'job'>, 'data': {'name': 'Pulling image layer', 'reference': 'a2feece4445c', 'uuid': 'df880854ebdc4bac819963a2756c3561', 'progress': 100, 'stage': 'Pull complete', 'done': False, 'parent_id': 'b8214eb029e0495aaa5971a249f263ea', 'errors': [], 'created': '2025-08-21T08:59:15.364813+00:00', 'extra': {'current': 21202276, 'total': 21202276}}}, 'id': 1138}
2025-08-21 11:01:32.122 DEBUG (MainThread) [supervisor.homeassistant.websocket] Sending: {'type': <WSType.SUPERVISOR_EVENT: 'supervisor/event'>, 'data': {'event': <WSEvent.JOB: 'job'>, 'data': {'name': 'Pulling image layer', 'reference': 'a2feece4445c', 'uuid': 'df880854ebdc4bac819963a2756c3561', 'progress': 100, 'stage': 'Pull complete', 'done': True, 'parent_id': 'b8214eb029e0495aaa5971a249f263ea', 'errors': [], 'created': '2025-08-21T08:59:15.364813+00:00', 'extra': {'current': 21202276, 'total': 21202276}}}, 'id': 1139}
```

supervisor/docker/interface.py
```diff
@@ -217,6 +219,117 @@ async def _docker_login(self, image: str) -> None:
...
+            return
+
+        if job.stage and stage < PullImageLayerStage.from_status(job.stage):
+            _LOGGER.debug(
+                "Received pull image log with status %s for job %s but job was already on stage %s, skipping",
+                reference.status,
+                job.uuid,
+                job.stage,
+            )
+            return
```
agners: Absent from a Docker bug, I can't see how this would happen. We connect through a unix socket which is not packet based but a stream, it guarantee things being in order (just like TCP). So unless Docker sends it in the wrong order, or our bus reorder things, this really should never happen!

I'd not make this a debug log, it is very unlikely we'd ever notice this. I'd raise severity to error here, and maybe even send to Sentry so we can start investigate why that happens.

supervisor/docker/interface.py
agners: > So yes, we can do this. But if we do then the job is constantly flipping between what progress its reporting. Because the layers are all pulled in parallel so the updates are all intermixed with each other.

Hm, I see the multiple layer in parallel makes it much nicer if handled with separate (sub) jobs. In a way it still feels weird that jobs, which usually are tied to the Python execution flow in Supervisor, suddenly get used manually as a data structure. But the main purpose of jobs is to convey progress, and this is what we use it here for. So with that in mind, and how progress reporting works, we can say this deviation in how the job gets managed is acceptable.

supervisor/docker/interface.py
mdegat01: Yea that's fair. I almost did that initially but `done` felt kind of special and seemed like it should be its own update. But I guess you're right, it just looks silly in the log of events.

supervisor/docker/interface.py
mdegat01: > Absent from a Docker bug, I can't see how this would happen.

I agree. This wasn't actually what I was worried about though. We are reading in these logs in the executor. We are then scheduling a task on the normal event loop per log called `fire_event`. Then that is scheduling another task on the normal event loop per listener listening for that event to process it.

My concern is that all this task scheduling will cause us to process them out of order. Not that docker will actually give them to us out of order.

> I'd raise severity to error here, and maybe even send to Sentry

I'd rather only send it to sentry tbh. Given what we're talking about I don't think the impact on the user is enough to raise it to their attention. If they miss a few of the 100 ms progress updates that really doesn't matter much. But we would like to know about it. Maybe time to make use of that `capture_event` API in Sentry rather then `capture_exception`

mdegat01: Ok added a new type of exception for this called `DockerLogOutOfOrder`. Raising and capturing it in sentry.

I'm still logging at debug level for now because I don't think users really need to know about this. Let me know if you want me to bump the severity in addition to capturing it in sentry.

agners (approved):
LGTM, thanks!

supervisor/docker/interface.py
agners: That is fine by me, with the Sentry capture we'll definitely notice if that happens in practice :+1: .

PR #6114, opened 2025-08-19
Revert .git missing check in store git

mdegat01:
## Proposed change

Effectively revert #5701 . At least all the logic that specifically checked for the presence of the `.git` subdirectory as it caused more trouble then it was worth.

I did change it so that if the repository's folder exists but is entirely empty it does a clone rather then trying to process it as a git directory. Because clone works fine on an empty directory and that is more sensible behavior here. Conveniently the tests added for #5701 properly test this behavior so we can keep them.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/utils/__init__.py
+++ b/supervisor/utils/__init__.py
@@ -168,3 +168,11 @@ def version_is_new_enough(
 ) -> bool:
     """Return True if the given version is new enough."""
     return version >= want_version
+
+
+def directory_missing_or_empty(path: Path) -> bool:
+    """Return true if path is not a directory or is empty.
+
+    Must be run in executor.
+    """
+    return not path.is_dir() or len(os.listdir(path)) == 0
--- a/supervisor/store/git.py
+++ b/supervisor/store/git.py
@@ -14,7 +14,7 @@ from ..coresys import CoreSys, CoreSysAttributes
 from ..exceptions import StoreGitCloneError, StoreGitError, StoreJobError
 from ..jobs.decorator import Job, JobCondition
 from ..resolution.const import ContextType, IssueType, SuggestionType, UnhealthyReason
-from ..utils import remove_folder
+from ..utils import directory_missing_or_empty, remove_folder
 from .validate import RE_REPOSITORY
 
 _LOGGER: logging.Logger = logging.getLogger(__name__)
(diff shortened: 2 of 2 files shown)


supervisor/utils/__init__.py
```diff
@@ -168,3 +168,11 @@ def version_is_new_enough(
...
     """Return True if the given version is new enough."""
     return version >= want_version
+
+
+def directory_missing_or_empty(path: Path) -> bool:
+    """Return true if path is not a directory or is empty.
+
+    Must be run in executor.
+    """
+    return not path.is_dir() or len(os.listdir(path)) == 0
```
Impact123, line 178: I'm not a proper python developer but I feel like `os.listdir` is unnecessarily expensive just to see if a directory contains **any** file.

agners: Actually, the code has a bug :see_no_evil: 

If path.is_dir() returns False (path is not a directory), then not path.is_dir() is True, which makes the function return True immediately (since there won't be a need to evaluate the second expression of the or statement).

However, if the path exists but is not a directory (e.g., it's a file), then calling os.listdir(path) is futile essentially, it would raise an `OSError`.

I'll open a PR to fix the function.

mdegat01: @Impact123 Some googling turned up that in python 3.5+ your supposed to use `scandir` now. Which is the same idea but a bit more performant because it does less stat calls. But asking python to list out the files directly in that folder and then checking if that list is empty is the only tool I'm aware of for this, I'll change it to use the newer `scandir` though

mdegat01: @agners I considered that. I was going to put in a check for that but then two things came to mind:
1. I'm not aware how that could happen short of someone going in and directly messing around in Supervisor's data folder. Which isn't something we support.
2. Even if it is a file, all we can do is throw an exception and create a suggestion to reset it right? That's what will happen anyway.

agners: > 1. I'm not aware how that could happen short of someone going in and directly messing around in Supervisor's data folder. Which isn't something we support.
> 1. Even if it is a file, all we can do is throw an exception and create a suggestion to reset it right? That's what will happen anyway.

Agree to both yes. But then the question would be why even use `os.listdir()` and not just raise.

Claude and I came up with this: #6116

PR #6133, opened 2025-08-26
Avoid duplicate evaluate_system() calls during resolution manager setup

agners:
## Proposed change

This PR optimizes the resolution manager initialization to avoid duplicate `evaluate_system()` calls during supervisor setup. 

Currently, when the resolution manager loads, it calls `healthcheck()` which internally calls `evaluate_system()`. However, during the setup stage, the supervisor's main setup process also calls `evaluate_system()`, creating redundant system evaluation work.

Remove the `evaluate_system()` in the setup stage. Also move the OS Agent diagnostics initialization into the start phase. This allows to leverage the broad exception handling. Finally check if OS Agent is connected at all, which prevents the common case of D-Bus not being available without causing exceptions. Supervisor already raises an error earlier if OS Agent is missing.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/resolution/module.py
+++ b/supervisor/resolution/module.py
@@ -215,8 +215,8 @@ class ResolutionManager(FileConfiguration, CoreSysAttributes):
 
     async def load(self):
         """Load the resoulution manager."""
-        # Initial system check when the manager is loaded
-        await self.check.check_system()
+        # Initial healthcheck check
+        await self.healthcheck()
 
         # Schedule the healthcheck
         self.sys_scheduler.register_task(self.healthcheck, SCHEDULED_HEALTHCHECK)
--- a/supervisor/resolution/module.py
+++ b/supervisor/resolution/module.py
@@ -215,8 +215,8 @@ class ResolutionManager(FileConfiguration, CoreSysAttributes):
 
     async def load(self):
         """Load the resoulution manager."""
-        # Initial healthcheck when the manager is loaded
-        await self.healthcheck()
+        # Initial system check when the manager is loaded
+        await self.check.check_system()
 
         # Schedule the healthcheck
         self.sys_scheduler.register_task(self.healthcheck, SCHEDULED_HEALTHCHECK)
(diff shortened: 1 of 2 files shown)


supervisor/resolution/module.py
```diff
@@ -215,8 +215,8 @@ def add_issue(
 
     async def load(self):
         """Load the resoulution manager."""
-        # Initial healthcheck when the manager is loaded
-        await self.healthcheck()
+        # Initial system check when the manager is loaded
+        await self.check.check_system()
```
mdegat01: Why not just drop this entirely and have the load only schedule the healthcheck task. And then run a full healthcheck here instead:
https://github.com/home-assistant/supervisor/blob/3d62c9afb1a1f45b02f72a46e7fa1c84e0f12fab/supervisor/core.py#L217

It seems more confusing to have the complete healthcheck split up into two places like this no?

agners: My worry is that something is using the `check_system()`, I want to avoid unnecessary reordering.

agners: In fact, having a closer look, the OS Agent diagnostics uses the supported flag, which is influenced by the system check.

https://github.com/home-assistant/supervisor/blob/3d62c9afb1a1f45b02f72a46e7fa1c84e0f12fab/supervisor/core.py#L204

Actually, this raises the question why we don't go the other direction: Remove the `evaluate_system()` call and leave the `healthcheck()` call in place.

While looking at it, I also don't really like how OS-Agent diagnostics is handled here: Why is such a broad Exception handling necessary? I guess we want to continue just in case OS Agent isn't available through D-Bus, however, the if statement's `self.sys_dbus.agent.diagnostics` actually accesses the OS Agent. So an unavailable D-Bus is not properly handled here :see_no_evil: .

Anyways, I realized that `setup()` has a broad exception handler which is suited for this. And to avoid things failing horrible just when OS Agent is not around, I've added `self.sys_dbus.agent.is_connected` to the if statement. This makes things work gracefully in the common case, and properly handled in the Exceptional case.

agners: Also this actually makes separate handling for development systems unnecessary.

PR #6134, opened 2025-08-27
Add background option to update/install APIs

mdegat01:
## Proposed change

Enable background update/install of Home Assistant and Addons. This way Home Assistant can listen in for progress updates on install/addon and present that to the user

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/api/homeassistant.py
+++ b/supervisor/api/homeassistant.py
@@ -20,6 +20,7 @@ from ..const import (
     ATTR_CPU_PERCENT,
     ATTR_IMAGE,
     ATTR_IP_ADDRESS,
+    ATTR_JOB_ID,
     ATTR_MACHINE,
     ATTR_MEMORY_LIMIT,
     ATTR_MEMORY_PERCENT,
@@ -37,8 +38,8 @@ from ..const import (
 from ..coresys import CoreSysAttributes
 from ..exceptions import APIDBMigrationInProgress, APIError
 from ..validate import docker_image, network_port, version_tag
-from .const import ATTR_FORCE, ATTR_SAFE_MODE
-from .utils import api_process, api_validate
+from .const import ATTR_BACKGROUND, ATTR_FORCE, ATTR_SAFE_MODE
+from .utils import api_process, api_validate, background_task
 
 _LOGGER: logging.Logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


supervisor/api/homeassistant.py
```diff
@@ -20,6 +20,7 @@
     ATTR_CPU_PERCENT,
     ATTR_IMAGE,
     ATTR_IP_ADDRESS,
+    ATTR_JOB_ID,
```
agners, line 23: We have `ATTR_JOB_ID` in `supervisor/const.py`, used in `supervisor/api/backups.py` already. Either use the existing, or if you want to have a separate constant for the API explicitly, we should also use it for in `backup.py` too.

supervisor/api/utils.py
```diff
@@ -198,3 +202,55 @@ async def api_validate(
         data_validated[origin_value] = data[origin_value]
 
     return data_validated
+
+
+async def background_task(
```
agners, line 206: We already have `_background_backup_task`, which does something similar I guess. We get away without static Job name strings there. Can we maybe get rid of that too? It feels error prone, and I don't know what it exactly buys us compared to the approach we use in Backup. Would be nice to have a generic way to do this background things.

mdegat01: Yea I wanted to do this too. The reason I didn't was because the backup one is a bit different. It looks for the system `FREEZE`. We did this because the validation phase is not cleanly divided between jobs but it all was finished as soon as we move core state to `FREEZE`.

But you're right, I can make this work. I just have to make this utility accept a bus event to look for and a filter function

supervisor/api/homeassistant.py
mdegat01: Yea cursor did this. AI wrote most of it and I picked it up from there. Will fix

mdegat01: Wait I'm confused, this is importing `ATTR_JOB_ID` from `supervisor/const.py` already. What's the suggestion here? I assumed that was being imported from the wrong place

agners: Hm, I am confused now too, I thought I saw another `ATTR_JOB_ID` introduced in this PR but it doesn't seem so. Anyhow, LGTM then.

mdegat01:
@agners refactored to use a specific `validation_complete` `asyncio.Event` as an arg rather then looking for bus events.

agners (approved):
Nice, LGTM! I like this explicit synchronization using `asyncio.Event` much more.

PR #6140, opened 2025-08-28
Add availability API for addons

mdegat01:
## Proposed change

Add API that determines availability of an add-on for the current system. It always checks against the latest version in the store so it can be used to check availability prior to an update of an existing add-on or install of a new one.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to documentation pull request: https://github.com/home-assistant/developers.home-assistant/pull/2769
- Link to cli pull request:
- Link to client library pull request: https://github.com/home-assistant-libs/python-supervisor-client/pull/172

## Checklist

### DIFF
--- a/tests/api/test_store.py
+++ b/tests/api/test_store.py
@@ -505,11 +505,27 @@ async def test_api_store_addons_addon_availability_success(
     assert resp.status == 200
 
 
-@pytest.mark.parametrize("supported_architectures", [["i386"], ["i386", "aarch64"]])
+@pytest.mark.parametrize(
+    ("supported_architectures", "api_action", "api_method", "installed"),
+    [
+        (["i386"], "availability", "get", False),
+        (["i386", "aarch64"], "availability", "get", False),
+        (["i386"], "install", "post", False),
+        (["i386", "aarch64"], "install", "post", False),
+        (["i386"], "update", "post", True),
+        (["i386", "aarch64"], "update", "post", True),
+    ],
+)
 async def test_api_store_addons_addon_availability_arch_not_supported(
-    api_client: TestClient, coresys: CoreSys, supported_architectures: list[str]
+    api_client: TestClient,
+    coresys: CoreSys,
+    supported_architectures: list[str],
+    api_action: str,
+    api_method: str,
+    installed: bool,
 ):
-    """Test /store/addons/{addon}/availability REST API - architecture not supported."""
+    """Test availability errors for /store/addons/{addon}/* REST APIs - architecture not supported."""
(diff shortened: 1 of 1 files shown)


tests/api/test_store.py
```diff
@@ -390,3 +393,131 @@ async def test_api_store_addons_changelog_corrupted(
...
+        "repository": "test",
+        "stage": "stable",
+        "version": "1.0.0",
+    }
+    coresys.store.data.addons[addon_obj.slug] = addon_config
+
+    # Mock the system architecture to be different
+    with patch.object(CpuArch, "supported", new=PropertyMock(return_value=["amd64"])):
+        resp = await api_client.get(f"/store/addons/{addon_obj.slug}/availability")
+        assert resp.status == 400
```
agners, line 554: Hm, since this is really an API to check, is `Bad Request` the right HTTP status? :thinking:

mdegat01:
@agners Reworked error to support specifying a key, message template and extra fields so we can make errors that are much more translation and client library friendly. Leveraged that for this work but this will require a pretty significant re-review and refactor of most of the errors in Supervisor going forward. As many should adopt this model.

tests/api/test_store.py
mdegat01: I dont think there's really a better option here other then using a 200 status instead. After discussion we decided to stick with this and instead significantly update error responses to have more useful info.

agners (approved):
This makes our error handling much more translation friendly: The key can be used to look up translations, and since we now pass parameters as separate fields the messages can easily formatted on client side without the need to parse and reformat our pre-processed messages. This makes any key plus extra field combination part of the API, as in if we want to change fields in an error we need to introduce a new key.

MartinHjelmare:
Does a Supervisor API user need to check the availability of an add-on before updating it? I'd expect that would happen automatically so no extra API call would be needed.

agners:
> Does a Supervisor user need to check the availability of an add-on before updating it? I'd expect that would happen automatically so no extra API call would be needed.

The problem is backups: If we simply raise an error on the actual update attempt, then the caller might already have created a backup. At least for the update flows under control by the hassio integration this will allow to first check availability, before taking a backup.

There is some more discussion in https://github.com/home-assistant/core/issues/148954.

MartinHjelmare:
Ok, that's good. I'd still expect the update call to raise an (the same) exception for an unavailable add-on.

mdegat01:
@MartinHjelmare Good point. I believe it does because both call the same method but i'll also add the same test cases there to make sure.

PR #6163, opened 2025-09-05
Add endpoint for complete logs of the latest container startup

### LINKED ISSUE #6147: Add support for container restart sensitive log history
### Description

Since we moved to the systemd-journal based advanced logging there is no longer a way to get the logs written since the last restart of Core/Add-ons. Lets investigate and an API which allows to get logs from the last restart onwards.

### Additional context

_No response_

sairon:
## Proposed change

Add endpoint that returns complete logs of the latest startup of container, which can be used for downloading Core logs in the frontend.

Realtime filtering header is used for the Journal API and StartedAt parameter from the Docker API is used as the reference point. This means that any other Range header is ignored for this parameter, yet the "lines" query argument can be used to limit the number of lines. By default "infinite" number of lines is returned.

For cases where HAOS is not present or it's older than v16.0, a suboptimal fallback method using Docker's internal `CONTAINER_LOG_EPOCH` field is used.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR fixes or closes issue: fixes #6147
- Link to documentation pull request: https://github.com/home-assistant/developers.home-assistant/pull/2785
- Link to cli pull request:

## Checklist

- [ ] [CLI][cli-repository] updated (if necessary) N/A
- [ ] [Client library][client-library-repository] updated (if necessary) N/A

### DIFF
--- a/supervisor/api/host.py
+++ b/supervisor/api/host.py
@@ -2,7 +2,6 @@
 
 import asyncio
 from contextlib import suppress
-import datetime
 import json
 import logging
 from typing import Any
@@ -225,7 +224,6 @@ class APIHost(CoreSysAttributes):
         if follow:
             params[PARAM_FOLLOW] = ""
 
-        since: datetime.datetime | None = None
         if latest:
             if not identifier:
                 raise APIError(
@@ -266,14 +264,8 @@ class APIHost(CoreSysAttributes):
                 # instead. Since this is really an edge case that doesn't matter much, we'll just
                 # return 2 lines at minimum.
                 lines = max(2, lines)
-            if since:
-                # realtime=[since]:[until][[:num_skip]:num_entries]
-                range_header = f"realtime={int(since.timestamp())}::0:{lines}"
-            else:
-                # entries=cursor[[:num_skip]:num_entries]
-                range_header = f"entries=:-{lines - 1}:{SYSTEMD_JOURNAL_GATEWAYD_LINES_MAX if follow else lines}"
-        elif since:
-            range_header = f"realtime={int(since.timestamp())}::0:{SYSTEMD_JOURNAL_GATEWAYD_LINES_MAX}"
+            # entries=cursor[[:num_skip]:num_entries]
+            range_header = f"entries=:-{lines - 1}:{SYSTEMD_JOURNAL_GATEWAYD_LINES_MAX if follow else lines}"
         elif latest:
             range_header = f"entries=0:{SYSTEMD_JOURNAL_GATEWAYD_LINES_MAX}"
         elif RANGE in request.headers:


abmantis:
Would it be much harder to have a `boot` api instead, similar to `-b` of journalctl? A 0 or missing value would return the latest one, while a positive value would return an older one.
This would be useful since getting the previous session is a common use case (we currently have the rotated log file for that).

sairon:
> Would it be much harder to have a `boot` api instead, similar to `-b` of journalctl? A 0 or missing value would return the latest one, while a positive value would return an older one.

Unfortunately there's not much we could use now for that in [systemd-journal-gatewayd](https://www.freedesktop.org/software/systemd/man/latest/systemd-journal-gatewayd.service.html). We don't have information about the previous epoch of the running container, if we use current `StartedAt` as the boundary for the end, we don't know when it started. There's `CONTAINER_LOG_EPOCH` in the metadata but that can't be split by container name, so that would mean we'd have to call the API multiple times to get list of container starts.

supervisor/api/host.py
```diff
@@ -239,8 +256,14 @@ async def advanced_logs_handler(
                 # instead. Since this is really an edge case that doesn't matter much, we'll just
                 # return 2 lines at minimum.
                 lines = max(2, lines)
-            # entries=cursor[[:num_skip]:num_entries]
-            range_header = f"entries=:-{lines - 1}:{SYSTEMD_JOURNAL_GATEWAYD_LINES_MAX if follow else lines}"
+            if since:
+                # realtime=[since]:[until][[:num_skip]:num_entries]
+                range_header = f"realtime={int(since.timestamp())}::0:{lines}"
```
agners: Is this safe to be used on older systemd-journal-gatewayd versions? :thinking:

sairon: Good point! It needs Systemd v256+, which is available only since HAOS 16. Providing the unknown header will lead to an error raised by the Journal Gateway - ideally we want to handle it more gracefully. And probably the client (frontend) should be aware of this limitation as well.

mdegat01: Hm, I'm not sure we've tried to handle that scenario before. We probably should've since it's definitely come up before. Like the recent `boards` API only worked on the latest version of the OS at the time it was released but doesn't appear to include any special handling for older OS's. I imagine other features relying on a newly developed OS feature could've used this handling as well 🙈 

Maybe we should set up something to use for this going forward as its unlikely to be the last time Supervisor adopts something new in OS. Maybe an annotation we could wrap API methods in? Or a standard helper method.

Looking at HTTP status codes, 501 Not Implemented might make the most sense? It has this description:
> The server either does not recognize the request method, or it lacks the ability to fulfill the request. Usually this implies future availability (e.g., a new feature of a web-service API).

It's not an exact match but that makes sense because our situation is pretty non-standard. Web services don't usually tell clients they need to update something because usually clients have no control over the web service, they are just users of it. But saying it is Not implemented right now because a backend update is required with a message telling them what to update seems fairly accurate for us in this spot.

sairon: Time is a flat circle :sweat_smile: 

https://github.com/home-assistant/supervisor/pull/5770#discussion_r2016085433

sairon:
After discussion with @agners, I have implemented the container epoch filter as a fallback. This needed more adjustments in tests, also silently changing behavior of `os_available` fixture to return version 16.2 as default. But since we override the returned version where needed, this doesn't cause any side-effects.

supervisor/api/host.py
```diff
@@ -213,6 +226,32 @@ async def advanced_logs_handler(
...
+                    "Latest logs can only be fetched for a specific identifier."
+                )
+
+            # Fallback is needed for older versions which don't support "realtime" Range header (Systemd<256)
+            use_fallback = (
+                not self.sys_os.available
+                or not self.sys_os.version
+                or self.sys_os.version < "16.0"
+            )
+            if use_fallback:
```
agners:
```suggestion
            if (
                not self.sys_os.available
                or not self.sys_os.version
                or self.sys_os.version < "16.0"
            ):
```

supervisor/api/host.py
```diff
@@ -336,3 +387,48 @@ async def disk_usage(self, request: web.Request) -> dict:
                 *known_paths,
             ],
         }
+
+    async def _get_container_start_time(
+        self, identifier: str
+    ) -> datetime.datetime | None:
+        """Get container start time for the given syslog identifier."""
+        container: Container = self.sys_docker.containers.get(identifier)
```
agners: This raises `docker.errors.NotFound` if the container does not exist. I think we don't need to handle None here, but not 100% sure.

supervisor/api/host.py
```diff
@@ -336,3 +387,48 @@ async def disk_usage(self, request: web.Request) -> dict:
...
+            return None
+
+        try:
+            return datetime.datetime.fromisoformat(started_at)
+        except ValueError:
+            _LOGGER.warning(
+                "Failed to parse StartedAt time of %s container, got: %s",
+                identifier,
+                started_at,
+            )
```
agners: I'd say not parsable start date is very unexpected (as in essentially impossible) here. I think I'd go for the offensive programming approach in this case and simply not handle at all. In the unlikely case that this happens in practise, we'll learn about it through Sentry and can implement a proper solution.

sairon: Yeah, good point. Removed the exception handling and test for it.

supervisor/api/host.py
sairon: Good catch! Fixed.

agners (approved):
LGTM!

supervisor/api/host.py
```diff
@@ -336,3 +387,41 @@ async def disk_usage(self, request: web.Request) -> dict:
...
         }
+
+    async def _get_container_start_time(
+        self, identifier: str
+    ) -> datetime.datetime | None:
+        """Get container start time for the given syslog identifier."""
+        try:
+            container: Container = self.sys_docker.containers.get(identifier)
+        except docker.errors.NotFound:
+            return None
```
mdegat01: I think this code has a bit of an issue right now. Whenever an addon is stopped successfully (i.e. did not crash) then Supervisor removes the container. That means you'll get `None` here for all successfully stopped addons. It will only work properly for currently running addons and addons which crashed.

Since this is a fallback maybe that's ok? But I would probably change the response error message asking "if its a docker container?" because I imagine that will be the more likely reason for getting `None` here then an incorrect name.

supervisor/api/host.py
```diff
@@ -336,3 +387,41 @@ async def disk_usage(self, request: web.Request) -> dict:
...
+    async def _get_container_last_epoch(self, identifier: str) -> str | None:
+        """Get Docker's internal log epoch of the latest log entry for the given identifier."""
+        try:
+            async with self.sys_host.logs.journald_logs(
+                params={"CONTAINER_NAME": identifier},
+                range_header="entries=:-1:2",  # -1 = next to the last entry
+                accept=LogFormat.JSON,
+                timeout=ClientTimeout(total=10),
+            ) as resp:
+                text = await resp.text()
```
mdegat01, line 380: Since we always use the same name for an addon's container, won't this always get the complete logs for all runs of that addon? Or I'm not understanding how this filters the log set to just the latest run of the addon.

supervisor/api/host.py
sairon: Damn, very good point! This is actually NOT the fallback - it's the default, and the fallback usage is determined by OS version (because it relies on the `Range: realtime` header which as added later). I guess our only option now is to always use the "fallback" method and ditch the idea of using container metadata. There's minor performance penalty (couple of hundreds of milliseconds with empty disk caches) but I'd say it's not a big issue given that it'll be used for one-off file download.

supervisor/api/host.py
sairon: Not really, this just gets the latest log entries (two of them, and the last is used - this is to workaround the limitations of the range header syntax which doesn't allow for equivalent of Python's `collection[-1]`) filtered by container name - from that `CONTAINER_LOG_EPOCH` is extracted and content of this field is unique for each start of the container. This is then used for filtering the logs in the actual request for log entries.

supervisor/api/host.py
mdegat01: Hm. Alternatively we were actually just talking about adding "last start time" for addons as persistent metadata to Supervisor. Which I think is exactly what you'd need here.

So maybe we leave it as is (though probably adjust the error message to account for this likely reason), and then when we add that in a follow-up PR we use it here? Perhaps with this prioritization:
1. Get start time from container attrs
2. If not found, ask Supervisor for that information
3. If still not found, use the other fallback

sairon: I think it's preferable to move on with this PR, as we mainly want to simplify troubleshooting of Core issues (people now often provide incomplete logs). The actual implementation of the functionality doesn't matter much. So I'll skip over the step 2 and add it once we have the Supervisor API for that.

agners: Yeah I agree with this, let's get this rolled out. We can (re)add the `realtime` approach which uses stored "last start time" later.

mdegat01: Yea that's basically what I was saying. Sorry maybe I was confusing. I agree with merging this and then following up later once we have last start time for addons to improve

supervisor/api/host.py
```diff
@@ -213,6 +227,31 @@ async def advanced_logs_handler(
...
+                or self.sys_os.version < "16.0"
+            ):
+                # Fallback is needed for older versions which don't support "realtime" Range header (Systemd<256)
+                try:
+                    epoch = await self._get_container_last_epoch(identifier)
+                    params["CONTAINER_LOG_EPOCH"] = epoch
+                except HostLogError as err:
+                    raise APIError(
+                        f"Cannot determine CONTAINER_LOG_EPOCH of {identifier}, is it a Docker container name?"
+                    ) from err
```
mdegat01: @sairon this message was the only thing I was saying we might want to change before merge. If a user tries to get latest logs for a currently stopped addon this is the message they will see. That seems confusing. I think we might want to adjust to something like "is the add-on stopped right now?" Because that's the most likely reason. The api user doesn't actually provide a docker container name and we check the slug to ensure an add-on with that slug exists before this point.

supervisor/api/host.py
```diff
@@ -213,6 +225,21 @@ async def advanced_logs_handler(
         if follow:
             params[PARAM_FOLLOW] = ""
 
+        since: datetime.datetime | None = None
```
agners: Seems since is dead code at this point. And since you removed `_get_container_start_time`, should we maybe cleanup this here too? We can readd it once we make use of it.

sairon: Right! Removed that too.

agners (approved):

supervisor/api/host.py
```diff
@@ -241,6 +266,8 @@ async def advanced_logs_handler(
                 lines = max(2, lines)
             # entries=cursor[[:num_skip]:num_entries]
             range_header = f"entries=:-{lines - 1}:{SYSTEMD_JOURNAL_GATEWAYD_LINES_MAX if follow else lines}"
+        elif latest:
+            range_header = f"entries=0:{SYSTEMD_JOURNAL_GATEWAYD_LINES_MAX}"
```
agners, line 270: This range header seems not to be the correct format for `entries`, at least not on current HAOS: Another colon seems to be required to make this working. Otherwise the `/core/logs/latest` endpoint errors out wiht:

```
2025-09-23 12:55:49.723 DEBUG (MainThread) [supervisor.api.middleware.security] /core/logs/latest access from Home Assistant
2025-09-23 12:55:49.745 ERROR (MainThread) [aiohttp.server] Error handling request from 172.30.32.1
  File "/usr/local/lib/python3.13/site-packages/aiohttp/web_protocol.py", line 510, in _handle_request
  File "/usr/local/lib/python3.13/site-packages/sentry_sdk/integrations/aiohttp.py", line 157, in sentry_app_handle
    reraise(*_capture_exception())
    ~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.13/site-packages/sentry_sdk/utils.py", line 1751, in reraise
    raise value
  File "/usr/local/lib/python3.13/site-packages/sentry_sdk/integrations/aiohttp.py", line 140, in sentry_app_handle
    response = await old_handle(self, request)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.13/site-packages/aiohttp/web_app.py", line 569, in _handle
    return await handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.13/site-packages/aiohttp/web_middlewares.py", line 117, in impl
    return await handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/supervisor/supervisor/api/middleware/security.py", line 198, in block_bad_requests
    return await handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/supervisor/supervisor/api/middleware/security.py", line 208, in system_validation
    return await handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/supervisor/supervisor/api/middleware/security.py", line 276, in token_validation
    return await handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/supervisor/supervisor/api/middleware/security.py", line 289, in core_proxy
    return await handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/supervisor/supervisor/api/utils.py", line 118, in wrap_api
    msg_data = await method(api, *args, **kwargs)
  File "/usr/src/supervisor/supervisor/api/host.py", line 326, in advanced_logs
    return await self.advanced_logs_handler(request, identifier, follow, latest)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/supervisor/supervisor/api/host.py", line 286, in advanced_logs_handler
    async for cursor, line in journal_logs_reader(resp, log_formatter):
    ...<22 lines>...
            break
  File "/usr/src/supervisor/supervisor/utils/systemd_journal.py", line 103, in journal_logs_reader
    length_raw = await resp.content.readexactly(8)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.13/site-packages/aiohttp/streams.py", line 487, in readexactly
    raise asyncio.IncompleteReadError(partial, len(partial) + n)
asyncio.exceptions.IncompleteReadError: 0 bytes read on a total of 8 expected bytes
```

agners: Fix in #6202

PR #6148, opened 2025-09-02
Check Core version and raise unsupported if older than 2 years

agners:
## Proposed change

Check the currently installed Core version relative to the last known Core version and if its older than 2 years, mark the system unsupported. Also add a Job condition to prevent automatic refreshing of the update information in this case. We use it to block fetching the version json as well as updating and adding add-on repositories.

Using current Core version as reference to mark the system unsupported is necessary so that an update of Core will mark the system as supported again, such that Supervisor will refresh the update information again. If we were to take the current date as a reference, then 2 years after a Core version of an installation gets declared unsupported, we'd only offer an update to the then newest known Core version, which would be 2 years old already and hence the system would remain unsupported.

## Type of change

- [ ] Dependency upgrade
- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (which adds functionality to the supervisor)
- [ ] Breaking change (fix/feature causing existing functionality to break)
- [ ] Code quality improvements to existing code or addition of tests

## Additional information

- This PR fixes or closes issue: fixes #
- This PR is related to issue:
- Link to documentation pull request: https://github.com/home-assistant/home-assistant.io/pull/40914
- Link to cli pull request:
- Link to client library pull request:

## Checklist

- [ ] The code change is tested and works locally.
- [ ] Local tests pass. **Your PR cannot be merged unless tests pass**
- [ ] There is no commented out code in this PR.
- [ ] I have followed the [development checklist][dev-checklist]
- [ ] The code has been formatted using Ruff (`ruff format supervisor tests`)
- [ ] Tests have been added to verify that the new code works.

If API endpoints or add-on configuration are added/changed:

- [ ] Documentation added/updated for [developers.home-assistant.io][docs-repository]
- [ ] [CLI][cli-repository] updated (if necessary)
- [ ] [Client library][client-library-repository] updated (if necessary)

[dev-checklist]: https://developers.home-assistant.io/docs/en/development_checklist.html
[docs-repository]: https://github.com/home-assistant/developers.home-assistant
[cli-repository]: https://github.com/home-assistant/cli
[client-library-repository]: https://github.com/home-assistant-libs/python-supervisor-client/

PR #6179, opened 2025-09-10
Persistent notifications to repairs and fix free_space check

### LINKED ISSUE #5961: Convert persistent notifications to repair issues
Supervisor currently creates persistent notification for some of the issues found by the resolution center. Convert those to actual repairs instead.

mdegat01:
## Proposed change

Supervisor continues to make persistent notifications rather then repairs for a small subset of issues. This no longer makes sense so remove this and make proper repairs for any we can.

The one exception is the security one targeted at people using custom components with Home Assistant <2021.1.5 because I don't think repairs existed then. Although I wonder if we really have to continue to leave this one around at this point? It's quite old.

In addition I changed the free space issue to stop suggesting to remove a full backup and bumped its lower limit to 2GB. I'm not sure if 2GB is enough either but its better then 1GB. And the "remove full backup" suggestion was always pretty odd and kind of the reason we didn't make this one into a repair already, people didn't like that. The repair will just link to our [clear up storage guide](https://www.home-assistant.io/more-info/free-space)

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR fixes or closes issue: fixes #5961
- Link to cli pull request:
- Link to core pull request: https://github.com/home-assistant/core/pull/152066

## Checklist


agners:
> The one exception is the security one targeted at people using custom components with Home Assistant <2021.1.5 because I don't think repairs existed then. Although I wonder if we really have to continue to leave this one around at this point? It's quite old.

Let's remove it once #6148 is merged.


> In addition I changed the free space issue to stop suggesting to remove a full backup and bumped its lower limit to 2GB. I'm not sure if 2GB is enough either but its better then 1GB. And the "remove full backup" suggestion was always pretty odd and kind of the reason we didn't make this one into a repair already, people didn't like that. The repair will just link to our [clear up storage guide](https://www.home-assistant.io/more-info/free-space)

At the time of writing, the x86-64 extracted image size (without layer sharing) 3.12GB, the layer sizes 651MB. In worst case, both are at the disk at the same time, so ideally we'd make sure that 4GB is available for a full Core update without any layer sharing. But with layer sharing, and unpacking while downloading we probably get away with quite a bit less.

In a quick test on aarch64 an update from 2025.9.1 to 2025.9.3 with 1.9GB disk space free worked. Looking at the image manifests shows that 112MB is shared, the rest is not shared (~547MB), so I'd expect that more than 2GB is required even in a good case :thinking: :shrug: . Maybe Docker automatically does layer by layer in case storage gets tight :thinking: .

In the end, it's a bit a tradeoff preventing users from starting an update which anyway doesn't work, and being overly cautions. Since we have better error handling now, folks will learn when the update doesn't complete due to space issues. I think going with 2GB for now is fine.

PR #6195, opened 2025-09-17
Add progress reporting to addon, HA and Supervisor updates

mdegat01:
## Proposed change

Provides progress updates for addon installs, addon updates, Home Assistant updates and Supervisor updates. Really just an extension of #6102 to wrap up the reports we get from docker into a single progress number on these 4 jobs.

To reduce noise we are no longer reporting progress updates from all the individual layers over WS to HA. And the main job will only provide an update when the % has changed by at least 1.

The couple catches here are these:
1. We assume (for now) that the docker pull is the entirety of the job. This isn't really true. There are a few steps around the actual image pull (some can be quite lengthy like making a backup). We can revisit this in a future PR if we want to be more accurate about progress
2. We assume download and extract take equal time and everything else is negligible. Also not true but a decent guess. We can refine in future
3. There may be other inaccuracies to iron out. But should still be a step forward

Depends on #6207

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/addons/addon.py
+++ b/supervisor/addons/addon.py
@@ -768,7 +768,7 @@ class Addon(AddonModel):
         on_condition=AddonsJobError,
         concurrency=JobConcurrency.GROUP_REJECT,
     )
-    async def install(self, *, progress_job_id: str | None = None) -> None:
+    async def install(self) -> None:
         """Install and setup this addon."""
         if not self.addon_store:
             raise AddonsError("Missing from store, cannot install!")
@@ -791,10 +791,7 @@ class Addon(AddonModel):
         # Install image
         try:
             await self.instance.install(
-                self.latest_version,
-                self.addon_store.image,
-                arch=self.arch,
-                progress_job_id=progress_job_id,
+                self.latest_version, self.addon_store.image, arch=self.arch
             )
         except DockerError as err:
             await self.sys_addons.data.uninstall(self)
@@ -877,9 +874,7 @@ class Addon(AddonModel):
         on_condition=AddonsJobError,
         concurrency=JobConcurrency.GROUP_REJECT,
     )
-    async def update(
-        self, *, progress_job_id: str | None = None
-    ) -> asyncio.Task | None:
+    async def update(self) -> asyncio.Task | None:
         """Update this addon to latest version.
 
         Returns a Task that completes when addon has state 'started' (see start)
@@ -893,12 +888,7 @@ class Addon(AddonModel):
         store = self.addon_store.clone()
 
         try:
-            await self.instance.update(
-                store.version,
-                store.image,
-                arch=self.arch,
-                progress_job_id=progress_job_id,
-            )
+            await self.instance.update(store.version, store.image, arch=self.arch)
         except DockerError as err:
             raise AddonsError() from err
 
--- a/supervisor/addons/addon.py
+++ b/supervisor/addons/addon.py
@@ -768,7 +768,7 @@ class Addon(AddonModel):
         on_condition=AddonsJobError,
         concurrency=JobConcurrency.GROUP_REJECT,
     )
-    async def install(self) -> None:
+    async def install(self, *, progress_job_id: str | None = None) -> None:
         """Install and setup this addon."""
         if not self.addon_store:
             raise AddonsError("Missing from store, cannot install!")
@@ -791,7 +791,10 @@ class Addon(AddonModel):
         # Install image
         try:
             await self.instance.install(
-                self.latest_version, self.addon_store.image, arch=self.arch
+                self.latest_version,
+                self.addon_store.image,
+                arch=self.arch,
+                progress_job_id=progress_job_id,
             )
         except DockerError as err:
             await self.sys_addons.data.uninstall(self)
@@ -874,7 +877,9 @@ class Addon(AddonModel):
         on_condition=AddonsJobError,
         concurrency=JobConcurrency.GROUP_REJECT,
     )
-    async def update(self) -> asyncio.Task | None:
+    async def update(
+        self, *, progress_job_id: str | None = None
+    ) -> asyncio.Task | None:
         """Update this addon to latest version.
 
         Returns a Task that completes when addon has state 'started' (see start)
@@ -888,7 +893,12 @@ class Addon(AddonModel):
         store = self.addon_store.clone()
 
         try:
-            await self.instance.update(store.version, store.image, arch=self.arch)
+            await self.instance.update(
+                store.version,
+                store.image,
+                arch=self.arch,
+                progress_job_id=progress_job_id,
+            )
         except DockerError as err:
             raise AddonsError() from err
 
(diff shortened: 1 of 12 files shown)


supervisor/docker/interface.py
```diff
@@ -325,17 +328,65 @@ def _process_pull_image_log(self, job_id: str, reference: PullLogEntry) -> None:
...
+        # To reduce noise, limit updates to when result has changed by an entire percent or when stage changed
+        if stage != install_job.stage or progress >= install_job.progress + 1:
+            install_job.update(stage=stage.status, progress=progress)
+
     @Job(
         name="docker_interface_install",
         on_condition=DockerJobError,
         concurrency=JobConcurrency.GROUP_REJECT,
+        internal=True,
     )
```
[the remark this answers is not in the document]
agners: Yes, that is correct, but the feature was quite new and hence it is not a big deal. Just to be sure lets mark the PR as breaking change.

supervisor/docker/interface.py
```diff
@@ -325,17 +328,65 @@ def _process_pull_image_log(self, job_id: str, reference: PullLogEntry) -> None:
...
+                stage = PullImageLayerStage.DOWNLOADING
+            elif (
+                stage == PullImageLayerStage.PULL_COMPLETE
+                and job_stage < PullImageLayerStage.PULL_COMPLETE
+            ):
+                stage = PullImageLayerStage.EXTRACTING
+
+        # Ensure progress is 100 at this point to prevent float drift
+        if stage == PullImageLayerStage.PULL_COMPLETE:
+            progress = 100
```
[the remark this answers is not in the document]
mdegat01: Percents go from 0-100, over 100 makes no sense. This is the kind of float drift I was trying to prevent 😆

supervisor/docker/interface.py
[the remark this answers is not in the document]
mdegat01: No one is listening for this right now, its safe.

mdegat01:
@agners breaking change because I marked `docker_interface_install` as internal or something else? Because only home assistant could be listening for that and we know it isn't right now.

I guess it does also hide it from the REST API though. Maybe that is fair, I really doubt anyone is listening considering how recently it is added. Would you prefer I change it so `internal` only affects what gets sent over WS and internal jobs still show up in the REST API? Or some kind of third state that's like "partially internal" (no WS messages but still in REST API).

agners:
> @agners breaking change because I marked `docker_interface_install` as internal or something else? Because only home assistant could be listening for that and we know it isn't right now.

Yes, just because of that.

> I guess it does also hide it from the REST API though. Maybe that is fair, I really doubt anyone is listening considering how recently it is added. Would you prefer I change it so `internal` only affects what gets sent over WS and internal jobs still show up in the REST API? Or some kind of third state that's like "partially internal" (no WS messages but still in REST API).


Yeah I also don't think that anyone is listening. But I prefer to add the label even if the change must likely won't be noticed by any consumer of the API.

supervisor/addons/addon.py
```diff
@@ -769,7 +769,7 @@ async def _check_ingress_port(self):
         on_condition=AddonsJobError,
         concurrency=JobConcurrency.GROUP_REJECT,
     )
-    async def install(self) -> None:
+    async def install(self, *, progress_job_id: str | None = None) -> None:
```
agners: Summarizing what we just discussed in a call:

Passing job ids around feels a bit weird, since jobs are otherwise implicitly bound to execution flow. The main idea here is to have the progress on a high level job, to make it easier to consume in Core, and once we want to extend the progress to cover backup, that no Core changes would be needed.

But we might be able to do that by leveraging the Job hierarchy in the Job manager, and implicitly wire up the progress up the chain. Let's try that approach and then rebase this PR if we can make this happen.

supervisor/docker/interface.py
```diff
@@ -220,10 +220,12 @@ async def _docker_login(self, image: str) -> None:
 
         await self.sys_run_in_executor(self.sys_docker.docker.login, **credentials)
 
-    def _process_pull_image_log(self, job_id: str, reference: PullLogEntry) -> None:
+    def _process_pull_image_log(
+        self, reference_id: str, progress_job_id: str, reference: PullLogEntry
+    ) -> None:
```
agners, line 225: This functions starts to get long. In essence, this extends on #6102, disabling the individual sub jobs reporting towards Core, and consolidates the progress on the provided progress job id based on the sub job.

  Docker Events → Bus Events → Progress Sub Jobs → Job Aggregation → Main Job Progress Update

Now I'd prefer more something along these lines, it would get rid of those [in my opinion](https://github.com/home-assistant/supervisor/pull/6102#discussion_r2290460834) somewhat awkward sub jobs.

  Docker Events → Bus Events → Direct Progress Aggregation → Main Job Progress Update

Granted, we probably have to use/invent some other data structure to do the progress aggregation, since multiple layers get pulled at once etc. But it does seem _to me_ that there is quite some complexity we could safe.

This is an implementation detail, so we could also change it later :shrug:.

mdegat01:
@agners ok re-wrote to leverage the features of #6207 . I dropped the stage information entirely, it just says `None` always. I can add some if we want, just like "Taking backup", "Pulling image", "Restarting", etc. But as nothing is looking at them currently I left it for now and just removed the Downloading and Extracting.

supervisor/addons/addon.py
```diff
@@ -227,6 +227,11 @@ async def load(self) -> None:
         )
 
         await self._check_ingress_port()
+
+        # Image expected not to be here during install and will be installed later
+        if initial_install:
+            return
+
```
mdegat01: This work actually found a small bug. Nothing the user would notice really but we were attempting to install the image twice. We expected to install the image after a number of other steps had completed but because the load now checks for the image and installs if not found it was actually installing it much earlier.

This fixes it so it installs the image in the proper sequence.

supervisor/jobs/__init__.py
```diff
@@ -282,8 +282,10 @@ def _on_job_change(
                 # reporting shouldn't raise and break the active job
                 continue
 
-            progress = sync.starting_progress + (
-                sync.progress_allocation * job_data["progress"]
+            progress = min(
+                100,
+                sync.starting_progress
+                + (sync.progress_allocation * job_data["progress"]),
```
mdegat01, line 288: This isn't actually what caught the bug. What caught the bug was attempting to sync to a parent job that was already 100% complete. I considered having it skip the parent sync when the parent was like this but decided since that should never happen and was always a bug it would be better to raise so we can fix it.

If progress goes over 100 here that is also probably a developer error due to an incorrect value for progress allocation which should be fixed. But it could also be a result of float drift which is not a bug so for that reason I decided to prevent it from going over 100 with a `min`.

supervisor/addons/addon.py
agners: Should we maybe instead simply drop the install? If the semantic of `load()` is that it makes sure the image is present, it feels that would be more coherent.

agners: https://github.com/home-assistant/supervisor/pull/6195/commits/bf427f7d6c5cf8eeca2bb957d005e84c3d110341 addresses this, nice :+1: .

PR #6207, opened 2025-09-24
Add progress syncing from child jobs

mdegat01:
## Proposed change

Add support for setting progress of a parent job from one or more child jobs automatically. Required for #6195

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist


supervisor/jobs/__init__.py
```diff
@@ -62,12 +65,35 @@ def _on_change(instance: "SupervisorJob", attribute: Attribute, value: Any) -> A
...
         raise ValueError("Field cannot be updated once job has started")
 
 
+@define(frozen=True)
+class ChildJobSyncFilter:
+    """Filter to identify a child job to sync progress from."""
+
+    name: str
+    reference: str | None | type[DEFAULT] = DEFAULT
+    starting_progress: float = field(default=0.0, validator=[ge(0.0), lt(100.0)])
```
agners: Is `starting_progress` really needed? :thinking:  Couldn't we just trigger on the job end event and set it according to the progress allocation (e.g. 100*progress_allocation). This way we'd make sure we end with what we expect, and then go from there again. I guess we'd have to store the starting point for each subjob :thinking:

mdegat01: > I guess we'd have to store the starting point for each subjob 🤔

This essentially. Each subjob as it began would have to store the progress of each parent its supposed to be syncing to right at its start so it knew what the base number was to add its progress to. It just seemed easier to me to have all that captured in the parent job for easy reference by the child. Since its really the one deciding its own progress.

mdegat01:
Actually come to think of it this is really a refactor not a new feature. Its completely internal and doesn't affect users at all.

supervisor/jobs/__init__.py
agners: But would that be a problem? The sub jobs need to run sequentially anyways. We just would need to register to start/end event of each sub job. 

E.g. assuming allocation 0.5:

```
progress = 0
First sub job starts (set progress base 0)
First sub job progress updates (progress = progress base + sub job progress (e.g. 50) * progress_allocation = 25)
First sub job ends (progress = progress base + 100 * progress_allocation = 50)
Second sub job starts (set progress base 50)
Second sub job progress updates (progress = progress base + sub job progress * progress_allocation)
Second sub job ends (progress = progress base + 100 * progress_allocation
```

This would also neatly make sure that we end up with 100% even if the last sub job das not complete with 100% :smile:

agners: We need to have some signal when the relevant child jobs start/end for this to work though, but I guess we have that? :thinking:

mdegat01: So the only issue with this plan is I kind of intentionally did not force jobs to end at 100%. Because it didn't really make sense if they ended in error.

I could do something like force them to 100% if the job did not end in error. But the challenge we would have there is what if a child job ended in error and the parent job caught that error and decided to continue anyway? Now its progress is in an odd state because the first child job ended at 50% or something.

What will probably have to happen then is your script will go like this:
```
progress = 0
try
  First sub job starts (set progress base 0)
  First sub job progress updates (progress = progress base + sub job progress (e.g. 50) * progress_allocation = 25)
  First sub job ends (progress = progress base + 100 * progress_allocation = 50)
catch handleable error
  progress = 50
try
  Second sub job starts (set progress base 50)
  Second sub job progress updates (progress = progress base + sub job progress * progress_allocation)
  Second sub job ends (progress = progress base + 100 * progress_allocation
catch handleable error
  progress = 100
```
I feel like you were hoping to remove all progress updates from the main job but I don't think we can avoid that really. They can either go in the `ChildJobSyncFilter` objects or they can go inline in the code as part of error handling. But can't skip them entirely.

mdegat01: Though it would work fine for the particular use case we're eyeing since there are no handleable errors from docker install/update. So I'll write it like this for now and we'll just say if this situation occurs in the future there may be some extra metadata coding required.

supervisor/jobs/__init__.py
```diff
@@ -152,8 +179,14 @@ def start(self) -> Generator[Self]:
         try:
             token = _CURRENT_JOB.set(self.uuid)
             yield self
+        # Cannot have an else without an except so we do nothing and re-raise
+        except:  # noqa: TRY203
+            raise
+        else:
+            self.update(progress=100, done=True)
```
agners, line 186: Seems a bit weird to have except just to have the else. Couldn't we just have the else outside of the try?

mdegat01: So the problem with that is when there's no error it'll do two updates. First the `finally` will do this `self.done = True`. Then after it'll do this `self.update(progress=100, done=True)`. Which ends up sending 2 messages to HA instead of 1.

I know this doesn't look pretty but it does work a little better. Now in case of no error we do the `update` call first and then the `finally` does nothing. And if there is an error the `finally` just sets `done` without touching `progress`.

PR #6212, opened 2025-09-25
Optimize API connection handling by removing redundant port checks

agners:
## Proposed change

This PR optimizes Home Assistant API connection handling by removing redundant port checks that were creating unnecessary socket connections. The main improvement eliminates duplicate socket creation during API connectivity checks.

**Key Performance Optimization:**
- **Remove redundant port checks**: Previously, each API connectivity check performed both a port check and then an actual API request, creating two socket connections per check
- **Significant socket reduction**: During a typical Supervisor startup when Home Assistant Core is running (e.g., after a Supervisor update), approximately 10 successful API checks are made. This change eliminates 10 unnecessary socket creations per startup since aiohttp connections are reused for the actual API calls
- **Simplified token management**: Replace Job-based token refresh with a simple asyncio.Lock to avoid Job interference in exception handling and reduce overhead. This avoid two separate connection attempt per API check in case Home Assistant Core http connection attempts are not yet responding (the token refresh attempt was silently ignored).

**Improved Error Handling:**
- **Contextual API error logging**: Move exception logging from the generic `make_request` method to call sites, providing users with specific context about what operation failed
- **Better diagnostics**: Enhanced error messages throughout the codebase (auth, discovery, ingress, resolution notifications) to help diagnose API connectivity issues

The changes maintain the same functionality while improving performance and providing better debugging information when API calls to Home Assistant Core fail.

## Type of change

## Additional information

- Link to cli pull request:

## Checklist

- [ ] Client library updated (if necessary)

### DIFF
--- a/supervisor/homeassistant/api.py
+++ b/supervisor/homeassistant/api.py
@@ -169,10 +169,7 @@ class HomeAssistantAPI(CoreSysAttributes):
         async with self.make_request("get", path) as resp:
             if resp.status in (200, 201):
                 return await resp.json()
-            else:
-                raise HomeAssistantAPIError(
-                    f"Home Assistant Core API return {resp.status}"
-                )
+            raise HomeAssistantAPIError(f"Home Assistant Core API return {resp.status}")
 
     async def get_config(self) -> dict[str, Any]:
         """Return Home Assistant config."""
--- a/supervisor/homeassistant/api.py
+++ b/supervisor/homeassistant/api.py
@@ -52,6 +52,8 @@ class HomeAssistantAPI(CoreSysAttributes):
             TimeoutError: On request timeouts
 
         """
+        # Fast path check without lock (avoid unnecessary locking
+        # for the majority of calls).
         if (
             self.access_token
             and self._access_token_expires
--- a/supervisor/homeassistant/api.py
+++ b/supervisor/homeassistant/api.py
@@ -2,7 +2,7 @@
 
 import asyncio
 from collections.abc import AsyncIterator
-from contextlib import asynccontextmanager, suppress
+from contextlib import asynccontextmanager
 from dataclasses import dataclass
 from datetime import UTC, datetime, timedelta
 import logging
@@ -15,7 +15,7 @@ from multidict import MultiMapping
 
 from ..coresys import CoreSys, CoreSysAttributes
 from ..exceptions import HomeAssistantAPIError, HomeAssistantAuthError
-from ..utils import check_port, version_is_new_enough
+from ..utils import version_is_new_enough
 from .const import LANDINGPAGE
 
 _LOGGER: logging.Logger = logging.getLogger(__name__)
@@ -101,7 +101,36 @@ class HomeAssistantAPI(CoreSysAttributes):
         params: MultiMapping[str] | None = None,
         headers: dict[str, str] | None = None,
     ) -> AsyncIterator[aiohttp.ClientResponse]:
-        """Async context manager to make a request with right auth."""
+        """Async context manager to make authenticated requests to Home Assistant API.
+
+        This context manager handles authentication token management automatically,
+        including token refresh on 401 responses. It yields the HTTP response
+        for the caller to handle.
+
+        Error Handling:
+        - HTTP error status codes (4xx, 5xx) are preserved in the response
+        - Authentication is handled transparently with one retry on 401
+        - Network/connection failures raise HomeAssistantAPIError
+        - No logging is performed - callers should handle logging as needed
+
+        Args:
+            method: HTTP method (get, post, etc.)
+            path: API path relative to Home Assistant base URL
+            json: JSON data to send in request body
+            content_type: Override content-type header
+            data: Raw data to send in request body
+            timeout: Request timeout in seconds
+            params: URL query parameters
+            headers: Additional HTTP headers
+
+        Yields:
+            aiohttp.ClientResponse: The HTTP response object
+
+        Raises:
+            HomeAssistantAPIError: When request cannot be completed due to
+                network errors, timeouts, or connection failures
+
+        """
         url = f"{self.sys_homeassistant.api_url}/{path}"
         headers = headers or {}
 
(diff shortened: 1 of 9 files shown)


supervisor/homeassistant/api.py
```diff
@@ -43,22 +41,33 @@ def __init__(self, coresys: CoreSys):
...
         if (
             self.access_token
             and self._access_token_expires
             and self._access_token_expires > datetime.now(tz=UTC)
         ):
             return
 
-        with suppress(asyncio.TimeoutError, aiohttp.ClientError):
+        async with self._token_lock:
+            # Double-check after acquiring lock (avoid race condition)
```
erwindouna, line 65: Wouldn't it make sense to only check the within the lock and not outside? Feels redundant and after acquiring the lock, you can check the attribute. That's exactly how you only want to check race conditioning, right? 😃

supervisor/homeassistant/api.py
```diff
@@ -120,23 +157,22 @@ async def make_request(
...
-        raise HomeAssistantAPIError()
+                _LOGGER.debug("Error on call %s: %s", url, err)
+                raise HomeAssistantAPIError(str(err)) from err
 
     async def _get_json(self, path: str) -> dict[str, Any]:
         """Return Home Assistant get API."""
         async with self.make_request("get", path) as resp:
             if resp.status in (200, 201):
                 return await resp.json()
             else:
```
erwindouna:
```suggestion
```

Redundant nesting; single nesting.

supervisor/homeassistant/api.py
```diff
@@ -43,22 +41,33 @@ def __init__(self, coresys: CoreSys):
...
+            aiohttp.ClientError: On network or connection errors
+            TimeoutError: On request timeouts
+
+        """
         if (
             self.access_token
             and self._access_token_expires
             and self._access_token_expires > datetime.now(tz=UTC)
         ):
             return
```
erwindouna, line 62:
```suggestion
```

Suggestion is to remove it here. Since the, vital, check is done within the lock.

supervisor/ingress.py
```diff
@@ -191,12 +192,15 @@ async def update_hass_panel(self, addon: Addon):
...
-                _LOGGER.warning(
-                    "Fails Ingress panel for %s with %i", addon.slug, resp.status
-                )
+        try:
+            async with self.sys_homeassistant.api.make_request(
+                method, f"api/hassio_push/panel/{addon.slug}"
+            ) as resp:
+                if resp.status in (200, 201):
+                    _LOGGER.info("Update Ingress as panel for %s", addon.slug)
+                else:
```
agners, line 201: But that would mean the warning below would not be pritned

supervisor/homeassistant/api.py
agners: Technically the check within the lock only would be sufficient, that is true.

However, if the comparison is safe to make outside of a lock, to do so is a common pattern to avoid lock contention and improve execution speed for the usual case (fast path). The expectation is that we have many, many calls where no token refresh is necessary. Acquiring the lock adds to the execution time of each of these calls. This double check adds a bit execution time for the unusual case where a token refresh is necessary.

supervisor/ingress.py
erwindouna: Indeed, skip my feedback - I was focused on early return patterns.

supervisor/homeassistant/api.py
erwindouna: Is this feedback directed to me, or are you referring to my point on the lock and race condition?

supervisor/homeassistant/api.py
```diff
@@ -120,23 +157,22 @@ async def make_request(
                         continue
                     yield resp
                     return
-            except TimeoutError:
-                _LOGGER.error("Timeout on call %s.", url)
-                break
+            except TimeoutError as err:
+                _LOGGER.debug("Timeout on call %s.", url)
```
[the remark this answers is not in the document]
agners: `TimeoutError` is usually an empty string, this adds no value to the message (just rather makes it confusing).

supervisor/homeassistant/api.py
```diff
@@ -120,23 +157,22 @@ async def make_request(
...
+                raise HomeAssistantAPIError(str(err)) from err
             except aiohttp.ClientError as err:
-                _LOGGER.error("Error on call %s: %s", url, err)
-                break
-
-        raise HomeAssistantAPIError()
+                _LOGGER.debug("Error on call %s: %s", url, err)
+                raise HomeAssistantAPIError(str(err)) from err
```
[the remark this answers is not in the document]
agners: With `from err` we preserve the original exception, so this is already addressed.

supervisor/homeassistant/api.py
agners: Sorry, somehow I messed up my feedback on the review. My reasoning is at https://github.com/home-assistant/supervisor/pull/6212#discussion_r2393744065.

supervisor/homeassistant/api.py
erwindouna: I see your reasoning. You're making the lock the slow path. On the scale of calls you're thinking, the outer-check is essentially "free", it's a cheap attribute check (like <200 nanoseconds scale we're talking here?), whereas indeed acquiring a lock can cost more microseconds.
If that’s the case, then the outer check is justified, since it avoids paying that overhead on the hot path and only incurs it in the rare refresh case. That does save precious microseconds for a high traffic system. That's the reasoning, right? :)

agners: Yes. That said, I did not measure/prove that this is really beneficial overall. But it felt like the right thing to do :shrug:

erwindouna: In terms of high traffic systems, I can totally relate to it: here the "redunant code" is actually to facilitate a fast path.
Maybe a comment in the upper check? Otherwise, you have a chance of other nerds like me pointing it out in a future. 🤓

erwindouna (approved):
This looks good to me, @agners!

agners:
@erwindouna thanks for the review!

PR #6225, opened 2025-10-01
Fix CID file handling to prevent directory creation

agners:
## Proposed change

It seems that under certain conditions Docker creates a directory instead of a file for the CID file. It is not quite clear how this exactly happens. The Sentry issues show that it is always the Observer container which is affected. It is probably related to the fact that the Observer has automatic restart policy, which likely starts the container on restart even after CID file creation failed (e.g. due to full disk).

This change ensures that the CID file is always created as a file, and any existing directory is removed before creating the file.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/docker/manager.py
+++ b/supervisor/docker/manager.py
@@ -326,10 +326,18 @@ class DockerAPI(CoreSysAttributes):
         if name:
             cidfile_path = self.coresys.config.path_cid_files / f"{name}.cid"
 
-            # Remove the file if it exists e.g. as a leftover from unclean shutdown
-            if cidfile_path.is_file():
-                with suppress(OSError):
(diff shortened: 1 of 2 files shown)


sairon (commented):
I saw something similar when I used Docker CLI and used `-v` instead of `--mount`. Maybe we should use the [`mounts` parameter](https://docker-py.readthedocs.io/en/stable/containers.html#docker.models.containers.ContainerCollection.run) here as well instead of `volumes`? 

Anyway, if the file doesn't exist then, the Docker autostart would fail completely, so I wonder whether it's not even worse then 🤔 But if it happens when you run out of space in the data partition or when it becomes unwritable for some reason, I guess this would be just one of many other problems you will face.

PR #6236, opened 2025-10-07
Remove Codenotary integrity check

agners:
## Proposed change
Formally deprecate CodeNotary build config. 

Remove CodeNotary specific integrity checking. The current code is specific to how Codenotary's integrity checking. The cosign signing mechanism only signs the container image digest (sha256). The container image digest in turn references the image manifest (which references the layers). This setup does not allow integrity checks on disk, but only during installation/download.

To ensure container integrity on disk other mechanism like EROFS with dm-verity can be used. Integrity is validated on the fly with such mechanism. Hence no Supervisor level implementation would be needed, but correct error handling at runtime if errors are detected.

In any case, a future container image integrity likely will look different then how Codenotary looked. Remove the current code to make way for a future implementations.

This can be seen as a follow up of #4217.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to documentation pull request: https://github.com/home-assistant/developers.home-assistant/pull/2993 and https://github.com/home-assistant/home-assistant.io/pull/44066
- Link to cli pull request: https://github.com/home-assistant/cli/pull/604
- Link to client library pull request: https://github.com/home-assistant-libs/python-supervisor-client/pull/206
- Link to core pull request: https://github.com/home-assistant/core/pull/165417/changes

## Checklist


supervisor/api/security.py
```diff
@@ -54,6 +49,9 @@ async def options(self, request: web.Request) -> None:
...
     async def integrity_check(self, request: web.Request) -> dict[str, Any]:
-        """Run backend integrity check."""
-        result = await asyncio.shield(self.sys_security.integrity_check())
-        return attr.asdict(result)
+        """Run backend integrity check.
+
+        CodeNotary integrity checking has been removed. This endpoint now returns
+        an error indicating the feature is currently non-functional.
+        """
+        return {"error": "No integrity checking available"}
```
mdegat01: This isn't actually an error response. This will be a 200 response with this content in `data`.

How about a 410 Gone status error response with this as the error message? Unless there's something in core/frontend that works better with it like this?

supervisor/security/module.py
```diff
@@ -67,30 +57,6 @@ def pwned(self, value: bool) -> None:
...
-            await cas_validate(signer, checksum)
-        except CodeNotaryUntrusted:
-            raise
-        except CodeNotaryError:
-            if self.force:
-                raise
-            self.sys_resolution.create_issue(
-                IssueType.TRUST,
-                ContextType.SYSTEM,
-                suggestions=[SuggestionType.EXECUTE_INTEGRITY],
```
mdegat01, line 86: `TRUST` should be removed from the `IssueType` enum and `EXECUTE_INTEGRITY` from `SuggestionType` right? That will also require a corresponding change in the client library to remove them from the enums there.

supervisor/resolution/checks/supervisor_trust.py
```diff
@@ -1,59 +0,0 @@
...
-        if not self.sys_security.content_trust:
-            _LOGGER.warning(
-                "Skipping %s, content_trust is globally disabled", self.slug
-            )
-            return
-
-        try:
-            await self.sys_supervisor.check_trust()
-        except CodeNotaryUntrusted:
-            self.sys_resolution.add_unhealthy_reason(UnhealthyReason.UNTRUSTED)
```
mdegat01, line 33: `UNTRUSTED` should be removed from `UnhealthyReason` as well right? That will require a few changes actually:
1. Remove it from the corresponding enum in client library
2. Remove it from the list of unhealthy reasons and their translations in core
3. Remove https://www.home-assistant.io/more-info/unhealthy/untrusted

supervisor/security/module.py
```diff
@@ -11,20 +11,10 @@
...
-)
-from ..jobs.const import JobConcurrency
-from ..jobs.decorator import Job, JobCondition
-from ..resolution.const import ContextType, IssueType, SuggestionType
-from ..utils.codenotary import cas_validate
+from ..exceptions import PwnedError
 from ..utils.common import FileConfiguration
 from ..utils.pwned import check_pwned_password
 from ..validate import SCHEMA_SECURITY_CONFIG
-from .const import ContentTrustResult, IntegrityResult
```
mdegat01, line 27: These two constants should probably be tossed now

supervisor/security/module.py
```diff
@@ -11,20 +11,10 @@
     FILE_HASSIO_SECURITY,
 )
 from ..coresys import CoreSys, CoreSysAttributes
-from ..exceptions import (
-    CodeNotaryError,
-    CodeNotaryUntrusted,
```
mdegat01, line 16: These exceptions should go

supervisor/api/security.py
```diff
@@ -1,19 +1,14 @@
...
 import voluptuous as vol
 
 from ..const import ATTR_CONTENT_TRUST, ATTR_FORCE_SECURITY, ATTR_PWNED
 from ..coresys import CoreSysAttributes
 from .utils import api_process, api_validate
 
-_LOGGER: logging.Logger = logging.getLogger(__name__)
-
 # pylint: disable=no-value-for-parameter
 SCHEMA_OPTIONS = vol.Schema(
```
mdegat01, line 15: The schema for `security/options` includes `content_trust` and `force_security`:
https://github.com/home-assistant/supervisor/blob/3c21a8b8efa171bdcfc5e0be33a4c77216a6097a/supervisor/api/security.py#L18-L24
https://github.com/home-assistant/supervisor/blob/3c21a8b8efa171bdcfc5e0be33a4c77216a6097a/supervisor/api/security.py#L39-L53

Neither of these options make sense anymore. They should be removed from here, the python properties removed from security module and removed from the schema for the persistent config file for the security module here:
https://github.com/home-assistant/supervisor/blob/3c21a8b8efa171bdcfc5e0be33a4c77216a6097a/supervisor/validate.py#L230-L237

That will also require a change in the client library, cli and documentation to remove those options.

The `security/info` API should not return their values either:
https://github.com/home-assistant/supervisor/blob/3c21a8b8efa171bdcfc5e0be33a4c77216a6097a/supervisor/api/security.py#L31-L37

That will require a change in the client library and documentation.

Should be able to toss those constants after as well.

pvizeli:
The reason why I did not removed it was, that wanted to just make a swap to cosign. It should be simple to replace it with using the cosign tool but I never found the time

agners:
> The reason why I did not removed it was, that wanted to just make a swap to cosign. It should be simple to replace it with using the cosign tool but I never found the time

We could come up with a way to use cosign as a drop-in replacement where we would check the file system content at runtime (similar to how it worked on cosign). But the next step then would be how can we act on it if there are issues... And unfortunately Docker is a bit annoying in that domain: If a layer is used by another image, and the corruption happened in a file which is part of a shared part, then removing and pulling the image does not fix the integrity issue (since the shared layer will not get fetched/written again) :disappointed: . In a way, the current approach would allow to detect integrity issues, but we have no way to fix it :cry: .

I'd like to leverage containerd integrated integrity checks, namely EROFS and fsverity (got introduced with https://github.com/containerd/containerd/pull/11352). But integration of this will be different, so I think it is simpler to remove the current code and then implement support for EROFS/fsverity from ground up.

supervisor/resolution/checks/supervisor_trust.py
agners: I think I'd stop short removing untrusted from API as it causes lots of churn, we probably can reuse this with cosign. We still can drop it if we come up with a different model to report untrusted containers, but I rather prefer to keep this and remove only if we really have different needs for cosign.

supervisor/security/module.py
agners: We might need something similar, but I rather prefer to remove it today and readd if needed. Also slight different wording might make sense (like `EXECUTE_INTEGRITY_CHECK`).

supervisor/api/security.py
agners: Initially, my intention was to leave the outside API intact, so we can reuse it for the cosign implementation if it feels right. But after a bit more consideration, I think it is better to remove the API as well. With cosign I'd like to use cosign specific wording, e.g. image signature verification for the verification process etc. Content verification will work quite a bit different with EROFS/fsverity, so the API does not really match.

The `force_security` flag is also used by HIBP, so I did not remove that part.

supervisor/docker/interface.py
```diff
@@ -425,17 +422,7 @@ async def process_pull_image_log(reference: PullLogEntry) -> None:
...
-                await self._validate_trust(cast(str, docker_image.id))
-            except CodeNotaryError:
-                with suppress(docker.errors.DockerException):
-                    await self.sys_run_in_executor(
-                        self.sys_docker.images.remove,
-                        image=f"{image}:{version!s}",
-                        force=True,
-                    )
-                raise
+            # CodeNotary content trust validation has been removed
```
mdegat01: Definitely a nit but it feels odd to comment code that's not there, seems like something history handles.

tests/resolution/check/test_docker_config_volume_fix.py
```diff
@@ -0,0 +1,141 @@
+"""Test docker config check fix for VOLUME mounts."""
```
mdegat01: This file is completely unchanged right? This looks like just an unrelated rename of the test file.

mdegat01 (approved):
Looks ready to me now. We'll need a PR in ~~cli, client library~~ and doc for the removal of that API option though.

EDIT: Ok I see the first two are done, whoops. Last thing is remove it from the doc here: https://developers.home-assistant.io/docs/api/supervisor/endpoints#security

supervisor/docker/interface.py
agners: Yeah fully agree. Claude Code made this on the initial PR, I removed a couple but seems I did not catch all of them :see_no_evil: .

tests/resolution/check/test_docker_config_volume_fix.py
agners: Oh I think that was a stale file on my disk from #6119. Tests are in existing file `tests/resolution/check/test_check_docker_config.py`. So this does not belong here really. Thanks for pointing out!

PR #6241, opened 2025-10-08
Fix WebSocket transport None race condition in proxy

agners:
## Proposed change

This PR fixes a race condition in the WebSocket proxy that occurs when clients disconnect during the WebSocket handshake.

The connection can be lost in the window between the Home Assistant API state check and the `server.prepare(request)` call. When that happens, `request.transport` becomes `None`, and aiohttp's WebSocket upgrade code (`web_ws.py` `_pre_start`) fails. Prior to aiohttp 3.14.0 it hit `assert transport is not None` (`AssertionError`); as of 3.14.0 (#6902) it raises `ConnectionResetError("Connection lost")` instead. In both cases the exception propagates out of the handler unhandled, producing a 500 response and a Sentry report for what is really just a client disconnect.

On Sentry we saw ~7k events daily, limited to ~90 users. Something on those systems seems to constantly poke the WebSocket API unsuccessfully.

The fix adds a transport validity check before attempting the WebSocket upgrade. If the connection is already closed (`request.transport is None`), we log a warning and raise `HTTPBadRequest` with a clear reason. aiohttp treats the resulting 4xx as a normal response, so the race no longer surfaces as an unhandled exception or a Sentry event.

**Behavior after fix:**
- Graceful `HTTPBadRequest` response with reason "Connection closed"
- Warning logged: "WebSocket connection lost before upgrade"
- No more 500s or Sentry noise from this path

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/api/proxy.py
+++ b/supervisor/api/proxy.py
@@ -224,6 +224,11 @@ class APIProxy(CoreSysAttributes):
             raise HTTPBadGateway
         _LOGGER.info("Home Assistant WebSocket API request initialize")
 
+        # Check if transport is still valid before WebSocket upgrade
+        if request.transport is None:
+            _LOGGER.warning("WebSocket connection lost before upgrade")
+            raise web.HTTPBadRequest(reason="Connection closed")
+
         # init server
         server = web.WebSocketResponse(heartbeat=30)
         await server.prepare(request)
--- a/tests/api/test_proxy.py
+++ b/tests/api/test_proxy.py
@@ -9,7 +9,7 @@ import logging
 from typing import Any, cast
 from unittest.mock import AsyncMock, patch
 
-from aiohttp import ClientPayloadError, ClientWebSocketResponse, WSCloseCode
+from aiohttp import ClientPayloadError, ClientWebSocketResponse, WSCloseCode, web
 from aiohttp.http_websocket import WSMessage, WSMsgType
 from aiohttp.test_utils import TestClient
 import pytest
@@ -221,6 +221,32 @@ async def test_proxy_auth_abort_log(
         assert (
             "Unexpected message during authentication for WebSocket API" in caplog.text
         )
+
+
+async def test_websocket_transport_none(
+    coresys,
+    caplog: pytest.LogCaptureFixture,
+):
+    """Test WebSocket connection with transport None is handled gracefully."""
+    # Get the API proxy instance from coresys
+    api_proxy = APIProxy.__new__(APIProxy)
+    api_proxy.coresys = coresys
+
+    # Create a mock request with transport set to None to simulate connection loss
+    mock_request = AsyncMock(spec=web.Request)
+    mock_request.transport = None
+
+    caplog.clear()
+    with caplog.at_level(logging.WARNING):
+        # This should raise HTTPBadRequest, not AssertionError
+        with pytest.raises(web.HTTPBadRequest) as exc_info:
+            await api_proxy.websocket(mock_request)
+
+        # Verify the error reason
+        assert exc_info.value.reason == "Connection closed"
+
+        # Verify the warning was logged
+        assert "WebSocket connection lost before upgrade" in caplog.text
 
 
 @pytest.mark.parametrize("path", ["", "mock_path"])


agners:
Can be reproduced easily here using the following reproducer:

<details>

```python
#!/usr/bin/env python3
"""Minimal reproducer for WebSocket transport None AssertionError.

This minimal reproducer focuses on the simplest way to trigger the race condition:
- Connect to /core/websocket
- Cancel the connection immediately after starting
- Repeat 20 times to hit the narrow timing window

The bug: request.transport becomes None between check_api_state() and server.prepare()
"""

import asyncio
import aiohttp
import os
import sys


async def main():
    """Minimal reproducer - just rapid connect/cancel cycles."""
    supervisor_host = os.getenv('SUPERVISOR', 'supervisor')
    supervisor_token = os.getenv('SUPERVISOR_TOKEN')

    if not supervisor_token:
        print("ERROR: Set SUPERVISOR_TOKEN environment variable")
        sys.exit(1)

    print(f"Connecting to: {supervisor_host}")
    print("Starting rapid connect/cancel test (20 attempts)...")
    print("Monitor logs: docker logs -f hassio_supervisor | grep -E 'AssertionError|transport|WebSocket'\n")

    headers = {'Authorization': f'Bearer {supervisor_token}'}

    for attempt in range(20):
        try:
            # Create a connection but close it immediately
            # This simulates network interruption during handshake
            async with aiohttp.ClientSession(headers=headers) as session:
                # Start WebSocket connection
                ws_task = asyncio.create_task(
                    session.ws_connect(
                        f"http://{supervisor_host}/core/websocket",
                        timeout=aiohttp.ClientTimeout(total=1.0)
                    )
                )

                # Give it a tiny moment to start, then cancel
                await asyncio.sleep(0.001)  # Very short delay
                ws_task.cancel()

                try:
                    await ws_task
                except asyncio.CancelledError:
                    pass

            print(f"  Attempt {attempt}: Connection cancelled")

        except Exception as e:
            print(f"  Attempt {attempt}: Error - {type(e).__name__}: {e}")

        # Small delay between attempts
        await asyncio.sleep(0.01)

    print("\n✅ Completed 20 attempts")
    print("\nCheck for AssertionError:")
    print("  docker logs hassio_supervisor | tail -100 | grep -A 5 AssertionError")
    print("\nOr check for fix working:")
    print("  docker logs hassio_supervisor | tail -100 | grep 'WebSocket connection lost'")


if __name__ == "__main__":
    try:
        asyncio.run(main())
    except KeyboardInterrupt:
        print("\n\nInterrupted")
```

</details>

erwindouna (commented):
Not to throw in confusion, a genuine question: wouldn't it be safer/earlier to also take `WebSocketResponse.prepare(`) in consideration? It is here where the handshake and writing to the transport starts, if I am correct.

If we're thinking in high-traffic systems, the connection might drop right between the prepare and `request.transport`. If my analysis is correct, then it might be worth wrapping the call or add a targeted test to confirm it behaves as graceful as you solve the `request.transport`.

agners:
> Not to throw in confusion, a genuine question: wouldn't it be safer/earlier to also take `WebSocketResponse.prepare(`) in consideration? It is here where the handshake and writing to the transport starts, if I am correct.

Hm, yeah that is a good point. The `await` means there is no guarantee that the `prepare()` coroutine gets executed right after the None check, opening up a race window still: If the connection drops right in that small window, it might cause the assertion still to happen.

What we could do is catch assertions (it is probably fine to assume that any assertion is probably connection related):

```python
  try:
      await server.prepare(request)
  except AssertionError as err:
      # Handle transport becoming None during prepare()
      raise web.HTTPBadRequest(reason="Connection closed")
```

I wonder if using assertion on aiohttp side is really the right approach :thinking: 

@bdraco thoughts?

erwindouna:
Thanks for taking it in consideration. Other than assertion, a `if server.prepare is None` will help anyhow to safe-guard, before the try/expect - but maybe that's nitpicking and overthinking...

agners:
> Thanks for taking it in consideration. Other than assertion, a `if server.prepare is None` will help anyhow to safe-guard, before the try/expect - but maybe that's nitpicking and overthinking...

Wouldn't `if server.prepare is None` evaluate to true always? That is a coroutine of the `WebSocketResponse` object, it should never be None. Did you meant to say `if request.transport is None`?

erwindouna:
I've made a nasty typo whilst typing on the smartphone. You are correct. ;)

erwindouna:
I believe this is not stale. A question is still pending to @bdraco from @agners.

agners:
Stack trace pre-aiohttp 3.14.0:

```
2025-10-08 13:41:18.229 ERROR (MainThread) [aiohttp.server] Error handling request from 172.30.32.1
  File "/usr/local/lib/python3.13/site-packages/aiohttp/web_protocol.py", line 510, in _handle_request
  File "/usr/local/lib/python3.13/site-packages/aiohttp/web_app.py", line 569, in _handle
    return await handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.13/site-packages/aiohttp/web_middlewares.py", line 117, in impl
    return await handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/supervisor/supervisor/api/middleware/security.py", line 198, in block_bad_requests
    return await handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/supervisor/supervisor/api/middleware/security.py", line 208, in system_validation
    return await handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/supervisor/supervisor/api/middleware/security.py", line 225, in token_validation
    return await handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/supervisor/supervisor/api/middleware/security.py", line 289, in core_proxy
    return await handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/supervisor/supervisor/api/proxy.py", line 227, in websocket
    await server.prepare(request)
  File "/usr/local/lib/python3.13/site-packages/aiohttp/web_ws.py", line 214, in prepare
    protocol, writer = self._pre_start(request)
                       ~~~~~~~~~~~~~~~^^^^^^^^^
  File "/usr/local/lib/python3.13/site-packages/aiohttp/web_ws.py", line 317, in _pre_start
    assert transport is not None
           ^^^^^^^^^^^^^^^^^^^^^
AssertionError
```

mdegat01 (approved):
Good to get rid of all these sentry events and clean this up 🎉

erwindouna (approved):
Looks good to me, thanks @agners!

agners:
I was wondering if the race even still exists (no reports on 2026.06.0 on beta channel in Sentry). Turns out, Sentry explicitly does not report `ConnectionResetError` to Sentry (filtered in `integrations/aiohttp.py:137-155`).

However, without this PR, it still prints a stack trace (was able to reproduce with the following script):

<details>

```
#!/usr/bin/env python3
"""Trigger the WebSocket transport-None race in the Supervisor core proxy.

Idea: send a complete WebSocket upgrade request to /core/websocket so the
handler is invoked and starts awaiting check_api_state() (a real roundtrip to
Core). Then close/abort the client socket *during* that window so that by the
time aiohttp's WebSocketResponse._pre_start() runs, request.transport is None.

On aiohttp >=3.14.0 this raises ConnectionResetError("Connection lost")
(previously: AssertionError).
"""

import argparse
import base64
import os
import socket
import struct
import threading
import time

HOST = "172.30.32.2"
PORT = 80
PATH = "/core/websocket"


def build_request(token: str) -> bytes:
    key = base64.b64encode(os.urandom(16)).decode()
    return (
        f"GET {PATH} HTTP/1.1\r\n"
        f"Host: {HOST}\r\n"
        f"Upgrade: websocket\r\n"
        f"Connection: Upgrade\r\n"
        f"Sec-WebSocket-Key: {key}\r\n"
        f"Sec-WebSocket-Version: 13\r\n"
        f"Authorization: Bearer {token}\r\n"
        f"\r\n"
    ).encode()


def attempt(token: str, delay: float, reset: bool) -> None:
    try:
        s = socket.create_connection((HOST, PORT), timeout=5)
    except OSError:
        return
    try:
        s.sendall(build_request(token))
        # Give the server time to read the request, invoke the handler, and
        # enter the check_api_state() roundtrip, then drop the connection.
        if delay:
            time.sleep(delay)
        if reset:
            # Abortive close -> RST
            s.setsockopt(
                socket.SOL_SOCKET, socket.SO_LINGER, struct.pack("ii", 1, 0)
            )
        s.close()  # FIN (or RST if linger set)
    except OSError:
        pass


def main() -> None:
    ap = argparse.ArgumentParser()
    ap.add_argument("--token", default=os.environ.get("SUPERVISOR_TOKEN", ""))
    ap.add_argument("--count", type=int, default=2000)
    ap.add_argument("--concurrency", type=int, default=40)
    ap.add_argument(
        "--max-delay-ms",
        type=float,
        default=8.0,
        help="upper bound of random pre-close delay",
    )
    ap.add_argument("--reset", action="store_true", help="use RST instead of FIN")
    args = ap.parse_args()
    if not args.token:
        raise SystemExit("need --token or SUPERVISOR_TOKEN")

    sem = threading.Semaphore(args.concurrency)
    threads: list[threading.Thread] = []

    # Deterministic spread of delays across [0, max] without RNG dependence.
    def worker(i: int) -> None:
        try:
            delay = (i % 17) / 17.0 * (args.max_delay_ms / 1000.0)
            attempt(args.token, delay, args.reset)
        finally:
            sem.release()

    start = time.time()
    for i in range(args.count):
        sem.acquire()
        t = threading.Thread(target=worker, args=(i,), daemon=True)
        t.start()
        threads.append(t)
    for t in threads:
        t.join()
    print(
        f"sent {args.count} attempts "
        f"(concurrency={args.concurrency}, max_delay={args.max_delay_ms}ms, "
        f"mode={'RST' if args.reset else 'FIN'}) in {time.time() - start:.1f}s"
    )


if __name__ == "__main__":
    main()
```

</details>

```
2026-06-05 09:38:54.416 ERROR (MainThread) [aiohttp.server] Error handling request from 172.30.32.1
  File "/usr/local/lib/python3.14/site-packages/aiohttp/web_protocol.py", line 521, in _handle_request
  File "/usr/local/lib/python3.14/site-packages/sentry_sdk/integrations/aiohttp.py", line 209, in sentry_app_handle
    response = await old_handle(self, request)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.14/site-packages/aiohttp/web_app.py", line 559, in _handle
    return await handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.14/site-packages/aiohttp/web_middlewares.py", line 117, in impl
    return await handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/supervisor/supervisor/api/middleware/security.py", line 288, in block_bad_requests
    return await handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/supervisor/supervisor/api/middleware/security.py", line 300, in system_validation
    return await handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/supervisor/supervisor/api/middleware/security.py", line 320, in token_validation
    return await handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/supervisor/supervisor/api/proxy.py", line 229, in websocket
    await server.prepare(request)
  File "/usr/local/lib/python3.14/site-packages/aiohttp/web_ws.py", line 271, in prepare
    protocol, writer = self._pre_start(request)
                       ~~~~~~~~~~~~~~~^^^^^^^^^
  File "/usr/local/lib/python3.14/site-packages/aiohttp/web_ws.py", line 373, in _pre_start
    raise ConnectionResetError("Connection lost")
ConnectionResetError: Connection lost
```

With this PR it's just a warning:
```
2026-06-05 09:44:00.006 WARNING (MainThread) [supervisor.api.proxy] WebSocket connection lost before upgrade
```

PR #6250, opened 2025-10-14
Improve error message clarity by specifying to check supervisor logs

cogneato:
Improve error message clarity by specifying to check supervisor logs with 'ha supervisor logs'

## Type of change
This addresses a common user experience issue where users don't understand what to do when they see "Unknown error, see supervisor" messages.

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/api/utils.py
+++ b/supervisor/api/utils.py
@@ -151,7 +151,7 @@ def api_return_error(
         if check_exception_chain(error, DockerAPIError):
             message = format_message(message)
     if not message:
-        message = "Unknown error, see supervisor logs (check with 'ha supervisor logs')"
+        message = "Unknown error, see Supervisor logs (check with 'ha supervisor logs')"
 
     match error_type:
         case const.CONTENT_TYPE_TEXT:
--- a/supervisor/jobs/__init__.py
+++ b/supervisor/jobs/__init__.py
@@ -98,7 +98,9 @@ class SupervisorJobError:
     """Representation of an error occurring during a supervisor job."""
 
     type_: type[HassioError] = HassioError
-    message: str = "Unknown error, see supervisor logs (check with 'ha supervisor logs')"
+    message: str = (
+        "Unknown error, see Supervisor logs (check with 'ha supervisor logs')"
+    )
     stage: str | None = None
 
     def as_dict(self) -> dict[str, str | None]:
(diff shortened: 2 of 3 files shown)


agners:
> Nice! Only thought, do we want to suggest they use the CLI here or include a my link to supervisor logs like `https://my.home-assistant.io/redirect/logs/?provider=supervisor`?

Yeah I was going to say, the more user friendly version are the logs available through the frontend, so we probably should suggest those primarily :thinking:? it seems that nowadays they are shown even without Advanced mode, so everyone should be able to navigate there. The problem is that navigation changed at one point, but I think it did not the last 2 years so at least right now it should be the same for all users of the latest Supervisor version (Settings > System > Logs > Select "Supervisor" in the top right menu). 

But maybe instead just get rid of the unspecific error in more/all cases? Do you happen to have a concrete case in which people run into commonly we could improve?

mdegat01:
@agners That's why I suggested a my link. It doesn't matter what the navigation is now, it'll just open the Supervisor logs in the UI wherever we're putting them now.

agners:
> @agners That's why I suggested a my link. It doesn't matter what the navigation is now, it'll just open the Supervisor logs in the UI wherever we're putting them now.

Yeah, but those exceptions get raised and pop up as a message on the bottom left. We can't use my links straight in the exception message.

We'd have to catch the exception and present it somehow differently in the frontend I guess :thinking:

mdegat01:
So leave it as a CLI direction then? If it just pops up as a toast they won't be able to copy and paste that. But there's a better chance they'll remember `ha supervisor logs` then a long URL.

sairon:
I think we're heading slightly the bikeshed direction here :sweat_smile: Any change here is welcome, it's not a big deal if we change the text later as nothing (not even docs I think) depend on it.

The only minor improvement I may suggest is to change the wording from `check with 'ha supervisor logs'` to something like `e.g. with 'ha supervisor logs'`, as the current version sounds too directive, like there's no other option. But that's just my feeling.

agners:
> Nice! Only thought, do we want to suggest they use the CLI here or include a my link to supervisor logs like `https://my.home-assistant.io/redirect/logs/?provider=supervisor`?

Let's try if we can use markdown here, if that works on frontend we can go with a my link. If not, then CLI is the best option.

But ideally we should eliminate such exceptions and use custom exception classes with proper messages to inform the user on what is wrong. I created a task https://github.com/home-assistant/supervisor/issues/6285 for this.

mdegat01:
Markdown doesn't work. Going to merge this and follow up with eliminating these types of error messages via #6285

Issue #6247, opened 2025-10-10
Migrate from dockerpy to aiodocker


mdegat01:
### Description

[Dockerpy](https://pypi.org/project/docker/) is not asyncio friendly, a bit of a hassle to use and no longer actively maintained as the Docker team does not use python for docker compose anymore.

[Aiodocker](https://pypi.org/project/aiodocker/) is asyncio friendly, still a bit of a hassle to use but is actively maintained by a new group.

We should migrate to the latter and begin contributing to it in order to improve the experience. Because we need a good docker client library in Supervisor, its a big part of what we do.

mdegat01:
Going to start to collect a running list of things I've found as I go through API by API.

Differences of note:
- `images.get` is equivalent `images.inspect` in aiodocker. In Dockerpy we received an `Image` object with the inspect details in `attrs` and some other common fields mapped into properties (`id`, `tags`, etc.)
- `images.list` in aiodocker pushes all kwargs directly to the `GET images/json` API. This makes it far less friendly to use then Dockerpy's `images.list` that accepted a few common params like `name` and handled making the query string json for you. [This code](https://github.com/docker/docker-py/blob/6e6a273573fe77f00776b30de0685162a102e43f/docker/api/image.py#L59-L100) is a reference for how that works.
- `images.delete` in aiodocker = `images.remove` in dockerpy
- containers are very different from images. `containers.get` and `containers.list` return a [`DockerContainer`](https://aiodocker.readthedocs.io/en/latest/_modules/aiodocker/containers.html#DockerContainer) instance. Unlike with dockerpy this instance does not have the inspect data, its only field is `id`. You must use `DockerContainer.show` to get the inspect info which is a coroutine and makes another API call to docker.
- aiodocker has even less definition for exceptions then dockerpy. For instance to determine if an image or container is "Not Found" you catch the generic `aiodocker.exceptions.DockerError` and check if `err.status == 404`. Compare to dockerpy which raises `docker.errors.NotFound`
- To tag an image, you use `images.tag` and just pass the image id as the first parameter. Whereas dockerpy did that for you since `Image` instances had a `tag` method. It's otherwise identical
[...]

PR #6252, opened 2025-10-15
Migrate images from dockerpy to aiodocker

mdegat01:
## Proposed change

First in a series of PRs to migrate us from [Dockerpy](https://pypi.org/project/docker/) to [Aiodocker](https://pypi.org/project/aiodocker/). See #6247 for details and reasoning. This is far too much work for 1 PR so it'll be broken down into chunks.

This PR covers the following:
1. Migrates all usage of dockerpy's `ImageCollection` object to aiodocker
2. Adds `aiodocker` to our requirements
3. Initializes both client objects at startup during this migration.
4. The `DockerAPI.docker` property will now return the aiodocker client object. The `DockerAPI.dockerpy` property will be used to access the dockerpy client until it is no longer needed

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR is related to issue: #6247
- Link to cli pull request:

## Checklist

### DIFF
--- a/tests/docker/test_interface.py
+++ b/tests/docker/test_interface.py
@@ -574,7 +574,7 @@ async def test_install_progress_handles_layers_skipping_download(
 
     # Reproduce EXACT sequence from SupervisorNoUpdateProgressLogs.txt:
     # Small layer (02a6e69d8d00) completes BEFORE normal layer (3f4a84073184) starts downloading
-    coresys.docker.docker.api.pull.return_value = [
+    logs = [
         {"status": "Pulling from test/image", "id": "latest"},
         # Small layer that skips downloading (02a6e69d8d00 in logs, 96 bytes)
         {"status": "Pulling fs layer", "progressDetail": {}, "id": "02a6e69d8d00"},
@@ -622,6 +622,7 @@ async def test_install_progress_handles_layers_skipping_download(
         {"status": "Digest: sha256:test"},
         {"status": "Status: Downloaded newer image for test/image:latest"},
     ]
+    coresys.docker.images.pull.return_value = AsyncIterator(logs)
 
     # Capture immutable snapshots of install job progress using job.as_dict()
     # This solves the mutable object problem - we snapshot state at call time


supervisor/docker/manager.py
```diff
@@ -487,35 +497,35 @@ def repair(self) -> None:
...
-            output = self.docker.api.prune_builds()
+            output = self.dockerpy.api.prune_builds()
             _LOGGER.debug("Builds prune: %s", output)
         except docker_errors.APIError as err:
             _LOGGER.warning("Error for builds prune: %s", err)
 
         _LOGGER.info("Prune stale volumes")
         try:
-            output = self.docker.api.prune_builds()
+            output = self.dockerpy.api.prune_volumes()
```
mdegat01, line 521: 🙈 

No idea how long this bug has been hidden in here. There were no test cases for this method.

mdegat01:
Looking at the failed build log, I'm thinking that the idea of doing this migration in stages may not work right now. Both libraries want the same unix socket and I don't think it can work like that. We may actually have to do this as one giant PR unless we can find a way to get both running simultaneously during the migration

mdegat01:
Ok I resolved the conflict with the unix sockets. In testing locally it seems to work fine as long as you aren't using literally the same file. So by binding one to `/run/docker.sock` and the other to `/var/run/docker.sock` things seem to work fine.

agners:
> Ok I resolved the conflict with the unix sockets. In testing locally it seems to work fine as long as you aren't using literally the same file. So by binding one to `/run/docker.sock` and the other to `/var/run/docker.sock` things seem to work fine.

To me this doesn't really make sense, since the two sockets are symlinked, so these end up to be the same files. Also, I would expect that the two libraries open a separate file handle, which should not cause any interference.

I tried to reproduce that by using `/run/docker.sock` for both libraries locally, and it did not seem to cause an issue.

Afaik, these were the two problematic "Run the Supervisor" CI runs:

- https://github.com/home-assistant/supervisor/actions/runs/18538493066/job/52840267827
- https://github.com/home-assistant/supervisor/actions/runs/18541341771/job/52928360530

I've reverted the change again, but this time the run went though fine
- https://github.com/home-assistant/supervisor/actions/runs/19238700199/job/54995906263?pr=6252

I did run the CI run a couple of times, it worked every time. So, its not reproducible anymore!? :thinking: :confused:  Maybe something was wrong with GitHub back then?

supervisor/docker/manager.py
```diff
@@ -756,46 +793,57 @@ def cleanup_old_images(
         """Clean up old versions of an image."""
         image = f"{current_image}:{current_version!s}"
         try:
-            keep = {cast(str, self.images.get(image).id)}
-        except ImageNotFound:
-            raise DockerNotFound(
-                f"{current_image} not found for cleanup", _LOGGER.warning
-            ) from None
-        except (DockerException, requests.RequestException) as err:
+            try:
```
agners, line 796: Any particular reason for this nested exception handler?

The outer `aiodocker.DockerError` except is most likely not used with that.

mdegat01: To be honest, probably over-engineered DRY-ness 😄 . I need to catch `DockerError` separately from `RequestException` so I can check its `status` to see if its a 404. If its not a 404 or if its a `RequestException` I log the same thing. So my two options were:
```py
try:
  try:
    image_attr = await self.images.inspect(image)
  except DockerError as err:
    if err.status == 404:
      raise DockerNotFound
    raise
except DockerError, RequestException as err:
  raise DockerError(f"Can't get {current_image} for cleanup", _LOGGER.warning) from err
```
Or this:
```py
try:
  image_attr = await self.images.inspect(image)
except DockerError as err:
  if err.status == 404:
    raise DockerNotFound
  raise DockerError(f"Can't get {current_image} for cleanup", _LOGGER.warning) from err
except RequestException as err:
  raise DockerError(f"Can't get {current_image} for cleanup", _LOGGER.warning) from err
```
I chose the first one for DRY reasons. I can switch it though.

agners: Ah I see, yeah it is a matter of preference. I am fine with the way its now :+1: .

PR #6265, opened 2025-10-23
Fix fetching host info on setups where UDisks is not fully functional

mnederlof:
## Proposed change

Fixes a bug on my NAS, where DBUS is limited. Whenever a host info request is done from the GUI (for example when listing addons), it would return a HTTP 500, because it got an exception on the disk part and prevents the GUI from proceeding to the addons page.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/hardware/disk.py
+++ b/supervisor/hardware/disk.py
@@ -9,7 +9,12 @@ from typing import Any
 from supervisor.resolution.const import UnhealthyReason
 
 from ..coresys import CoreSys, CoreSysAttributes
-from ..exceptions import DBusError, DBusObjectError, HardwareNotFound
+from ..exceptions import (
+    DBusError,
+    DBusNotConnectedError,
+    DBusObjectError,
+    HardwareNotFound,
+)
 from .const import UdevSubsystem
 from .data import Device
 
@@ -207,6 +212,8 @@ class HwDisk(CoreSysAttributes):
         try:
             block_device = self.sys_dbus.udisks2.get_block_device_by_path(device_path)
             drive = self.sys_dbus.udisks2.get_drive(block_device.drive)
+        except DBusNotConnectedError:
+            return None
         except DBusObjectError:
             _LOGGER.warning(
                 "Unable to find UDisks2 drive for device at %s", device_path.as_posix()
--- a/tests/hardware/test_disk.py
+++ b/tests/hardware/test_disk.py
@@ -376,3 +376,14 @@ async def test_try_get_nvme_life_time_missing_percent_used(
         coresys.config.path_supervisor
     )
     assert lifetime is None
+
+
+async def test_try_get_nvme_life_time_dbus_not_connected(coresys: CoreSys):
+    """Test getting lifetime info from an NVMe when DBUS is not connected."""
+    # Set the dbus for udisks2 bus to be None, to make it forcibly disconnected.
+    coresys.dbus.udisks2.dbus = None
+
+    lifetime = await coresys.hardware.disk.get_disk_life_time(
+        coresys.config.path_supervisor
+    )
+    assert lifetime is None


agners:
> Fixes a bug on my NAS, where DBUS is limited. Whenever a host info request is done from the GUI (for example when listing addons), it would return a HTTP 500, because it got an exception on the disk part and prevents the GUI from proceeding to the addons page.

Can you share the stack trace? I guess the reason is missing UDisks in this case?

mnederlof:
Well, that was hard to find, i have to say.., there is no stacktrace to be found normally, unless you really start debugging..

Homeassistant core only has this:
```
2025-10-23 14:12:41.756 ERROR (MainThread) [homeassistant.components.hassio] Failed to to call /host/info - Unknown error, see supervisor
```
With home assistant cli i get this:
```
root@NAS:~# docker exec -it hassio_cli bash
hassio-cli:/# ha --log-level debug host info
DEBU[0000] Adding homedir to searchpath                  homedir=/root
INFO[0000] No configfile found
DEBU[0000] Debug flags                                   apiToken=... cfgFile= endpoint=supervisor logLevel=debug noProgress=false rawJSON=false
DEBU[0000] host info                                     args="[]"
DEBU[0000] [GenerateURI]                                 base=supervisor command=info section=host
DEBU[0000] [GenerateURI] Result                          uri="http://supervisor/host/info" url="http://supervisor/host/info" url(string)="http://supervisor/host/info"
DEBU[0000] Response                                      body="{\"result\":\"error\",\"message\":\"Unknown error, see supervisor\"}" fields.time=23.009206ms headers="map[Content-Length:[60] Content-Type:[application/json; charset=utf-8] Date:[Thu, 23 Oct 2025 12:22:43 GMT] Server:[Python/3.13 aiohttp/3.12.15]]" received-at="2025-10-23 14:22:43.719970025 +0200 CEST m=+0.029589023" request="&{GET http://supervisor/host/info HTTP/1.1 1 1 map[Accept:[application/json] Authorization:[Bearer ... ] User-Agent:[go-resty/2.16.5 (https://github.com/go-resty/resty)]] <nil> <nil> 0 [] false supervisor map[] map[] <nil> map[]   <nil> <nil> <nil>  {{}} <nil> [] map[]}" status="400 Bad Request" statuscode=400
Error: Unknown error, see supervisor
```

In the logs from supervisor, there is nothing, unless i adapt the api to actually raise an exception, then i see this error in supervisor logs:

```
2025-10-23 14:20:09.080 ERROR (MainThread) [aiohttp.server] Error handling request from 172.30.32.5
  File "/usr/src/supervisor/supervisor/api/host.py", line 115, in info
    ATTR_DISK_LIFE_TIME: await self.sys_host.info.disk_life_time(),
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ^
  File "/usr/src/supervisor/supervisor/hardware/disk.py", line 284, in get_disk_life_time
    nvme_lifetime = await self._try_get_nvme_lifetime(mount_source_device_name)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/supervisor/supervisor/hardware/disk.py", line 208, in _try_get_nvme_lifetime
    block_device = self.sys_dbus.udisks2.get_block_device_by_path(device_path)
    raise DBusNotConnectedError()
supervisor.exceptions.DBusNotConnectedError
```

mnederlof:
And yes, the udisks dbus is missing on my nas

mnederlof:
Updated the test, now it should hit coverage.

supervisor/hardware/disk.py
```diff
@@ -207,6 +212,8 @@ async def _try_get_nvme_lifetime(self, device_name: str) -> float | None:
         try:
             block_device = self.sys_dbus.udisks2.get_block_device_by_path(device_path)
             drive = self.sys_dbus.udisks2.get_drive(block_device.drive)
+        except DBusNotConnectedError:
+            return None
```
mdegat01, line 216: Kind of a nitpick rather then a functional issue but I'd prefer not to rely on exception handling unless necessary. So in this case think it would be better to just do this at the top of `_try_get_nvme_lifetime`:
```py
        if not self.sys_dbus.udisks2.is_connected:
            return None
```

mnederlof: Yeah, either could work, though i'd think this is_connected should be implemented elsewhere then, given there is no usage for it yet in this class, and this class mostly relies on the exceptions anyway, which is why i'd use this approach for the current fix.

agners:
> Small nitpick, I'm ok with it either way though. I think missing UDisk2 probably means an unsupported system but I suppose this could happen to anyone if UDisks2 crashed somehow so it's fine.

Yeah it seems that in today's codebase it is a feature. I think we added UDisks2 to HAOS at one point, so we support it being missing. Officially we only support recent versions of HAOS and Supervised with UDisks2, so operating without UDisks2 is not in the supported range. But since this adds minimal code, I don't mind merging this.

agners (approved):
LGTM, thanks!

mnederlof:
awesome thanks for the quick followup!

PR #6266, opened 2025-10-23
Avoid adding Content-Type to non-body responses

agners:
## Proposed change

The current code sets the content-type header for all responses to the result's content_type property if upstream does not set a content_type. The default value for content_type is "application/octet-stream".

For responses that do not have a body (like 204 No Content or 304 Not Modified), setting a content-type header is unnecessary and potentially misleading. Follow HTTP standards by only adding the content-type header to responses that actually contain a body.

This also avoid reading the body for such requests. It seems that reading the body does not cause issues, but the corresponding Core code avoided reading at all (see https://github.com/home-assistant/core/pull/147899).

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/api/ingress.py
+++ b/supervisor/api/ingress.py
@@ -254,20 +254,24 @@ class APIIngress(CoreSysAttributes):
         ) as result:
             headers = _response_header(result)
 
-            # Empty body responses (304, 204, HEAD, etc.) should not be streamed,
-            # otherwise aiohttp < 3.9.0 may generate an invalid "0\r\n\r\n" chunk
-            # This also avoids setting content_type for empty responses.
-            if must_be_empty_body(request.method, result.status):
-                return web.Response(
-                    headers=headers,
-                    status=result.status,
-                )
-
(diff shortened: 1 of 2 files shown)


agners:
I've extended the PR to preserve `Content-Type` header for head requests. [RFC 9110 says](https://www.rfc-editor.org/rfc/rfc9110#name-head):

> The server SHOULD send the same header fields in response to a HEAD request as it would have sent if the request method had been GET. 

Note that we don't send things like `Content-Length`, which is allowed for efficency:

> However, a server MAY omit header fields for which a value is determined only while generating the content. For example, some servers buffer a dynamic response to GET until a minimum amount of data is generated so that they can more efficiently delimit small responses or make late decisions with regard to content selection. Such a response to GET might contain [Content-Length](https://www.rfc-editor.org/rfc/rfc9110#field.content-length) and [Vary](https://www.rfc-editor.org/rfc/rfc9110#field.vary) fields, for example, that are not generated within a HEAD response.

PR #6272, opened 2025-10-27
Reject Core Backup if sending backup start WebSocket message fails

agners:
## Proposed change

If sending the WebSocket message to Home Assistant Core to inform it about the start of a backup fails, we should reject the backup process entirely. This prevents taking a backup without Home Assistant Core being aware of it.

As a side effect, this likely prevents situation where Core does not learn about the backup progress since we also won't be able to send WebSocket messages about Job progress updates.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/.github/workflows/builder.yml
+++ b/.github/workflows/builder.yml
@@ -320,6 +320,15 @@ jobs:
             exit 1
           fi
 
+      - name: Wait for Home Assistant Core to start
+        run: |
+          echo "Waiting for Home Assistant Core to start"
+          timeout 10m ha supervisor logs -f -n 10000 -b 0 | grep -q "Detect a running Home Assistant instance"
+          if [ "$?" != "0" ]; then
+            echo "Home Assistant Core did not start within 10 minutes"
+            exit 1
+          fi
+
       - name: Create full backup
         id: backup
         run: |


agners:
In cases where Core API does not respond (e.g. after an update or restart) we still allow to take a backup. It seems that in such cases we won't be able to inform the Core about backup starting via WebSocket either (`backup_start` WebSocket message does not get delivered). As a consequence, Job updates also do not get delivered, which the new Backup integration relies on (an example error report of such a system seems to be https://github.com/home-assistant/core/issues/143158). With this change, we will reject a backup entirely if informing the Core fails.

A downside will be that on such systems we won't take a backup. From Sentry it seems there are quite some system where the Core API does not respond properly (`HomeAssistantStartupTimeout` etc.). So I am not sure if a hanging backup is maybe the better situation then rejecting from a users perspective :thinking: 

Ideally we'd find the root cause of why Core API does not respond, so the WebSocket messages get delivered again.

/cc @emontnemery

agners:
This is superseded by #6553: Now if `begin_backup` cannot be delivered to Core, the `async_send_command` no longer returns `None` but instead raises `HomeAssistantWSError`. This is converted into a `HomeAssistantBackupError` hence leading to the same outcome this PR tried to implement.

Issue #6285, opened 2025-11-04
Eliminate exceptions without message reaching api_return_error


agners:
### Description

The generic API handler `api_return_error` converts exceptions without message into "Unknown error, see supervisor". Such errors are not very useful to users, we should replace them with custom exception classes along with a user friendly message telling what is wrong.

Let's eliminate obvious/known cases where we raise Exceptions without messages first, then capture such events in Sentry to find other commonly occurring errors without messages.

agners:
One of the issues seen in Support is when people are using `ha auth reset --username test --password <newpw>`, so far this lead to:

```
Error: Unknown error, see Supervisor logs
```

Which very much sounds like a problem with the command in first place. Since we have `ha auth reset --interactive` which lists users as well, the message should make clear that the username is wrong.

agners:
We eliminated blank errors (without information at all) and made a few common ones which were raised as unknown with translatable error. At this point we didn't see more complaints about spurious unknown errors.

PR #6303, opened 2025-11-10
Remove unknown errors from addons and auth

mdegat01:
## Proposed change

So the driving force behind this is fixing #6285. Because frankly we never want to return that unknown error message with no other information to users.

But the problem with tracking down these message-less errors is then you have to figure out what was the actual source we're not saying. And then when improving those errors we have an ongoing effort to remap all errors output by Supervisor to use the new `error_key` and `extra_fields` concept to support localization.

So this is going to be the first in a series of PRs about this. Going to go a chunk at a time with the message-less errors we raise (this one starts with addons and auth). Along the way i will also be improving other errors that are connected to these spots that technically have a message but it isn't particularly informative. As you can see there are a lot of changes to errors in docker-related code, not all of which began message-less.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request: https://github.com/home-assistant/cli/pull/620

## Checklist

### DIFF
--- a/supervisor/exceptions.py
+++ b/supervisor/exceptions.py
@@ -568,15 +568,11 @@ class AuthError(HassioError):
     """Auth errors."""
 
 
-# This one uses the check logs rider even though its not a 500 error because it
-# is bad practice to return error specifics from a password reset API.
 class AuthPasswordResetError(AuthError, APIError):
     """Auth error if password reset failed."""
 
     error_key = "auth_password_reset_error"
-    message_template = (
-        f"Unable to reset password for '{{user}}'. {MESSAGE_CHECK_SUPERVISOR_LOGS}"
-    )
+    message_template = "Username '{user}' does not exist. Check list of users using '{auth_list_command}'."
 
     def __init__(
         self,
@@ -585,7 +581,7 @@ class AuthPasswordResetError(AuthError, APIError):
         user: str,
     ) -> None:
         """Initialize exception."""
-        self.extra_fields = {"user": user} | EXTRA_FIELDS_LOGS_COMMAND
+        self.extra_fields = {"user": user, "auth_list_command": "ha auth list"}
         super().__init__(None, logger)
 
 
--- a/tests/api/test_auth.py
+++ b/tests/api/test_auth.py
@@ -139,12 +139,12 @@ async def test_failed_password_reset(
     body = await resp.json()
     assert (
         body["message"]
-        == "Unable to reset password for 'john'. Check supervisor logs for details (check with 'ha supervisor logs')"
+        == "Username 'john' does not exist. Check list of users using 'ha auth list'."
     )
     assert body["error_key"] == "auth_password_reset_error"
     assert body["extra_fields"] == {
         "user": "john",
-        "logs_command": "ha supervisor logs",
+        "auth_list_command": "ha auth list",
     }
     assert expected_log in caplog.text
 


mdegat01:
@agners I made this in draft because tbh its ominous that all the tests passed given all the errors I changed. More *should* have failed 🙈 

So I'm going to add a bunch of tests before I call this ready. But since I know this was taking too long I wanted to get what I had in front of you so you could take a look. And if you have any fundamental issues with how I'm doing this let me know.

supervisor/addons/addon.py
```diff
@@ -793,9 +799,17 @@ def setup_data():
             await self.instance.install(
                 self.latest_version, self.addon_store.image, arch=self.arch
             )
+        except AddonsError:
+            await self.sys_addons.data.uninstall(self)
+            raise
+        except DockerBuildError as err:
+            _LOGGER.error("Could not build image for addon %s: %s", self.slug, err)
+            await self.sys_addons.data.uninstall(self)
+            raise AddonUnknownError(addon=self.slug) from err
```
agners: Shouldn't this be an error with actual texts so the user can see the issue?

agners: The current code leads to:

Image

If we were to raise `DockerBuildError` directly, it would lead to:

Image

Now granted, the second option is maybe not very user friendly since it contains all the error data. But this is also a bit a rendering issue, if newlines in the error message would be converted to line breaks, it would be quite readable.

The error itself would not be translatable, I guess that is a problem.

mdegat01: Yea I mean I thought that was our decision process here. Focus on making errors we display in the UI useful and translatable. If the only way the user can solve the problem is by looking at the log or turning on debugging tools on their system then just have the error message say that. Don't dump a giant untranslatable stack trace on the user in a pop-up, just display that we don't know what happened and direct them to the log with a translatable error.

But as we agreed I'll make one that specifically says the build failed rather then just "something is wrong with the addon". It'll still basically just direct them to the log for all the rest of the info though.

supervisor/exceptions.py
```diff
@@ -41,6 +41,94 @@ class HassioNotSupportedError(HassioError):
...
+
+
+class APIDBMigrationInProgress(APIError):
+    """Service is unavailable due to an offline DB migration is in progress."""
+
+    status = 503
+
+
+class APIUnknownSupervisorError(APIError):
+    """Unknown error occurred within supervisor. Adds supervisor check logs rider to mesage template."""
```
agners:
```suggestion
    """Unknown error occurred within supervisor. Adds supervisor check logs rider to message template."""
```

supervisor/exceptions.py
```diff
@@ -346,13 +568,68 @@ class AuthError(HassioError):
     """Auth errors."""
 
 
-class AuthPasswordResetError(HassioError):
+# This one uses the check logs rider even though its not a 500 error because it
+# is bad practice to return error specifics from a password reset API.
+class AuthPasswordResetError(AuthError, APIError):
```
agners, line 571: I get what you mean, but this only really applies if the API is accessible by a less privileged user, to avoid information leak (e.g. email field to reset password on the open Internet). But this does not apply here: The API is only accessible to admin users (CLI and Home Assistant Core). And since we have `/auth/list` too (introduced in #4912), it is pointless to avoid leaking information, if there is just an API next to it which allows to list all users 🤷 . At this point its only a user inconvenience, they have to jump through another hoops to understand what's wrong with the entered username.

So, I'd go for user friendliness here, and straight tell the user what's on:

```
Error: Username 'test2' does not exist. Check list of users using 'ha auth list'.
```

If we want to prevent our API from leaking user information, we'd have to revert the `/auth/list` API (and remove the interactive mode). But from what I can tell, the this is currently by design. Since user database is stored unencrypted on that system anyways. With physical access, it's game over anyways.

This particular error case has actually been pointed out previously (I've added a note in the task too, see  https://github.com/home-assistant/supervisor/issues/6285#issuecomment-3606652613).

PR #6311, opened 2025-11-13
Check frontend availability after Home Assistant Core updates

agners:
## Proposed change

This PR enhances the Home Assistant Core update verification process to ensure the frontend is actually accessible and serving properly after updates, not just that the API endpoints respond.

Currently, after a core update, Supervisor only checks if the API endpoints (`/api/core/state` or `/api/config`) respond and whether the `frontend` component is loaded. This could miss cases where the API is responsive but the frontend fails to serve the web UI, leaving users with a seemingly successful update but an inaccessible interface.

This PR adds a `check_frontend_available()` method that fetches the root path (`/`) and verifies it returns HTML content with a 200 status code. This check runs after confirming the frontend component is loaded. If the frontend is inaccessible, the update automatically triggers a rollback to the previous version.

Additionally, this PR fixes blocking I/O operations in the rollback log file handling by using the async executor pattern.

## Type of change

- [ ] Dependency upgrade

## Additional information

## Checklist

### DIFF
--- a/tests/homeassistant/test_api.py
+++ b/tests/homeassistant/test_api.py
@@ -89,3 +89,119 @@ async def test_check_frontend_available_api_error(coresys: CoreSys):
         result = await coresys.homeassistant.api.check_frontend_available()
 
     assert result is False
+
+
+async def test_get_config_success(coresys: CoreSys):
+    """Test get_config returns valid config dictionary."""
+    coresys.homeassistant.version = AwesomeVersion("2025.8.0")
+
+    expected_config = {
+        "latitude": 32.87336,
+        "longitude": -117.22743,
+        "elevation": 0,
+        "unit_system": {
+            "length": "km",
+            "mass": "g",
+            "temperature": "°C",
+            "volume": "L",
+        },
+        "location_name": "Home",
+        "time_zone": "America/Los_Angeles",
+        "components": ["frontend", "config"],
+        "version": "2025.8.0",
+    }
+
+    mock_response = MagicMock()
+    mock_response.status = 200
+
+    async def mock_json():
+        return expected_config
+
+    mock_response.json = mock_json
+
+    @asynccontextmanager
+    async def mock_make_request(*_args, **_kwargs):
+        yield mock_response
+
+    with patch.object(
+        type(coresys.homeassistant.api), "make_request", new=mock_make_request
+    ):
+        result = await coresys.homeassistant.api.get_config()
+
+    assert result == expected_config
+
+
+async def test_get_config_returns_none(coresys: CoreSys):
+    """Test get_config raises error when None is returned."""
+    coresys.homeassistant.version = AwesomeVersion("2025.8.0")
+
+    mock_response = MagicMock()
+    mock_response.status = 200
+
+    async def mock_json():
+        return None
+
+    mock_response.json = mock_json
+
+    @asynccontextmanager
+    async def mock_make_request(*_args, **_kwargs):
+        yield mock_response
+
+    with (
+        patch.object(
+            type(coresys.homeassistant.api), "make_request", new=mock_make_request
+        ),
+        pytest.raises(
+            HomeAssistantAPIError, match="No config received from Home Assistant API"
+        ),
+    ):
+        await coresys.homeassistant.api.get_config()
+
+
+async def test_get_config_returns_non_dict(coresys: CoreSys):
+    """Test get_config raises error when non-dict is returned."""
+    coresys.homeassistant.version = AwesomeVersion("2025.8.0")
+
+    mock_response = MagicMock()
+    mock_response.status = 200
+
+    async def mock_json():
+        return ["not", "a", "dict"]
+
+    mock_response.json = mock_json
+
+    @asynccontextmanager
+    async def mock_make_request(*_args, **_kwargs):
+        yield mock_response
+
+    with (
+        patch.object(
+            type(coresys.homeassistant.api), "make_request", new=mock_make_request
+        ),
+        pytest.raises(
+            HomeAssistantAPIError, match="No config received from Home Assistant API"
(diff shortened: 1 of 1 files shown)


agners:
When trying to update to todays nightly build with a bad frontend manually like so:

```
ha core update --version 2025.12.0.dev202511130240
```

The Supervisor rolls back the Core to the previous version

```
2025-11-13 09:38:35.853 WARNING (MainThread) [supervisor.homeassistant.api] Frontend returned status 500
2025-11-13 09:38:35.853 ERROR (MainThread) [supervisor.homeassistant.core] Frontend component loaded but frontend is not accessible
2025-11-13 09:38:35.853 CRITICAL (MainThread) [supervisor.homeassistant.core] HomeAssistant update failed -> rollback!
2025-11-13 09:38:35.855 INFO (MainThread) [supervisor.resolution.module] Create new issue update_rollback - core / None
```

agners:
What I did notice is that we delete the old image _before_ rollback. This means we clean the image, then redownload it again.

This always used to be the case, and maybe we could even make the case that this makes sure the previous image is properly downloaded again. On the other hand, this redownload might fail. If Core worked before, the image on disk is most likely in a good state, so this only adds the risk that rollback fails :thinking:  I'll create a separate PR for this.

```
2025-11-13 09:38:32.372 INFO (MainThread) [supervisor.homeassistant.core] Successfully started Home Assistant 2025.12.0.dev202511130240
...
2025-11-13 09:38:32.396 INFO (MainThread) [supervisor.docker.manager] Cleanup images: ['ghcr.io/home-assistant/qemux86-64-homeassistant:2025.12.0.dev202511120239']
...
2025-11-13 09:38:35.853 WARNING (MainThread) [supervisor.homeassistant.api] Frontend returned status 500
2025-11-13 09:38:35.853 ERROR (MainThread) [supervisor.homeassistant.core] Frontend component loaded but frontend is not accessible
2025-11-13 09:38:35.853 CRITICAL (MainThread) [supervisor.homeassistant.core] HomeAssistant update failed -> rollback!
...
2025-11-13 09:38:35.856 INFO (MainThread) [supervisor.homeassistant.core] Updating Home Assistant to version 2025.12.0.dev202511120239
```

tests/homeassistant/test_api.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+async def test_check_frontend_available_api_error(coresys: CoreSys):
+    """Test frontend availability check handles API errors gracefully."""
+    coresys.homeassistant.version = AwesomeVersion("2025.8.0")
+
+    @asynccontextmanager
+    async def mock_make_request(*args, **kwargs):
+        raise HomeAssistantAPIError("Connection failed")
+        yield  # pragma: no cover
```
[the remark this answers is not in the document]
agners: @copilot this is needed otherwise the test fails with:

```
self = <contextlib._AsyncGeneratorContextManager object at 0x7f9811c042b0>

    async def __aenter__(self):
        # do not keep args and kwds alive unnecessarily
        # they are only needed for recreation, which is not possible anymore
        del self.args, self.kwds, self.func
        try:
>           return await anext(self.gen)
                         ^^^^^^^^^^^^^^^
E           TypeError: 'coroutine' object is not an async iterator

../.pyenv/versions/3.13.5/lib/python3.13/contextlib.py:214: TypeError
```

supervisor/homeassistant/core.py
```diff
@@ -303,12 +303,22 @@ async def _update(to_version: AwesomeVersion) -> None:
             except HomeAssistantError:
                 # The API stoped responding between the up checks an now
                 self._error_state = True
-                data = None
+                return
+
+            if data is None:
+                self._error_state = True
+                return
```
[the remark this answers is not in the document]
agners: Technically, `get_config()` can return None if Core returns with an empty body etc, the type annotations are not really accurate.

But rather than checking this on call site, it is better to make a basic validation in `get_config()` .

supervisor/homeassistant/api.py
```diff
@@ -219,3 +222,36 @@ async def check_api_state(self) -> bool:
...
+        Returns:
+            True if the frontend responds successfully, False otherwise.
+
+        """
+        # Skip check on landingpage
+        if (
+            self.sys_homeassistant.version is None
+            or self.sys_homeassistant.version == LANDINGPAGE
+        ):
+            return False
```
mdegat01: Actually wait, should this return `True`? I get why you want to skip the check in this case but auto returning `False` here seems like it could be harmful since it tells the caller the check failed. Normally when a check is skipped you either return `True` or something to tell the caller to ignore this check this time.

agners: We actually call `check_frontend_available` after fetching the config via Core API `api/config` and checking that `frontend` got loaded. So at this point its pretty much a given that we are dealing with actually Core (not landing page).

Let's simply defer this problem to the caller. I remove the check and adjust the doc string accordingly.

supervisor/homeassistant/api.py
```diff
@@ -175,7 +175,10 @@ async def _get_json(self, path: str) -> dict[str, Any]:
 
     async def get_config(self) -> dict[str, Any]:
         """Return Home Assistant config."""
-        return await self._get_json("api/config")
+        config = await self._get_json("api/config")
+        if config is None or not isinstance(config, dict):
+            raise HomeAssistantAPIError("No config received from Home Assistant API")
+        return config
```
mdegat01, line 181: I guess now that this has some more logic in it we should probably have at least one test that doesn't mock the method entirely and uses mock requests to return a fixture, since codecov says this block is totally uncovered.  We can come back to it though, still pretty simple code here.

agners: Good point, just added test coverage for it!

PR #6324, opened 2025-11-19
Fix call_at to use event loop time base instead of Unix timestamp

agners:
## Proposed change

This PR fixes a critical bug in `CoreSys.call_at()` where Unix timestamps were incorrectly passed to `asyncio.loop.call_at()`, which expects times in the event loop's monotonic time base. This caused scheduled jobs to be scheduled approximately 55 years in the future.

The bug was previously masked by time-machine 2.19.0, which patched `time.monotonic()` and made `loop.time()` return Unix timestamps. Time-machine 3.0.0 removed this patching (as it caused event loop freezes), exposing the bug.

**Changes:**
- Fixed `CoreSys.call_at()` to properly convert datetime to event loop time base
- Simplified `test_job_scheduled_at` to avoid time-machine async context managers
- Test now follows the same pattern as `test_job_scheduled_delay`

**Root cause:**
- `asyncio.loop.time()` is based on `time.monotonic()` (~66,000 seconds since boot)
- `datetime.timestamp()` returns Unix time (~1,763,000,000 seconds since 1970)
- Difference: ~55 years

**Fix:**
```python
delay = when.timestamp() - time.time()  # Calculate delay
loop_time = self.loop.time() + delay    # Convert to loop time base
loop.call_at(loop_time, funct, ...)     # Correct!
```

This fix is necessary regardless of time-machine version and prevents the bug from manifesting in production.

This got uncovered by the dependabot time-machine bump to 3.0.0 with #6321.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR is related to issue: Time-machine 3.0.0 upgrade

## Checklist

### DIFF
--- a/supervisor/coresys.py
+++ b/supervisor/coresys.py
@@ -657,6 +657,8 @@ class CoreSys:
             funct = partial(funct, **kwargs)
 
         # Convert datetime to event loop time base
+        # If datetime is in the past, delay will be negative and call_at will
+        # schedule the call as soon as possible.
         delay = when.timestamp() - time.time()
         loop_time = self.loop.time() + delay
 


supervisor/coresys.py
```diff
@@ -655,8 +656,12 @@ def call_at(
         if kwargs:
             funct = partial(funct, **kwargs)
 
+        # Convert datetime to event loop time base
+        delay = when.timestamp() - time.time()
+        loop_time = self.loop.time() + delay
```
sairon, line 663: Unlike the previous (broken) implementation, this can potentially result in negative `when` argument for the `call_at` function if you provide a datetime before the application start. What happens (and what _should_ happen in our scheduler) then? Same, though less edge-casey, situation is when the time would be before "now".

sairon: FWIW a quick test shows that anything that's not in the future (including negative time) results in the callback to run immediately. So if it's okay and we don't want to prevent from it, even a negative `loop_time`  is fine and no handling of that is needed.

agners: Yeah that sounds like a useful default behavior. Let me add a comment to point that out. Thanks for testing!

sairon (approved):
LGTM!

PR #6360, opened 2025-11-27
Use Docker's official registry domain detection logic

agners:
## Proposed change

Replace the custom `IMAGE_WITH_HOST` regex with a proper implementation based on Docker's reference parser (`vendor/github.com/distribution/reference/normalize.go`).

The previous regex was inaccurate - it only matched domains with a TLD suffix (e.g., `ghcr.io`) but missed valid registry domains like:
- `localhost` and `localhost:5000`
- Registries with ports only (e.g., `myregistry:5000`)
- IPv6 addresses (e.g., `[::1]:5000`)

This PR also changes the Docker Hub identifier from `hub.docker.com` to `docker.io` (the official default registry), while maintaining backward compatibility with the legacy identifier.

**Changes:**
- Change `DOCKER_HUB` from `"hub.docker.com"` to `"docker.io"` (official default)
- Add `DOCKER_HUB_LEGACY` for backward compatibility with `"hub.docker.com"`
- Add `IMAGE_DOMAIN_REGEX` and `get_domain()` function based on Docker's normalize.go
- Update credential handling to support both `docker.io` and `hub.docker.com`
- Add comprehensive tests for domain detection

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/docker/manager.py
+++ b/supervisor/docker/manager.py
@@ -52,7 +52,7 @@ from ..validate import SCHEMA_DOCKER_CONFIG
 from .const import DOCKER_HUB, DOCKER_HUB_LEGACY, LABEL_MANAGED
 from .monitor import DockerMonitor
 from .network import DockerNetwork
-from .utils import get_domain_from_image
+from .utils import get_registry_from_image
 
 _LOGGER: logging.Logger = logging.getLogger(__name__)
 
@@ -216,12 +216,12 @@ class DockerConfig(FileConfiguration):
             return None
 
         # Check if image uses a custom registry (e.g., ghcr.io/org/image)
-        domain = get_domain_from_image(image)
-        if domain:
-            if domain in self.registries:
-                return domain
+        registry = get_registry_from_image(image)
+        if registry:
+            if registry in self.registries:
+                return registry
         else:
-            # No domain prefix means Docker Hub
+            # No registry prefix means Docker Hub
             # Support both docker.io (official) and hub.docker.com (legacy)
             if DOCKER_HUB in self.registries:
                 return DOCKER_HUB
--- a/supervisor/docker/utils.py
+++ b/supervisor/docker/utils.py
@@ -14,11 +14,11 @@ import re
 # - Contains ":" (like myregistry:5000)
 # - IPv6 addresses in brackets (like [::1]:5000)
 #
-# Note: Docker also treats uppercase letters as domain indicators since
+# Note: Docker also treats uppercase letters as registry indicators since
 # namespaces must be lowercase, but this regex handles lowercase matching
-# and the get_domain_from_image() function validates the domain rules.
-IMAGE_DOMAIN_REGEX = re.compile(
-    r"^(?P<domain>"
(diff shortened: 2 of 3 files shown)


supervisor/docker/const.py
```diff
@@ -15,11 +15,64 @@
...
+        get_domain("myregistry.com/nginx")         -> "myregistry.com"
+        get_domain("localhost/myimage")            -> "localhost"
+        get_domain("localhost:5000/myimage")       -> "localhost:5000"
+        get_domain("registry.io:5000/org/app:v1")  -> "registry.io:5000"
+        get_domain("[::1]:5000/myimage")           -> "[::1]:5000"
+
+    """
+    match = IMAGE_DOMAIN_REGEX.match(image_ref)
+    if match:
+        domain = match.group(1)
```
sairon: Using named group at least for the domain part would be nice here.

supervisor/docker/const.py
```diff
@@ -15,11 +15,64 @@
...
+    r"localhost(?::[0-9]+)?|"  # localhost with optional port
+    r"(?:[a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9-]*[a-zA-Z0-9])"  # domain component
+    r"(?:\.(?:[a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9-]*[a-zA-Z0-9]))*"  # more components
+    r"(?::[0-9]+)?|"  # optional port
+    r"\[[a-fA-F0-9:]+\](?::[0-9]+)?"  # IPv6 with optional port
+    r")/"  # must be followed by /
+)
+
+
+def get_domain(image_ref: str) -> str | None:
```
sairon: Isn't a function in the `const.py` module kind of antipattern? Something like `utils.get_domain_for_image` might be nicer (to make it similar to `get_registry_for_image` and preserve easy testability).

sairon (commented):
Looks better, but now when looking at the individual capture groups in the regexp, I realized one more thing - we're not extracting _domain_, we're extracting _registry URL_ from the image name. So `get_registry_from_image` could be more appropriate?

agners:
> Looks better, but now when looking at the individual capture groups in the regexp, I realized one more thing - we're not extracting _domain_, we're extracting _registry URL_ from the image name. So `get_registry_from_image` could be more appropriate?

But is URL correct? That usually includes protocol as well no?

Probably FQDN would be correct. But I think domain is a bit "Docker speak" here, see https://github.com/moby/moby/blob/v28.3.3/vendor/github.com/distribution/reference/normalize.go#L11-L31 :shrug.

sairon:
I think FQDN isn't correct either, as this identifier includes port as well. I think that just "registry" (without "URL") is appropriate for this helper function too. In the end we do:
```
def get_registry_for_image(self, image: str) -> str | None:
(...)
    domain = get_domain_from_image(image)
    if domain:
        if domain in self.registries:
            return domain
```
So semantically the helper function returns "registry" as well.

PR #6371, opened 2025-12-01
Use count-based progress for Docker image pulls

agners:
## Proposed change

Refactor Docker image pull progress to use a simpler count-based approach where each layer contributes equally (100% / total_layers) regardless of size.

The core issue was that Docker rate-limits concurrent downloads (by default 3 at a time, see Docker [DefaultMaxConcurrentDownloads](https://github.com/moby/moby/blob/bea959c7b793b32a893820b97c4eadc7c87fabb0/daemon/config/config.go#L32) constant) and reports layer sizes only when downloading starts. With the previous size-weighted progress calculation, large layers appearing late would cause progress to drop dramatically (e.g., 59% -> 29%) as the total size increased. We prevented the progress from going backwards, but in practice that meant the progress would stale for an extended amount of time.

The new approach:
- Each layer contributes equally to overall progress
- Per-layer progress: 70% download weight, 30% extraction weight
- Progress only starts after first "Downloading" event (when full layer count is known)
- Always caps at 99% - job completion handles final 100%

This moves progress tracking to a dedicated module (`pull_progress.py`) and removes the complex size-based scaling logic that tried to account for unknown layer sizes. With this, there is always some progress. Unfortunately, it also means that progress slows down towards the completion, since the larger layers are then still downloaded/extracted. But this is the best we can do currently.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

- [ ] Client library updated (if necessary)

### DIFF
--- a/supervisor/docker/pull_progress.py
+++ b/supervisor/docker/pull_progress.py
@@ -237,9 +237,8 @@ class ImagePullProgress:
     def calculate_progress(self) -> float:
         """Calculate overall progress 0-100.
 
-        Uses count-based progress where each layer contributes equally.
-        Each layer's individual progress (0-100) is weighted by 1/total_layers.
-        This ensures progress never goes backwards when large layers appear late.
+        Uses count-based progress where each layer that needs pulling contributes
+        equally. Layers that already exist locally are excluded from the calculation.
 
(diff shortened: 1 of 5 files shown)


agners:
With #6379 landed this is obsolete. Closing.

PR #6374, opened 2025-12-01
Use unpublished local wheels during PR builds

sairon:
## Proposed change

Refactor wheel building to use the new `local-wheels-repo-path` and move wheels building into a separate CI job. Wheels are only published on published (i.e. release or merged dev), for PR builds they are passed as artifacts to the build job instead.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/.github/workflows/builder.yml
+++ b/.github/workflows/builder.yml
@@ -72,13 +72,12 @@ jobs:
 
       - name: Get changed files
         id: changed_files
-        if: steps.version.outputs.publish == 'false'
         uses: masesgroup/retrieve-changed-files@491e80760c0e28d36ca6240a27b1ccb8e1402c13 # v3.0.0
 
       - name: Check if requirements files changed
         id: requirements
         run: |
-          if [[ "${{ steps.changed_files.outputs.all }}" =~ (requirements.txt|build.yaml|builder.yml) ]]; then
+          if [[ "${{ steps.changed_files.outputs.all }}" =~ (requirements\.txt|build\.yaml|\.github/workflows/builder\.yml) ]]; then
             echo "changed=true" >> "$GITHUB_OUTPUT"
           fi
 
@@ -131,7 +130,7 @@ jobs:
           wheels-host: ""
           wheels-user: ""
           wheels-key: ""
-          local-wheels-repo-path: "wheels"
+          local-wheels-repo-path: "wheels/"
           abi: ${{ env.ABI }}
           tag: ${{ env.TAG }}
           arch: ${{ matrix.arch }}


.github/workflows/builder.yml
```diff
@@ -78,49 +78,101 @@ jobs:
       - name: Check if requirements files changed
         id: requirements
         run: |
-          if [[ "${{ steps.changed_files.outputs.all }}" =~ (requirements.txt|build.yaml) ]]; then
+          if [[ "${{ steps.changed_files.outputs.all }}" =~ (requirements.txt|build.yaml|builder.yml) ]]; then
```
agners: Should we use full path for builder.yaml? Probably doesn't matter much, but makes the intention a bit clearer:

```suggestion
          if [[ "${{ steps.changed_files.outputs.all }}" =~ (requirements.txt|build.yaml|.github/workflows/builder.yml) ]]; then
```

.github/workflows/builder.yml
```diff
@@ -78,49 +78,101 @@ jobs:
...
           requirements: "requirements.txt"
 
+      - name: Build local wheels
+        uses: home-assistant/wheels@e5742a69d69f0e274e2689c998900c7d19652c21 # 2025.12.0
+        if: needs.init.outputs.publish == 'false'
+        with:
+          wheels-host: ""
+          wheels-user: ""
+          wheels-key: ""
+          local-wheels-repo-path: "wheels"
```
agners: I often add a slash in the end just to make it obvious that we are dealing with a path (and at times, it helps to avoid surprises when a target can get interpreted as a file or directory).

```suggestion
          local-wheels-repo-path: "wheels/"
```

.github/workflows/builder.yml
```diff
@@ -78,49 +78,101 @@ jobs:
...
-      contents: read
-      id-token: write
-      packages: write
+    if: needs.init.outputs.requirements == 'true'
+    runs-on: ${{ matrix.runs-on }}
     strategy:
       matrix:
         arch: ${{ fromJson(needs.init.outputs.architectures) }}
+        include:
+          - runs-on: ubuntu-24.04
```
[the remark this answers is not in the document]
agners: > The matrix `include` configuration is malformed. Line 94 creates an entry with only `runs-on: ubuntu-24.04` without specifying an `arch`, which doesn't match any matrix combination and will either be ignored or cause an error.

@copilot  isn't the `arch` set on matrix level setting `arch` for entries which are not set explicitlyi by `include`?

.github/workflows/builder.yml
```diff
@@ -78,49 +78,101 @@ jobs:
...
+          - arch: aarch64
+            runs-on: ubuntu-24.04-arm
```
agners, line 95: Copilot might be right here, I think this is the right syntax:

```suggestion
      matrix:
        arch: ${{ fromJson(needs.init.outputs.architectures) }}
        runs-on: [ubuntu-24.04]
        include:
          - arch: aarch64
            runs-on: ubuntu-24.04-arm
```

sairon: No, it's not: https://github.com/home-assistant/docker/blob/5a138e0b1b57875ec900589ce52eba3f849745a9/.github/workflows/builder.yml#L38-L49

.github/workflows/builder.yml
sairon: I would ultimately like to switch to [tj-actions/changed-files](https://github.com/tj-actions/changed-files) which we migrated to in some other places already.

.github/workflows/builder.yml
agners: Hm, I see, since `runs-on` is not defined above, it can get added through `include` to all combinations. In the end, both then should work work :thinking: . Anyways, but I agree this is correct.

.github/workflows/builder.yml
```diff
@@ -78,49 +78,101 @@ jobs:
...
+        uses: actions/upload-artifact@330a01c490aca151604b8cf639adc76d48f6c5d4 # v5.0.0
+        with:
+          name: wheels-${{ matrix.arch }}
+          path: wheels
+          retention-days: 1
+
+  build:
+    name: Build ${{ matrix.arch }} supervisor
+    needs: [init, build_wheels]
+    if: ${{ !cancelled() && !failure() }}
```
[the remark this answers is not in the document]
agners: It seems that `always()` might be implied, see this [StackOverflow comment](https://stackoverflow.com/questions/69354003/github-action-job-fire-when-previous-job-skipped#comment138343758_77066140). Let's see.

PR #6379, opened 2025-12-02
Refactor Docker pull progress with registry manifest fetcher

agners:
## Proposed change

Refactor Docker image pull progress tracking for more accurate and reliable progress updates.

**Problem:**
The previous progress tracking had issues where:
- Progress would stay stuck when large layers were discovered late (Docker rate-limits concurrent downloads to ~3 layers, if large layers get discovered late, total would get bumped late in the image pull)
- Progress would jump to 100% prematurely when cached layers were reported before downloads started
- Size-weighted progress was unreliable because layer sizes weren't known until each layer started downloading

**Solution:**
This PR introduces a registry manifest fetcher that retrieves layer sizes directly from container registries before pulling. This enables accurate size-based progress tracking:

1. **Registry Manifest Fetcher** (`supervisor/docker/manifest.py`):
   - Fetches manifests directly from registries (ghcr.io, Docker Hub, private registries)
   - Discovers auth endpoints via WWW-Authenticate headers
   - Handles multi-arch manifest lists by selecting the correct platform
   - Uses credentials from Docker config for private registries

2. **Size-weighted progress**:
   - Each layer contributes to progress proportionally to its byte size
   - Large layers (e.g., 500MB) have more weight than small layers (e.g., 100 bytes)
   - Falls back to count-based progress (equal weight per layer) if manifest fetch fails

3. **Exclude cached layers**:
   - Layers that "Already exist" locally are excluded from progress calculation
   - Progress only tracks bytes that actually need to be downloaded

4. **Wait for downloads to start**:
   - Progress returns 0% until first "Downloading" event
   - This prevents premature progress jumps from cached layer events

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/tests/api/test_supervisor.py
+++ b/tests/api/test_supervisor.py
@@ -448,6 +448,6 @@ async def test_supervisor_api_stats_failure(
     assert body["error_key"] == "supervisor_unknown_error"
     assert body["extra_fields"] == {"logs_command": "ha supervisor logs"}
     assert (
-        "Could not inspect container 'hassio_supervisor': DockerError(500, 'fail')"
+        "Could not inspect container 'hassio_supervisor': [500] {'message': 'fail'}"
         in caplog.text
     )
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -12,6 +12,7 @@ from uuid import uuid4
 from aiodocker.containers import DockerContainer, DockerContainers
 from aiodocker.docker import DockerImages
 from aiodocker.execs import Exec
+from aiodocker.networks import DockerNetwork, DockerNetworks
 from aiohttp import ClientSession, web
 from aiohttp.test_utils import TestClient
 from awesomeversion import AwesomeVersion
@@ -125,26 +126,53 @@ async def docker() -> DockerAPI:
         "State": {"ExitCode": 0, "Status": "stopped", "Running": False},
         "Image": "abc123",
     }
+    network_inspect = {
+        "Name": "hassio",
+        "Id": "hassio123",
+        "EnableIPv4": True,
+        "EnableIPv6": False,
+        "IPAM": {
+            "Driver": "default",
+            "Options": None,
+            "Config": [
+                {
+                    "Subnet": "172.30.32.0/23",
+                    "IPRange": "172.30.33.0/24",
+                    "Gateway": "172.30.32.1",
+                }
+            ],
+        },
+        "Containers": {},
+    }
 
     with (
         patch("supervisor.docker.manager.DockerClient", return_value=MagicMock()),
-        patch("supervisor.docker.manager.DockerAPI.api", return_value=MagicMock()),
         patch("supervisor.docker.manager.DockerAPI.info", return_value=MagicMock()),
         patch("supervisor.docker.manager.DockerAPI.unload"),
-        patch("supervisor.docker.manager.aiodocker.Docker", return_value=MagicMock()),
+        patch(
+            "supervisor.docker.manager.aiodocker.Docker",
+            return_value=(
+                docker_client := MagicMock(
+                    networks=MagicMock(spec=DockerNetworks),
+                    images=(docker_images := MagicMock(spec=DockerImages)),
+                    containers=(docker_containers := MagicMock(spec=DockerContainers)),
+                )
+            ),
+        ),
         patch(
             "supervisor.docker.manager.DockerAPI.images",
-            new=PropertyMock(
-                return_value=(docker_images := MagicMock(spec=DockerImages))
-            ),
+            new=PropertyMock(return_value=docker_images),
         ),
         patch(
             "supervisor.docker.manager.DockerAPI.containers",
-            new=PropertyMock(
-                return_value=(docker_containers := MagicMock(spec=DockerContainers))
-            ),
+            new=PropertyMock(return_value=docker_containers),
         ),
     ):
+        docker_client.networks.get.return_value = docker_network = MagicMock(
+            spec=DockerNetwork
+        )
+        docker_network.show.return_value = network_inspect
+
         docker_obj = await DockerAPI(MagicMock()).post_init()
         docker_obj.config._data = {"registries": {}}
         with patch("supervisor.docker.monitor.DockerMonitor.load"):
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -34,7 +34,7 @@ jobs:
           python-version: ${{ env.DEFAULT_PYTHON }}
       - name: Restore Python virtual environment
         id: cache-venv
-        uses: actions/cache@8b402f58fbc84540c8b491a91e594a4576fec3d7 # v5.0.2
+        uses: actions/cache@cdf6c1fa76f9f475f3d7449005a359c84ca0f306 # v5.0.3
         with:
           path: venv
           key: |
@@ -48,7 +48,7 @@ jobs:
           pip install -r requirements.txt -r requirements_tests.txt
       - name: Restore pre-commit environment from cache
         id: cache-precommit
-        uses: actions/cache@8b402f58fbc84540c8b491a91e594a4576fec3d7 # v5.0.2
+        uses: actions/cache@cdf6c1fa76f9f475f3d7449005a359c84ca0f306 # v5.0.3
         with:
           path: ${{ env.PRE_COMMIT_CACHE }}
           lookup-only: true
@@ -76,7 +76,7 @@ jobs:
           python-version: ${{ needs.prepare.outputs.python-version }}
       - name: Restore Python virtual environment
         id: cache-venv
-        uses: actions/cache@8b402f58fbc84540c8b491a91e594a4576fec3d7 # v5.0.2
+        uses: actions/cache@cdf6c1fa76f9f475f3d7449005a359c84ca0f306 # v5.0.3
         with:
           path: venv
           key: |
@@ -88,7 +88,7 @@ jobs:
           exit 1
       - name: Restore pre-commit environment from cache
         id: cache-precommit
-        uses: actions/cache@8b402f58fbc84540c8b491a91e594a4576fec3d7 # v5.0.2
+        uses: actions/cache@cdf6c1fa76f9f475f3d7449005a359c84ca0f306 # v5.0.3
         with:
           path: ${{ env.PRE_COMMIT_CACHE }}
           key: |
@@ -119,7 +119,7 @@ jobs:
           python-version: ${{ needs.prepare.outputs.python-version }}
       - name: Restore Python virtual environment
         id: cache-venv
-        uses: actions/cache@8b402f58fbc84540c8b491a91e594a4576fec3d7 # v5.0.2
+        uses: actions/cache@cdf6c1fa76f9f475f3d7449005a359c84ca0f306 # v5.0.3
         with:
           path: venv
           key: |
@@ -131,7 +131,7 @@ jobs:
           exit 1
       - name: Restore pre-commit environment from cache
         id: cache-precommit
-        uses: actions/cache@8b402f58fbc84540c8b491a91e594a4576fec3d7 # v5.0.2
+        uses: actions/cache@cdf6c1fa76f9f475f3d7449005a359c84ca0f306 # v5.0.3
         with:
           path: ${{ env.PRE_COMMIT_CACHE }}
           key: |
@@ -177,7 +177,7 @@ jobs:
           python-version: ${{ needs.prepare.outputs.python-version }}
       - name: Restore Python virtual environment
         id: cache-venv
-        uses: actions/cache@8b402f58fbc84540c8b491a91e594a4576fec3d7 # v5.0.2
+        uses: actions/cache@cdf6c1fa76f9f475f3d7449005a359c84ca0f306 # v5.0.3
         with:
           path: venv
           key: |
@@ -189,7 +189,7 @@ jobs:
           exit 1
       - name: Restore pre-commit environment from cache
         id: cache-precommit
-        uses: actions/cache@8b402f58fbc84540c8b491a91e594a4576fec3d7 # v5.0.2
+        uses: actions/cache@cdf6c1fa76f9f475f3d7449005a359c84ca0f306 # v5.0.3
         with:
           path: ${{ env.PRE_COMMIT_CACHE }}
           key: |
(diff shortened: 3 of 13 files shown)


supervisor/docker/manifest.py
```diff
@@ -0,0 +1,354 @@
+"""Docker registry manifest fetcher.
+
+Fetches image manifests directly from container registries to get layer sizes
+before pulling an image. This enables accurate size-based progress tracking.
```
agners, line 4: Rather than implementing registry manifest fetcher as part of Supervisor we should either call the Docker CLI (we'd need to create a config to get authentication right, but we have the [necessary code already](https://github.com/home-assistant/supervisor/blob/50d31202aef6de2ee4a31902a210ad22dcd8c27c/supervisor/addons/build.py#L131)) or move the logic into aiodocker (I would prefer this approach).

supervisor/docker/interface.py
```diff
@@ -374,33 +214,55 @@ async def install(
...
+            )
+            if manifest:
+                pull_progress.set_manifest(manifest)
+                _LOGGER.debug(
+                    "Using manifest for progress: %d layers, %d bytes",
+                    manifest.layer_count,
+                    manifest.total_size,
+                )
+        except (aiohttp.ClientError, TimeoutError) as err:
+            _LOGGER.debug("Could not fetch manifest for progress: %s", err)
```
mdegat01: I would log this at at least warn level. I get that we proceed anyway but this will visibly mess up progress so users will probably look for what went wrong.

supervisor/docker/manifest.py
```diff
@@ -0,0 +1,352 @@
...
+        """Get or create aiohttp session."""
+        if self._session is None or self._session.closed:
+            self._session = aiohttp.ClientSession()
+        return self._session
+
+    async def close(self) -> None:
+        """Close the session."""
+        if self._session and not self._session.closed:
+            await self._session.close()
+            self._session = None
```
mdegat01:
```suggestion
    def __init__(self, coresys: CoreSys, session: aiohttp.ClientSession) -> None:
        """Initialize the fetcher."""
        self.coresys = coresys
        self._session: aiohttp.ClientSession = session
```
We should pass in an existing session here for performance unless there some reason we really need a new one. The one we use for almost everything else is available from within `DockerManager` at `self.sys_websession` since it inherits from `CoreSysAttributes`.

Alternatively we could reuse the one we make for `aiodocker` in the constructor but that one has a modified default timeout. Which could be ok here I suppose? Either way lets not make another.

supervisor/docker/manifest.py
```diff
@@ -0,0 +1,352 @@
...
+                if resp.status != 401:
+                    _LOGGER.warning(
+                        "Unexpected status %d from registry %s", resp.status, registry
+                    )
+                    return None
+
+                www_auth = resp.headers.get("WWW-Authenticate", "")
+        except aiohttp.ClientError as err:
+            _LOGGER.warning("Failed to connect to registry %s: %s", registry, err)
+            return None
```
mdegat01, line 133: There's no way around trying and failing to figure out if we need auth and how to provide it? We could just call `self._get_credentials(registry)` up front but it seems like we can't figure out the `realm` and `service` without making the call and failing which is unfortunate for performance...

supervisor/docker/manifest.py
```diff
@@ -0,0 +1,352 @@
...
+        # Check if this is a manifest list (multi-arch image)
+        if "list" in media_type or "index" in media_type:
+            manifests = manifest.get("manifests", [])
+            if not manifests:
+                _LOGGER.warning("Empty manifest list for %s/%s", registry, repository)
+                return None
+
+            # Find matching platform
+            target_os = "linux"
+            target_arch = "amd64"  # Default
```
mdegat01: This feels odd. If we list a default here it seems like it should be based on the system information. But in practice I don't think this will ever actually be used because `platform` is never `None` in here:
https://github.com/home-assistant/supervisor/blob/620c90676c8f80cf958a0d0bcd537d93a23def2e/supervisor/docker/interface.py#L217

Maybe just make `platform` a required parameter to this method instead to avoid this oddity?

supervisor/docker/manifest.py
```diff
@@ -0,0 +1,352 @@
...
+                    break
+
+            if not platform_manifest:
+                # Fall back to first manifest
+                _LOGGER.debug(
+                    "Platform %s/%s not found, using first manifest",
+                    target_os,
+                    target_arch,
+                )
+                platform_manifest = manifests[0]
```
mdegat01: Shouldn't this just raise at this point? Or is this likely to work for some reason? Seems like if the arch we need is missing its just not going to work...

supervisor/docker/pull_progress.py
```diff
@@ -0,0 +1,371 @@
...
+            self._seen_downloading = True
+            if (
+                entry.progress_detail
+                and entry.progress_detail.current is not None
+                and entry.progress_detail.total is not None
+            ):
+                layer.download_current = entry.progress_detail.current
+                # Only set total_size if not already set or if this is larger
+                # (handles case where total changes during download)
+                layer.total_size = max(layer.total_size, entry.progress_detail.total)
```
mdegat01:
```suggestion
            if entry.progress_detail and entry.progress_detail.current is not None:
                layer.download_current = entry.progress_detail.current
            if entry.progress_detail and entry.progress_detail.total is not None:
                # Only set total_size if not already set or if this is larger
                # (handles case where total changes during download)
                layer.total_size = max(layer.total_size, entry.progress_detail.total)
```
Maybe separate these two sets? I expect any changing of total in here now is generally going to be an edge case. Given docker has already changed this on us once lets make sure we aren't depending on total in here to set current as that's the more important part now.

supervisor/docker/pull_progress.py
```diff
@@ -0,0 +1,371 @@
...
+            return
+
+        # Handle "Extracting" - extraction in progress
+        if status is LayerPullStatus.EXTRACTING:
+            # For overlay2: progressDetail has {current, total} in bytes
+            # For containerd: progressDetail has {current, units: "s"} (time elapsed)
+            # We can only use byte-based progress (overlay2)
+            layer.download_complete = True
+            if layer.total_size > 0:
+                layer.download_current = layer.total_size
```
mdegat01, line 218:
```suggestion
        # Handle "Download complete" - download phase done
        if not layer.download_complete and status >= LayerPullStatus.DOWNLOAD_COMPLETE:
            layer.download_complete = True
            if layer.total_size > 0:
                layer.download_current = layer.total_size
            elif layer.total_size == 0:
                # Small layer that skipped downloading phase
                # Set minimal size so it doesn't distort weighted average
                layer.total_size = 1
                layer.download_current = 1
            return

        # Handle "Extracting" - extraction in progress
        if status is LayerPullStatus.EXTRACTING:
            # For overlay2: progressDetail has {current, total} in bytes
            # For containerd: progressDetail has {current, units: "s"} (time elapsed)
            # We can only use byte-based progress (overlay2)
```
I assume we can still use the comparison operators here right? Should prevent us from having to duplicate code into later statuses like this.

supervisor/docker/pull_progress.py
```diff
@@ -0,0 +1,371 @@
...
+                    else 0,
+                )
+            return
+
+        # Handle "Pull complete" - layer is fully done
+        if status is LayerPullStatus.PULL_COMPLETE:
+            layer.download_complete = True
+            layer.extract_complete = True
+            if layer.total_size > 0:
+                layer.download_current = layer.total_size
```
mdegat01, line 246: Comparison operators should prevent the need to keep duplicating this as mentioned above.

supervisor/docker/pull_progress.py
```diff
@@ -0,0 +1,371 @@
...
+            layer for layer in self.layers.values() if not layer.already_exists
+        ]
+
+        if not layers_to_pull:
+            # All layers already exist, nothing to download
+            return 100.0
+
+        # Use size-weighted progress if manifest sizes are available
+        if self._manifest_layer_sizes:
+            return min(100, self._calculate_size_weighted_progress(layers_to_pull))
```
mdegat01, line 286: Should we put `max(0, ` in here too? Or is there no chance float drift will cause us to cross negative? Like if we got a very small progress report on a huge image.

tests/api/test_supervisor.py
```diff
@@ -359,6 +359,8 @@ async def test_api_progress_updates_supervisor_update(
         and evt.args[0]["data"]["event"] == WSEvent.JOB
         and evt.args[0]["data"]["data"]["name"] == "supervisor_update"
     ]
+    # Count-based progress: 2 layers need pulling (each worth 50%)
```
mdegat01, line 363: None of this is count-based progress anymore right?

tests/conftest.py
```diff
@@ -154,6 +154,9 @@ async def docker() -> DockerAPI:
         docker_obj.info.storage = "overlay2"
         docker_obj.info.version = AwesomeVersion("1.0.0")
 
+        # Mock manifest fetcher to return None (falls back to count-based progress)
+        docker_obj._manifest_fetcher.get_manifest = AsyncMock(return_value=None)
+
```
mdegat01, line 207: This doesn't seem accurate? Removing the manifest doesn't force us to count-based progress, it just means the progress reporter creates and tracks layers as it sees them. More things would have to go wrong to force us to used count-based progress I think?

tests/docker/test_manifest.py
```diff
@@ -0,0 +1,166 @@
...
+
+from supervisor.docker.manifest import (
+    DOCKER_HUB,
+    ImageManifest,
+    RegistryManifestFetcher,
+    parse_image_reference,
+)
+
+
+class TestParseImageReference:
```
mdegat01: This style of organizing tests is legacy I believe? Any particular reason we're using Test classes here when everything else uses individual test methods?

tests/api/test_homeassistant.py
```diff
@@ -305,6 +305,8 @@ async def test_api_progress_updates_home_assistant_update(
         and evt.args[0]["data"]["event"] == WSEvent.JOB
         and evt.args[0]["data"]["data"]["name"] == "home_assistant_core_update"
```
mdegat01, line 307: I see we added a lot of unit tests for the new stuff but wouldn't it make sense to have at least one of these higher level tests pulling and using a manifest to help more accurately report progress? Like how most of our existing tests would stumble initially because they make layers on demand as they see them and progress may go backwards but with a manifest it would be accurate from the beginning.

mdegat01 (changes requested):
Few things to note in here. The session reuse one in particular seems like something we need to fix as we go to a lot of effort to avoid making new sessions elsewhere.

supervisor/docker/interface.py
```diff
@@ -374,33 +214,55 @@ async def install(
...
+            pull_progress.process_event(event)
 
-            listener = self.sys_bus.register_event(
-                BusEvent.DOCKER_IMAGE_PULL_UPDATE, process_pull_image_log
-            )
+            # Update job if progress changed significantly (>= 1%)
+            should_update, progress = pull_progress.should_update_job()
+            if should_update:
+                stage = pull_progress.get_stage()
+                current_job.update(progress=progress, stage=stage)
```
mdegat01, line 250: Also it might be good to have some info level logging around here for #6390 . Removing the timeout fixes part of the issue but they're right that no logs for 15 minutes will still look like a crash. Some logs here would add visibility to what's going on until onboarding can add a progress bar of some kind.

supervisor/docker/manifest.py
agners: Agreed, we should reuse the existing session.

> Alternatively we could reuse the one we make for `aiodocker` in the constructor but that one has a modified default timeout. Which could be ok here I suppose? Either way lets not make another.

The `aiodocker` session really serves for a specific purpose, using a local Unix socket (which is why we use the modified default timeout). I rather prefer to use the global `self.sys_websession`, as its geared for "Internet usage".

supervisor/docker/manifest.py
agners: Yeah good point, ideally add a warning message just like as if the manifest fetching fails, and go down the "no-manifest" route.

supervisor/docker/interface.py
agners: Yeah agreed. But I rather have that in a follow up PR, then we can also discuss the exact rate of logging etc.

supervisor/docker/pull_progress.py
agners: We only adding positive values (or dividing by positive values), which can only ever lead to positive floats. This seems impossible to me.

supervisor/docker/manifest.py
agners: Yeah... But given that auth is really not all that common and more of a corner case I don't think it is a big deal :shrug: .

tests/conftest.py
agners: No this is still accurate.

This is the whole reason I've moved to the manifest approach: If a couple small layers are downloaded first, and we calculate the overall progress based on that, we quickly arrive at a high progress procentage (like 60% or so), then if a huge layer follows, and we "rescale" progress, the progress stays stuck potentially for minutes! This is very confusing for the user. More confusing than a progress bar which moves a bit faster at times, and then slower at other times. But if we have 10 layers, it will move the progress bar during a large download still.

So, either we can get the layer sizes ahead of time (99.9% of the cases with the manifest fetcher), and have a progress bar which tracks overall image download progress by known overall size ahead of time, or we fallback with the layer count method.

tests/api/test_homeassistant.py
agners: I've checked my log and Docker events captures locally, I do have a couple of events captures, but I think I only have a log captured. Another problem is that maybe on that (virtual) machine the 70/30 split for downloading/extracting is probably off:
[layer-progress-non-linear-containerd-snapshotter.txt](https://github.com/user-attachments/files/23964590/layer-progress-non-linear-containerd-snapshotter.txt)

But anyhow, an end-to-end test would be nice indeed. I mostly encountered this with Home Assistant Core bumps. But it depends a bit if base layers got updated or not. So 2025.11.3->2025.12.0 might behave quite a bit different from 2025.11.2->2025.11.3.

supervisor/docker/manifest.py
mdegat01: Fair enough. Yea I mean reading how the code works I don't see an obvious way around it if the realm and service can change. So this is what we have to do I suppose.

tests/conftest.py
mdegat01: Ah right, since its designed to generally know all sizes up front we start calculating as soon as we see some progress, not when we see downloading for all layers. So without a manifest its just going to shoot right up. Ok that's fine then.

Issue #6395, opened 2025-12-04
Update core 180s start timeout is too short for Synology based VMs


jprates:
### Describe the issue you are experiencing

Please see https://github.com/home-assistant/core/issues/157987#issue-3695477370

Update to v2025.12.0 always rolls back due to timeout.

### What type of installation are you running?

Home Assistant OS

### Which operating system are you running on?

Home Assistant Operating System

### Steps to reproduce the issue

1. Update 2025.11.3 to 2025.12.0 on Synology virtual machine running HA OS.

### Anything in the Supervisor logs that might be useful for us?

```txt
2025-12-04 17:45:05.796 INFO (MainThread) [supervisor.homeassistant.core] Updating Home Assistant to version 2025.12.0
2025-12-04 17:45:05.798 INFO (MainThread) [supervisor.docker.interface] Updating image ghcr.io/home-assistant/qemux86-64-homeassistant:2025.11.3 to ghcr.io/home-assistant/qemux86-64-homeassistant:2025.12.0
2025-12-04 17:45:05.799 INFO (MainThread) [supervisor.docker.interface] Downloading docker image ghcr.io/home-assistant/qemux86-64-homeassistant with tag 2025.12.0.
2025-12-04 17:52:07.698 INFO (SyncWorker_3) [supervisor.docker.manager] Stopping homeassistant application
2025-12-04 17:53:03.972 INFO (SyncWorker_3) [supervisor.docker.manager] Cleaning homeassistant application
2025-12-04 17:53:07.187 INFO (MainThread) [supervisor.homeassistant.module] Update pulse/client.config: /data/tmp/homeassistant_pulse
2025-12-04 17:53:14.231 INFO (MainThread) [supervisor.docker.homeassistant] Starting Home Assistant ghcr.io/home-assistant/qemux86-64-homeassistant with version 2025.12.0
2025-12-04 17:53:14.232 INFO (MainThread) [supervisor.homeassistant.core] Wait until Home Assistant is ready
2025-12-04 17:56:14.886 ERROR (MainThread) [supervisor.homeassistant.core] No Home Assistant Core response, assuming a fatal startup error
2025-12-04 17:56:14.887 CRITICAL (MainThread) [supervisor.homeassistant.core] HomeAssistant update failed -> rollback!
2025-12-04 17:56:14.913 INFO (MainThread) [supervisor.resolution.module] Create new issue update_rollback - core / None
[...]

PR #6404, opened 2025-12-06
Increase timeout waiting for Core API to work around 2025.12.x issues

hendrikb:
This change provides a fix for a problem that various people have reported:

- https://github.com/home-assistant/supervisor/issues/6395
- https://github.com/home-assistant/core/issues/157987
- https://community.home-assistant.io/t/homeassistant-update-failed-rollback-2025-12-0/958651/44
- and even some other smart home community forums somewhere on the internet ([example, in German](https://community.simon42.com/t/probleme-mit-home-assistant-update-ab-version-2025-12-aktuell-2025-12-0-video-von-simon42/71137/72))...

A too low timeout makes the supervisor think that an upgrade to HA Core 2025.12.x that actually goes as planned but hasn't finished in time and is therefor rolled back. Increasing that timeout fixes the problem.

## Proposed change

With Home Assistant Core 2025.12.x updates available the STARTUP_API_RESPONSE_TIMEOUT that HA supervisor is willing to wait (before assuming a startup failure and rolling back the entire core update) seems to be too low on not-so-beefy hosts. The problem has been seen on many Synology NAS machines running Home Assistant on the side (like in my case). I have doubled the timeout from 3 to 6 minutes and the upgrade to Core 2025.12.1 then works on my Synology DS723+. My update took 4min 56s -- hence the timeout increase was proven necessary.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR fixes or closes issue: fixes https://github.com/home-assistant/supervisor/issues/6395
- This PR is related to issue: https://github.com/home-assistant/core/issues/157987

## Checklist

- [ ] ~~Tests have been added to verify that the new code works.~~ (there is no _new_ code)

Note: ~~I have not touched any tests/specs since the change is so trivial and I have not set up a home assistant development environment for it. Feel free to help out here if strictly necessary.~~
EDIT: I have fixed a failed test that was caused by my proposed code change. See details here: https://github.com/home-assistant/supervisor/pull/6404#issuecomment-3621871667


agners:
IMHO, it would be good if we'd get to why Core suddenly takes that much longer to answer the first API requests, as mentioned in https://github.com/home-assistant/core/issues/157987#issuecomment-3617100272.

But I am fine merging this today so affected systems can upgrade to 2025.12.0.

hendrikb:
_Of course_ I fell for the timeout being asserted in a test, which resulted in a failed test in `tests/homeassistant/test_core.py::test_api_check_timeout` -- I have adjusted that test. The entire test suite runs fine now, locally. Sorry for being "so quick". :woozy_face: 

> IMHO, it would be good if we'd get to why Core suddenly takes that much longer to answer the first API requests (...)

I absolutely agree with you on that. IMHO there is _something_ going on with Synology + QEMU. Also my personal perception tells me that booting up the system takes very long, and apparently with Core 2025.12.x we must have made a big leap in initial loading time from "somewhere less than 3 minutes" to "almost 5 minutes" (like in my case) -- but investigating this should not happen in this issue.

As mentioned above, the goal for this change is to provide a _reasonable increase_ for that timeout so that people can update to Core 2025.12.x without needing to wait for a longer investigation on what causes these longer loading times.

supervisor/homeassistant/core.py
```diff
@@ -48,7 +48,7 @@
 
 SECONDS_BETWEEN_API_CHECKS: Final[int] = 5
 # Core Stage 1 and some wiggle room
-STARTUP_API_RESPONSE_TIMEOUT: Final[timedelta] = timedelta(minutes=3)
+STARTUP_API_RESPONSE_TIMEOUT: Final[timedelta] = timedelta(minutes=6)
```
sairon: Let's go with a slightly larger timeout:
```suggestion
STARTUP_API_RESPONSE_TIMEOUT: Final[timedelta] = timedelta(minutes=10)
```

tests/homeassistant/test_core.py
```diff
@@ -387,7 +387,7 @@ async def mock_sleep(*args):
         ):
             await coresys.homeassistant.core.start()
 
-    assert coresys.homeassistant.api.get_api_state.call_count == 3
+    assert coresys.homeassistant.api.get_api_state.call_count == 6
```
sairon:
```suggestion
    assert coresys.homeassistant.api.get_api_state.call_count == 10
```

VladimirPchelko:
Maybe I'm misunderstanding something - but the current executable code does not contain this fix, and the problem is updating to new versions of the same component including this same fix....???

and that doesn't even fix the problem! this is workaround....

PR #5605, opened 2025-02-05
Drop Docker config from Supervisor backup

agners:
## Proposed change
The Docker config is part of the main backup metadata. It allows to setup repository specific authentication information (e.g. to use a login for Docker Hub credentials for  `hub.docker.com` can be entered). The password for these login information gets encrypted. Because we consolidate encrypted and unencrypted backups today on Supervisor side, we end up with only one of the two metadata in the cache.

We could implement a fix by storing this particular metadata per location. However, this adds complexity and code for a feature which is not much used.

Instead, we should move to a Supervisor specific inner tar (`supervisor.tar[.gz]`). 

## Type of change

- [ ] Dependency upgrade
- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (which adds functionality to the supervisor)
- [x] Breaking change (fix/feature causing existing functionality to break)
- [x] Code quality improvements to existing code or addition of tests

## Additional information

- This PR fixes or closes issue: fixes #
- This PR is related to issue:
- Link to documentation pull request:
- Link to cli pull request:
- Link to client library pull request:

## Checklist

- [ ] The code change is tested and works locally.
- [ ] Local tests pass. **Your PR cannot be merged unless tests pass**
- [ ] There is no commented out code in this PR.
- [ ] I have followed the [development checklist][dev-checklist]
- [ ] The code has been formatted using Ruff (`ruff format supervisor tests`)
- [ ] Tests have been added to verify that the new code works.

If API endpoints or add-on configuration are added/changed:

- [ ] Documentation added/updated for [developers.home-assistant.io][docs-repository]
- [ ] [CLI][cli-repository] updated (if necessary)
- [ ] [Client library][client-library-repository] updated (if necessary)

[dev-checklist]: https://developers.home-assistant.io/docs/en/development_checklist.html
[docs-repository]: https://github.com/home-assistant/developers.home-assistant
[cli-repository]: https://github.com/home-assistant/cli
[client-library-repository]: https://github.com/home-assistant-libs/python-supervisor-client/

## Summary by CodeRabbit

- **Refactor**
  - The backup and restore processes have been streamlined to focus on essential system data. Management of Docker registry credentials during these operations has been removed, simplifying the overall workflow. This update ensures a more straightforward and reliable backup experience, with a focus on maintaining only the most critical data.

PR #6411, opened 2025-12-08
Include network storage mount configurations in backups

ptarjan:
## Proposed change

When creating backups, now stores mount configurations (CIFS/NFS shares) including server, share, credentials, and other settings. When restoring from a backup, mount configurations are automatically restored.

This fixes the issue where network storage definitions for backups were lost after restoring from a backup, requiring users to manually reconfigure their network storage mounts.

Changes:
- Add ATTR_MOUNTS schema to backup validation
- Add store_mounts() method to save mount configs during backup
- Add restore_mounts() method to restore mount configs during restore
- Add MOUNTS stage to backup/restore job stages
- Update BackupManager to call mount backup/restore methods
- Add tests for mount backup/restore functionality

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR fixes or closes issue: Fixes home-assistant/core#148663
- Link to cli pull request:

## Checklist

### DIFF
--- a/.github/workflows/builder.yml
+++ b/.github/workflows/builder.yml
@@ -38,6 +38,7 @@ env:
   CRANE_SHA256: "8ef3564d264e6b5ca93f7b7f5652704c4dd29d33935aff6947dd5adefd05953e"
   BUILD_NAME: supervisor
   BUILD_TYPE: supervisor
+  ARCHITECTURES: '["amd64", "aarch64"]'
 
 concurrency:
   group: "${{ github.workflow }}-${{ github.ref }}"
@@ -48,7 +49,7 @@ jobs:
     name: Initialize build
     runs-on: ubuntu-latest
     outputs:
-      architectures: ${{ steps.info.outputs.architectures }}
+      architectures: ${{ env.ARCHITECTURES }}
       version: ${{ steps.version.outputs.version }}
       channel: ${{ steps.version.outputs.channel }}
       publish: ${{ steps.version.outputs.publish }}
@@ -58,10 +59,6 @@ jobs:
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           fetch-depth: 0
-
-      - name: Get information
-        id: info
-        uses: home-assistant/actions/helpers/info@master
 
       - name: Get version
         id: version
@@ -174,7 +171,7 @@ jobs:
 
       - name: Install Cosign
         if: needs.init.outputs.publish == 'true'
-        uses: sigstore/cosign-installer@ba7bc0a3fef59531c69a25acd34668d6d3fe6f22 # v4.1.0
+        uses: sigstore/cosign-installer@cad07c2e89fa2edd6e2d7bab4c1aa38e53f76003 # v4.1.1
         with:
           cosign-release: ${{ env.COSIGN_VERSION }}
 
@@ -467,7 +464,7 @@ jobs:
           password: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Install Cosign
-        uses: sigstore/cosign-installer@ba7bc0a3fef59531c69a25acd34668d6d3fe6f22 # v4.1.0
+        uses: sigstore/cosign-installer@cad07c2e89fa2edd6e2d7bab4c1aa38e53f76003 # v4.1.1
         with:
           cosign-release: ${{ env.COSIGN_VERSION }}
 
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -346,7 +346,7 @@ jobs:
         with:
           python-version: ${{ needs.prepare.outputs.python-version }}
       - name: Install Cosign
-        uses: sigstore/cosign-installer@ba7bc0a3fef59531c69a25acd34668d6d3fe6f22 # v4.1.0
+        uses: sigstore/cosign-installer@cad07c2e89fa2edd6e2d7bab4c1aa38e53f76003 # v4.1.1
         with:
           cosign-release: "v2.5.3"
       - name: Restore Python virtual environment
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,7 +9,7 @@ brotli==1.2.0
 ciso8601==2.3.3
 colorlog==6.10.1
 cpe==1.3.1
-cryptography==46.0.5
+cryptography==46.0.6
 debugpy==1.8.20
 deepmerge==2.0
 dirhash==0.5.0
@@ -21,7 +21,6 @@ orjson==3.11.7
 pulsectl==24.12.0
 pyudev==0.24.4
 PyYAML==6.0.3
-requests==2.32.5
 securetar==2026.2.0
 sentry-sdk==2.56.0
 setuptools==82.0.1
--- a/requirements_tests.txt
+++ b/requirements_tests.txt
@@ -11,5 +11,4 @@ pytest==9.0.2
 ruff==0.15.7
 time-machine==3.2.0
 types-pyyaml==6.0.12.20250915
-types-requests==2.32.4.20260324
 urllib3==2.6.3
--- a/supervisor/docker/interface.py
+++ b/supervisor/docker/interface.py
@@ -16,7 +16,6 @@ import aiodocker
 import aiohttp
 from awesomeversion import AwesomeVersion
 from awesomeversion.strategy import AwesomeVersionStrategy
-import requests
 
 from ..const import (
     ATTR_PASSWORD,
@@ -34,7 +33,6 @@ from ..exceptions import (
     DockerHubRateLimitExceeded,
     DockerJobError,
     DockerNotFound,
-    DockerRequestError,
 )
 from ..jobs.const import JOB_GROUP_DOCKER_INTERFACE, JobConcurrency
 from ..jobs.decorator import Job
@@ -328,7 +326,7 @@ class DockerInterface(JobGroup, ABC):
 
     async def exists(self) -> bool:
         """Return True if Docker image exists in local repository."""
-        with suppress(aiodocker.DockerError, requests.RequestException):
+        with suppress(aiodocker.DockerError):
             await self.sys_docker.images.inspect(f"{self.image}:{self.version!s}")
             return True
         return False
@@ -343,10 +341,6 @@ class DockerInterface(JobGroup, ABC):
                 return None
             raise DockerAPIError(
                 f"Docker API error occurred while getting container information: {err!s}"
-            ) from err
-        except requests.RequestException as err:
-            raise DockerRequestError(
-                f"Error communicating with Docker to get container information: {err!s}"
             ) from err
 
     async def is_running(self) -> bool:
@@ -368,7 +362,7 @@ class DockerInterface(JobGroup, ABC):
         self, version: AwesomeVersion, *, skip_state_event_if_down: bool = False
     ) -> None:
         """Attach to running Docker container."""
-        with suppress(aiodocker.DockerError, requests.RequestException):
+        with suppress(aiodocker.DockerError):
             docker_container = await self.sys_docker.containers.get(self.name)
             self._meta = await docker_container.show()
             self.sys_docker.monitor.watch_container(self._meta)
@@ -386,7 +380,7 @@ class DockerInterface(JobGroup, ABC):
                     ),
                 )
 
-        with suppress(aiodocker.DockerError, requests.RequestException):
+        with suppress(aiodocker.DockerError):
             if not self._meta and self.image:
                 self._meta = await self.sys_docker.images.inspect(
                     f"{self.image}:{version!s}"
@@ -489,7 +483,7 @@ class DockerInterface(JobGroup, ABC):
         if self.image == expected_image:
             try:
                 image = await self.sys_docker.images.inspect(image_name)
-            except (aiodocker.DockerError, requests.RequestException) as err:
+            except aiodocker.DockerError as err:
                 raise DockerError(
                     f"Could not get {image_name} for check due to: {err!s}",
                     _LOGGER.error,
@@ -611,10 +605,6 @@ class DockerInterface(JobGroup, ABC):
         except (aiodocker.DockerError, ValueError) as err:
             raise DockerNotFound(
                 f"No version found for {self.image}", _LOGGER.info
-            ) from err
-        except requests.RequestException as err:
-            raise DockerRequestError(
-                f"Communication issues with dockerd on Host: {err}", _LOGGER.warning
             ) from err
 
         _LOGGER.info("Found %s versions: %s", self.image, available_version)
--- a/supervisor/docker/manager.py
+++ b/supervisor/docker/manager.py
@@ -23,7 +23,6 @@ from aiodocker.stream import Stream
 from aiodocker.types import JSONObject
 from aiohttp import ClientTimeout, UnixConnector
 from awesomeversion import AwesomeVersion, AwesomeVersionCompareException
-import requests
 
 from ..const import (
     ATTR_ENABLE_IPV6,
@@ -44,7 +43,6 @@ from ..exceptions import (
     DockerError,
     DockerNoSpaceOnDevice,
     DockerNotFound,
-    DockerRequestError,
 )
 from ..utils.common import FileConfiguration
 from ..validate import SCHEMA_DOCKER_CONFIG
@@ -589,12 +587,6 @@ class DockerAPI(CoreSysAttributes):
             raise DockerAPIError(
                 f"Can't start {name or container.id}: {err}", _LOGGER.error
             ) from err
-        except requests.RequestException as err:
-            raise DockerRequestError(
-                f"Dockerd connection issue for {name or container.id}: {err}",
-                _LOGGER.error,
-            ) from err
-
         return container
 
     async def run(
@@ -609,10 +601,6 @@ class DockerAPI(CoreSysAttributes):
         except aiodocker.DockerError as err:
             raise DockerAPIError(
                 f"Can't inspect started container {name}: {err}", _LOGGER.error
-            ) from err
-        except requests.RequestException as err:
-            raise DockerRequestError(
-                f"Dockerd connection issue for {name}: {err}", _LOGGER.error
             ) from err
 
         return container_attrs
@@ -992,7 +980,7 @@ class DockerAPI(CoreSysAttributes):
                 if err.status != HTTPStatus.NOT_FOUND:
                     raise
 
-        except (aiodocker.DockerError, requests.RequestException) as err:
+        except aiodocker.DockerError as err:
             raise DockerError(
                 f"Can't remove image {image}: {err}", _LOGGER.warning
             ) from err
@@ -1041,7 +1029,7 @@ class DockerAPI(CoreSysAttributes):
 
         try:
             return await self.images.inspect(docker_image_list[0])
-        except (aiodocker.DockerError, requests.RequestException) as err:
+        except aiodocker.DockerError as err:
             raise DockerError(
                 f"Could not inspect imported image due to: {err!s}", _LOGGER.error
             ) from err
@@ -1095,7 +1083,7 @@ class DockerAPI(CoreSysAttributes):
                         f"{current_image} not found for cleanup", _LOGGER.warning
                     ) from None
                 raise
-        except (aiodocker.DockerError, requests.RequestException) as err:
+        except aiodocker.DockerError as err:
             raise DockerError(
                 f"Can't get {current_image} for cleanup", _LOGGER.warning
             ) from err
@@ -1129,7 +1117,7 @@ class DockerAPI(CoreSysAttributes):
             images_list = await self.images.list(
                 filters=json.dumps({"reference": image_names})
             )
-        except (aiodocker.DockerError, requests.RequestException) as err:
+        except aiodocker.DockerError as err:
             raise DockerError(
                 f"Corrupt docker overlayfs found: {err}", _LOGGER.warning
             ) from err
@@ -1138,6 +1126,6 @@ class DockerAPI(CoreSysAttributes):
             if docker_image["Id"] in keep:
                 continue
 
-            with suppress(aiodocker.DockerError, requests.RequestException):
+            with suppress(aiodocker.DockerError):
                 _LOGGER.info("Cleanup images: %s", docker_image["RepoTags"])
                 await self.images.delete(docker_image["Id"], force=True)
--- a/supervisor/exceptions.py
+++ b/supervisor/exceptions.py
@@ -859,10 +859,6 @@ class DockerAPIError(DockerError):
     """Docker API error."""
 
 
-class DockerRequestError(DockerError):
-    """Dockerd OS issues."""
-
-
 class DockerTrustError(DockerError):
     """Raise if images are not trusted."""
 


supervisor/backups/backup.py
```diff
@@ -920,3 +927,69 @@ def restore_repositories(self, replace: bool = False) -> Awaitable[None]:
...
+            ATTR_DEFAULT_BACKUP_MOUNT: (
+                self.sys_mounts.default_backup_mount.name
+                if self.sys_mounts.default_backup_mount
+                else None
+            ),
+            ATTR_MOUNTS: [
+                mount.to_dict(skip_secrets=False) for mount in self.sys_mounts.mounts
+            ],
+        }
+        self._data[ATTR_MOUNTS] = mounts_data
```
mdegat01: We can't put this information here. The backup's config file is actually always stored unencrypted even if you make an encrypted backup with a password. It's the only file that is special like this. Everything else in the backup is encrypted if you set a password but not this one file. That's how all the UIs are able to work and show you metadata about them even if the backups are encrypted. Since this can contain sensitive password information, we shouldn't store it in here unencrypted.

Instead we should just add a `mounts.tar` to the backup using securetar and add the mounts config as a JSON file to it. That way it is stored properly encrypted. The config can be the same as what you're already doing and to restore you just extract and then use the JSON to restore all the mount info.

supervisor/backups/backup.py
```diff
@@ -920,3 +927,69 @@ def restore_repositories(self, replace: bool = False) -> Awaitable[None]:
...
+                # Skip if mount already exists
+                if mount_name in self.sys_mounts:
+                    _LOGGER.info(
+                        "Mount %s already exists, skipping restore", mount_name
+                    )
+                    continue
+
+                # Create mount from backup data
+                mount = Mount.from_dict(self.coresys, mount_data)
+                await self.sys_mounts.create_mount(mount)
```
mdegat01: We shouldn't use this method to do the restore of mounts, at least not as is. This method will only succeed if the network device is available right at that moment. But that's not really what we want here. These mounts were already validated, we know they work even if they don't work at that moment.

So we should always restore the config here and then as a separate task we try to bring that mount online. If that second part fails we report that to the user as an error but it doesn't affect the success state of the backup or our ability to restore that mount and others.

Take a look at addons, I might try to do something similar here. Where we complete the restore and create and return an `asyncio.Task` for starting the mount. Collect all these tasks and then after everything else in the backup has completed await them and log any errors encountered. I think a new `restore_mount` method would probably be the cleanest approach here for this.

supervisor/backups/backup.py
```diff
@@ -920,3 +927,69 @@ def restore_repositories(self, replace: bool = False) -> Awaitable[None]:
...
+                success = False
+
+        # Restore default backup mount if specified and mount exists
+        default_mount_name = self.mounts.get(ATTR_DEFAULT_BACKUP_MOUNT)
+        if default_mount_name and default_mount_name in self.sys_mounts:
+            try:
+                self.sys_mounts.default_backup_mount = self.sys_mounts.get(
+                    default_mount_name
+                )
+                _LOGGER.info("Restored default backup mount: %s", default_mount_name)
```
mdegat01: I feel like there's two use cases here:
1. If the default mount on this system is `None` this is fine, overwrite it with whatever was in the backup
2. If the default mount on this system is not `None` this feels possibly incorrect, I'm not sure overwriting it without checking is welcome...

Maybe a new restore option is warranted here, a boolean for whether to change the default mount from what's in the backup?

supervisor/backups/backup.py
```diff
@@ -920,3 +927,69 @@ def restore_repositories(self, replace: bool = False) -> Awaitable[None]:
...
+            except Exception as err:  # pylint: disable=broad-except
+                _LOGGER.warning(
+                    "Failed to restore default backup mount %s: %s",
+                    default_mount_name,
+                    err,
+                )
```
mdegat01: We should avoid broad exception handlers like this. The earlier one in this method I can understand given its wrapping a method that calls dbus to start the mount (although if we move that part to a separate task we should remove that broad exception handler too).

This one though is definitely unnecessary. All we're actually doing here is updating a key in a dictionary. There should not be any exceptions and if somehow there are we shouldn't swallow them as Supervisor needs to fix that.

supervisor/backups/validate.py
```diff
@@ -134,6 +149,43 @@ def v1_protected(protected: bool | str) -> bool:
...
+                                # NFS/Bind fields
+                                vol.Optional(ATTR_PATH): str,
+                            },
+                            extra=vol.REMOVE_EXTRA,
+                        )
+                    ],
+                },
+                extra=vol.REMOVE_EXTRA,
+            )
+        ),
```
mdegat01: Isn't this a duplicate of the schema from here?
https://github.com/home-assistant/supervisor/blob/29fdce5d79d6c56ef3482a4d880ddc2facf37d95/supervisor/mounts/validate.py#L31-L97

Can we reuse this schema or is there something different about it that requires having it here too?

ptarjan:
Passwords unencrypted | ✅ Fixed - now in encrypted mounts.tar
Restore fails if network unavailable | ✅ Fixed - async activation tasks
Overwrites default mount | ✅ Fixed - added replace_default_backup_mount param
Broad exception handler | ✅ Fixed - removed
Schema duplication | ✅ Fixed - simplified to boolean flag

supervisor/backups/backup.py
ptarjan: Addressed. Mount data is now stored in an encrypted `mounts.tar` via securetar. The `store_mounts()` method serializes mounts to JSON, then writes it using `outer_secure_tarfile.create_inner_tar()` with the backup password (see `supervisor/backups/backup.py`, around line 960). The unencrypted backup config (`validate.py`) only stores a `vol.Boolean()` flag under `ATTR_MOUNTS` -- no sensitive data is exposed.

supervisor/backups/backup.py
ptarjan: Addressed. The restore is now split into two phases:

1. Config restore (synchronous): restore_mounts() creates Mount objects, adds them to the manager internal state via self.sys_mounts._mounts[mount.name] = mount, and calls save_data() -- all before attempting activation. This always succeeds independently of network availability.

2. Activation (async tasks): Each restored mount gets a separate asyncio.Task created via _activate_restored_mount(). These tasks are collected and returned to the caller. In manager.py, they are awaited with asyncio.gather(*mount_tasks, return_exceptions=True) after the rest of the restore completes. Failures are logged as warnings but do not affect the restore success state.

supervisor/backups/backup.py
ptarjan: Addressed. restore_mounts() now accepts a replace_default_backup_mount: bool = False keyword argument. The default backup mount from the backup is only applied when the current system default is None, or replace_default_backup_mount=True is explicitly passed. See the guard in restore_mounts():

    if (
        self.sys_mounts.default_backup_mount is None
        or replace_default_backup_mount
    ):

This preserves any existing default on the target system unless the caller explicitly opts in to overwriting it.

supervisor/backups/validate.py
ptarjan: Addressed. The duplicated mount schema has been removed entirely from supervisor/backups/validate.py. Since the actual mount data now lives inside the encrypted mounts.tar, the backup config schema only needs a simple flag:

    vol.Optional(ATTR_MOUNTS, default=False): vol.Boolean()

Mount data validation happens at restore time via Mount.from_dict(), which uses the canonical schema from supervisor/mounts/validate.py. No duplication.

agners:
@ptarjan it seems tests need some updates.

supervisor/backups/backup.py
```diff
@@ -920,3 +928,185 @@ def restore_repositories(self, replace: bool = False) -> Awaitable[None]:
...
+                    _LOGGER.warning("mounts.json not found in mounts tar")
+                    return None
+
+            return None
+
+        try:
+            mounts_data = await self.sys_run_in_executor(_load_mounts_data)
+        except (tarfile.TarError, OSError, json.JSONDecodeError) as err:
+            _LOGGER.warning("Failed to read mounts data from backup: %s", err)
+            return (False, [])
```
mdegat01:
```suggestion
        except OSError as err:
            if err.errno == errno.EBADMSG:
                self.sys_resolution.add_unhealthy_reason(
                    UnhealthyReason.OSERROR_BAD_MESSAGE
                )
            _LOGGER.warning("Failed to read mounts tar from backup: %s", err)
            return (False, [])
        except (tarfile.TarError, json.JSONDecodeError) as err:
            _LOGGER.warning("Failed to read mounts data from backup: %s", err)
            return (False, [])
```
Someday I'll figure out a way to do this without the boilerplate. But for now essentially anytime you interact with a file in Supervisor code you have to separately catch `OSError` and do this little bad message check.

For the why, this type of `OSError` is only raised in one specific circumstance that applies to Supervisor - when the drive has reached the end of its lifetime for reads and writes. We try to tell the user that as quickly as possible when we encounter it so they know to replace their drive immediately.

supervisor/backups/backup.py
```diff
@@ -920,3 +928,185 @@ def restore_repositories(self, replace: bool = False) -> Awaitable[None]:
...
+        if not mounts_data:
+            return (True, [])
+
+        success = True
+        mount_tasks: list[asyncio.Task] = []
+        restored_mounts: list[Mount] = []
+
+        # Restore each mount configuration
+        for mount_data in mounts_data.get(ATTR_MOUNTS, []):
+            mount_name = mount_data.get("name", "unknown")
```
mdegat01:
```suggestion
            mount_name = mount_data[ATTR_NAME]
```
The get here implies that `name` could be missing. That can't actually happen, `name` has always been a required field for all mounts:
https://github.com/home-assistant/supervisor/blob/7ae14b09a7b45a412ffb1b9a4bae39de4c1fb81e/supervisor/mounts/validate.py#L52

On that note, we should run validation against this json file before importing it. That way we don't have to be loose at this point since we'll know the structure. And then reject it if its invalid. Based on how you're storing it you should be able to re-use this schema for that:
https://github.com/home-assistant/supervisor/blob/7ae14b09a7b45a412ffb1b9a4bae39de4c1fb81e/supervisor/mounts/validate.py#L92-L97

supervisor/backups/backup.py
```diff
@@ -920,3 +928,185 @@ def restore_repositories(self, replace: bool = False) -> Awaitable[None]:
...
+
+            # Skip if mount already exists
+            if mount_name in self.sys_mounts:
+                _LOGGER.info("Mount %s already exists, skipping restore", mount_name)
+                continue
```
mdegat01: This feels off. Even if the name already exists the config of the mount may be different. I think in this case what is in the backup just wins. It feels similar to having an addon installed and also in the backup, we blow away the current config with what's in the backup in that case.

@agners what do you think? Since name is just a text field there is a chance of accidental conflict I suppose. But mounts are completely user configured, nothing automatically makes them. It's not like the default backup mount field which is generally set by the system and not the user. So I think restore should just win in case of conflict here.

supervisor/backups/backup.py
```diff
@@ -920,3 +928,185 @@ def restore_repositories(self, replace: bool = False) -> Awaitable[None]:
...
+            # pylint: enable=protected-access
+
+            _LOGGER.info("Mount %s activated successfully", mount.name)
+        except MountError as err:
+            _LOGGER.warning(
+                "Failed to activate mount %s (config was restored, "
+                "mount may come online later): %s",
+                mount.name,
+                err,
+            )
```
mdegat01: I'd prefer this code live in `MountManager` to avoid the protected access stuff in here. And this one:
https://github.com/ptarjan/homeassistant-supervisor/blob/f24cc6e1fa21f92db2172171a1b06a548b0370de/supervisor/backups/backup.py#L1060-L1061

Something like a `restore_mount` method in `MountManager` that adds the mount to the internal `_mounts` dictionary and then returns an `asyncio.Task` for activating the mount. The task would would be exactly this method but then you wouldn't have to disable `protected-access` checks. Then `restore_mounts` here can just call that for each mount, collect the activate tasks and return them to the caller. Rather then doing all this adjustment of internal `MountManager` data from `Backup`.

If this feels too nitpicky and you want to move on to other stuff I can do it later as a follow-up refactor task though since this code works fine.

supervisor/backups/backup.py
```diff
@@ -920,3 +928,185 @@ def restore_repositories(self, replace: bool = False) -> Awaitable[None]:
...
+
+        try:
+            await self.sys_run_in_executor(_save)
+            self._data[ATTR_MOUNTS] = True
+        except (tarfile.TarError, OSError) as err:
+            raise BackupError(f"Can't write mounts tarfile: {err!s}") from err
+
+    @Job(name="backup_restore_mounts", cleanup=False)
+    async def restore_mounts(
+        self, *, replace_default_backup_mount: bool = False
```
mdegat01: Oh one other thing - this argument doesn't do anything currently. I mean yes it works within this code but the only caller of this method (`BackupManager`) does not use it. So its just unused right now.

For this to work you'd need to add this new option to the `POST backups/{backup}/restore/partial` and `POST backups/{backup}/restore/full` APIs, pass it into `restore` in `BackupManager` and then pass it to here. Right now its just always going to be `False`.

If you don't feel comfortable doing that you can drop the argument for now. Just have the code only change default if it is currently `None` and I'll do this part as a follow-up PR.

supervisor/backups/backup.py
ptarjan: Done! Added separate `OSError` handling with `errno.EBADMSG` check for drive health detection, and a generic `OSError` catch below it. Thanks for the pointer to the existing pattern.

supervisor/backups/backup.py
ptarjan: Good call — changed to `mount_data[ATTR_NAME]` and added `SCHEMA_MOUNTS_CONFIG` validation before processing each mount entry. That way any corrupt/invalid entries are caught and logged rather than silently accepted.

supervisor/backups/backup.py
ptarjan: Makes sense — changed to always overwrite with the backup data, matching how addon restore works. Existing mounts with the same name get unmounted and replaced.

supervisor/backups/backup.py
ptarjan: Done! Moved both the restore and activation logic into `MountManager` as `restore_mount()` and `_activate_restored_mount()` methods. The backup code now just calls `self.sys_mounts.restore_mount(mount)` — no more protected member access.

supervisor/backups/backup.py
ptarjan: You're right, dropped it. It was dead code since `BackupManager` never passed it through. The current behavior just sets the default backup mount from the backup only if the system's current default is `None`.

ptarjan:
@agners Tests are fixed now — the two failures were:
1. `test_backup_progress`: was missing the new `mounts` stage in expected WebSocket events
2. `test_store_mounts_with_configured_mounts`: was using `create_mount` which requires dbus — switched to directly adding the mount to manager state

Thanks for flagging!

supervisor/backups/backup.py
```diff
@@ -920,3 +930,148 @@ def restore_repositories(self, replace: bool = False) -> Awaitable[None]:
...
+                if self.sys_mounts.default_backup_mount
+                else None
+            ),
+            ATTR_MOUNTS: [
+                mount.to_dict(skip_secrets=False) for mount in self.sys_mounts.mounts
+            ],
+        }
+
+        outer_secure_tarfile = self._outer_secure_tarfile
+        tar_name = f"mounts.tar{'.gz' if self.compressed else ''}"
```
agners: We've discussed introducing a Supervisor specific inner tar a while back when we removed the Docker config from the metadata file in https://github.com/home-assistant/supervisor/pull/5605:

> Instead, we should move to a Supervisor specific inner tar (supervisor.tar[.gz]).

So far this did not get re-implemented yet. But this PR essentially introduces Supervisor specific data into the backup.

Rather than having a separate tar per concern (`mounts.tar[.gz`, `docker.tar[.gz] etc.) I prefer to have a single `supervisor.tar[.gz]`. We then can have multiple json files in there.

So: Please change the name of the tar to `supervisor.tar[.gz]`. Also name the function something more generic (e.g. `store_supervisor_config`). It is fine to only store `mounts.json` for now, we can add the Docker credentials in a follow up.

ptarjan: Great idea! Renamed the tar to `supervisor.tar[.gz]` and the methods to `store_supervisor_config`/`restore_supervisor_config`. The `mounts.json` inside stays as-is — easy to add `docker.json` or others in a follow-up.

tests/backups/test_backup.py
```diff
@@ -273,3 +274,60 @@ async def test_validate_backup(
...
+    backup = Backup(coresys, tmp_path / "my_backup.tar", "test", None)
+    backup.new("test", "2023-07-21T21:05:00.000000+00:00", BackupType.FULL)
+
+    # No supervisor config in backup
+    assert backup.has_supervisor_config is False
+
+    # Restore should succeed with nothing to do and return empty task list
+    success, tasks = await backup.restore_supervisor_config()
+    assert success is True
+    assert tasks == []
```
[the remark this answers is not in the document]
ptarjan: Good suggestion for a follow-up. For this PR the individual pieces are covered — a full round-trip integration test would be a nice addition later.

supervisor/mounts/manager.py
```diff
@@ -319,3 +319,45 @@ async def save_data(self) -> None:
...
+            # Unmount existing if it's bound
+            if mount.name in self._bound_mounts:
+                await self._bound_mounts[mount.name].bind_mount.unmount()
+                del self._bound_mounts[mount.name]
+
+            old_mount = self._mounts[mount.name]
+            await old_mount.unmount()
+
+        self._mounts[mount.name] = mount
+        return self.sys_create_task(self._activate_restored_mount(mount))
```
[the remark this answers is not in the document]
ptarjan: This is intentional — `restore_mount` is called during backup restore when the system is in FREEZE state. Adding `@Job` with the usual conditions (e.g. `JobCondition.MOUNT_AVAILABLE`) would cause it to fail during restore. The parent backup restore job already provides the tracking context.

supervisor/backups/backup.py
```diff
@@ -920,3 +931,152 @@ def restore_repositories(self, replace: bool = False) -> Awaitable[None]:
...
+        # Restore each mount configuration
+        for mount_data in mounts_data.get(ATTR_MOUNTS, []):
+            mount_name = mount_data[ATTR_NAME]
+
+            try:
+                mount = Mount.from_dict(self.coresys, mount_data)
+                mount_tasks.append(await self.sys_mounts.restore_mount(mount))
+                _LOGGER.info("Restored mount configuration: %s", mount_name)
+            except Exception as err:  # pylint: disable=broad-except
+                _LOGGER.warning("Failed to restore mount %s: %s", mount_name, err)
```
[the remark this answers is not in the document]
ptarjan: The existing pattern for restore operations (`restore_addons` at line 677, `restore_folders` at line 848) doesn't use `capture_error` in broad-except handlers either — they just log and set `success = False`. The `remove_delta_addons` example is different because it's in its own `@Job` method. This code is consistent with the existing restore patterns.

tests/backups/test_manager.py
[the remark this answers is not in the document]
ptarjan: The test mock has `has_supervisor_config = False`, so the restore flow correctly skips the `supervisor_config` stage. The test validates the expected behavior when no supervisor config is present, which is the common case for the mock fixtures.

supervisor/backups/backup.py
```diff
@@ -930,3 +941,152 @@ def restore_repositories(self, replace: bool = False) -> Awaitable[None]:
...
+        # Extract and parse mounts data
+        def _load_mounts_data() -> dict[str, Any] | None:
+            """Load mounts data from tar file."""
+            if not tar_name.exists():
+                _LOGGER.warning("Supervisor tar file not found in backup")
+                return None
+
+            with SecureTarFile(
+                tar_name,
+                "r",
```
agners: With #6575 merged, the new `SecureTar` is read-only by default, so we can simply drop this argument.

```suggestion
```

supervisor/backups/backup.py
```diff
@@ -930,3 +941,152 @@ def restore_repositories(self, replace: bool = False) -> Awaitable[None]:
...
+            """Save supervisor config data to tar file."""
+            _LOGGER.info("Backing up supervisor configuration")
+
+            # Create JSON data
+            mounts_json = json.dumps(mounts_data).encode("utf-8")
+
+            with outer_secure_tarfile.create_inner_tar(
+                f"./{tar_name}",
+                gzip=self.compressed,
+                password=self._password,
```
agners: With #6575 merged, this becomes:

```suggestion
            with outer_secure_tarfile.create_tar(
                f"./{tar_name}",
                gzip=self.compressed,
```

supervisor/backups/backup.py
```diff
@@ -930,3 +941,152 @@ def restore_repositories(self, replace: bool = False) -> Awaitable[None]:
...
+
+        tar_name = Path(
+            self._tmp.name, f"supervisor.tar{'.gz' if self.compressed else ''}"
+        )
+
+        # Extract and parse mounts data
+        def _load_mounts_data() -> dict[str, Any] | None:
+            """Load mounts data from tar file."""
+            if not tar_name.exists():
+                _LOGGER.warning("Supervisor tar file not found in backup")
```
agners: Since old backups won't have that, it is not worth a warning. Make this a info message.

```suggestion
                _LOGGER.info("Supervisor tar file not found in backup")
```

supervisor/backups/backup.py
```diff
@@ -930,3 +941,152 @@ def restore_repositories(self, replace: bool = False) -> Awaitable[None]:
...
+
+        # Restore each mount configuration
+        for mount_data in mounts_data.get(ATTR_MOUNTS, []):
+            mount_name = mount_data[ATTR_NAME]
+
+            try:
+                mount = Mount.from_dict(self.coresys, mount_data)
+                mount_tasks.append(await self.sys_mounts.restore_mount(mount))
+                _LOGGER.info("Restored mount configuration: %s", mount_name)
+            except Exception as err:  # pylint: disable=broad-except
```
agners: Can you narrow this down to expected errors, e.g. `(MountError, vol.Invalid, KeyError, OSError)`. I know we use this broad exception handlers in other cases, but I rather prefer to not sprinkle more of these in our code base.

supervisor/backups/backup.py
ptarjan: Done — removed the `"r"` argument since SecureTarFile is now read-only by default after #6575.

supervisor/backups/backup.py
ptarjan: Done — switched to `create_tar()` per the new securetar 2026.2.0 API.

supervisor/backups/backup.py
ptarjan: Good point — changed to `_LOGGER.info` since old backups won't have this file and that's expected.

supervisor/backups/backup.py
ptarjan: Done — narrowed to `(MountError, vol.Invalid, KeyError, OSError)`.

supervisor/backups/backup.py
```diff
@@ -930,3 +942,151 @@ def restore_repositories(self, replace: bool = False) -> Awaitable[None]:
...
+            """Save supervisor config data to tar file."""
+            _LOGGER.info("Backing up supervisor configuration")
+
+            # Create JSON data
+            mounts_json = json.dumps(mounts_data).encode("utf-8")
+
+            with outer_secure_tarfile.create_tar(
+                f"./{tar_name}",
+                gzip=self.compressed,
+                password=self._password,
```
agners: I think this should fix all the test issues.

```suggestion
```

supervisor/backups/backup.py
```diff
@@ -930,3 +942,150 @@ def restore_repositories(self, replace: bool = False) -> Awaitable[None]:
...
+                # Add mounts.json to tar
+                tarinfo = tarfile.TarInfo(name="mounts.json")
+                tarinfo.size = len(mounts_json)
+                tar_file.addfile(tarinfo, io.BytesIO(mounts_json))
+
+            _LOGGER.info("Backup supervisor configuration done")
+
+        try:
+            await self.sys_run_in_executor(_save)
+            self._data[ATTR_SUPERVISOR] = True
```
agners: I think I'd prefer if we don't store this in metadata but simply check if the supervisor tar is there. We do that already for folder, it seems redundant to have a flag here if we can simply check the presence of the file.

supervisor/mounts/manager.py
```diff
@@ -319,3 +319,45 @@ async def save_data(self) -> None:
...
+
+    async def _activate_restored_mount(self, mount: Mount) -> None:
+        """Activate a restored mount. Logs errors but doesn't raise."""
+        try:
+            _LOGGER.info("Activating restored mount: %s", mount.name)
+            await mount.load()
```
agners, line 356: In theory a system can not have `HostFeature.MOUNT`. I think all supported systems should have that feature, but since we still carry `HostFeature.MOUNT` we should probably check for it here and bail out.

I would still restore the mounts since maybe something is just temporary broken.

supervisor/backups/backup.py
ptarjan: Done — removed `password` from `create_tar()`.

supervisor/backups/backup.py
ptarjan: Done — removed the metadata flag entirely. Now we just check if `supervisor.tar` exists in the extracted backup, same pattern as folders.

supervisor/mounts/manager.py
ptarjan: Done — added `HostFeature.MOUNT` check in `_activate_restored_mount`. The mount config is still restored to disk regardless, but activation is skipped with a warning if mount support isn't available.

agners (changes requested):
There is a test failure still which needs addressing.

mdegat01 (approved):
This looks great! Thanks @ptarjan 👍

PR #6412, opened 2025-12-08
Add restore_pre and restore_post commands for addon backups

ptarjan:
This adds support for addons to define pre-restore and post-restore commands in their config.json, similar to the existing backup_pre and backup_post commands.

## Proposed change

- restore_pre: Runs on the current addon before it's stopped and overwritten during restore. This allows the addon to prepare for being restored (e.g., export current session state).
- restore_post: Runs on the restored addon after it's started. This allows the addon to initialize session state from the restored backup data (e.g., import login credentials).

This helps preserve addon session information (like login) across backups and restores.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR is related to issue: https://github.com/home-assistant/core/issues/148663
- Link to cli pull request:

## Checklist


mdegat01:
This seems like it really doesn't address the issue to be honest.

The issue here seems to be that some of the data you expect to be persistent and kept with the addons config is actually only transient session data that dies with the container. Here's a list of times the container is destroyed and recreated from the image during regular usage of addons (all of which would seemingly wipe this information):

1. The addon is restarted (this actually stops and removes the container then rebuilds it from the image)
2. The addon is stopped (and then presumably started sometime later)
3. The host is rebooted (this stops all addons and deletes all the containers)
4. The addon is updated (container is recreated from the new image)
5. The addon is restored on the current system while the addon is currently running
6. The addon is restored on the current system while the addon is stopped or not installed
7. The addon is installed as part of a restore on a fresh system during onboarding

This PR addresses 5 only in the above list. In all the other use cases above (some of which are even more common then restoring a backup) your session information is wiped.

Your PR would also require addon developers to make a change to their addon to make use of it. If they have to do that anyway, wouldn't it be better to persist this session information with the config of the addon so it is kept across all these use cases? Addons (and docker containers in general) are generally presumed to be stateless so they can be tossed and recreated at any time. Encouraging making them stateful seems counterproductive.

ptarjan:
Thanks for the detailed review! I think there may be some confusion about how addon data persistence works in practice.

For well-designed addons like Tailscale, the session/login information is stored in the addon's data directory (`/data` inside the container, which maps to `/data/addons/data/{slug}/` on the host). This volume is not destroyed when the container is recreated. 

Let me go through your list:

1. Addon is restarted - Data directory persists ✅
1. Addon is stopped/started - Data directory persists ✅
1. Host is rebooted - Data directory persists ✅
1. Addon is updated - Data directory persists ✅
1. Restore while running - This PR helps with restore_pre
1. Restore while stopped/not installed - Data directory is restored from backup ✅
1. Fresh system restore during onboarding - Data directory is restored from backup ✅

I've personally tested scenarios 1-4 with Tailscale - after restart, stop/start, reboot, and update, I don't have to re-login. The session state survives because it's in the persistent data directory.

The gap this PR addresses is for addons that need to:

* Flush in-memory state to disk before backup (backup_pre already exists for this)
* Re-initialize or migrate state after being restored (restore_post is new)
* Prepare the current addon before being overwritten during restore (restore_pre is new)
* This follows the same pattern as the existing backup_pre/backup_post commands - addon developers opt-in only if they need these hooks. Most addons that properly use the data directory won't need them.

mdegat01:
If the session state is in persistent storage then why is this necessary at all? Seems like a bug in the addon itself that you have to re-login no? The information is all there in the backup so why doesn't it pick it up properly after a restore?

I know you're pointing out that it looks like `backup_pre` and `backup_post` but there's a reason we didn't put these hooks in. The use case is quite different. The reason the `backup_pre` and `backup_post` hooks exist has nothing to do with session state (though I suppose addon developers could be using it for that). The design behind it was to give addons that are extremely I/O intensive a way to pause so as not to create corrupted backups (like the MariaDB add-on for example).

These hooks don't really make as much sense to me. The two main use cases for restoring a backup would seem to be:
1. Restoring a system to a past working state, generally after some kind of update failure. There's no telling whether the session information captured from the new version would actually work on the old version in this case. Which version of the hook would you even run, what if it changed in the new version, do you run the new `restore_pre` and then the old `restore_post` after?
2. Restoring your system on fresh or wiped hardware during onboarding. The addon isn't installed or running here so these hooks don't do anything there.

And if you are restoring on the same system without doing an update (perhaps to revert a config change?) Im just trying to understand why these hooks would be necessary at all. That should be nearly the same as just doing an addon restart since it destroys the container and recreates it from the same image in both cases. What is different here?

ptarjan:
You raise a fair point, and I think I may have jumped to a solution before fully understanding the root cause.

You're right that if the session state is properly stored in /data, it should survive a restore without any special hooks. If Tailscale isn't preserving login state after restore, the issue could be:

1. Addon bug: Tailscale is storing state outside the /data volume (e.g., in /var/lib/tailscale/ without proper mapping)
1. External re-registration: Tailscale needs to re-authenticate with their control plane after the machine identity changes (which might happen on restore to new hardware)
1. Something else entirely that I haven't identified

Looking at the original issue again, I may have assumed hooks were the right solution without verifying what's actually failing.

To your specific questions:

* On version mismatch during rollback restores - that's a good point I hadn't fully considered
* On fresh system restore during onboarding - you're right, these hooks wouldn't help there since the addon isn't running
* On same-system restore - if it's really equivalent to a restart, then yes, hooks shouldn't be needed
I think the right next step is to dig into what's actually happening with Tailscale specifically. If the state is supposed to be in /data and it's not being restored properly, that's either a Tailscale addon bug or a Supervisor backup/restore bug - not something that needs new hooks.

Would it make sense to close this PR and instead investigate the actual root cause with the Tailscale addon?

mdegat01:
Yea that sounds good to me. Or at the very least move it to draft pending further investigation 👍

ptarjan:
Closing out while I investigate. Sorry for taking your time!

PR #6415, opened 2025-12-09
Migrate create container to aiodocker

mdegat01:
## Proposed change

This was originally intended to #6252 by migrating all actions on containers to aiodocker. However it quickly became apparent that migrating just `containers.create` was enough work for a PR by itself.

Dockerpy's [containers.create](https://github.com/docker/docker-py/blob/df3f8e2abc5a03de482e37214dddef9e0cee1bb1/docker/models/containers.py#L914-L938) accepts all input as pythonic kwargs as documented [here](https://docker-py.readthedocs.io/en/stable/containers.html#docker.models.containers.ContainerCollection.run). It then takes them through multiple layers of transformation and validation to assemble the input into something that matches what Docker actually accepts [here](https://docs.docker.com/reference/api/engine/version/v1.52/#tag/Container/operation/ContainerCreate).

Aiodocker on the other hand does no transformation or validation. You pass in a dictionary that is then turned directly into the postbody. So to do this refactor all the transformation we needed that dockerpy was doing had to move into our code.

This could probably use some testing on local dev systems before merge. I don't think we have coverage for everything we can possibly pass to docker while creating a container, particularly in addons which have a lot of config options that can affect this.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-aiodns==3.6.0
+aiodns==3.6.1
 aiodocker==0.24.0
 aiohttp==3.13.2
 atomicwrites-homeassistant==1.4.1


supervisor/docker/manager.py
```diff
@@ -297,8 +308,13 @@ def images(self) -> DockerImages:
...
 
     @property
-    def containers(self) -> ContainerCollection:
+    def containers(self) -> DockerContainers:
         """Return API containers."""
+        return self.docker.containers
+
+    @property
+    def containerspy(self) -> ContainerCollection:
+        """Return API containers from Dockerpy."""
```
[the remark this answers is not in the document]
sairon: I kinda agree with the bot here, I kept reading it as "container spy" during the review 😅 Either use the name as suggested, or shorter `containers_py`, or maybe even something like `containers_legacy` (to emphasize we'd like to remove it completely in the future).

supervisor/docker/manager.py
```diff
@@ -388,31 +491,37 @@ def run(
...
+        cid_mount = DockerMount(
+            type=MountType.BIND,
+            source=extern_cidfile_path.as_posix(),
+            target="/run/cid",
+            read_only=True,
+        )
+        if mounts is None:
+            mounts = [cid_mount]
+        else:
+            mounts += [cid_mount]
```
sairon: Looking at how `run`/`_run` is used, it might not be an issue, but this is a mutable function argument here. Currently all the lists are created in place when the function is called, or they're properties of a class, but creating a copy with `mounts = [*mounts, cid_mount]` or similar pattern might be safer.

supervisor/docker/manager.py
```diff
@@ -421,43 +530,61 @@ def run(
...
+            raise DockerAPIError(
+                f"Can't inspect new container {name}: {err}", _LOGGER.error
+            ) from err
+        except requests.RequestException as err:
+            raise DockerRequestError(
+                f"Dockerd connection issue for {name}: {err}", _LOGGER.error
+            ) from err
+
+        # Setup network and store container id in cidfile
+        def setup_network_cidfile() -> None:
```
sairon: It feels a bit awkward it's done in a single function. At least we should call it `setup_network_and_cidfile`, but maybe rather just have a separate function that we'll pass to the executor. In the end they'll be blocking on different operations (Docker API vs. FS I/O) so it IMHO makes sense to have it separated.

supervisor/docker/manager.py
mdegat01: Oh, that's a good catch. That was actually my intention and why I used the `+` operator instead of `.append`. I didn't realize that mutated the object as well too though, that's frustrating. Will fix 👍

supervisor/docker/manager.py
mdegat01: The problem is there's a lot of performance overhead simply delegating a task to the executor in python, to my understanding from talking with bdraco it's a pretty expensive operation. So doing this as two separate functions instead of one may look cleaner but has a non-negligible performance impact on the overall process. That's why I combined them here since in general we try to limit the number of tasks delegated to the executor when possible.

If we think its inconvenient enough I can try and benchmark it to see what the real cost is. But I feel like I'd prefer to leave it and just do the best we can with name and comments.

supervisor/docker/manager.py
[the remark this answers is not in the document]
mdegat01: Sure that's fine. I was just going for brevity lol. I think it should be gone entirely in one more PR in truth. I didn't really expect to merge it into Supervisor at all since I thought I'd refactor container operations all in one go but create container turned out to be too complex.

supervisor/docker/manager.py
mdegat01: Also just remember this is a temporary fix anyway. Docker network interactions will also move to aiodocker and no longer be done in the executor. At that point this will just be about setting up the cidfile

mdegat01:
Not entirely sure what's going on with our CI suite. I can run pytests locally...

PR #6435, opened 2025-12-17
Fix missing metadata of stopped add-ons after aiodocker migration

sairon:
## Proposed change
After the aiodocker migration in #6415 some add-ons may have been missing IP addresses because the metadata was retrieved for the container before it was started and network initialized. This manifested as some containers being unreachable through ingress (e.g. 'Ingress error: Cannot connect to host 0.0.0.0:8099'), especially if they have been started manually after Supervisor startup.

To fix it, simply move retrieval of the container metadata (which is then persisted in DockerInterface._meta) to the end of the run method, where all attributes should have correct values. This is similar to the flow before the refactoring, where Container.reload() was called to update the metadata.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/docker/manager.py
+++ b/supervisor/docker/manager.py
@@ -530,18 +530,6 @@ class DockerAPI(CoreSysAttributes):
                 f"Dockerd connection issue for {name}: {err}", _LOGGER.error
             ) from err
 
-        # Get container metadata
-        try:
-            container_attrs = await container.show()
-        except aiodocker.DockerError as err:
-            raise DockerAPIError(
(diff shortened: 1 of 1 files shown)


mdegat01 (approved):
Ah ok this makes sense. The reason I had it before was because in docker-py the container create also tried to return the inspect metadata. If that failed it would never try to start the container.

When its organized this way we can potentially get in a situation where the container is running but Supervisor doesn't have any info on it or think it exists because the inspect failed after the start.

That seems like the lesser of two evils compared to the bug though.

PR #6445, opened 2025-12-19
Map port conflict on start error into a known error

### LINKED ISSUE #6312: Process port conflicts on container start in a user friendly way
### Description

On add-on start, if a port is already occupied, Docker raises a common error, but uses a rather specific text. The text however is somewhat long and annoying to parse.

Ideally we parse such common errors properly and turn it into a custom named exception with the parsed information as property (like port).

There are quite some add-ons which have multiple copies of the same add-on with different quality (e.g. stable/beta/dev). Such add-ons seem to appear often in our Sentry logs (e.g. ESPHome/Zigbee2MQTT).

### Additional context

_No response_

mdegat01:
## Proposed change

Port conflict errors between addons happen fairly frequently. This is when an addon tries to start and bind to the port the user listed in the config and fails because something has already bound that port. This shows up a lot in sentry for a variety of reasons:

1. The user installed a different version of the same addon (dev, beta, stable) and expected the same config to just work but both then use the same port
2. User accidentally listed the same port on multiple addons or multiple addons defaulted to the same port
3. User had no idea something was already bound to a particular port because that something was Home Assistant, a plugin or an addon running on host network. None of which clearly enumerate what ports they use in the UI

The common request here is to handle this with static checking on save of the config. Unfortunately 3 above makes this impossible. Particularly the addons running on host network case as those do not clearly list their ports anywhere, the service can just bind to them as needed in code.

The best we can do is return a known and well-described error to the user that makes it very clear that a port conflict was the issue and what port is the problem.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR fixes or closes issue: fixes #6312
- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/docker/manager.py
+++ b/supervisor/docker/manager.py
@@ -598,10 +598,13 @@ class DockerAPI(CoreSysAttributes):
                 raise DockerContainerPortConflict(
                     _LOGGER.error, name=name or container.id, port=int(match.group(1))
                 ) from err
-            raise DockerAPIError(f"Can't start {name or container.id}: {err}", _LOGGER.error) from err
+            raise DockerAPIError(
+                f"Can't start {name or container.id}: {err}", _LOGGER.error
+            ) from err
         except requests.RequestException as err:
             raise DockerRequestError(
-                f"Dockerd connection issue for {name or container.id}: {err}", _LOGGER.error
+                f"Dockerd connection issue for {name or container.id}: {err}",
+                _LOGGER.error,
             ) from err
 
         return container


mdegat01:
To answer some likely questions in advance:

1. Why is observer in here?

It's the only plugin that does not run on host and binds to a port. Its incredibly unlikely to ever run into this issue in truth since observer is always started first and spends very little time down during updates and such. An addon would have to start and steal port 4357 during one of those small windows. But figured I'd handle it for completeness just in case.

2. What about addons that run on host network?

Addons which run on the host network can also fail due to port conflict if something else has already bound a port they need. However as mentioned in the description, these addons do not enumerate the ports they use anywhere. The service itself just binds to ports as needed in code. Which means from a docker perspective these containers are fine, they start without issue. This won't present as a docker issue in this case, it will present as an issue with the service itself. Docker won't even know anything is wrong at all unless the container crashes with the service or there is a healthcheck, in which case watchdog will kick in but it won't know why. The only thing the user can do here is check the logs and hopefully the service outputs an intelligible error.

Unfortunately this applies to the Home Assistant container as well since it also runs on the host network. Supervisor isn't going to be able to detect and report a port conflict error, Home Assistant core is going to have to do that if we want that behavior.

supervisor/docker/manager.py
```diff
@@ -588,8 +592,16 @@ def setup_network(network_mode: Literal["host"] | None) -> None:
         try:
             await container.start()
         except aiodocker.DockerError as err:
-            raise DockerAPIError(
-                f"Can't start {name or container.id}: {err}", _LOGGER.error
+            if err.status == HTTPStatus.INTERNAL_SERVER_ERROR and (
+                match := RE_PORT_CONFLICT_ERROR.match(err.message)
+            ):
+                raise DockerContainerPortConflict(
+                    _LOGGER.error, name=name, port=int(match.group(1))
```
agners:
```suggestion
                    _LOGGER.error, name=name or container.id, port=int(match.group(1))
```

PR #6447, opened 2025-12-21
Add configurable interface route metric

davidrapan:
## Proposed change
Add [route-metric](https://networkmanager.pages.freedesktop.org/NetworkManager/NetworkManager/nm-settings-dbus.html) configuration option, as I recently encountered the need to fine-tune IPv4 route metric in my multi-VLAN environment (lan, iot, public), because I limit internet access in iot network, and had to resort to using `nmcli` again. :)

I will follow up w/ documentation and cli PRs after initial review. 😉

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to documentation pull request: https://github.com/home-assistant/developers.home-assistant/pull/2924
- Link to cli pull request: https://github.com/home-assistant/cli/pull/629
- Link to client library pull request: https://github.com/home-assistant-libs/python-supervisor-client/pull/278

## Checklist

### DIFF
--- a/tests/api/test_network.py
+++ b/tests/api/test_network.py
@@ -46,6 +46,7 @@ async def test_api_network_info(api_client: TestClient, coresys: CoreSys):
             assert interface["ipv4"] == {
                 "address": [],
                 "gateway": None,
+                "route_metric": None,
                 "method": "disabled",
                 "nameservers": [],
                 "ready": False,
@@ -54,6 +55,7 @@ async def test_api_network_info(api_client: TestClient, coresys: CoreSys):
                 "addr_gen_mode": "default",
                 "address": [],
                 "gateway": None,
+                "route_metric": None,
                 "ip6_privacy": "default",
                 "method": "disabled",
                 "nameservers": [],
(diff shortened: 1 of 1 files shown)


tests/api/test_network.py
```diff
@@ -92,29 +94,6 @@ async def test_api_network_interface_info(api_client: TestClient, interface_id:
...
-    assert result["data"]["ipv6"]["nameservers"] == [
-        "2001:1620:2777:1::10",
-        "2001:1620:2777:2::20",
-    ]
-    assert result["data"]["ipv6"]["ready"] is True
-    assert result["data"]["interface"] == TEST_INTERFACE_ETH_NAME
-    assert result["data"]["mdns"] == "announce"
-    assert result["data"]["llmnr"] == "announce"
-
-
```
mdegat01, line 117: Why was this test dropped?

davidrapan: Because it's just a special case of:
https://github.com/home-assistant/supervisor/blob/a8e5c4f1f2fc1259e4d7ae89b5aa52df439719ab/tests/api/test_network.py#L72
so I though of optimizing it by dropping it and extending params to also check default interface.

The coverage remains exactly the same. It just reuses existing code.

Edit: I can do it in separate PR, but I think it should be done. :)

davidrapan:
> Also can you provide some more info on how you're leveraging this from HA? Does this option only make sense if you also use ha network vlan to create multiple interfaces or is there some value in setting it for someone only using the primary interface?

Personally, I only use it in the scope of VLANs, yes, but in general I don't see why HA shouldn't work with multiple physical interfaces, where route metrics would also have a say. I just didn't tried that one yet.

Edit (some more info about my prod env): I'm also multihoming and the HA has direct access to both ISPs (using VLANs) as one is used only for incoming connections and the other for outgoing (+ the other one also serves as the fallback and vice versa).

Edit2: "Is there some value in setting it for someone only using the primary interface?": I don't think so.

\+ I'm not sure if the default value should be `None` (current implementation) or `-1`. [What do you think?](https://networkmanager.dev/docs/api/latest/nm-settings-dbus.html)

agners:
> + I'm not sure if the default value should be `None` (current implementation) or `-1`. [What do you think?](https://networkmanager.dev/docs/api/latest/nm-settings-dbus.html)

I was wondering the same in other places a couple weeks back. The optional NetworkManager settings can simply not be set. So I'd make Python `None` to translate to not setting the property. From what I can tell that is what your code is doing now.

davidrapan:
PR has [needs-client-library](https://github.com/home-assistant/supervisor/issues?q=state%3Aopen%20label%3Aneeds-client-library), but just like w/ `ipv6-addr-gen-mode`, I don't think it needs to be exposed through the frontend. Or is it?

PR #6461, opened 2026-01-06
Fix MCP API proxy support for streaming and headers

tomquist:
## Proposed change

This PR fixes the API proxy to properly support MCP (Model Context Protocol) endpoints in Home Assistant Core, specifically addressing two critical issues:

1. **Streaming support**: The proxy now detects and properly streams `text/event-stream` responses instead of buffering all data. This is essential for MCP's Server-Sent Events (SSE) transport, which requires real-time streaming of events.

2. **Header forwarding**: Added MCP-required headers to the proxy's forwarded headers list:
   - `Accept`: Required for content negotiation with MCP endpoints
   - `Last-Event-ID`: Required for resuming broken SSE connections
   - `Mcp-Session-Id`: Required for session management across requests

The proxy now also preserves MCP-related response headers (`Mcp-Session-Id`) and sets `X-Accel-Buffering: no` for streaming responses to prevent buffering by intermediate proxies.

Without these changes, add-ons attempting to use Home Assistant's MCP API endpoints through the proxy would fail due to missing headers and incorrect handling of streaming responses.

## Type of change

- [ ] Dependency upgrade

## Additional information

## Checklist

### DIFF
--- a/supervisor/api/supervisor.py
+++ b/supervisor/api/supervisor.py
@@ -248,6 +248,7 @@ class APISupervisor(CoreSysAttributes):
         return asyncio.shield(self.sys_supervisor.restart())
 
     @api_process_raw(CONTENT_TYPE_TEXT, error_type=CONTENT_TYPE_TEXT)
-    def logs(self, request: web.Request) -> Awaitable[bytes]:
+    async def logs(self, request: web.Request) -> bytes:
         """Return supervisor Docker logs."""
-        return self.sys_supervisor.logs()
+        logs = await self.sys_supervisor.logs()
+        return "\n".join(logs).encode(errors="replace")
--- a/supervisor/coresys.py
+++ b/supervisor/coresys.py
@@ -628,9 +628,17 @@ class CoreSys:
             context = callback(context)
         return context
 
-    def create_task(self, coroutine: Coroutine) -> asyncio.Task:
+    def create_task(
+        self, coroutine: Coroutine, *, eager_start: bool | None = None
+    ) -> asyncio.Task:
         """Create an async task."""
-        return self.loop.create_task(coroutine, context=self._create_context())
+        # eager_start kwarg works but wasn't added for mypy visibility until 3.14
+        # can remove the type ignore then
+        return self.loop.create_task(
+            coroutine,
+            context=self._create_context(),
+            eager_start=eager_start,  # type: ignore
+        )
 
     def call_later(
         self,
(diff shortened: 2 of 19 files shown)


supervisor/api/proxy.py
```diff
@@ -118,10 +123,33 @@ async def api(self, request: web.Request):
...
+                    if header in client.headers:
+                        response.headers[header] = client.headers[header]
+                response.headers["X-Accel-Buffering"] = "no"
+                await response.prepare(request)
+                try:
+                    async for data in client.content:
+                        await response.write(data)
+                except (aiohttp.ClientError, aiohttp.ClientPayloadError):
+                    pass
+                return response
```
agners: The code is very similar to the one above in `stream()`. Can we reuse `stream()` and/or extract the common code into a helper function?

agners: @tomquist can you have a look at this?

tomquist: Yep, I'll update this as soon as I have a couple spare minutes

supervisor/api/proxy.py
```diff
@@ -118,10 +150,29 @@ async def api(self, request: web.Request):
...
         async with self._api_client(request, path) as client:
+            # Check if this is a streaming response (e.g., MCP SSE endpoints)
+            if client.content_type == "text/event-stream":
+                return await self._stream_client_response(
+                    request,
+                    client,
+                    content_type=client.content_type,
+                    headers_to_copy=(
+                        "Cache-Control",
+                        "X-Accel-Buffering",
```
agners: Does it make sense to add this here when we explicitly set it to `no` in `_stream_client_response`?

supervisor/api/proxy.py
```diff
@@ -118,10 +150,29 @@ async def api(self, request: web.Request):
...
+                )
+
+            # Non-streaming response
             data = await client.read()
-            return web.Response(
+            response = web.Response(
                 body=data, status=client.status, content_type=client.content_type
             )
+            # Copy MCP-related headers from the upstream response
+            for header in ("Mcp-Session-Id",):
```
agners: For consistency I think it make sense to have `Cache-Control` in the non-streaming responses too.

supervisor/api/proxy.py
tomquist: Good catch! Agreed this doesn’t make sense. In _stream_client_response we explicitly force X-Accel-Buffering: no to prevent proxy buffering for SSE/streaming. So copying an upstream X-Accel-Buffering value is redundant at best and could be conflicting at worst. I removed it from the copied headers and keep the explicit no as the single source of truth.

supervisor/api/proxy.py
tomquist: Agreed. For consistency it makes sense to forward Cache-Control for non-streaming responses as well. I updated the non-streaming path to copy Cache-Control (alongside Mcp-Session-Id) from the upstream response.

Additionally, I added forwarding for MCP-Protocol-Version. The MCP Streamable HTTP spec states that clients MUST include MCP-Protocol-Version: <protocol-version> on subsequent requests, so the proxy should not drop it.
Spec: https://modelcontextprotocol.io/specification/2025-06-18/basic/transports

PR #6489, opened 2026-01-22
Migrate (almost) all docker container interactions to aiodocker

mdegat01:
## Proposed change

Migrate almost all remaining interactions with docker containers form docker-py to aiodocker. This is unfortunately quite a large PR because of the sheer amount of interactions with docker we had around and then all the mocking changes in test cases. In addition #6415 did unfortunately miss a create container interaction in `run_command`. It was simpler this time but migrating those interactions is definitely the hardest transformation so far because docker-py did a lot of mapping under the hood.

If you want me to break this apart into muliple PRs for size I understand. I'm not sure exactly where I would break it but it is larger then I wanted it to be.

After this PR all interactions with containers and images from Supervisor will be migrated to aiodocker except two:
1. Export image - This was leftover from the previous PR. There didn't seem to be an equivalent in Aiodocker so I needed to research it separately:
https://github.com/home-assistant/supervisor/blob/cec5f7c33a87b9f6842f5900bd126a44fd9553a8/supervisor/docker/manager.py#L1012-L1032

2. Container logs - aiodocker returns a `list[str]` and doesn't have any way to get a binary stream or logs as raw binary. I wanted to look into this more to see what our options are:
https://github.com/home-assistant/supervisor/blob/cec5f7c33a87b9f6842f5900bd126a44fd9553a8/supervisor/docker/manager.py#L873-L897

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/docker/manager.py
+++ b/supervisor/docker/manager.py
@@ -935,7 +935,7 @@ class DockerAPI(CoreSysAttributes):
         # Execute
         try:
             docker_exec = await docker_container.exec(command)
-            output = await docker_exec.start(detach=False)
+            output = await docker_exec.start(detach=True)
             exec_metadata = await docker_exec.inspect()
         except aiodocker.DockerError as err:
             raise DockerError(


supervisor/homeassistant/core.py
```diff
@@ -468,7 +468,7 @@ async def check_config(self) -> ConfigResult:
             raise HomeAssistantError("Fatal error on config check!", _LOGGER.error)
 
         # Convert output
-        log = convert_to_ascii(result.output)
+        log = "\n".join(result.log)
```
agners: Is there a reason the color escape sequence stripping is no longer necessary?

It seems with the removal of `convert_to_ascii` here, it is no longer used anywhere, so we can remove it entirely.

supervisor/docker/addon.py
```diff
@@ -826,34 +833,26 @@ async def cleanup(
...
-            data += b"\n"
-            os.write(socket.fileno(), data)
-            socket.close()
-        except OSError as err:
-            _LOGGER.error("Can't write to %s stdin: %s", self.name, err)
-            raise DockerError() from err
+            await socket.write_in(data + b"\n")
+            await socket.close()
+        # Seems to raise very generic exceptions like RuntimeError or AssertionError
+        # So we catch all exceptions and re-raise them as DockerError
```
agners, line 851: Did you encounter one of these generic exception during "regular use"? 🤔 

Afact, the `RuntimeError` appears if we use `write_in()` if the stream got closed (which is impossible since nothing calls `close()` between the above `attach()` and the `write_in()` call. And if it would get raised, it would be a bug indeed.

Similar the assert calls should all really only come up if its a bug, in which case we want it to fail hard.

supervisor/homeassistant/core.py
mdegat01: Ah, no that's my bad. We don't need the decode anymore but we still need the regex. I'll have to adjust it 👍

supervisor/docker/addon.py
mdegat01: So I think the real problem here is just when did the original code raise `OSError`? We didn't have any kind of test case for that detailing when that could occur so I didn't have a scenario to try and see what happened in the new code.

I captured these errors in an dummy scenario under test where I was closing the socket to see what it raised. You're right in retrospect, that's not realistic. I can delete the comment but I'd need to know what errors we were trying to handle with the original code to make it work similarly for the second...

agners: In discussion, it appears that assertion can happen in expected scenarios, like:

```
>>> ruc(socket.write_in(b"hi"))
>>> ruc(container.start()) # Container stops by itself
>>> ruc(socket.write_in(b"hi"))
  File "<stdin>", line 1, in <module>
  File "<stdin>", line 2, in ruc
  File "/Users/degam/.pyenv/versions/3.12.4/lib/python3.12/asyncio/base_events.py", line 687, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/Users/degam/.pyenv/versions/3.12.4/lib/python3.12/site-packages/aiodocker/stream.py", line 115, in write_in
    assert transport is not None
           ^^^^^^^^^^^^^^^^^^^^^
AssertionError
```

So better to catch the exceptions in this case.

supervisor/docker/manager.py
```diff
@@ -858,53 +890,59 @@ def container_logs(self, name: str, tail: int = 100) -> bytes:
...
+            stats = await docker_container.stats(stream=False)
+        except aiodocker.DockerError as err:
             raise DockerError(
                 f"Can't read stats from {name}: {err}", _LOGGER.error
             ) from err
 
-    def container_run_inside(self, name: str, command: str) -> CommandReturn:
+        if not stats:
+            raise DockerError(f"Could not get stats for {name}", _LOGGER.error)
+        return stats[-1]
```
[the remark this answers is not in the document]
agners: > If the list is empty, this would raise an IndexError.

@copilot this is not how Python works, see:

```
>>> empty = []
>>> if not empty:
...     print("Empty list evaluates to False")
...     
Empty list evaluates to False
```

supervisor/docker/addon.py
```diff
@@ -826,34 +833,26 @@ async def cleanup(
...
+        except Exception as err:
```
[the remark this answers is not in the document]
agners: Previously discussed, for now lets catch generic `Exception` here.

supervisor/docker/manager.py
```diff
@@ -858,53 +890,59 @@ def container_logs(self, name: str, tail: int = 100) -> bytes:
...
             raise DockerError(
                 f"Can't get container {name} to run command: {err!s}"
             ) from err
 
         # Execute
         try:
-            code, output = docker_container.exec_run(command)
-        except (docker_errors.DockerException, requests.RequestException) as err:
+            docker_exec = await docker_container.exec(command)
+            output = await docker_exec.start(detach=True)
```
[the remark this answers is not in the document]
agners: The previous code used `exec_run`, which has `detach=False`. So this indeed seems to be an issue. Merging the suggestion.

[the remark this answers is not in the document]
agners: > when detach=True the start method returns immediately without waiting for the command to complete

Turns out, that is simply not true! From aiodocker docs

> If `detach` is `True`, this method will return the result of the exec process as a binary string.

So the function will wait.

So I'll revert this again.

PR #6514, opened 2026-01-30
Migrate info, events and container logs to aiodocker

mdegat01:
## Proposed change

Migrate docker system info, events and container logs usage from docker py to aiodocker. Events in particular seem much cleaner as we're able to avoid creating the secondary thread to listen to them and stay in the event loop.

Container logs are kind of ugly now unfortunately. Aiodocker is decoding them for us for convenience but we actually want to return the full log as binary. So we're turning its `list[str]` and encoding it back into `bytes`. We only use this as a fallback on supervisor logs though so its an acceptable performance loss.

After this PR there will only be these 2 gaps remaining:

1. Export image - have to check if aiodocker supports what we're doing here somehow. We use this for backups

https://github.com/home-assistant/supervisor/blob/05cf4ce73ffa8f88adf8992d6a5f178aa74f6789/supervisor/docker/manager.py#L1008-L1028

2. Docker Repair - I'm not seeing any reference to `prune` in aiodocker at all. Not sure if we can migrate that part of our repair process. Will look into it

https://github.com/home-assistant/supervisor/blob/05cf4ce73ffa8f88adf8992d6a5f178aa74f6789/supervisor/docker/manager.py#L699-L734

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/tests/backups/test_manager.py
+++ b/tests/backups/test_manager.py
@@ -1836,47 +1836,6 @@ async def test_reload_error(
         assert coresys.core.healthy is healthy_expected
 
 
-@pytest.mark.usefixtures("supervisor_internet", "install_addon_ssh")
-async def test_monitoring_after_full_restore(
-    coresys: CoreSys, full_backup_mock: Backup
-):
-    """Test monitoring of addon state still works after full restore."""
-    await coresys.core.set_state(CoreState.RUNNING)
-    coresys.hardware.disk.get_disk_free_space = lambda x: 5000
-    coresys.homeassistant.core.start = AsyncMock(return_value=None)
-    coresys.homeassistant.core.stop = AsyncMock(return_value=None)
-    coresys.homeassistant.core.update = AsyncMock(return_value=None)
-
-    manager = await BackupManager(coresys).load_config()
-
-    backup_instance = full_backup_mock.return_value
-    backup_instance.protected = False
-    assert await manager.do_restore_full(backup_instance)
-
-    backup_instance.restore_addons.assert_called_once_with([TEST_ADDON_SLUG])
-    assert coresys.core.state == CoreState.RUNNING
-    coresys.docker.unload.assert_not_called()
-
-
-@pytest.mark.usefixtures("supervisor_internet", "install_addon_ssh")
-async def test_monitoring_after_partial_restore(
-    coresys: CoreSys, partial_backup_mock: Backup
-):
-    """Test monitoring of addon state still works after full restore."""
-    await coresys.core.set_state(CoreState.RUNNING)
-    coresys.hardware.disk.get_disk_free_space = lambda x: 5000
-
-    manager = await BackupManager(coresys).load_config()
-
-    backup_instance = partial_backup_mock.return_value
-    backup_instance.protected = False
-    assert await manager.do_restore_partial(backup_instance, addons=[TEST_ADDON_SLUG])
-
-    backup_instance.restore_addons.assert_called_once_with([TEST_ADDON_SLUG])
-    assert coresys.core.state == CoreState.RUNNING
-    coresys.docker.unload.assert_not_called()
-
-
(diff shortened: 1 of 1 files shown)


supervisor/api/supervisor.py
```diff
@@ -248,6 +248,7 @@ def restart(self, request: web.Request) -> Awaitable[None]:
         return asyncio.shield(self.sys_supervisor.restart())
 
     @api_process_raw(CONTENT_TYPE_TEXT, error_type=CONTENT_TYPE_TEXT)
-    def logs(self, request: web.Request) -> Awaitable[bytes]:
+    async def logs(self, request: web.Request) -> bytes:
         """Return supervisor Docker logs."""
-        return self.sys_supervisor.logs()
+        logs = await self.sys_supervisor.logs()
+        return "\n".join(logs).encode(errors="replace")
```
agners, line 254: Ideally we would do some type of streaming but since this is almost exclusively used in fallback I think the performance downsides are acceptable.

mdegat01: Yea I think we have to leave this as is for now because its more important we match the signature of what is there. This is a backup source for logs, signature is defined by the primary source.

supervisor/docker/monitor.py
```diff
@@ -88,13 +112,38 @@ def run(self) -> None:
...
+
+        await self._event_tasks.put(None)
+
+    async def _await_event_tasks(self):
+        """Await event callback tasks to clean up and capture output."""
+        while (event := await self._event_tasks.get()) is not None:
+            try:
+                await event.task
+            except HassioError:
+                pass
```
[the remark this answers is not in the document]
mdegat01: If an exception is raised that inherits from `HassioError` that means whatever went wrong was already handled and wrapped into one of our internal types of exceptions. And generally always logged unless there was a good reason not to.

Considering this runs as an asynchronous agent there's nothing further to do with these types of errors, this catch is just to prevent them from failing through to the generic handling for unhandled types of exceptions. I'll add a comment explaining the pass and leave it at that.

supervisor/docker/monitor.py
```diff
@@ -47,24 +63,32 @@ def watch_container(self, container_metadata: dict[str, Any]):
...
+    async def _subscribe(self) -> None:
         """Monitor and process docker events."""
-        if not self._events:
-            raise RuntimeError("Monitor has not been loaded!")
+        events = self.docker.events.subscribe()
+
+        while True:
+            event: dict[str, Any] | None = await events.get()
+            if event is None:
+                break
```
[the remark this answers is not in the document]
mdegat01: Missed this one initially and think some context/suggestion got lost after the rebase. But I get what this is saying and yea that makes sense 👍

tests/backups/test_manager.py
```diff
@@ -1836,47 +1836,6 @@ def mock_is_dir(path: Path) -> bool:
...
-    manager = await BackupManager(coresys).load_config()
-
-    backup_instance = partial_backup_mock.return_value
-    backup_instance.protected = False
-    assert await manager.do_restore_partial(backup_instance, addons=[TEST_ADDON_SLUG])
-
-    backup_instance.restore_addons.assert_called_once_with([TEST_ADDON_SLUG])
-    assert coresys.core.state == CoreState.RUNNING
-    coresys.docker.unload.assert_not_called()
-
```
mdegat01, line 1878: These tests no longer work now that we aren't mocking `DockerAPI.unload` in all tests. But they really don't make any sense. Conceptually testing that the monitor still functions for an addon through a restore is sensible but they aren't testing that, they're testing if the entire docker manager was unloaded or not, which wouldn't happen.

We also can't actually test that the monitor works through a restore as is because the entire Addon object is a mock in these tests. So I'm just going to remove them.

PR #6520, opened 2026-02-03
Fix Docker exec exit code handling by using detach=False

### LINKED ISSUE #6518: Backup fails to include MAriaDB
### The problem

Repair says:

----

Not all add-ons could be included in automatic backup
Warning · Reported by Backup

Add-ons MariaDB could not be included in automatic backup. Please check the Supervisor logs for more information. Another attempt will be made at the next scheduled time if a backup schedule is configured.

----

My database is about 2.9GB in size and this is the first time this has happened. 

### What version of Home Assistant Core has the issue?

core-2026.2.0b1

### What was the last working version of Home Assistant Core?

core-2025.1.3

Home Assistant OS

### Integration causing the issue

Backup

### Link to integration documentation on our website

https://www.home-assistant.io/integrations/backup/

### Diagnostics information

_No response_

### Example YAML snippet

```yaml

```

### Anything in the logs that might be useful for us?

```txt
2026-02-03 02:45:15.738 ERROR (MainThread) [supervisor.addons.addon] Pre-/Post backup command for add-on core_mariadb returned error code: None. Please report this to the addon developer. Enable debug logging to capture complete command output using ha supervisor options --logging debug
```

### Additional information

_No response_

agners:
## Proposed change

This PR fixes an issue where Docker exec commands executed inside containers returned `None` as the exit code, causing add-on backup/restore commands to fail even when they succeeded.

**Root Cause:**
The `container_run_inside()` method was using `detach=True` when calling Docker's exec API. While this returns an awaitable, the `await` only waits for the HTTP/REST call to complete, NOT for the actual command execution to finish. The command continues running in the background, so when `inspect()` is called immediately after, the `ExitCode` field is still `None`.

**Solution:**
Switch to `detach=False` which returns a Stream object. Reading from this stream automatically waits for the exec to complete, ensuring the exit code is available when needed. As a bonus, we now capture the actual command output instead of just empty bytes.

**Changes:**
- Use `detach=False` with Stream-based output reading in `container_run_inside()`
- Add defensive validation to ensure `ExitCode` is never `None`
- Update test mocks to properly simulate the Stream interface using `AsyncMock`
- Add debug logging to show exit codes after command execution

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR fixes or closes issue: fixes #6518
- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/addons/addon.py
+++ b/supervisor/addons/addon.py
@@ -24,8 +24,6 @@ from securetar import AddFileError, atomic_contents_add, secure_path
 import voluptuous as vol
 from voluptuous.humanize import humanize_error
 
-from supervisor.docker.manager import ExecReturn
-
 from ..bus import EventListener
 from ..const import (
     ATTR_ACCESS_TOKEN,
@@ -63,6 +61,7 @@ from ..const import (
 from ..coresys import CoreSys
 from ..docker.addon import DockerAddon
 from ..docker.const import ContainerState
+from ..docker.manager import ExecReturn
 from ..docker.monitor import DockerContainerStateEvent
 from ..docker.stats import DockerStats
 from ..exceptions import (
--- a/supervisor/docker/manager.py
+++ b/supervisor/docker/manager.py
@@ -18,6 +18,7 @@ from typing import Any, Final, Self, cast
 import aiodocker
 from aiodocker.containers import DockerContainer, DockerContainers
 from aiodocker.images import DockerImages
+from aiodocker.stream import Stream
 from aiodocker.types import JSONObject
 from aiohttp import ClientTimeout, UnixConnector
 import attr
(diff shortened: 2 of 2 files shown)


supervisor/addons/addon.py
```diff
@@ -24,6 +24,8 @@
 import voluptuous as vol
 from voluptuous.humanize import humanize_error
 
+from supervisor.docker.manager import ExecReturn
```
sairon: This should be a relative import - we got rid of absolute ones in #6446.

supervisor/docker/manager.py
```diff
@@ -967,17 +967,41 @@ async def container_run_inside(self, name: str, command: str) -> ExecReturn:
                 f"Can't get container {name} to run command: {err!s}"
             ) from err
 
-        # Execute
+        # Execute - use detach=False to wait for completion and capture output
         try:
             docker_exec = await docker_container.exec(command)
-            output = await docker_exec.start(detach=True)
+            # start() with detach=False returns a Stream object (not a coroutine)
+            stream = docker_exec.start(detach=False)
```
sairon: Nit, but typehint would be better than just a comment, as the method returns `Any`. `from aiodocker.stream import Stream` would also have to be added.
```suggestion
            stream: Stream = docker_exec.start(detach=False)
```

Issue #6543, opened 2026-02-09
Create Enum base classes tolerant of missing values for D-Bus/Docker API usage


agners:
### Description

Today we use `IntEnum` and `StrEnum` for various D-Bus values (e.g. `ConnectivityState`,  `MulticastProtocolEnabled` or `RaucState`) and Docker API values (e.g. `RestartPolicy`). We don't use explicit type conversion for values since a missing value would lead to exceptions potentially breaking Supervisor functionality.

Let's try to use a Supervisor specific base class which accepts missing values and report those to Sentry.

agners:
With #6545 merged I think we have in place what this task requires. Closing.

PR #6533, opened 2026-02-05
Add explicit StrEnum conversions for D-Bus and Docker properties

agners:
Convert string values from D-Bus and Docker API to their respective StrEnum types explicitly to satisfy typeguard runtime type checking.

The conversions will raise ValueError if an unknown enum value is encountered, this will be caught in Sentry. Since D-Bus and the particular property from Docker API are not mission critical for Supervisor update functionality, it will be possible to add new enum values in the future as they arise.

## Type of change

- [ ] Dependency upgrade
- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (which adds functionality to the supervisor)
- [ ] Breaking change (fix/feature causing existing functionality to break)
- [x] Code quality improvements to existing code or addition of tests

## Additional information

- This PR fixes or closes issue: fixes #
- This PR is related to issue:
- Link to documentation pull request:
- Link to cli pull request:
- Link to client library pull request:

## Checklist

- [ ] The code change is tested and works locally.
- [ ] Local tests pass. **Your PR cannot be merged unless tests pass**
- [ ] There is no commented out code in this PR.
- [ ] I have followed the [development checklist][dev-checklist]
- [ ] The code has been formatted using Ruff (`ruff format supervisor tests`)
- [ ] Tests have been added to verify that the new code works.

If API endpoints or add-on configuration are added/changed:

- [ ] Documentation added/updated for [developers.home-assistant.io][docs-repository]
- [ ] [CLI][cli-repository] updated (if necessary)
- [ ] [Client library][client-library-repository] updated (if necessary)

[dev-checklist]: https://developers.home-assistant.io/docs/en/development_checklist.html
[docs-repository]: https://github.com/home-assistant/developers.home-assistant
[cli-repository]: https://github.com/home-assistant/cli
[client-library-repository]: https://github.com/home-assistant-libs/python-supervisor-client/

PR #6545, opened 2026-02-09
Add D-Bus tolerant enum base classes to prevent crashes on unknown values

agners:
## Proposed change

D-Bus services (systemd, NetworkManager, RAUC, UDisks2) can introduce new enum values at any time via OS updates. Standard Python enum construction raises `ValueError` for unknown values, which would crash the Supervisor.

This PR introduces `DBusStrEnum` and `DBusIntEnum` base classes that use Python's `_missing_` hook to create pseudo-members for unknown values. These pseudo-members:

- Pass `isinstance` checks (satisfying typeguard runtime type checking)
- Preserve the original value (`str(x)`, `int(x)`)
- Don't pollute `list(MyEnum)` or `__members__`
- Log at WARNING level and report to Sentry (deduplicated per class+value)

17 D-Bus enums are migrated to the new base classes. Enums only sent TO D-Bus (`StopUnitMode`, `StartUnitMode`) and bitmask/socket enums (`ConnectionStateFlags`, `DNSAddressFamily`) are left unchanged.

Additionally, the manual `try/except` workaround in `NetworkInterface.type` is removed (now handled automatically), and 6 systemd-resolved properties gain explicit enum conversions that are now safe to add.

Supersedes #6533 for the D-Bus portion (Docker `RestartPolicy` conversion is not included).

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR is related to issue: #6543
- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/utils/sentry.py
+++ b/supervisor/utils/sentry.py
@@ -95,7 +95,7 @@ def fire_and_forget_capture_message(
     def _capture() -> None:
         try:
             sentry_sdk.capture_message(msg, level=level)
-        except Exception:
+        except Exception:  # pylint: disable=broad-except
             _LOGGER.debug("Failed to send message to Sentry: %s", msg)
 
     try:


supervisor/dbus/enum.py
```diff
@@ -0,0 +1,56 @@
...
+    """IntEnum that tolerates unknown values from D-Bus."""
+
+    @classmethod
+    def _missing_(cls, value: object) -> "DBusIntEnum | None":
+        if not isinstance(value, int):
+            return None
+        _report_unknown_value(cls, value)
+        obj = int.__new__(cls, value)
+        obj._name_ = f"UNKNOWN_{value}"
+        obj._value_ = value
```
mdegat01, line 55: This is pretty creative. I didn't know you could add pseudo members like this to handle unknowns but still keep typing intact 👍

supervisor/utils/sentry.py
```diff
@@ -78,6 +79,28 @@ async def async_capture_exception(err: BaseException) -> None:
...
+    if not sentry_sdk.is_initialized():
+        return
+    try:
+        loop = asyncio.get_running_loop()
+    except RuntimeError:
+        sentry_sdk.capture_message(msg, level=level)
+    else:
+        loop.run_in_executor(
+            None, partial(sentry_sdk.capture_message, msg, level=level)
+        )
```
mdegat01: If this raises its going to end up with an ugly "coroutine wasn't awaited and raised error..." dump in the log. And since its calling a network API that can definitely happen. For a fire and forget style we might want to wrap this in something that handles all like:

```py
        def capture_message_suppress_errors():
            try:
                loop.run_in_executor(
                    None, partial(sentry_sdk.capture_message, msg, level=level)
                )
            except Exception as err:
                _LOGGER.warning("Could not report message to sentry: %s, %s", msg, err)

        loop.run_in_executor(None, capture_message_suppress_errors)
```
Or literally just use `with suppress(Exception):` if we don't even want the logging.

tests/dbus/test_enum.py
```diff
@@ -0,0 +1,306 @@
...
+    """Sample IntEnum for testing."""
+
+    ONE = 1
+    TWO = 2
+
+
+# -- DBusStrEnum tests --
+
+
+class TestDBusStrEnum:
```
mdegat01: My understanding is TestClasses like this is a legacy style. We generally don't do this anymore, we just write funcitons that start with `test_` and pytest discovers them. I don't think its harmful necessarily if you don't want to re-write it but we do avoid doing it elsewhere.

supervisor/dbus/enum.py
```diff
@@ -0,0 +1,56 @@
...
+import logging
+
+from ..utils.sentry import fire_and_forget_capture_message
+
+_LOGGER: logging.Logger = logging.getLogger(__name__)
+
+_reported: set[tuple[str, object]] = set()
+
+
+def _report_unknown_value(cls: type, value: object) -> None:
```
mdegat01: When I read this typing my initial concern was allowing any object was a bad idea here since we risk memory leak build-up in `_reported` and sentry won't accept something non-serializable anyway. But in practice value can only be `string | int` based on below code. Could re-type it to match that although if you want to leave it to match signature of value in `_missing_` I suppose that also makes sense.

mdegat01 (changes requested):
I think we should suppress errors as I suggested in the one feedback below because a network API will definitely raise once in a while and that'll leave an ugly log the way it is now. The others are more nitpicks you can decide whether you like or want to ignore.

supervisor/utils/sentry.py
agners: `loop.run_in_executor` returns a future, and if we'd were to await that it would raise the Exception. However, since we don't await it, your suggested code does not handle the exception.

However, this does:

```py
  def _capture() -> None:                                                                                                                                                                                                         
      try:                   
          sentry_sdk.capture_message(msg, level=level)
      except Exception:
          _LOGGER.debug("Failed to send message to Sentry: %s", msg)

  loop.run_in_executor(None, _capture)
```

FWIW, since #5697 Supervisor captures warnings in Sentry, including the "coroutine was never awaited". But since this is during a Sentry capture, it does not make sense to rely on that mechanism. So I agree, capturing all Exceptions when ceiling `sentry_sdk.capture_message` make sense here.

supervisor/dbus/enum.py
agners: Claude came up with this, and I was actually considering removing it again: Sentry deduplicates reports on its own as well.

But then, it might actually lead to quite some messages, so limit them on Client side seems not a bad idea.

> But in practice value can only be `string | int` based on below code. Could re-type it to match that although if you want to leave it to match signature of value in `_missing_` I suppose that also makes sense.

Seems a good idea, with the `isinstance` type check below it will work without explicit casting.

tests/dbus/test_enum.py
agners: Hm, Claude always reverts to these class style. #6552 adjusts the `copilot-instructions.md`/`CLAUDE.md` accordingly.

supervisor/utils/sentry.py
```diff
@@ -78,6 +79,33 @@ async def async_capture_exception(err: BaseException) -> None:
...
+    Safe to call from sync code running in the event loop. The executor future
+    is intentionally not awaited (fire-and-forget).
+    """
+    if not sentry_sdk.is_initialized():
+        return
+
+    def _capture() -> None:
+        try:
+            sentry_sdk.capture_message(msg, level=level)
+        except Exception:
```
agners:
```suggestion
        except Exception:  # pylint: disable=broad-except
```

supervisor/utils/sentry.py
mdegat01: right my bad, I didn't mean to have the run in executor inside the run in executor function. Meant to put what you wrote but forgot to adjust the copy paste part🤦

mdegat01 (approved):
LGTM 👍

PR #6553, opened 2026-02-11
Simplify HomeAssistantWebSocket and raise on connection errors

agners:
## Proposed change

Refactor `HomeAssistantWebSocket` to simplify the confusing "message" vs "command" distinction and properly surface connection errors.

Previously, `async_send_command` silently returned `None` when Core was unreachable (via `_can_send` returning `False`), causing callers to get misleading errors like *"Home Assistant returned invalid response of 'None' instead of a list of users"*. The "message" vs "command" split also didn't reflect Core's architecture where everything is just a WS command.

**Key changes:**

- `_can_send` → `_ensure_connected`: Only handles connection logic, raises `HomeAssistantWSError` on failure instead of returning `False`
- `async_send_command`: Now properly raises on connection errors so callers can handle them appropriately
- `async_send_message` / `send_message` → `_async_send_command` / `send_command`: Aligned naming with Core's nomenclature (everything is a "command")
- Removed dead `WSClient.async_send_message` method (was never called)
- Removed `MIN_VERSION` gating — minimum supported Core is now 2024.2+ (per #6148), well past all gated versions (2021.2.4, 2022.1.0)
- Added proper error handling in callers (`_hardware_events`, `tasks.py` addon updates)

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/.github/workflows/builder.yml
+++ b/.github/workflows/builder.yml
@@ -293,16 +293,10 @@ jobs:
       - name: Start the Supervisor
         run: docker start hassio_supervisor
 
-      - name: Get Supervisor IP
-        id: supervisor
-        run: |
-          SUPERVISOR=$(docker inspect --format='{{.NetworkSettings.IPAddress}}' hassio_supervisor)
-          echo "ip=${SUPERVISOR}" >> "$GITHUB_OUTPUT"
-
       - &wait_for_supervisor
         name: Wait for Supervisor to come up
         run: |
-          SUPERVISOR=${{ steps.supervisor.outputs.ip }}
+          SUPERVISOR=$(docker inspect --format='{{.NetworkSettings.IPAddress}}' hassio_supervisor)
           echo "Waiting for Supervisor API at http://${SUPERVISOR}/supervisor/ping"
           timeout=300
           elapsed=0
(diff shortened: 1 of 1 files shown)


agners:
Hm, it seems this actually uncovered an issue during backup. The Supervisor integration tests fail when trying to create a backup, the backup creation now fails with:


```
2026-02-11T15:07:05.9417105Z [31m2026-02-11 09:07:05.393 ERROR (MainThread) [supervisor.backups.manager] Backup 886dd571 error: Preparing backup of Home Assistant Core failed. Failed to inform HA Core: Can't connect to Home Assistant Core WebSocket, the API is not reachable.[0m
```

This is because Core is not yet ready at this point. It seems so far we simply created a backup and silently ignored the fact that Core was not aware of the backup being taken (`backup/start` WebSocket message not being delivered).

supervisor/homeassistant/websocket.py
```diff
@@ -212,92 +197,88 @@ async def _get_ws_client(self) -> WSClient:
...
-        return None
 
-    def send_message(self, message: dict[str, Any]) -> None:
-        """Send a supervisor/event message."""
+    def send_command(self, message: dict[str, Any]) -> None:
+        """Send a fire-and-forget command via WebSocket."""
         if self.sys_core.state in CLOSING_STATES:
             return
-        self.sys_create_task(self.async_send_message(message))
+        self.sys_create_task(self._async_send_command(message))
```
mdegat01, line 274: Since we were already leaving all these dangling tasks this isn't new behavior and not going to request a change for it. But if we're doing clean up here, could be a time to address this. We could leverage a similar pattern to what I did in the `DockerMonitor`:

1. `load` creates an `asyncio.Queue` of tasks and a run forever waiter task
2. Waiter task just keeps pulling tasks out of queue and awaiting them with a `while True`, nicely logging and/or reporting to sentry any errors it finds. Exits the loop when it gets `None` from the queue
3. All these fire and forget methods add the task they create to the queue with a `put_nowait`. Suppress any queue full errors and let it dangle in that edge case
4. Add an `unload` which just puts `None` in the queue and awaits the waiter task to let it end. Have that called from `supervisor.core.Core.stop` as part of Supervisor stop tasks

mdegat01: One thing I found out with the docker monitor changes is this approach is also really nice for testing. In order to test stuff that is doing fire and forget with dangling tasks you basically have no choice but to rely on `asyncio.sleep` calls to give it time to complete these tasks before proceeding to your assertions.

If you have a task waiting loop like that then you can just await the unload after you've called the code under test. That only returns once all the tasks have been completed so then you can assert that was done correctly. No sleep guessing.

agners: I have some more refactoring ideas which also eliminate calls and/or wait for the result. I don't see a good reason why we use fire-and-forget today.

agners:
> I guess its ok because its only caller handles `HomeAssistantAPIError` today and `HomeAssistantWSError` inherits from that. But it does feel a bit odd that this one method forces callers to handle the exception when the others don't.

The code already was like this. We also have a second function which calls the same WebSocket API: `list_users` in `supervisor/auth.py`. I'll refactor that in a follow up.

PR #6558, opened 2026-02-12
Unify Core user handling with HomeAssistantUser model

agners:
## Proposed change

Unify Core user handling around a single `HomeAssistantUser` dataclass, replacing
the split between `IngressSessionDataUser` (ingress) and raw `dict[str, Any]` (auth).

**`HomeAssistantUser` dataclass** (`const.py`): Models the fields Supervisor actually
uses from Core's `config/auth/list` response. Replaces `IngressSessionDataUser` and
the untyped dicts that `Auth.list_users()` returned.

**`HomeAssistant.list_users()`** (`homeassistant/module.py`): New single entry point.
Replaces the old `get_users()` (which was `@Job`-throttled and returned
`IngressSessionDataUser`) and the inline WS call in `Auth.list_users()`. Both callers
now go through this method. Core's `config/auth/list` always returns a list, never
`None` — since #6553 the WS layer raises on connection errors instead of returning
`None`, so the old `if not raw: return []` guard was dead code hiding a meaningful
distinction. Replaced with an assert.

**`display_name` removal**: The `display_name` field was introduced in #4152 but was
never part of Core's `config/auth/list` response — Core returns `name`, not
`display_name`. The ingress header `X-Remote-User-Display-Name` now correctly reads
from `user.name`. Legacy session data stored with the old `displayname` key is still
handled via `from_dict()` migration.

**`Auth.list_users()`** (`auth.py`): Now delegates to `HomeAssistant.list_users()` and
returns typed `list[HomeAssistantUser]` instead of `list[dict[str, Any]]`.
`AuthListUsersNoneResponseError` removed as dead code (see above).

**Ingress** (`api/ingress.py`): `_find_user_by_id()` uses the new
`list_users()`. On WS failure, falls back to the cached user list instead of returning
`None`.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/api/ingress.py
+++ b/supervisor/api/ingress.py
@@ -75,12 +75,6 @@ def status_code_must_be_empty_body(code: int) -> bool:
 class APIIngress(CoreSysAttributes):
     """Ingress view to handle add-on webui routing."""
 
-    _list_of_users: list[HomeAssistantUser]
-
-    def __init__(self) -> None:
-        """Initialize APIIngress."""
-        self._list_of_users = []
-
     def _extract_addon(self, request: web.Request) -> Addon:
         """Return addon, throw an exception it it doesn't exist."""
         token = request.match_info["token"]
@@ -309,13 +303,12 @@ class APIIngress(CoreSysAttributes):
     async def _find_user_by_id(self, user_id: str) -> HomeAssistantUser | None:
         """Find user object by the user's ID."""
         try:
-            self._list_of_users = await self.sys_homeassistant.list_users()
+            users = await self.sys_homeassistant.list_users()
         except HomeAssistantAPIError as err:
-            _LOGGER.error(
-                "%s error occurred while requesting list of users: %s", type(err), err
-            )
+            _LOGGER.warning("Could not fetch list of users: %s", err)
+            return None
 
-        return next((user for user in self._list_of_users if user.id == user_id), None)
+        return next((user for user in users if user.id == user_id), None)
 
 
 def _init_header(


supervisor/api/ingress.py
```diff
@@ -306,18 +306,14 @@ async def _handle_request(
...
         """Find user object by the user's ID."""
         try:
-            list_of_users = await self.sys_homeassistant.get_users()
-        except (HomeAssistantAPIError, TypeError) as err:
+            self._list_of_users = await self.sys_homeassistant.list_users()
+        except HomeAssistantAPIError as err:
             _LOGGER.error(
                 "%s error occurred while requesting list of users: %s", type(err), err
             )
-            return None
```
mdegat01, line 317: So the challenge here is we're now using the cache in a new use case. The `config/auth/list` API never returned `None` as a successful response so the only time we'd use our cache was when the old `_can_send` method returned `False`. Which meant Supervisor was unable to talk to core for one of the reasons in there. Now in addition to those we'll also be using the cache when the `config/auth/list` API responds with an error.

More usage of the cache means increased chance of a stale cache issue. Where we allow a client through that was supposed to have their access revoked but that wasn't reflected in our cache. Maybe we should simply raise and respond with an error here rather then using the cache? Or change `_ensure_connected` to raise a new type of error that inherits from `HomeAssistantWSError` but is distinct so we can continue to use the cache in that case but not when the `config/auth/list` API actually responds with an error.

Idk its probably ok but since this is auth code it makes me a little nervous to change how we're using a cache.

supervisor/homeassistant/module.py
```diff
@@ -570,21 +567,13 @@ def _restore_home_assistant() -> Any:
...
+        Raises HomeAssistantWSError on WebSocket connection/communication failure.
+        """
+        raw: list[dict[str, Any]] | None = await self.websocket.async_send_command(
             {ATTR_TYPE: "config/auth/list"}
         )
-
-        if list_of_users:
-            return [IngressSessionDataUser.from_dict(data) for data in list_of_users]
-        return []
+        assert raw is not None
```
mdegat01: Same question as the other code, is this a real use case? If so, in what situation does it respond without error and `None` as the response? Ideally we would handle this situation by raising an error that inherits from `HassioError` here and better describes what happened rather then an unhandled `AssertionError`.

mdegat01: Why did we keep the `| None` typing on `async_send_command`? I'm not seeing a situation where we ever return `None` unless the API we're calling has that a a possible response. Could we change the signature to just `T` and then drop this unnecessary check?

mdegat01 (approved):
Approved but please take a look at the comments and see if you want to act on them. Changing how we use that cache in particular makes me a bit nervous.

supervisor/api/ingress.py
agners: > Now in addition to those we'll also be using the cache when the `config/auth/list` API responds with an error.

I am a bit confused what you mean by that, #6553 largely replaced `None` with `HomeAssistantWSError`, and with this change this essentially leads to the same outcome: Use cached list_of_users.

That said, I am not sure if caching is a good idea here in general. I think we should just fail to resolve a user id in case we can't get the list of users. 🤔

agners: Oh I see, the WebSocket command can raise an exception if we have an issue on Core side, and its currently `HomeAssistantWSError`, the same as raised in `_ensure_connected`. Yeah I think that's not ideal. There is `HomeAssistantWSConnectionError`, I think it can work for this case.

supervisor/homeassistant/module.py
agners: Agreed, we can remove the `| None` typing now.

supervisor/api/ingress.py
agners: > More usage of the cache means increased chance of a stale cache issue. Where we allow a client through that was supposed to have their access revoked but that wasn't reflected in our cache. Maybe we should simply raise and respond with an error here rather then using the cache?

I've followed that approach. The cache doesn't seem to add a performance benefit, and the distinction of caching on connection error and not caching on WebSocket errors originating from Core seems rather arbitrary. I also did not found anything in the discussion when this got introduced in https://github.com/home-assistant/supervisor/pull/4152.

> Or change `_ensure_connected` to raise a new type of error that inherits from `HomeAssistantWSError` but is distinct so we can continue to use the cache in that case but not when the `config/auth/list` API actually responds with an error.

I've still implemented this change: It feels more fitting that `_ensure_connected` uses `HomeAssistantWSConnectionError`.

mdegat01: So the only use case I could think of here is the main reason `_can_send` used to return `None` was when core was shutting down. And that's kind of the only use case I can think of when using the cache made sense, because nothing is modified during shutdown so we probably have up to date info.

But I just checked and the only place this is used is the `/auth/list` API which is only called from the host shell to help users reset lost passwords. So handling errors by always raising and never using the cache seems safe and preferable 👍

PR #6559, opened 2026-02-12
Harden backup tar extraction with Python `tar_filter`

agners:
## Proposed change

Replace `filter="fully_trusted"` in all backup tar extraction sites with
Python's built-in [`filter="tar"`](https://docs.python.org/3/library/tarfile.html#tarfile.tar_filter).

The `fully_trusted` filter was added in 2da27937a purely to suppress Python
3.12 deprecation warnings — it's a no-op that passes all members unchanged.

Importantly, the `tar` filter **preserves uid/gid and file permissions**,
which is critical for add-ons that run as non-root users. The stricter `data`
filter resets uid/gid and enforces minimum permissions, which would break
file ownership after restore.

This also removes the now-redundant `secure_path` member filtering from all
extraction sites. The `tar` filter is strictly stronger than `secure_path`
for name validation (`os.path.realpath` is stateful vs `secure_path`'s purely
lexical check against `/fake`). Note that `_folder_restore` was not using
`secure_path` at all — it passed the raw tar iterator with no member
filtering.

**Important**: this is a defense-in-depth improvement, not a full security
boundary. Backups should still be considered trusted input — apps and custom
integrations have broad system access and can include arbitrary file content in
backed-up directories.

**Note:** A first revision of this PR used the `data_filter`. But this proved to be too invasive (as it stops restoring user/group permission settings).

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/tests/backups/test_backup_security.py
+++ b/tests/backups/test_backup_security.py
@@ -9,6 +9,7 @@ from securetar import SecureTarFile
 
 from supervisor.addons.addon import Addon
 from supervisor.backups.backup import Backup
+from supervisor.backups.const import BackupType
 from supervisor.coresys import CoreSys
 from supervisor.exceptions import BackupInvalidError
 
@@ -210,14 +211,13 @@ async def test_addon_restore_rejects_symlink_escape(
 async def test_folder_restore_rejects_path_traversal(
     coresys: CoreSys, tmp_supervisor_data: Path
 ):
-    """Test that folder restore raises BackupInvalidError for path traversal."""
+    """Test that folder restore rejects path traversal in backup tar."""
     traversal_info = tarfile.TarInfo(name="../../etc/passwd")
     traversal_info.size = 9
 
     # Create backup with a malicious share folder tar inside
     backup_tar_path = tmp_supervisor_data / "backup.tar"
     with tarfile.open(backup_tar_path, "w:") as outer_tar:
-        # Create a malicious share.tar.gz and add it to the backup
         share_tar_path = tmp_supervisor_data / "share.tar.gz"
         _create_tar_gz(
             share_tar_path, [traversal_info], {"../../etc/passwd": b"malicious"}
@@ -225,16 +225,15 @@ async def test_folder_restore_rejects_path_traversal(
         outer_tar.add(share_tar_path, arcname="./share.tar.gz")
 
     backup = Backup(coresys, backup_tar_path, "test", None)
-    backup._data["compressed"] = True
+    backup.new("test", "2025-01-01", BackupType.PARTIAL, compressed=True)
     async with backup.open(None):
-        with pytest.raises(BackupInvalidError):
-            await backup._folder_restore("share")
+        assert await backup.restore_folders(["share"]) is False
 
 
 async def test_folder_restore_rejects_symlink_escape(
     coresys: CoreSys, tmp_supervisor_data: Path
 ):
-    """Test that folder restore raises BackupInvalidError for symlink escape."""
+    """Test that folder restore rejects symlink escape in backup tar."""
     link_info = tarfile.TarInfo(name="escape")
     link_info.type = tarfile.SYMTYPE
     link_info.linkname = "../outside"
@@ -253,7 +252,6 @@ async def test_folder_restore_rejects_symlink_escape(
         outer_tar.add(share_tar_path, arcname="./share.tar.gz")
 
     backup = Backup(coresys, backup_tar_path, "test", None)
-    backup._data["compressed"] = True
+    backup.new("test", "2025-01-01", BackupType.PARTIAL, compressed=True)
     async with backup.open(None):
-        with pytest.raises(BackupInvalidError):
-            await backup._folder_restore("share")
+        assert await backup.restore_folders(["share"]) is False


supervisor/backups/utils.py
```diff
@@ -10,3 +14,19 @@ def create_slug(name: str, date_str: str) -> str:
...
+    """Filter for backup tar extraction.
+
+    Applies tarfile.data_filter for security (rejects dangerous symlinks,
+    device nodes, resets uid/gid) but skips problematic entries with a
+    warning instead of aborting the entire extraction.
+    """
+    try:
+        return tarfile.data_filter(member, dest_path)
+    except tarfile.FilterError as err:
+        _LOGGER.warning("Skipping %s: %s", member.name, err)
```
mdegat01: Should we continue in this situation rather then abort? The obvious use case I can think of here is a user was handed a manipulated backup that is using the restore process as an attack vector on their system. In which case we'd want to abort.

And if there's a legit use case where an addon or something includes files which fail this filter in its backup then its probably not going to run successfully without them right? So we should also probably abort in that case as well.

What's the situation where the desired behavior would be to restore everything else?

agners: So my fear is that we break currently working backups.

Looking at [data_filter documentation](https://docs.python.org/3/library/tarfile.html#tarfile.data_filter), it seems there are two cases where an exceptions is thrown:

* Refuse to extract links (hard or soft) that link to absolute paths, or ones that link outside the destination.
  This raises `AbsoluteLinkError` or `LinkOutsideDestinationError`.
  Note that such files are refused even on platforms that do not support symbolic links.
*  Refuse to extract device files (including pipes). This raises `SpecialFileError`.

Previously we simply printed a warning and skipped those files, so this implements the same.

> And if there's a legit use case where an addon or something includes files which fail this filter in its backup then its probably not going to run successfully without them right? So we should also probably abort in that case as well.

True, or even worse, it works somewhat but the user only realizes later that something is wrong.

The question is, would we make the whole backup fail (raise an `BackupError` exception which aborts it all) or only cause a partial error (specific to the inner tar)? 🤔

emontnemery: I think we should only reject restoring what we reject backing up.

There's obviously no way to back up files outside of what we back up, so we should reject any files with absolute paths or files which otherwise will end up outside the restore destination, for example because they include `..` to escape the destination.

I'm not at all convinced it's a good idea to mess with permissions or rejecting non regular files.

Maybe the built in [`tar_filter`](https://docs.python.org/3/library/tarfile.html#tarfile.tar_filter) is a better choice?
That filter mainly guards against files escaping the destination, which seems to be the relevant threat?

agners:
> What is the problem that we're trying to guard against?

As the PR description says, it's meant as a defense-in-depth approach. I trust the Python developers to have implemented something sensible to use (the [data filter](https://docs.python.org/3/library/tarfile.html#tarfile.data_filter)) :shrug: Instead of implement our own filter it does seem like a good idea to use a built-in feature, which likely got much more scrutiny.

If this breaks actual use cases it might be to defensive for our case. That is why I put up this PR: Let's have a discussion what/if this is too much.

supervisor/backups/utils.py
agners: > I'm not at all convinced it's a good idea to mess with permissions or rejecting non regular files.

Hm, yeah after reconsidering, not restoring permissions can actually be problematic: Add-ons can (and some do) run with user permissions. the `data_filter` strips user/group permission and restores files as the user id extracting the backup, which in Supervisor's case is `root`. So the files won't be writeable anymore :cry: 

> Maybe the built in [`tar_filter`](https://docs.python.org/3/library/tarfile.html#tarfile.tar_filter) is a better choice?

It is closer to what we have today, but still strips some permission bits, namely:

> * Clear high mode bits (setuid, setgid, sticky) and group/other write bits ([S_IWGRP](https://docs.python.org/3/library/stat.html#stat.S_IWGRP) | [S_IWOTH](https://docs.python.org/3/library/stat.html#stat.S_IWOTH)).

The first two are really useful for system level executable, which do not belong into a (data) backup. The sticky bit could be used in a shared directory, but I don't think it is commonly used.

mdegat01: > The first two are really useful for system level executable, which do not belong into a (data) backup. 

We probably have to be very loose with the term data here. The SSH addons for example are very low level and customizable. Users could definitely be including custom executables they want used in their startup/shutdown/login scripts. I also personally use an addon called Caddy 2 for reverse proxy that allows me to build my own custom binary using their tooling instead of using the one in the image if you want to include their plugins and things.

I agree with the assessment above, messing with permissions at all here seems like a bad idea.

tests/backups/test_backup_security.py
```diff
@@ -0,0 +1,139 @@
...
+
+    dest = tmp_path / "out"
+    dest.mkdir()
+    with tarfile.open(tar_path, "r:gz") as tar:
+        try:
+            tar.extractall(path=dest, filter="tar")
+        except tarfile.OutsideDestinationError:
+            pass
+        else:
+            raise AssertionError("Expected OutsideDestinationError")
```
emontnemery: Why don't you use `pytest.raises`?

supervisor/addons/addon.py
```diff
@@ -1444,10 +1444,11 @@ def _extract_tarfile() -> tuple[TemporaryDirectory, dict[str, Any]]:
             tmp = TemporaryDirectory(dir=self.sys_config.path_tmp)
             try:
                 with tar_file as backup:
+                    # The tar filter rejects path traversal and absolute names,
+                    # aborting restore of potentially crafted backups.
```
emontnemery: Maybe like this?
Change also in the other cases.

```suggestion
                    # aborting restore of malicious backups with such exploits.
```

supervisor/addons/addon.py
```diff
@@ -1444,10 +1444,11 @@ def _extract_tarfile() -> tuple[TemporaryDirectory, dict[str, Any]]:
...
                     backup.extractall(
```
emontnemery, line 1450: Where are we catching errors? Same question for the other locations where we extract

agners: The exceptions like `tarfile.OutsideDestinationError` are derrived from `tarfile.FilterError`, which in turn is derrived from `tarfile.TarError`. It seems `tarfile.TarError` is handled explicitly in all cases but the outer tar. I'll add it there too.

agners: I've went a bit different direction by handling `tarfile.FilterError` explicitly, turning them into `BackupInvalidError. This should it a bit clearer that it is backup file related problem.

emontnemery:
Codecov says coverage of the modified code is poor. Maybe we should add some more tests?

PR #6568, opened 2026-02-13
Cache Core config on startup and validate on Supervisor restart

agners:
## Proposed change

Cache Home Assistant Core's config (`get_config()`) once when Core reaches
the RUNNING state, instead of fetching it on every use.

Previously, `get_config()` was called via REST API during update verification
(to check "frontend" in components) and via WebSocket during hardware events
(to check "usb" in components). This PR fetches the config once in
`_block_till_run()` when Core is detected as running and caches it for
subsequent use.

A new `ensure_started()` method is added to `HomeAssistantCore` which handles
the Supervisor restart case — when the Supervisor container restarts but Core
is already running, it calls `_block_till_run()` to validate API readiness and
populate the cached config. Without this, the cache would remain empty after a
Supervisor restart since `start()` is never called.

I considered using the Docker container state event watchdog to detect when
Core is already running (the `attach()` during `load()` fires a RUNNING event),
but it seemed not fitting for this change. Probably best to converge all the
logic around Core "up" detection at one point in a future refactor.

**Cache lifecycle:**
- Cleared at the top of `_block_till_run()` and in `stop()`
- Populated in `_block_till_run()` after Core reports RUNNING state
- Used by update verification and hardware event handling
- Non-fatal if the fetch fails (suppressed, cache stays None)

**Trade-off:** If a user installs a new integration (e.g., USB) at runtime,
the cache won't reflect it until the next Core restart. This is acceptable
since USB discovery typically requires a restart anyway, and the cache
eliminates a WebSocket/API round-trip on every hardware event.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/homeassistant/core.py
+++ b/supervisor/homeassistant/core.py
@@ -77,7 +77,7 @@ class HomeAssistantCore(JobGroup):
         super().__init__(coresys, JOB_GROUP_HOME_ASSISTANT_CORE)
         self.instance: DockerHomeAssistant = DockerHomeAssistant(coresys)
         self._error_state: bool = False
-        self._core_config: dict[str, Any] | None = None
+        self._cached_core_config: dict[str, Any] | None = None
         self._watchdog_listener: EventListener | None = None
 
     @property
@@ -86,9 +86,13 @@ class HomeAssistantCore(JobGroup):
         return self._error_state
 
     @property
-    def core_config(self) -> dict[str, Any] | None:
-        """Return cached Core config or None if not available."""
-        return self._core_config
+    def cached_core_config(self) -> dict[str, Any] | None:
+        """Return Core config captured when Core reached RUNNING state.
+
+        This is a snapshot from startup and does not reflect integrations
+        loaded at runtime (e.g. a user installing an integration after boot).
+        """
+        return self._cached_core_config
 
     async def ensure_started(self) -> None:
         """Ensure Home Assistant Core is running and setup is complete.
@@ -347,7 +351,7 @@ class HomeAssistantCore(JobGroup):
             await _update(to_version)
 
         if not self.error_state and rollback:
-            config = self._core_config
+            config = self.cached_core_config
 
             # Verify that the frontend is loaded
             if config is None:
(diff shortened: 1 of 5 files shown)


agners:
I was hoping this would simplify things more, but instead it added another wrinkle to Core lifecycle management :see_no_evil: . At this point I am not sure yet if that make sense, any opinion welcome.

agners:
In discussion with @mdegat01 we should be careful that config can be incomplete (since integrations can be loaded after Home Assistant startup).

supervisor/homeassistant/core.py
```diff
@@ -548,6 +573,10 @@ async def _block_till_run(self) -> None:
                 if state.core_state == "RUNNING":
                     _LOGGER.info("Detect a running Home Assistant instance")
                     self._error_state = False
+                    with suppress(HomeAssistantAPIError):
+                        self._cached_core_config = (
+                            await self.sys_homeassistant.api.get_config()
+                        )
```
mdegat01, line 579: So this means that if this call fails then we simply have no core config available until the next time `_block_till_run` is called? Which presumably will be the next time core is updated or restarted right? And no hardware events will be processed until then because the cache is `None`. That seems problematic 🤔 

I'd suggest an alternate pattern here. Rather then having a get config method and a cache property like this, make the get config method smarter. Lets add a `get_config` method to this class that only asks Home Assistant for the config if its not already cached, otherwise it simply returns the cache. Then the other callers just use it and don't care whether they get cached results or not.

This also allows us to make other nice enhancements, such as:
1. Always using WSS for this and only making a call to the REST API as a fallback option if the WSS connection is down (so all places we get HA config in the code do the same thing)
2. Adding a listener to this class as a future enhancement to listen for config changes and automatically invalidate its cache so the next check pulls the up to date config

agners: The call is made right after we successfully connected through the API. Chances are relatively low that this call will fail.

But even if so, something is probably wong, and I'd rather have no config and assume things are not loaded than making wrong assumptions about what is going on on Core side...

Anyhow, I realized that the USB events are no longer used in Core since https://github.com/home-assistant/core/pull/136676. So I wonder if we should just remove the logic.

mdegat01: Yea seems like the right move. Core's not going to return to use these since they found a faster way that works on all installation types.

agners:
Closing for the discussed reasons.

PR #6582, opened 2026-02-19
Ensure uuid of dismissed suggestion/issue matches an existing one

### LINKED ISSUE #6193: Reload Network storage doesn't work

Scenario: I have a network storage configured as CIFS to expand HA storage for backup. All is working ok until I restart the NAS. 
Correctly a notification appears telling me that I have a problem with the newtork storage, but if I try to click on "Reload" to try a new connection (I know that the NAS is ready), it fails immediately suggesting me to check the Supervisor logs. 

But nothing is logged there , except the first error, when the NAS was rebooting and they are the only logs I can attach to this issue. 

After 2 hours I'still trying to manually reload it, but it won't work. It fails and nothing is logged. 

Probably the network storage was already connected, but the notification remaining active and for that reason the Reload doesn't work. To remove the notification I have to restart HA. 

Home Assistant OS

1. Connect a CIFS network storage
2. Unplug the network storage so HA will find it disconnected
3. Try to repair/reload the network storage after having it plugged in again. 
...

```txt
2025-09-16 21:36:38.856 INFO (MainThread) [supervisor.mounts.manager] Reloading mount: HABackup
2025-09-16 21:36:38.880 ERROR (MainThread) [supervisor.mounts.mount] Reloading HABackup did not succeed. Check host logs for errors from mount or systemd unit mnt-data-supervisor-mounts-HABackup.mount for details.
2025-09-16 21:36:38.882 INFO (MainThread) [supervisor.resolution.module] Create new suggestion execute_reload - mount / HABackup
2025-09-16 21:36:38.882 INFO (MainThread) [supervisor.resolution.module] Create new suggestion execute_remove - mount / HABackup
2025-09-16 21:36:38.882 INFO (MainThread) [supervisor.resolution.module] Create new issue mount_failed - mount / HABackup
```

## System Information

version | core-2025.9.3
-- | --
dev | false
hassio | true
docker | true
container_arch | aarch64
user | root
virtualenv | false
python_version | 3.13.7
os_name | Linux
os_version | 6.12.34-haos-raspi
arch | aarch64
timezone | Europe/Rome
config_dir | /config

GitHub API | ok
-- | --
GitHub Content | ok
GitHub Web | ok
HACS Data | ok
GitHub API Calls Remaining | 4987
Installed Version | 2.0.5
Stage | running
Available Repositories | 2235
Downloaded Repositories | 8

</details>

logged_in | false
-- | --
can_reach_cert_server | ok
can_reach_cloud_auth | ok
can_reach_cloud | ok

</details>

host_os | Home Assistant OS 16.1
-- | --
update_channel | stable
supervisor_version | supervisor-2025.09.0
agent_version | 1.7.2
docker_version | 28.3.3
disk_total | 228.5 GB
disk_used | 24.3 GB
nameservers | 192.168.80.1
healthy | true
supported | true
host_connectivity | true
supervisor_connectivity | true
ntp_synchronized | true
virtualization | 
board | rpi5-64
supervisor_api | ok
version_api | ok
installed_addons | Studio Code Server (5.19.3), lxp-bridge (v0.13.0), InfluxDB (5.0.2), Advanced SSH & Web Terminal (21.0.3), Log Viewer (0.17.1), Mosquitto broker (6.5.2), Grafana (11.0.0), DSS VoIP Notifier for ARM (4.1.0), SQLite Web (4.4.0), Piper (1.6.4), Whisper (2.6.0), Music Assistant Server (2.5.8)

</details>

dashboards | 3
-- | --
resources | 2
views | 12
mode | storage

</details>

<details><summary>Network Configuration</summary>

adapters | lo (disabled), end0 (enabled, default, auto), docker0 (disabled), hassio (disabled), veth00fe7ac (disabled), veth9d5c14d (disabled), vethe16c9f8 (disabled), vethc01d6e6 (disabled), veth629f4d1 (disabled), veth742ce46 (disabled), veth774ef82 (disabled), vethb5ccf80 (disabled), veth4eb381d (disabled), veth839d767 (disabled), veth450bd63 (disabled), veth0e715ff (disabled), veth08ec2a9 (disabled)
-- | --
ipv4_addresses | lo (127.0.0.1/8), end0 (192.168.80.10/24), docker0 (172.30.232.1/23), hassio (172.30.32.1/23), veth00fe7ac (), veth9d5c14d (), vethe16c9f8 (), vethc01d6e6 (), veth629f4d1 (), veth742ce46 (), veth774ef82 (), vethb5ccf80 (), veth4eb381d (), veth839d767 (), veth450bd63 (), veth0e715ff (), veth08ec2a9 ()
ipv6_addresses | lo (::1/128), end0 (fe80::3ccc:8ec4:e3eb:4b10/64), docker0 (fe80::a81c:feff:fedb:deb4/64), hassio (fe80::f0c6:e3ff:fef8:d01a/64), veth00fe7ac (fe80::9c5e:65ff:fee6:f953/64), veth9d5c14d (fe80::dc7b:69ff:fe7c:6db3/64), vethe16c9f8 (fe80::30ef:76ff:fece:74da/64), vethc01d6e6 (fe80::6cd8:3fff:fedc:ed55/64), veth629f4d1 (fe80::dc77:35ff:fe32:f981/64), veth742ce46 (fe80::e463:1bff:fedd:5027/64), veth774ef82 (fe80::b84c:27ff:fe24:42ec/64), vethb5ccf80 (fe80::f02d:93ff:fec5:a4/64), veth4eb381d (fe80::8cc5:d9ff:fecd:b161/64), veth839d767 (fe80::e4ad:5bff:fe23:8f88/64), veth450bd63 (fe80::3c47:7bff:fe4f:3f86/64), veth0e715ff (fe80::74e1:b8ff:fe9b:4df4/64), veth08ec2a9 (fe80::8cc:caff:fe84:4808/64)
announce_addresses | 192.168.80.10, fe80::3ccc:8ec4:e3eb:4b10

</details>

oldest_recorder_run | 15 agosto 2025 alle ore 20:08
-- | --
current_recorder_run | 13 settembre 2025 alle ore 16:36
estimated_db_size | 1658.51 MiB
database_engine | sqlite
database_version | 3.48.0

</details>

### Supervisor diagnostics

[config_entry-hassio-3f7684dded3f8b47ce6dc9b4dd040e14.json](https://github.com/user-attachments/files/22371446/config_entry-hassio-3f7684dded3f8b47ce6dc9b4dd040e14.json)

### Additional information

_No response_

mdegat01:
## Proposed change

In Supervisor we need to check a lot if an issue/suggestion is present to prevent duplicating them. To facilitate this (and because no one else was listening), our caches were built where the equality check is based on type, context and reference rather then the generated UUID.

Now that core is listening to status of all issues and suggestions for repairs this caused some subtle bugs that went unnoticed. In a number of places we were using a pattern like this:

1. Class had a cached instance of a particular issue it needed to create and check for that was instance specific (ex. "Mount Failed" issue in mount objects or "Boot Failed" issue in addons)
2. Code would check for the presence of this cached instance when it detected the situation was resolved. This check would pass even though the UUID didn't match because context members did
3. Code would call dismiss issue/suggestion method with the cached instance. This would remove the issue/suggestion from cache without error
4. Code would tell Core that an issue/suggestion was removed. However it had an ID that did not match Core's cache and so Core raised errors or ignored the faulty message

The most common manifestation of this was mounts. Supervisor would self-repair them and dismiss the issue with the above pattern. Core would not know what to do with this and leave the orphaned repairs in place even though the mounts worked. They could not be dismissed or applied anymore and were only cleared with a restart.

This fixes all these spots and requires that a suggestion/issue passed to `apply_suggestion`, `dismiss_suggestion`, or `dismiss_issue` matches on UUID what's in our cache or it raises. That will prevent this going forward.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR fixes or closes issue: fixes #6193
- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/addons/addon.py
+++ b/supervisor/addons/addon.py
@@ -197,11 +197,12 @@ class Addon(AddonModel):
             self.sys_resolution.dismiss_issue(issue)
 
         # Dismiss device access missing issue if present and we stopped
-        if (
-            new_state == AddonState.STOPPED
-            and self.device_access_missing_issue in self.sys_resolution.issues
+        if new_state == AddonState.STOPPED and (
+            issue := self.sys_resolution.get_issue_if_present(
+                self.device_access_missing_issue
+            )
         ):
-            self.sys_resolution.dismiss_issue(self.device_access_missing_issue)
+            self.sys_resolution.dismiss_issue(issue)
 
         self.sys_homeassistant.websocket.supervisor_event_custom(
             WSEvent.ADDON,
--- a/tests/addons/test_addon.py
+++ b/tests/addons/test_addon.py
@@ -34,6 +34,8 @@ from supervisor.exceptions import (
 )
 from supervisor.hardware.helper import HwHelper
 from supervisor.ingress import Ingress
+from supervisor.resolution.const import ContextType, IssueType, SuggestionType
+from supervisor.resolution.data import Issue
 from supervisor.utils.dt import utcnow
 
 from .test_manager import BOOT_FAIL_ISSUE, BOOT_FAIL_SUGGESTIONS
@@ -994,16 +996,40 @@ async def test_addon_manual_only_boot(install_addon_example: Addon):
     assert install_addon_example.boot == "manual"
 
 
-async def test_addon_start_dismisses_boot_fail(
-    coresys: CoreSys, install_addon_ssh: Addon
+@pytest.mark.parametrize(
+    ("initial_state", "target_state", "issue", "suggestions"),
+    [
+        (
+            AddonState.ERROR,
+            AddonState.STARTED,
+            BOOT_FAIL_ISSUE,
+            [suggestion.type for suggestion in BOOT_FAIL_SUGGESTIONS],
+        ),
+        (
+            AddonState.STARTED,
+            AddonState.STOPPED,
+            Issue(
+                IssueType.DEVICE_ACCESS_MISSING,
+                ContextType.ADDON,
+                reference=TEST_ADDON_SLUG,
+            ),
+            [SuggestionType.EXECUTE_RESTART],
+        ),
+    ],
+)
+async def test_addon_state_dismisses_issue(
+    coresys: CoreSys,
+    install_addon_ssh: Addon,
+    initial_state: AddonState,
+    target_state: AddonState,
+    issue: Issue,
+    suggestions: list[SuggestionType],
 ):
-    """Test a successful start dismisses the boot fail issue."""
-    install_addon_ssh.state = AddonState.ERROR
-    coresys.resolution.add_issue(
-        BOOT_FAIL_ISSUE, [suggestion.type for suggestion in BOOT_FAIL_SUGGESTIONS]
-    )
+    """Test an addon state change dismisses the issues."""
+    install_addon_ssh.state = initial_state
+    coresys.resolution.add_issue(issue, suggestions)
 
-    install_addon_ssh.state = AddonState.STARTED
+    install_addon_ssh.state = target_state
     assert coresys.resolution.issues == []
     assert coresys.resolution.suggestions == []
 


supervisor/resolution/module.py
```diff
@@ -165,21 +169,37 @@ def _make_issue_message(self, issue: Issue) -> dict[str, Any]:
...
-        raise ResolutionNotFound()
+        raise ResolutionIssueNotFound(uuid=uuid)
+
+    def get_issue_if_present(self, issue: Issue) -> Issue | None:
+        """Get issue matching provided one if it exists in resolution manager."""
+        for i in self._issues:
+            if i != issue:
+                continue
+            return i
+        return None
```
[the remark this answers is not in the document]
mdegat01: Added API level tests to check the exception was being returned correctly. Which they weren't actually because there was an extra wrapper in the API module so good catch. Think we're ok without the others as these are all internal.

supervisor/resolution/module.py
```diff
@@ -165,21 +169,37 @@ def _make_issue_message(self, issue: Issue) -> dict[str, Any]:
...
 
-    def get_issue(self, uuid: str) -> Issue:
+    def get_issue_by_id(self, uuid: str) -> Issue:
         """Return issue with uuid."""
         for issue in self._issues:
             if issue.uuid != uuid:
                 continue
             return issue
-        raise ResolutionNotFound()
+        raise ResolutionIssueNotFound(uuid=uuid)
```
agners, line 194: `ResolutionIssueNotFound` is a `APINotFound` error, hence an `API` related exception. This is a helper function in the resolution module, and its consumer may or may not be the API. Therefor I think this should be a non-API exception raised here, and translated into API exception only at the API layer 🤔

mdegat01:
@agners Yep, you're right about the low confidence review. Looks like I missed one and we didn't have test coverage for it. Added a test case and fixed it.

supervisor/resolution/module.py
mdegat01: So this used to be our direction yes. If you remember, we discussed during #6303 how I had to change from this directive though. Or well didn't have to but it would be extremely inconvenient and lead to a lot of code bloat when coupled with plan to make errors translatable.

If we did what you are suggesting here it would effectively lead to this being added to exceptions:
```py
class APIResolutionIssueNotFound(APINotFound):
    """Raise if issue does not exist."""

    error_key = "api_resolution_issue_not_found_error"
    message_template = "Issue {uuid} does not exist"

    def __init__(self, logger: Callable[..., None] | None = None, *, uuid: str) -> None:
        """Initialize exception."""
        self.extra_fields = {"uuid": uuid}
        super().__init__(None, logger)

class ResolutionIssueNotFound(ResolutionNotFound):
    """Raise if issue does not exist."""

    error_key = "resolution_issue_not_found_error"
    message_template = "Issue {uuid} does not exist"

    def __init__(self, logger: Callable[..., None] | None = None, *, uuid: str) -> None:
        """Initialize exception."""
        self.extra_fields = {"uuid": uuid}
        super().__init__(None, logger)
```
We'd essentially have to duplicate every error we wanted to add a message key to.

Now I mean technically the second one in this case would not need a message key because its never coming out of the API. This is true. But we still need the class to exist and be this specific. Because in the API layer we have to catch and translate from one error type to the other and there's only one very specific error circumstance that translates to `APIResolutionIssueNotFound`. If we used our old pattern of something like `raise ResolutionNotFound(f"Issue {uuid} does not exist", _LOGGER.error)` in the resolution manager code then it would be very challenging for the API layer to know when to translate to `APIResolutionIssueNotFound` without parsing the message.

This is what I realized in #6303 and why I came to the conclusion that we should stop doing this translation to API specific errors in the API layer. If an error can be returned in the API we should define as an APIError to begin with. It doesn't harm anything if the code is being run in a headless context and doesn't reach the API and significantly simplifies the code when we're in an API context.

mdegat01: Now if we do encounter a situation where one type of error results in different status codes in different APIs, then we might want to use separated API and non-API errors with a translation I suppose. We'll cross that bridge when we get to it though, I don't think we have any situations like that today.

agners: Hm, ok, I see this is consistent with what we started in #6303.

It does make sense to have the messages and translation key present in the domain layer exception. Messages and translations are also not necessarily http API specific: Translatable exceptions could also be useful if part of the code were used in a library or interfaced in some other means (non-http).

The mix of concerns really is the `APIError` multi-inheritance in the domain exceptions. What probably would be cleaner is to have `@api_process` list all domain exceptions which translate to 404 etc.

But let's not make this hold up this PR: We already have that pattern, and maybe its a mix of concern we accept for simplicity sake.

agners (approved):

supervisor/resolution/module.py
mdegat01: @agners yea lets talk about it. I'm not in love with this pattern and I certainly don't love that I have to keep adding pylint ignores for too many ancestors. I prefer it over error type duplication and bloat but there's probably a better way

PR #6590, opened 2026-02-23
Use Unix socket for Supervisor to Core communication

agners:
## Proposed change

Switch internal Supervisor→Core HTTP and WebSocket communication from TCP (port 8123) to a Unix domain socket when the installed Core version supports it.

The existing `/run/supervisor` directory on the host (already mounted at `/run/os` inside the Supervisor container) is bind-mounted into the Core container as `/run/supervisor`. Core receives the `SUPERVISOR_CORE_API_SOCKET` environment variable with the socket path, creates the socket there, and Supervisor connects to it via `aiohttp.UnixConnector` at `/run/os/core.sock`.

Since the Unix socket is only reachable by processes on the same host, requests arriving over it are implicitly trusted and authenticated as the existing "Supervisor" system user. This removes the current token round-trip where Core creates a refresh token, hands it to Supervisor, and Supervisor sends it back as a Bearer token on every Core API call. WebSocket connections are likewise authenticated implicitly, skipping the `auth_required`/`auth` handshake.

This reduces attack surface by removing the need for network-based communication between Supervisor and Core, avoids potential port conflicts or `http.server_host` configuration issues, and removes authentication overhead for internal IPC.

**Key design decisions:**
- Session, URL, and transport management lives in `HomeAssistantAPI` as public properties (`session`, `api_url`, `ws_url`) used by WebSocket and proxy code
- Version-gated with `CORE_UNIX_SOCKET_MIN_VERSION` so older Core versions continue using TCP transparently
- LANDINGPAGE version is explicitly excluded from the version comparison (it's not CalVer)
- Hard-fails with a clear error if the Unix socket file is unexpectedly missing when expected
- `WSClient.connect()` for Unix socket (no auth) and `WSClient.connect_with_auth()` for TCP (token auth) cleanly separate the two connection modes
- Token refresh always uses the TCP websession since it is inherently a TCP/Bearer-auth operation
- Proxy `_websocket_client` reuses `WSClient` instead of duplicating the WebSocket auth handshake
- Logs which transport (Unix socket vs TCP) is being used on first request
- Reuses the existing `/run/os` ↔ `/run/supervisor` host mount rather than creating new paths

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR fixes or closes issue: https://github.com/home-assistant/supervisor/issues/6626
- Link to cli pull request:
- Related Core PR: https://github.com/home-assistant/core/pull/163907

## Checklist

### DIFF
--- a/tests/homeassistant/test_api.py
+++ b/tests/homeassistant/test_api.py
@@ -1,65 +1,39 @@
 """Test Home Assistant API."""
 
 from contextlib import asynccontextmanager
-from unittest.mock import MagicMock, patch
+from unittest.mock import AsyncMock, MagicMock, patch
 
 from aiohttp import hdrs
 from awesomeversion import AwesomeVersion
 import pytest
 
 from supervisor.coresys import CoreSys
+from supervisor.docker.const import ContainerState
+from supervisor.docker.monitor import DockerContainerStateEvent
 from supervisor.exceptions import HomeAssistantAPIError
+from supervisor.homeassistant.api import APIState, HomeAssistantAPI
+from supervisor.homeassistant.const import LANDINGPAGE
+
+from tests.common import MockResponse
+
+# --- check_frontend_available ---
 
 
-async def test_check_frontend_available_success(coresys: CoreSys):
-    """Test frontend availability check succeeds with valid HTML response."""
-    coresys.homeassistant.version = AwesomeVersion("2025.8.0")
-
-    mock_response = MagicMock()
-    mock_response.status = 200
-    mock_response.headers = {hdrs.CONTENT_TYPE: "text/html; charset=utf-8"}
-
-    @asynccontextmanager
-    async def mock_make_request(*args, **kwargs):
-        yield mock_response
-
-    with patch.object(
-        type(coresys.homeassistant.api), "make_request", new=mock_make_request
-    ):
-        result = await coresys.homeassistant.api.check_frontend_available()
-
-    assert result is True
-
-
-async def test_check_frontend_available_wrong_status(coresys: CoreSys):
-    """Test frontend availability check fails with non-200 status."""
-    coresys.homeassistant.version = AwesomeVersion("2025.8.0")
-
-    mock_response = MagicMock()
-    mock_response.status = 404
-    mock_response.headers = {hdrs.CONTENT_TYPE: "text/html"}
-
-    @asynccontextmanager
-    async def mock_make_request(*args, **kwargs):
-        yield mock_response
-
-    with patch.object(
-        type(coresys.homeassistant.api), "make_request", new=mock_make_request
-    ):
-        result = await coresys.homeassistant.api.check_frontend_available()
-
-    assert result is False
-
-
-async def test_check_frontend_available_wrong_content_type(
-    coresys: CoreSys, caplog: pytest.LogCaptureFixture
+@pytest.mark.parametrize(
+    ("status", "content_type", "expected"),
+    [
+        (200, "text/html; charset=utf-8", True),
+        (404, "text/html", False),
+        (200, "application/json", False),
+    ],
+)
+async def test_check_frontend_available(
+    coresys: CoreSys, status: int, content_type: str, expected: bool
 ):
-    """Test frontend availability check fails with wrong content type."""
-    coresys.homeassistant.version = AwesomeVersion("2025.8.0")
-
+    """Test frontend availability based on HTTP status and content type."""
     mock_response = MagicMock()
-    mock_response.status = 200
-    mock_response.headers = {hdrs.CONTENT_TYPE: "application/json"}
+    mock_response.status = status
+    mock_response.headers = {hdrs.CONTENT_TYPE: content_type}
 
     @asynccontextmanager
     async def mock_make_request(*args, **kwargs):
@@ -68,15 +42,11 @@ async def test_check_frontend_available_wrong_content_type(
     with patch.object(
         type(coresys.homeassistant.api), "make_request", new=mock_make_request
     ):
-        result = await coresys.homeassistant.api.check_frontend_available()
-
-    assert result is False
-    assert "unexpected content type" in caplog.text
+        assert await coresys.homeassistant.api.check_frontend_available() is expected
 
 
 async def test_check_frontend_available_api_error(coresys: CoreSys):
     """Test frontend availability check handles API errors gracefully."""
-    coresys.homeassistant.version = AwesomeVersion("2025.8.0")
 
     @asynccontextmanager
     async def mock_make_request(*args, **kwargs):
@@ -86,15 +56,14 @@ async def test_check_frontend_available_api_error(coresys: CoreSys):
     with patch.object(
         type(coresys.homeassistant.api), "make_request", new=mock_make_request
     ):
-        result = await coresys.homeassistant.api.check_frontend_available()
+        assert await coresys.homeassistant.api.check_frontend_available() is False
 
-    assert result is False
+
+# --- get_config / get_core_state ---
 
 
 async def test_get_config_success(coresys: CoreSys):
     """Test get_config returns valid config dictionary."""
-    coresys.homeassistant.version = AwesomeVersion("2025.8.0")
-
     expected_config = {
         "latitude": 32.87336,
         "longitude": -117.22743,
@@ -113,11 +82,7 @@ async def test_get_config_success(coresys: CoreSys):
 
     mock_response = MagicMock()
     mock_response.status = 200
-
-    async def mock_json():
-        return expected_config
-
-    mock_response.json = mock_json
+    mock_response.json = AsyncMock(return_value=expected_config)
 
     @asynccontextmanager
     async def mock_make_request(*_args, **_kwargs):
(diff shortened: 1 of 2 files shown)


mdegat01:
What's this waiting on btw @agners ? It looks like both PRs are in draft and there's a good number of comments on the other. Is this dependent on something else in the epic or is it just a time issue?

agners:
> What's this waiting on btw @agners ? It looks like both PRs are in draft and there's a good number of comments on the other. Is this dependent on something else in the epic or is it just a time issue?

Yeah it's waiting on Core part to complete. I've prioritized some release things and other bug fixes so just a time thing. I'll get back to it next week.

supervisor/docker/homeassistant.py
```diff
@@ -136,6 +138,8 @@ def mounts(self) -> list[DockerMount]:
                             propagation=PropagationMode.RSLAVE
                         ),
                     ),
+                    # Supervisor <-> Core communication socket
+                    MOUNT_CORE_RUN,
```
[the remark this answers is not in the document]
agners: Does it matter much :shrug:  I think I'd just always mount it for simplicity.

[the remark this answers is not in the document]
agners: It is simple enough, so I've added it.

supervisor/api/proxy.py
```diff
@@ -179,57 +178,17 @@ async def api(self, request: web.Request):
...
-                self.sys_homeassistant.api.access_token = None
-                return await self._websocket_client()
-
-            raise HomeAssistantAuthError()
-
-        except (RuntimeError, ValueError, TypeError, ClientConnectorError) as err:
+            return ws_client.client
+        except HomeAssistantAPIError as err:
+            _LOGGER.error("Error connecting to Home Assistant WebSocket: %s", err)
+            raise APIError() from err
```
mdegat01:
```suggestion
            raise APIError(f"Error connecting to Home Assistant WebSocket: {err}", _LOGGER.error) from err
```
Since we're trying to avoid returning no message errors from the API this seems preferrable. Alternatively we can make a `HomeAssistantWSUnknownError` class with an error key and something like "Error connecting to Home Assistant WebSocket. Check Supervisor logs for details".

supervisor/api/proxy.py
```diff
@@ -179,57 +178,17 @@ async def api(self, request: web.Request):
...
+        except (RuntimeError, ValueError, TypeError) as err:
             _LOGGER.error("Client error on WebSocket API %s.", err)
-        except HomeAssistantAuthError:
-            _LOGGER.error("Failed authentication to Home Assistant WebSocket")
-
-        raise APIError()
+            raise APIError() from err
```
mdegat01: Same comment as above about empty message error responses from the API. We should probably also log something here else there's not much hope of debug if it happens. Does it actually happen today?

supervisor/api/proxy.py
```diff
@@ -179,57 +178,17 @@ async def api(self, request: web.Request):
...
-                return await self._websocket_client()
-
-            raise HomeAssistantAuthError()
-
-        except (RuntimeError, ValueError, TypeError, ClientConnectorError) as err:
+            return ws_client.client
+        except HomeAssistantAPIError as err:
+            _LOGGER.error("Error connecting to Home Assistant WebSocket: %s", err)
+            raise APIError() from err
+        except (RuntimeError, ValueError, TypeError) as err:
```
mdegat01:
```suggestion
        except RuntimeError, ValueError, TypeError as err:
```
Looks like the AI is still not doing what you want here despite the instructions change. Is this a mypy problem by any chance? When it did it to me I noticed it was driven from a mypy error about it which was really because mypy was out of date in that environment.

supervisor/docker/const.py
```diff
@@ -89,6 +89,7 @@ class MountBindOptions:
 
     propagation: PropagationMode | None = None
     read_only_non_recursive: bool | None = None
+    create_mountpoint: bool | None = None
```
mdegat01: It doesn't seem like this is used anywhere?

supervisor/homeassistant/api.py
```diff
@@ -86,11 +183,50 @@ async def ensure_access_token(self) -> None:
...
+                return await WSClient.connect_with_auth(
+                    self.session,
+                    self.ws_url,
+                    self._access_token,
+                    max_msg_size=max_msg_size,
+                )
+            except HomeAssistantAPIError:
+                if attempt == 2:
+                    raise
+                self._access_token = None
```
mdegat01:
```suggestion
            except HomeAssistantAPIError:
                self._access_token = None
                if attempt == 2:
                    raise
```
I think we want to do this right? Otherwise next time we try to connect `_ensure_access_token` will see an existing one and try to reuse it, even though it didn't work.

supervisor/homeassistant/api.py
```diff
@@ -86,11 +183,50 @@ async def ensure_access_token(self) -> None:
...
+
+        # Unreachable, but satisfies type checker
+        raise HomeAssistantAPIError("Failed to connect WebSocket")
```
mdegat01: It does but codecov won't like this, permanently uncoverable code. I can't really think of a better way to do this without unreachable code so maybe we just end with a comment that says there's no return because it can't reach here and tell the type checker to ignore here?

agners:
> Some small changes, looks great though. Should definitely be a performance boost, have you measured it at all?

Just did some rough test, this is the time it takes for the first connection (in `get_api_state`). The Unix socket doesn't need to do the auth dance, I guess that is mostly where the difference is coming from:

```
2026-04-08 23:07:17.317 INFO (MainThread) [supervisor.homeassistant.api] Connected to Core via Unix socket /run/os/core.sock (52ms)
2026-04-08 23:07:36.237 INFO (MainThread) [supervisor.homeassistant.api] Connected to Core via Unix socket /run/os/core.sock (31ms)
2026-04-08 23:07:48.488 INFO (MainThread) [supervisor.homeassistant.api] Connected to Core via Unix socket /run/os/core.sock (24ms)
2026-04-08 23:08:01.720 INFO (MainThread) [supervisor.homeassistant.api] Connected to Core via Unix socket /run/os/core.sock (45ms)
2026-04-08 23:08:56.262 INFO (MainThread) [supervisor.homeassistant.api] Connected to Core via TCP http://172.30.32.1:80 (87ms)
2026-04-08 23:09:45.408 INFO (MainThread) [supervisor.homeassistant.api] Connected to Core via TCP http://172.30.32.1:80 (44ms)
2026-04-08 23:10:04.411 INFO (MainThread) [supervisor.homeassistant.api] Connected to Core via TCP http://172.30.32.1:80 (85ms)
2026-04-08 23:10:18.631 INFO (MainThread) [supervisor.homeassistant.api] Connected to Core via TCP http://172.30.32.1:80 (53ms)
```

Measuring 10'000 API calls barley shows a difference:

```json
{
  "result": "ok",
  "data": {
    "transport": "unix_socket",
    "requests": 10000,
    "errors": 0,
    "min_ms": 2.2,
    "max_ms": 157.8,
    "avg_ms": 2.9,
    "median_ms": 2.8,
    "p95_ms": 3.5
  }
}
{
  "result": "ok",
  "data": {
    "transport": "tcp",
    "requests": 10000,
    "errors": 0,
    "min_ms": 2.3,
    "max_ms": 124.8,
    "avg_ms": 2.8,
    "median_ms": 2.7,
    "p95_ms": 3.3
  }
}
```

From literature, the expected latency difference is single digit microseconds, Python being interpreted is just much to slow for the transport to make a difference.

I also tested throughput with a test endpoint returning 1GB of data:

```json
    "throughput": {
      "transport": "tcp",
      "size_bytes": 1073741824,
      "duration_ms": 1798.4,
      "throughput_mb_s": 569.4
    }
    "throughput": {
      "transport": "unix_socket",
      "size_bytes": 1073741824,
      "duration_ms": 2057.8,
      "throughput_mb_s": 497.6
    }
```

It actually appears that Unix socket is slightly slower. Not sure if the aiohttp implementation is not quite optimized :thinking: .

supervisor/api/proxy.py
agners: Might be that this code predates my `CLAUDE.md` file change. That said, I wonder why we need to capture `RuntimeError` here, that usually indicates a logic error in our code. I wonder if we should remove this from the list :thinking:

agners: Yeah with the centralized authentication logic in `websocket.py:connect()` we capture all exceptions there. Which is probably also not all that nice.

`RuntimeError` only really seems to be raised if the connection is closed. From what I can tell, this can't really happen as we open a new connection in that case. `RuntimeError` has been handled in proxy.py since it's inception, its not clear to me if that ever happened in practice.

Let's remove `RuntimeError`. In case it happens in practice, we'll discover it through Sentry then. I've also got rid of the broad `except Exception` in `connect()`, and convert those into `HomeAssistantAPIError`. I think with that we have quite a bit better exception handling than before.

supervisor/docker/const.py
agners: Yeah this came from an earlier iteration, where I used different paths in `/run`. But we need to have access to that location in Supervisor too, so it must exist at time we use in Supervisor. Hence I don't see any use case for that create mountpoint flag in this PR anymore. I drop it.

supervisor/homeassistant/api.py
agners: We have just a 80% requirement now, and I think this would bypass it as well: `# pragma: no cover` :shrug: . I made this a `RuntimeError` so it would be noticed.

mdegat01 (approved):
This looks good. Kind of thought we'd get more performance boost out of it but its a good change either way.

However I should note something. So I installed this on a local dev system to test. First off, I had no issues 👍 But it did take me a sec to figure out how to get it actually used. Just updating supervisor to the branch code isn't enough, you also have to rebuild the core container after the supervisor update. Otherwise the socket isn't mounted in core's container so it won't use it.

I fixed this for myself easily enough with `ha core rebuild` so I could do my testing. But our users aren't going to do that. Most likely the only time the core container is rebuilt for users is when they update core. Even a host reboot does not cause a rebuild of core container because unlike the addon containers we don't remove core's container automatically when we stop it. We only remove it if there's a recurring issue, an update or the user specifically asks us to with `ha core rebuild` or a safe mode restart.

This means you may not actually get real-world usage of this new code path for quite a while after we ship it unless core happens to need to ship a patch after. Maybe we want to consider a change to kind of force the issue there? Like something that detects the core container's config does not match expectations and sets a flag to make the next restart a rebuild automatically? This can be a separate PR if you want, its a related problem but not really the same task.

agners:
> Just updating supervisor to the branch code isn't enough, you also have to rebuild the core container after the supervisor update. Otherwise the socket isn't mounted in core's container so it won't use it.

Yes, I am aware, and probably should have noted that in the PR description.

It does have the added side effect that the feature will be used slowly more and more, like a slow rollout. Which in case we see issues, we can fix them before all users are affected, or worst case even revert.

> . Maybe we want to consider a change to kind of force the issue there? Like something that detects the core container's config does not match expectations and sets a flag to make the next restart a rebuild automatically?

So that would mean it will still take a while until the feature is rolled out. In my mind, we do point releases fairly often, so it will get adopted quite quickly. Not sure if it is really worth the additional complexity just for a slightly faster rollout.

mdegat01:
I was thinking restarts happen more frequently then updates but yea I guess that's fair. The main issue I see is just if you're hoping to collect info from beta before promoting to stable it may be tough to count on that. But I'm good with it either way, LGTM 👍

PR #6598, opened 2026-02-26
Add default verbose timestamps for plugin logs

agners:
## Proposed change

Add support for default verbose formatting in advanced logs route registration and use it for plug-in log endpoints so timestamps are included by default.

This PR:
- Adds a `default_verbose` option to advanced logs route wiring and host log handling.
- Enables `default_verbose=True` for `/dns/logs`, `/audio/logs`, and `/multicast/logs` (including follow/latest/boot variants wired via `_register_advanced_logs`).
- Uses the same `default_verbose` route-level behavior for host routes that use default identifiers (`/host/logs`, `/host/logs/follow`, `/host/logs/boots/{bootid}`, `/host/logs/boots/{bootid}/follow`) and removes the previous in-handler special case.
- Updates API tests to validate formatter expectations for these endpoints.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR fixes or closes issue: N/A
- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/api/__init__.py
+++ b/supervisor/api/__init__.py
@@ -193,10 +193,13 @@ class RestAPI(CoreSysAttributes):
         self.webapp.add_routes(
             [
                 web.get("/host/info", api_host.info),
-                web.get("/host/logs", api_host.advanced_logs),
+                web.get(
+                    "/host/logs",
+                    partial(api_host.advanced_logs, default_verbose=True),
+                ),
                 web.get(
                     "/host/logs/follow",
-                    partial(api_host.advanced_logs, follow=True),
+                    partial(api_host.advanced_logs, follow=True, default_verbose=True),
                 ),
                 web.get("/host/logs/identifiers", api_host.list_identifiers),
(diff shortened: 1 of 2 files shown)


agners:
Host logs are special today too, so why can't plug-ins be? :thinking: :shrug: 

As for apps, we could make that a configuration.

agners:
FWIW, I prefer this method over adding timestamps inline for all plug-ins. Timestamp is stored as metadata anyway, and formatting the timestamp when the log message gets actually read is more flexible then statically log with timestamp.

If we have a better way to transfer metadata to the frontend then sure we can also render it on the frontend. But we are not there today. So this is the next best option IMHO.

sairon:
> Host logs are special today too, so why can't plug-ins be? 🤔 🤷

Well, the verbose format is supposed to mimic journal logs (with process names and PIDs), without that Host logs would make little sense.

But nevermind, let's start with this and once we have better support in frontend (just a toggle for the verbose logs would be nice) we can revisit.

PR #6597, opened 2026-02-26
Pin plug-in version temporarily when manually rolling back to previous versions

agners:
## Proposed change

This PR makes it easier to temporarily test a previous plugin version for regressions without the Supervisor immediately upgrading it again.

When a plugin is updated with an explicit `--version`, Supervisor now stores an in-memory runtime pin for that plugin version. As long as Supervisor stays up, automatic plugin update paths skip that plugin and log a warning. The pin is intentionally temporary and is cleared on Supervisor restart.

To keep behavior consistent and easier to maintain, auto-update decision and logging were centralized in `PluginBase.auto_update()`. Scheduled plugin update tasks, plugin-manager startup checks, and the `PLUGINS_UPDATED` job condition now use the same code path.

Tests were added/updated to cover runtime pin behavior and auto-update skipping.

## Type of change

- [ ] Dependency upgrade
- [x] New feature (which adds functionality to the supervisor)
- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] Breaking change (fix/feature causing existing functionality to break)
- [ ] Code quality improvements to existing code or addition of tests

## Additional information

- This PR fixes or closes issue: fixes #
- This PR is related to issue:
- Link to documentation pull request:
- Link to cli pull request:
- Link to client library pull request:

## Checklist

- [x] The code change is tested and works locally.
- [x] Local tests pass. **Your PR cannot be merged unless tests pass**
- [x] There is no commented out code in this PR.
- [ ] I have followed the [development checklist][dev-checklist]
- [x] The code has been formatted using Ruff (`ruff format supervisor tests`)
- [x] Tests have been added to verify that the new code works.

If API endpoints or add-on configuration are added/changed:

- [ ] Documentation added/updated for [developers.home-assistant.io][docs-repository]
- [ ] [CLI][cli-repository] updated (if necessary)
- [ ] [Client library][client-library-repository] updated (if necessary)

[dev-checklist]: https://developers.home-assistant.io/docs/en/development_checklist.html
[docs-repository]: https://github.com/home-assistant/developers.home-assistant
[cli-repository]: https://github.com/home-assistant/cli
[client-library-repository]: https://github.com/home-assistant-libs/python-supervisor-client/

PR #6606, opened 2026-03-02
Respect auto-update setting for plug-in auto-updates

agners:
## Proposed change

Respect the existing Supervisor auto-update toggle for plug-in auto-updates.

Previously, plug-ins (DNS, CLI, Audio, Observer, Multicast) were always auto-updated regardless of the Supervisor auto-update setting. This adds `JobCondition.AUTO_UPDATE` to the scheduled plug-in update tasks and checks `auto_update` in the plug-in manager startup path, so disabling auto-updates in Supervisor settings also prevents plug-ins from being auto-updated.

This is useful when manually rolling back a plug-in to a previous version for regression testing — disabling auto-updates prevents the plug-in from being immediately updated back to the latest version.

Manual plug-in updates (via API) and the `PLUGINS_UPDATED` job pre-condition are not affected.

This is an alternative to #6597.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/jobs/decorator.py
+++ b/supervisor/jobs/decorator.py
@@ -458,10 +458,9 @@ class Job(CoreSysAttributes):
             ]
         ):
             if not coresys.sys_updater.auto_update:
-                _LOGGER.debug(
-                    "Skipping plugin auto-updates because Supervisor auto-update is disabled"
+                raise JobConditionException(
+                    f"'{method_name}' blocked from execution, plugin(s) {', '.join(plugin.slug for plugin in out_of_date)} are not up to date and auto-update is disabled"
                 )
-                return
 
             errors = await asyncio.gather(
                 *[plugin.update() for plugin in out_of_date], return_exceptions=True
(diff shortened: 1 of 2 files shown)


supervisor/jobs/decorator.py
```diff
@@ -457,6 +457,12 @@ async def check_conditions(
                 if plugin.need_update
             ]
         ):
+            if not coresys.sys_updater.auto_update:
+                _LOGGER.debug(
+                    "Skipping plugin auto-updates because Supervisor auto-update is disabled"
+                )
+                return
```
mdegat01:
```suggestion
                raise JobConditionException(
                    f"'{method_name}' blocked from execution, plugin(s) {', '.join(out_of_date)} must be updated first and auto update is disabled"
                )
```
If auto update is disabled then that means we cannot attempt to update them in this spot. That doesn't mean the job condition passes though. If a job requires plugins are up to date then this needs to raise a `JobConditionException` if they aren't, regardless of whether or not we were allowed to update them first.

We also can't use `return` here either way because there may be other job conditions to check after.

mdegat01 (changes requested):

supervisor/jobs/decorator.py
agners: Right! I've added the exception. `out_of_date` is a list of `PluginBase`, we need to use the slug of the plug-ins for the message, so it got slightly more complicated.

mdegat01 (approved):
LGTM 👍

PR #6610, opened 2026-03-03
Fix apps build using wrong architecture for non-native arch apps

agners:
## Proposed change

When building a locally-built app (one without an `image` tag in its config, requiring a Dockerfile build), the architecture was always set to `sys_arch.default` (e.g. `amd64` on x86_64) regardless of the app's declared architectures. This meant an app declaring only `i386` support would incorrectly build with `amd64` as the architecture, producing a wrong image name (e.g. `local/amd64-addon-foo` instead of `local/i386-addon-foo`), wrong `--platform`, wrong `BUILD_ARCH` build arg, and wrong base image.

The fix uses `sys_arch.match()` against the app's declared arch list in all code paths, consistent with how image-based apps already worked.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -508,8 +508,8 @@ async def coresys(
         "HostConfig": {"Privileged": True},
     }
     coresys_obj.arch._default_arch = CpuArch.AMD64
-    coresys_obj.arch._supported_arch = [CpuArch.AMD64, CpuArch.I386]
-    coresys_obj.arch._supported_set = {CpuArch.AMD64, CpuArch.I386}
+    coresys_obj.arch._supported_arch = [CpuArch.AMD64]
+    coresys_obj.arch._supported_set = {CpuArch.AMD64}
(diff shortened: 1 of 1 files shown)


sairon (approved):

sairon (changes requested):
The test failures are relevant to the conftest changes.

agners:
> The test failures are relevant to the conftest changes.

Hm, I see, adding `_supported_arch` exposed a oddity in our tests: `load()` extents the `_supported_arch` existing list. This is only really a problem in tests, where we prepopulate it in conftest, but some tests call `load()` explicitly, which then causes problems.

Making `load()` idempotent seems safer anyways. Maybe we should even not set `_supported_arch` in `__init__()` to cause hard failure when trying to use architecture information while they are not properly loaded. But that seems out of scope for this PR.

PR #6614, opened 2026-03-04
Deprecate advanced mode option in addon config

### LINKED ISSUE #6589: Removal of advanced mode
### Description

Advanced mode needs to be removed, meaning that any flags that are used to hide features behind advanced mode need addressing.
This task is to track both the discovery of where flags are used, and the progress for them to be updated.

### Additional context

Epic: https://github.com/home-assistant/epics/issues/26

mdegat01:
## Proposed change

Advanced mode is being removed from across Home Assistant. This deprecates the config option by ignoring it and always returning `False` and logging any time it is seen in config.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR fixes or closes issue: fixes #6589
- Link to documentation pull request: https://github.com/home-assistant/developers.home-assistant/pull/2982
- Link to cli pull request:
- Link to client library pull request: https://github.com/home-assistant-libs/python-supervisor-client/pull/279

## Checklist

### DIFF
--- a/supervisor/addons/model.py
+++ b/supervisor/addons/model.py
@@ -254,7 +254,7 @@ class AddonModel(JobGroup, ABC):
 
     @property
     def advanced(self) -> bool:
-        """Return advanced mode of add-on."""
+        """Return False; advanced mode is deprecated and no longer supported."""
         # Deprecated since Supervisor 2026.03.0; always returns False and can be
         # removed once that version is the minimum supported.
         return False


mdegat01:
@agners is breaking-change the right label here? It's not actually breaking yet but it will be in the future. I don't know what other label makes sense of the ones we went through the other day though 🤔

mdegat01:
@agners I added needs-client-library but not exactly sure what to do there. There's no functional change there because its not being removed at this time. Do you think its proper to add a commit with comments to get the deprecation in the patch nodes of the next release of aiohasupervisor separately from supervisor? Or is this PR here enough to start the deprecation clock there too?

agners:
> @agners is breaking-change the right label here? It's not actually breaking yet but it will be in the future. I don't know what other label makes sense of the ones we went through the other day though 🤔

I'd say yes, we should mark this PR already as breaking. The label is more informational, but we want to get the attention of release notes readers early.

> @agners I added needs-client-library but not exactly sure what to do there. There's no functional change there because its not being removed at this time. Do you think its proper to add a commit with comments to get the deprecation in the patch nodes of the next release of aiohasupervisor separately from supervisor? Or is this PR here enough to start the deprecation clock there too?

Hm, we left out other deprecated fields from the beginning, but that was different :thinking:  Just an empty commit I would not do, we can also manually add a note to the release if we want to inform client library users.

The question maybe is if we could use the Python warnings feature to warn users on access or something? :thinking:

agners:
> Hm, we left out other deprecated fields from the beginning, but that was different 🤔 Just an empty commit I would not do, we can also manually add a note to the release if we want to inform client library users.

As discussed, it actually seems very sensible to add a Python comment to the client library model which says that the field is deprecated. With that we have a changelog entry as well.

PR #6618, opened 2026-03-09
Treat empty string password as None in backup restore

agners:
## Proposed change

When an empty string `""` is passed as the backup restore password via `hassio.restore_full` or `hassio.restore_partial` (or possibly other callers of the Supervisor REST API), securetar 2026.2.0 crashes with `AttributeError: 'NoneType' object has no attribute 'restore_key_material'`.

This is caused by an inconsistency in securetar 2026.2.0 where `password is not None` sets `encrypted=True` but `if password:` (truthiness) is used to derive the key, leaving `_root_key_context` as `None` for empty
strings. This is being fixed in securetar via https://github.com/home-assistant-libs/securetar/pull/136, which will make
securetar consistently use `is not None` checks and thereby support `""` as a valid password.

However, backup creation in Supervisor traditionally uses a truthiness check (`if password:`) to decide whether to encrypt, meaning a backup encrypted with `""` as the password can never be created. To stay consistent with
that behavior, this PR normalizes `""` to `None` in `set_password()`. This is also how the Supervisor behaved before securetar 2025.12.0 when key derivation was done in the Supervisor itself and `set_password` had its own
truthiness check.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR is related to issue: https://home-assistant-io.sentry.io/issues/SUPERVISOR-1EH7
- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/backups/backup.py
+++ b/supervisor/backups/backup.py
@@ -335,9 +335,8 @@ class Backup(JobGroup):
         """Set the password for an existing backup.
 
         Treat empty string as None to stay consistent with backup creation
-        which also uses a truthiness check and to work around a securetar
-        2026.2.0 bug where empty string password sets encrypted=True but fails
-        to derive a key, leading to an AttributeError on restore.
+        and Supervisor behavior before #6402, independent of SecureTar
+        behavior in this regard.
         """
         self._password = password or None
 


supervisor/backups/backup.py
```diff
@@ -331,8 +332,14 @@ def new(
...
 
     def set_password(self, password: str | None) -> None:
-        """Set the password for an existing backup."""
-        self._password = password
+        """Set the password for an existing backup.
+
+        Treat empty string as None to stay consistent with backup creation
+        which also uses a truthiness check and to work around a securetar
+        2026.2.0 bug where empty string password sets encrypted=True but fails
+        to derive a key, leading to an AttributeError on restore.
```
emontnemery: I don't think this comment is accurate. Yes, there's a bug in securetar, but we're not exactly working around that bug, we're making sure an empty password is treated like no password. I'd suggest to write that instead.

tests/backups/test_backup.py
```diff
@@ -310,6 +310,12 @@ async def test_validate_backup(
                 BackupInvalidError, match="Invalid password for backup f92f0339"
             ),
         ),
+        (
+            "",
+            pytest.raises(
+                BackupInvalidError, match="Invalid password for backup f92f0339"
+            ),
+        ),
```
emontnemery, line 318: Is this same as the None-case?

emontnemery (changes requested):

tests/backups/test_backup.py
agners: Since we treat "" now like no password, yes.

PR #6623, opened 2026-03-11
Remove obsolete persistent notification system and core_security check

agners:
## Proposed change

Remove the obsolete persistent notification system (`ResolutionNotify`) and the `core_security` resolution check.

The `core_security` check detected Home Assistant Core versions below 2021.1.5 running with custom components and created a persistent notification via `ResolutionNotify` to warn users. Since #6148 bumped the minimum supported Core version well past 2021.1.5, this check can never trigger, making both the check and the entire notification module dead code.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:
- Link to client library pull request: https://github.com/home-assistant-libs/python-supervisor-client/pull/276

## Checklist


agners:
> We should remove `core_security` from here but that would be a breaking change. So how do we want to handle that in this case 🤔

Good catch. It seems that `core_security` is not used in Core, and highly doubt that anyone ever used that string constant. So it is a "tree fell in the woods" situation: Technically breaking, but no-one will ever notice, so why bother :shrug: .

In the very unlikely case where a client is using `core_security`, the resolution API `/api/resolution/check/{check}/options` or `/api/resolution/check/{check}/run` will just return 404s. In the very unlikely case that a user used the client library enum, the check will be gone from the enum and hopefully a static checker will raise an error.

Issue #6549, opened 2026-02-10
Report if time is not synchronized to the user


agners:
### Description

If systemd-timesyncd reports that time is not synchronized we should inform the user that time is (possibly) wrong. 

We need to take air gapped systems into account, e.g. if there is no Internet connectivity NTP sync will not be possible, but the time still might be right. Is there information if systemd-timesyncd was able to read the RTC? 🤔 If the RTC reports a seemingly valid time, we should trust it.

sairon:
There's now #6625 which provides some feedback when the time had to be synced using the fallback method.

Now, do we want to also raise unhealthy flag if NTP is broken? To account for airgapped systems, it would mean we'd have to compare the time for example to some build time timestamp to assess if the time is "seemingly valid" (it IMHO doesn't matter if it was set from RTC, which doesn't seem to be exposed in DBus anyway). Otherwise the suggestion/fixup might be sufficient to address this issue as well.

agners:
> Now, do we want to also raise unhealthy flag if NTP is broken? To account for airgapped systems, it would mean we'd have to compare the time for example to some build time timestamp to assess if the time is "seemingly valid" (it IMHO doesn't matter if it was set from RTC, which doesn't seem to be exposed in DBus anyway). Otherwise the suggestion/fixup might be sufficient to address this issue as well.

I think I wouldn't go as far. If it is air gaped, we just can't tell really, so just don't check anything. Or in other words, have the NTP check only active if host internet = true.

PR #6625, opened 2026-03-12
Fix fallback time sync, create repair issue if time is out of sync

### LINKED ISSUE #6015: `DBusFatalError: Automatic time synchronization is enabled` on initial setup (blocking all subsequent setup steps)

I'm attempting to set up HA on a Raspberry Pi 3, using the image from github (`haos_rpi3-64-16.0.img.xz`) flashed directly onto an SD card (using `dd`).

The system boots, but the initial setup fails / isn't able to progress.

Looking in the logs there are many errors related to the supervisor not having internet access etc (despite me viewing this from the web viewer, so it definitely has a network connection), but the initial/underlying error is that the system time is out of date, but it's unable to set it because automatic time synchronization is enabled:

```
2024-10-09 02:43:53.650 WARNING (MainThread) [supervisor.core] System time/date shift over more than 3 days found!
2024-10-09 02:43:53.650 INFO (MainThread) [supervisor.host.control] Setting new host datetime: 2025-07-12T09:26:28+00:00
2024-10-09 02:43:53.676 CRITICAL (MainThread) [supervisor.core] Fatal error happening on load Task [...] supervisor.exceptions.DBusFatalError: Automatic time synchronization is enabled
```

Presumably this is because NTP is enabled.

I can't disable NTP, because I get an error about a read-only filesystem.

I was able to set the correct date myself using `date -s`; I couldn't find any sort of override option for `datetimectl` (which is the cli interface to the same underlying functions the supervisor is using).

I'm not sure whether you'd be able to mitigate this by disabling NTP immediately prior to https://github.com/home-assistant/supervisor/blob/9a0f530a2f639af81d896cb35a9c12012d1e1a42/supervisor/core.py#L424 and then re-enabling it again if it was previously enabled, or whether that would also fail due to the read-only file system.

To be clear: I have successfully worked around this problem, so I don't need immediate help, but I imagine this would have tripped up anybody who didn't know how to drop into a login shell and directly adjust the time using `date`.

Home Assistant OS

1. Have the system clock set to 2024 (it's currently 2025) for some reason (I don't think the raspi3 has a RTC, so I'm not sure where it got that date from?)
2. Do a standard install of HA, following the documented process
3. Watch as it's unable to make any progress, due to a lot of errors about the supervisor not having internet access
...

```txt
(I don't have access to this because I haven't finished setup yet)
```

### System information

(I haven't "configured" it yet so don't have system information, sorry -- I'm using a Raspberry Pi 3, using the image `haos_rpi3-64-16.0.img`)

### Supervisor diagnostics

(again, I don't have access to this yet because I haven't finished setup)

### Additional information

_No response_

sairon:
The "poor man's NTP" using the whois service didn't work because it attempted to sync the time when the NTP service was enabled, which is rejected by the timedated service. To fix this, Supervisor now first disables the systemd-timesyncd service and creates a repair issue before adjusting the time. The timesyncd service stays disabled until submitting the fixup. Theoretically, if the time moves backwards from an invalid time in the future, systemd-timesyncd could otherwise restore the wrong time from a timestamp if we did that after the time was set.

Also, the sync is now performed if the time is more that 1 hour off and in both directions (previously it only intervened if it was more than 3 days in the past).

Fixes #6015, refs #6549

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR fixes or closes issue: fixes #6015
- This PR is related to issue: #6549 
- Link to cli pull request:
- Link to client library pull request: home-assistant-libs/python-supervisor-client#282

## Checklist


sairon:
Tested in a VM with `no-kvmclock` and invalid RTC time. After the system startup with an old version of Supervisor:

```
# timedatectl
               Local time: Sat 2026-03-07 13:00:54 CET
           Universal time: Sat 2026-03-07 12:00:54 UTC
                 RTC time: Sat 2026-03-07 12:00:54
                Time zone: Europe/Prague (CET, +0100)
System clock synchronized: no
              NTP service: active
          RTC in local TZ: no
```

With this PR:

```
2026-03-07 13:01:33.023 WARNING (MainThread) [supervisor.core] System time/date shift over more than 1 hour detected!
2026-03-07 13:01:33.023 INFO (MainThread) [supervisor.core] Stopping systemd-timesyncd to allow manual time adjustment
2026-03-07 13:01:33.026 INFO (MainThread) [supervisor.resolution.module] Create new suggestion enable_ntp - system / None
2026-03-07 13:01:33.026 INFO (MainThread) [supervisor.resolution.module] Create new issue ntp_sync_failed - system / None
2026-03-07 13:01:34.028 INFO (MainThread) [supervisor.host.control] Setting new host datetime: 2026-03-12T09:51:01+00:00
2026-03-12 10:51:01.019 ...
```


```
# systemctl status systemd-timesyncd | cat
○ systemd-timesyncd.service - Network Time Synchronization
     Loaded: loaded (/usr/lib/systemd/system/systemd-timesyncd.service; enabled; preset: enabled)
    Drop-In: /usr/lib/systemd/system/systemd-timesyncd.service.d
             └─hassos.conf, ro.conf
     Active: inactive (dead) since Sat 2026-03-07 13:01:33 CET; 4 days ago
   Duration: 1min 9.856s
 Invocation: 996fc43e10e74b2a9e7f7952abad25aa
       Docs: man:systemd-timesyncd.service(8)
    Process: 499 ExecStart=/usr/lib/systemd/systemd-timesyncd (code=exited, status=0/SUCCESS)
   Main PID: 499 (code=exited, status=0/SUCCESS)
     Status: "Idle."
   Mem peak: 2.2M
        CPU: 22ms

Mar 07 13:01:00 homeassistant systemd-timesyncd[499]: Network configuration changed, trying to establish connection.
Mar 07 13:01:00 homeassistant systemd-timesyncd[499]: Network configuration changed, trying to establish connection.
Mar 07 13:01:10 homeassistant systemd-timesyncd[499]: Timed out waiting for reply from 162.159.200.123:123 (time.cloudflare.com).
Mar 07 13:01:20 homeassistant systemd-timesyncd[499]: Timed out waiting for reply from 162.159.200.1:123 (time.cloudflare.com).
Mar 07 13:01:28 homeassistant systemd-timesyncd[499]: Network configuration changed, trying to establish connection.
Mar 07 13:01:32 homeassistant systemd-timesyncd[499]: Network configuration changed, trying to establish connection.
Mar 07 13:01:32 homeassistant systemd-timesyncd[499]: Network configuration changed, trying to establish connection.
Mar 07 13:01:33 homeassistant systemd[1]: Stopping Network Time Synchronization...
Mar 07 13:01:33 homeassistant systemd[1]: systemd-timesyncd.service: Deactivated successfully.
Mar 07 13:01:33 homeassistant systemd[1]: Stopped Network Time Synchronization.
```

```
# timedatectl
               Local time: Thu 2026-03-12 10:51:01 CET
           Universal time: Thu 2026-03-12 09:51:01 UTC
                 RTC time: Thu 2026-03-12 09:51:02
                Time zone: Europe/Prague (CET, +0100)
System clock synchronized: no
              NTP service: inactive
          RTC in local TZ: no
```

supervisor/core.py
```diff
@@ -423,11 +424,27 @@ async def _adjust_system_datetime(self) -> None:
...
+                "systemd-timesyncd.service", StopUnitMode.REPLACE
+            )
+            # Keep service disabled and create a repair issue
+            self.sys_resolution.create_issue(
+                IssueType.NTP_SYNC_FAILED,
+                ContextType.SYSTEM,
+                suggestions=[SuggestionType.ENABLE_NTP],
+            )
+            # We need to wait a bit for the service to stop.
+            await asyncio.sleep(1)
```
agners, line 446: Hm, I see we don't have a easy way to check, the logic in mount seems a bit complicated. Maybe it would be nice to have some helpers for systemd unit control 🤔 

Anyways, should be a separate PR anyways.

supervisor/core.py
```diff
@@ -423,11 +424,27 @@ async def _adjust_system_datetime(self) -> None:
...
+            _LOGGER.info("Stopping systemd-timesyncd to allow manual time adjustment")
+            await self.sys_dbus.systemd.stop_unit(
+                "systemd-timesyncd.service", StopUnitMode.REPLACE
+            )
+            # Keep service disabled and create a repair issue
+            self.sys_resolution.create_issue(
+                IssueType.NTP_SYNC_FAILED,
+                ContextType.SYSTEM,
+                suggestions=[SuggestionType.ENABLE_NTP],
+            )
```
agners, line 444: Maybe worth explaining why we consider NTP sync failed in this case.

Are we sure that by that point at least 1 or 2 attempts got made? 🤔

agners (changes requested):
Maybe worth adjusting `test_adjust_system_datetime_if_time_behind` to check that NTP gets disabled first.

supervisor/core.py
sairon: There's attempt to sync at boot, then the `systemd-timesyncd` keeps retrying, so at least a few attempts have been made, as you can also see in the log from the unit above. I'd say that it's rather obvious that the sync failed if time is off with the service enabled, isn't it?

agners: Ok, yeah, but this means we rely on the OS timesync being enabled and run before Supervisor starts (which is HAOS specific behavior). This wasn't always the case afaik. Also NetworkManager restart systemd-timesyncd when we get a NTP server via DHCP. Just saying, a more robust way of learning when NTP syncs happen(ed) wouldn't be wrong. Or rather, clarifying what the assumptions are here (systemd-timescynd got started and attempted to sync, but seemingly failed to do so).

agners: Btw, this also means we really trust whoami to be right. Technically the whoami service could be wrong too... So if NTP claims it got synced, I'd rather trust that.

sairon: > Btw, this also means we really trust whoami to be right. Technically the whoami service could be wrong too... So if NTP claims it got synced, I'd rather trust that.

We already do that, the function bails out on several places if `self.sys_host.info.dt_synchronized`.

supervisor/core.py
mdegat01: Is this info available from the systemd unit service? Maybe its state property? Because we do already have a helper for this:
https://github.com/home-assistant/supervisor/blob/7ac900da834c9f28742909bc96db9c805bb135fc/supervisor/dbus/systemd.py#L54-L78

If the state tells us what we need to know here something like this could work:
```py
timesync_service = self.sys_dbus.systemd.get_unit('systemd-timesyncd.service')
async with timesync_service.properties_changed() as listener:
  while (props := await listener.wait_for_signal()):
    if props.get("ActiveState") == UnitActiveState.INACTIVE:
      break
```
Not totally sure if that's the right property name but you get the idea, at least as long as something in the unit object tells us what we need to know. Probably should wrap it in a timeout so it doesn't wait indefinitely as well.

Should also wrap the call to stop the unit in the `async with` that creates the listener so we don't have a race condition where we miss the signal if it stops too fast. Or just add a check to see if the state is already stopped before starting the loop.

supervisor/core.py
agners: Oh right. Nice!

supervisor/core.py
agners: Yeah we'd make sure to listen before we send the command to avoid race conditions.

The service stopped within 300ms on my Raspberry Pi 3, so I think 1 second is pretty safe here.

agners (approved):
LGTM, thanks!

supervisor/core.py
agners: Turns out there are quiet some users which run into issues with the fixed sleep. I've implemented the signal based wait for systemd unit as a helper and made use of it for this case as well: https://github.com/home-assistant/supervisor/pull/6715.

agners:
This also shows up with Home Assistant Green shipped with HAOS 16.3, when NTP is not available:

```
2025-06-25 18:38:45.146 INFO (MainThread) [supervisor.os.manager] Detect Home Assistant Operating System 16.3 / BootSlot A
2025-06-25 18:38:45.146 DEBUG (MainThread) [supervisor.utils.whoami] Check whoami to verify connectivity/system with: https://services.home-assistant.io/whoami/v1
2025-06-25 18:38:45.225 WARNING (MainThread) [supervisor.utils.whoami] Whoami service failed with SSL verification: Cannot connect to host services.home-assistant.io:443 ssl:True [SSLCertVerificationError: (1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: certificate is not yet valid (_ssl.c:1032)')]
2025-06-25 18:38:45.226 INFO (MainThread) [supervisor.core] Whoami service SSL error
2025-06-25 18:38:45.226 DEBUG (MainThread) [supervisor.utils.whoami] Check whoami to verify connectivity/system with: http://services.home-assistant.io/whoami/v1
2025-06-25 20:38:45.251 INFO (MainThread) [supervisor.host.control] Setting host timezone: Europe/Zurich
2025-06-25 20:38:45.251 DEBUG (MainThread) [supervisor.utils.dbus] D-Bus call - org.freedesktop.timedate1.call_set_timezone on /org/freedesktop/timedate1
2025-06-25 20:38:45.256 DEBUG (MainThread) [supervisor.utils.dbus] Property change for org.freedesktop.timedate1-/org/freedesktop/timedate1: ['Timezone'] changed & [] invalidated
2025-06-25 20:38:45.257 DEBUG (MainThread) [supervisor.utils.dbus] D-Bus call - org.freedesktop.DBus.Properties.call_get_all on /org/freedesktop/timedate1
2025-06-25 20:38:45.273 WARNING (MainThread) [supervisor.core] System time/date shift over more than 3 days found!
2025-06-25 20:38:45.273 INFO (MainThread) [supervisor.host.control] Setting new host datetime: 2026-04-23T12:10:30+00:00
2025-06-25 20:38:45.274 DEBUG (MainThread) [supervisor.utils.dbus] D-Bus call - org.freedesktop.timedate1.call_set_time on /org/freedesktop/timedate1
2025-06-25 20:38:45.288 DEBUG (MainThread) [supervisor.utils.dbus] D-Bus fast error on call - org.freedesktop.timedate1.call_set_time on /org/freedesktop/timedate1: Automatic time synchronization is enabled
2025-06-25 20:38:45.289 CRITICAL (MainThread) [supervisor.core] Fatal error happening on load Task <coroutine object Core._adjust_system_datetime at 0xffff9dd96020>: Automatic time synchronization is enabled
  File "/usr/src/supervisor/supervisor/core.py", line 188, in setup
    await setup_task
  File "/usr/src/supervisor/supervisor/core.py", line 431, in _adjust_system_datetime
    await self.sys_host.control.set_datetime(data.dt_utc)
  File "/usr/src/supervisor/supervisor/host/control.py", line 83, in set_datetime
    await self.sys_dbus.timedate.set_time(new_time)
  File "/usr/src/supervisor/supervisor/dbus/timedate.py", line 107, in set_time
    await self.connected_dbus.call(
        "set_time", int(utc.timestamp() * 1000000), False, False
    )
  File "/usr/src/supervisor/supervisor/utils/dbus.py", line 141, in call_dbus
    raise DBus.from_dbus_error(err) from None
supervisor.exceptions.DBusFatalError: Automatic time synchronization is enabled

```

With no access to NTP on these systems time never gets into a valid range, causing Supervisor to fail the connectivity check and therefor skip downloading Core forever:

```
2025-06-25 20:48:51.868 DEBUG (MainThread) [supervisor.supervisor] Supervisor Connectivity check failed: Cannot connect to host checkonline.home-assistant.io:443 ssl:True [SSLCertVerificationError: (1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: certificate is not yet valid (_ssl.c:1032)')]
2025-06-25 20:48:51.869 DEBUG (MainThread) [supervisor.bus] Fire event 'supervisor_job_end' with 'SupervisorJob(created=datetime.datetime(2025, 6, 25, 18, 48, 51, 765581, tzinfo=datetime.timezone.utc), uuid='d7eca8dff8df4fba9b1e8c16473dee59', name='supervisor_check_connectivity', reference=None, progress=100, stage=None, parent_id='4897578329234b7c9b9fca0e858e43fa', done=False, on_change=<bound method JobManager._on_job_change of <supervisor.jobs.JobManager object at 0xffff9c8b0830>>, internal=False, errors=[], release_event=None, extra=None, child_job_syncs=None, parent_job_syncs=[])'
```

PR #6650, opened 2026-03-19
Add firewall rules to protect Docker gateway from external access

agners:
## Proposed change

Add iptables rules via a systemd transient unit to drop traffic addressed to the bridge gateway IP from non-bridge interfaces.

The firewall manager waits for the transient unit to complete and verifies success via D-Bus property change signals. On failure, the system is marked unhealthy and host-network add-ons are prevented from booting.

## Type of change

- [ ] Dependency upgrade
- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (which adds functionality to the supervisor)
- [ ] Breaking change (fix/feature causing existing functionality to break)
- [ ] Code quality improvements to existing code or addition of tests

## Additional information

- This PR fixes or closes issue: fixes #
- This PR is related to issue: 
- Link to documentation pull request: https://github.com/home-assistant/home-assistant.io/pull/44251
- Link to cli pull request: https://github.com/home-assistant-libs/python-supervisor-client/pull/296
- Link to client library pull request:

## Checklist

- [ ] The code change is tested and works locally.
- [ ] Local tests pass. **Your PR cannot be merged unless tests pass**
- [ ] There is no commented out code in this PR.
- [ ] I have followed the [development checklist][dev-checklist]
- [ ] The code has been formatted using Ruff (`ruff format supervisor tests`)
- [ ] Tests have been added to verify that the new code works.

If API endpoints or add-on configuration are added/changed:

- [ ] Documentation added/updated for [developers.home-assistant.io][docs-repository]
- [ ] [CLI][cli-repository] updated (if necessary)
- [ ] [Client library][client-library-repository] updated (if necessary)

[dev-checklist]: https://developers.home-assistant.io/docs/en/development_checklist.html
[docs-repository]: https://github.com/home-assistant/developers.home-assistant
[cli-repository]: https://github.com/home-assistant/cli
[client-library-repository]: https://github.com/home-assistant-libs/python-supervisor-client/

PR #6658, opened 2026-03-23
Disable hassio gateway protection rules in dev mode

sairon:
## Proposed change

Since there's no Systemd guaranteed to be available for Devcontainer to create the transient unit, we need to skip creating of the firewall rules when Supervisor is started in dev mode.

See: 

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR is related to issue: https://github.com/home-assistant/supervisor/pull/6650#issuecomment-4097583552
- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/host/firewall.py
+++ b/supervisor/host/firewall.py
@@ -153,6 +153,10 @@ class FirewallManager(CoreSysAttributes):
 
     async def apply_gateway_firewall_rules(self) -> None:
         """Apply gateway firewall rules, marking unsupported on failure."""
+        if self.sys_dev:
+            _LOGGER.info("Skipping gateway firewall rules in development mode")
+            return
+
         if await self._apply_gateway_firewall_rules():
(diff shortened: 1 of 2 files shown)


agners (approved):
It is fine to have the rules not apply for development environment, so this fixes devcontainer for now.

That said, I don't like that our devcontainer environment diverges from real deployments more and more. The reason the rules don't apply is that there is no systemd:

```
2026-03-23 08:57:15.772 ERROR (MainThread) [supervisor.host.firewall] Systemd not available, cannot apply gateway firewall rules
```

`iptables` is present and it would be possible to apply the rules from a container permission standpoint. Also, since the rules are limited to `hassio` network, I think it would be acceptable on a dev system as well.

We should explore to enable systemd in devcontainer.

But for now, this fixes the devcontainer, so let's move forward with this.

PR #6674, opened 2026-03-27
Include Docker registry configurations in backups

### LINKED ISSUE #6671: Use Supervisor specific inner tar to backup Docker registries
### Description

With #5605 we removed the registry configuration from the main backup metadata file so we don't have to extra encrypt these information. But with that, we no longer backup any stored Docker registry configuration.

With the introduction of backups for network mounts in PR #6411, we've added a Supervisor specific tar to the backup, just like the PR description in #5605 suggested. So with that, we can add a Docker specific json file into that inner tar, backing up Docker configuration as well.

### Additional context

_No response_

agners:
## Proposed change

With #5605 Docker registry credentials were removed from backup metadata (`backup.json`) to avoid exposing secrets in unencrypted data. With the introduction of the encrypted `supervisor.tar` inner archive in #6411 for network mount configurations, we now have a secure place to store sensitive Supervisor configuration.

This PR adds `docker.json` alongside `mounts.json` inside the encrypted `supervisor.tar`, restoring the ability to backup and restore Docker registry configurations.

On restore, registries from the backup are merged with any existing registries. Old backups without `docker.json` are handled gracefully via backward-compatible extraction.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR fixes or closes issue: fixes #6671
- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/backups/backup.py
+++ b/supervisor/backups/backup.py
@@ -984,7 +984,7 @@ class Backup(JobGroup):
             ],
         }
 
-        docker_data = {ATTR_REGISTRIES: dict(registries)}
+        docker_data = {ATTR_REGISTRIES: registries}
 
         outer_secure_tarfile = self._outer_secure_tarfile
         tar_name = f"supervisor.tar{'.gz' if self.compressed else ''}"
--- a/supervisor/docker/manager.py
+++ b/supervisor/docker/manager.py
@@ -224,7 +224,7 @@ class DockerConfig(FileConfiguration):
     @property
     def registries(self) -> dict[str, Any]:
         """Return credentials for docker registries."""
-        return self._data.get(ATTR_REGISTRIES, {})
+        return self._data[ATTR_REGISTRIES]
 
     def get_registry_for_image(self, image: str) -> str | None:
         """Return the registry name if credentials are available for the image.


supervisor/backups/backup.py
```diff
@@ -980,6 +984,8 @@ async def store_supervisor_config(self) -> None:
             ],
         }
 
+        docker_data = {ATTR_REGISTRIES: dict(registries)}
```
mdegat01: Why do we need the `dict` here? The typing for `self.sys_docker.config.registries` says it returns `dict[str, Any]`

supervisor/backups/backup.py
```diff
@@ -1056,44 +1082,64 @@ def _load_mounts_data() -> dict[str, Any] | None:
...
+                docker_data = SCHEMA_DOCKER_CONFIG(docker_data)
+            except vol.Invalid as err:
+                _LOGGER.warning("Invalid docker data in supervisor config: %s", err)
+                success = False
+                docker_data = None
+
+        if docker_data:
+            registries = docker_data.get(ATTR_REGISTRIES, {})
+            if registries:
+                self.sys_docker.config.registries.update(registries)
```
mdegat01, line 1139: So this is interesting. I was going to say that this may not work because the property does this:
https://github.com/home-assistant/supervisor/blob/1fd78dfc4e04562ff7a31125d200878eb06147dc/supervisor/docker/manager.py#L224-L228

And if `ATTR_REGISTRIES` is not in `_data` you're updating a new temporary object and not `_data`. But when I looked at the API to see how adding a registry that way is handled I realized it has the same issue:
https://github.com/home-assistant/supervisor/blob/1fd78dfc4e04562ff7a31125d200878eb06147dc/supervisor/api/docker.py#L114-L123

I guess the reason this was never a bug is because the schema never allowed that field to go missing:
https://github.com/home-assistant/supervisor/blob/1fd78dfc4e04562ff7a31125d200878eb06147dc/supervisor/validate.py#L182-L198

So we probably should just update the property to this:
```py
    @property
    def registries(self) -> dict[str, Any]:
        """Return credentials for docker registries."""
        return self._data[ATTR_REGISTRIES]
```
Because if that field somehow was actually missing its going to be broken regardless of whether we raise `ValueError` here or not.

agners: Uh yeah, nice catch! Updated accordingly!

In general, I don't like all those properties which do addional logic as part of the getter or setter. It's considered bad practice. At one point we should go through them en eliminate as much as reasonably possible.

PR #6694, opened 2026-04-01
Make app builds work without build.yaml

sairon:
## Proposed change

The builds are moving build configuration into the Dockerfile itself (base image defaults via `ARG`, labels via `LABEL`). This change makes `build.yaml` optional for local app builds while preserving backward compatibility for apps that still define it.

Key changes:
* Track whether `build.yaml` was found, log a warning if it is.
* Skip `BUILD_FROM` build arg when no build file exists, letting the Dockerfile's own `ARG BUILD_FROM=...` default take effect.
* Always include all configured registry credentials in docker config instead of matching only the base image's registry.
* Only set `io.hass.name` and `io.hass.description` labels when non-empty, as they could be defined in the Dockerfile directly.
* Log a deprecation warning when `build.yaml` is present.

Refs home-assistant/epics#33

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,6 +1,8 @@
 {
   "name": "Supervisor dev",
-  "image": "ghcr.io/home-assistant/devcontainer:3-supervisor",
+  "image": "ghcr.io/home-assistant/devcontainer:4-supervisor",
+  "overrideCommand": false,
+  "remoteUser": "vscode",
   "containerEnv": {
     "WORKSPACE_DIRECTORY": "${containerWorkspaceFolder}"
   },
@@ -46,6 +48,8 @@
   },
   "mounts": [
     "type=volume,target=/var/lib/docker",
-    "type=volume,target=/mnt/supervisor"
+    "type=volume,target=/var/lib/containerd",
+    "type=volume,target=/mnt/supervisor",
+    "type=tmpfs,target=/tmp"
   ]
 }
--- a/.github/workflows/builder.yml
+++ b/.github/workflows/builder.yml
@@ -35,8 +35,6 @@ on:
 env:
   DEFAULT_PYTHON: "3.14.3"
   COSIGN_VERSION: "v2.5.3"
-  CRANE_VERSION: "v0.20.7"
-  CRANE_SHA256: "8ef3564d264e6b5ca93f7b7f5652704c4dd29d33935aff6947dd5adefd05953e"
   BUILD_NAME: supervisor
   BUILD_TYPE: supervisor
   IMAGE_NAME: hassio-supervisor
@@ -161,12 +159,6 @@ jobs:
           path: wheels
           retention-days: 1
 
-      - name: Set version
-        if: needs.init.outputs.publish == 'true'
-        uses: home-assistant/actions/helpers/version@master
-        with:
-          type: ${{ env.BUILD_TYPE }}
-
       - name: Set up Python ${{ env.DEFAULT_PYTHON }}
         if: needs.init.outputs.publish == 'true'
         uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
@@ -276,7 +268,7 @@ jobs:
           image: ghcr.io/home-assistant/amd64-hassio-supervisor
           image-tags: runner
           load: true
-          version: runner
+          version: ${{ needs.init.outputs.version }}
 
       # Pull the Supervisor for publish runs to test the published image
       - name: Pull Supervisor
--- a/Dockerfile
+++ b/Dockerfile
@@ -40,9 +40,11 @@ RUN \
         ${LOCAL_WHEELS:+--find-links $LOCAL_WHEELS}
 
 # Install Home Assistant Supervisor
+ARG BUILD_VERSION="9999.09.9.dev9999"
 COPY . supervisor
 RUN \
-    uv pip install --no-cache -e ./supervisor \
+    sed -i "s/^SUPERVISOR_VERSION =.*/SUPERVISOR_VERSION = \"${BUILD_VERSION}\"/g" /usr/src/supervisor/supervisor/const.py \
+    && uv pip install --no-cache -e ./supervisor \
     && python3 -m compileall ./supervisor/supervisor
 
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -25,5 +25,5 @@ securetar==2026.2.0
 sentry-sdk==2.57.0
 setuptools==82.0.1
 voluptuous==0.16.0
-dbus-fast==4.0.0
+dbus-fast==4.0.4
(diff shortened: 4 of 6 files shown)


supervisor/addons/build.py
```diff
@@ -48,20 +54,29 @@ def __init__(self, coresys: CoreSys, addon: AnyAddon) -> None:
...
         """
         try:
-            return find_one_filetype(
+            result = find_one_filetype(
                 self.addon.path_location, "build", FILE_SUFFIX_CONFIGURATION
             )
+            self._has_build_file = True
+            return result
         except ConfigurationFileError:
+            self._has_build_file = False
```
agners: Ugh yeah all bit meh here.

Maybe we should consider moving away from `FileConfiguration` for `AddonBuild`, since that is actually what this PR is trying to do: Not rely on `build.json`/`build.yaml` anymore 🤔 .

supervisor/addons/build.py
```diff
@@ -144,43 +161,33 @@ def build_is_valid() -> bool:
...
+            f"{stored[ATTR_USERNAME]}:{stored[ATTR_PASSWORD]}".encode()
+        ).decode()
 
-        Raises:
-            HassioArchNotFound: If the add-on is not supported on the current architecture.
+    def get_docker_config_json(self) -> str | None:
+        """Generate Docker config.json content with all configured registry credentials.
 
+        Returns a JSON string with registry credentials, or None if no registries
+        are configured.
```
agners, line 201: Exposes more credentials then strictly necessary to the builder, but works, so I am fine with it 🤷 

I wonder if there is really no other way to pass credentials to builder. But since that is rather recent, i guess this was really the only way to implement it.

supervisor/docker/addon.py
```diff
@@ -684,6 +684,13 @@ async def _build(self, version: AwesomeVersion, image: str | None = None) -> Non
         # Check if the build environment is valid, raises if not
         await build_env.is_valid()
 
+        if build_env.has_build_file:
+            _LOGGER.warning(
+                "Add-on %s uses build.yaml which is deprecated. "
+                "Move build parameters into the Dockerfile directly.",
```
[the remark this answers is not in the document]
sairon: And it can be `.yml` as well. We refer to the file as to `build.yaml` in docs too, I think it should be obvious and I'd prefer to keep it simple.

supervisor/addons/build.py
```diff
@@ -81,10 +96,12 @@ def arch(self) -> CpuArch:
...
 
     @property
-    def base_image(self) -> str:
-        """Return base image for this add-on."""
+    def base_image(self) -> str | None:
+        """Return base image for this add-on, or None to use Dockerfile default."""
         if not self._data[ATTR_BUILD_FROM]:
-            return f"ghcr.io/home-assistant/{self.arch!s}-base:latest"
+            if self._has_build_file:
+                return "ghcr.io/home-assistant/base:latest"
```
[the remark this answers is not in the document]
sairon: This is intentional, the images are now guaranteed to be multi-arch.

supervisor/addons/build.py
```diff
@@ -144,43 +161,33 @@ def build_is_valid() -> bool:
...
-        )
-
-        config = {"auths": {registry_key: {"auth": auth_string}}}
-
-        return json.dumps(config)
+        auths = {
+            self._registry_key(registry): {"auth": self._registry_auth(registry)}
+            for registry in self.sys_docker.config.registries
+        }
+        return json.dumps({"auths": auths})
```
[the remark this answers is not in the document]
sairon: IMHO there are potentially more risks if someone installs builds shady Dockerfiles and I don't see a simple solution for this (unless we want to parse the Dockerfile to extract the image). In the end the authentication to a single registry may not be needed at all and we still supply the credentials.

supervisor/addons/build.py
```diff
@@ -203,27 +210,35 @@ def get_docker_args(
...
+            LABEL_TYPE: META_APP,
             **self.additional_labels,
         }
 
+        # Set name only if non-empty, could have been set in Dockerfile
+        if name := self._fix_label("name"):
+            labels[LABEL_NAME] = name
+
+        # Set description only if non-empty, could have been set in Dockerfile
+        if description := self._fix_label("description"):
```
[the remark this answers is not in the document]
sairon: This seems like an unlikely edge case, and in the end it's IMO better for the dedicated fields from the config to take precedence.

supervisor/addons/build.py
```diff
@@ -203,27 +210,35 @@ def get_docker_args(
...
             build_cmd.extend(["--label", f"{key}={value}"])
 
         build_args = {
-            "BUILD_FROM": self.base_image,
             "BUILD_VERSION": version,
             "BUILD_ARCH": self.arch,
             **self.additional_args,
         }
 
+        if self.base_image is not None:
```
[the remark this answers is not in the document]
sairon: Also quite unlikely, supplying `BUILD_FROM` in `args` would be an anti-pattern.

supervisor/addons/build.py
sairon: I wonder if this is really a problem. The builder is actually the official `docker` image from the Docker Hub and the config file with credentials is supplied to that. When you run an image build on a local machine, the BuildKit builder has access to all the credentials in Docker config too, but that doesn't mean it can exfiltrate them - or does it? It should only allow to do operations on the registry that require authentication. 

The credentials are copied to a temporary file on the host but there you have limited accessibility as well, and you can access other files containing them, so I don't see this an issue either.

supervisor/addons/build.py
sairon: Not sure I follow but the meh try-except allowing the file not to be present was here before as well. Majority of apps will keep using the structured config for quite a while, so not using `FileConfiguration` now would require us to handle it in different manner anyway.

agners: Well, a `build.json`/`build.yaml` was so far essentially mandatory. The code handled a graceful fallback which lead to using a empty dict, which ultimately applied the default schema of that config file.

In my books the old code was already a bit funky, since `FileConfiguration` is mostly used for Supervisor internal configuration files. But whatever, reuse I guess 🤷 

However, now we want to get rid of `build.json`/`build.yaml`. Still relying on `FileConfiguration` always, let the system always come up with a empty build config, and then not use it seems just the wrong way around.

So I'd say, lets bite the bullet, and move away from `FileConfiguration` for `AddonBuild` now.

supervisor/addons/build.py
agners: I guess the only way you could (miss)use the other credentials if you use multi-staged builds with other `FROM`s... But you could argue that is a feature, that we support multi-stage builds from different private repositories 😅 🤷 .

As I said, I am fine with it. Just thoughts.

PR #6700, opened 2026-04-02
Prevent Sentry event feedback loop from warning handler

agners:
## Proposed change

The custom warning_handler sends warnings to Sentry via capture_exception. When the Sentry SDK's own HTTP transport triggers a warning (e.g. urllib3 InsecureRequestWarning due to broken CA certs), this creates a self-inducing feedback loop: the warning is captured, sent to Sentry, which triggers another warning, and so on.

Skip capture_exception for warnings originating from Sentry SDK background threads (identified by the "sentry-sdk." thread name prefix). Warnings are still logged normally via _LOGGER.warning.

A dedicated test validates that the SDK's BackgroundWorker thread uses the expected naming convention, so any future SDK change will be caught.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/bootstrap.py
+++ b/supervisor/bootstrap.py
@@ -7,6 +7,7 @@ from importlib import import_module
 import logging
 import os
 import signal
+import threading
 import warnings
 
 from colorlog import ColoredFormatter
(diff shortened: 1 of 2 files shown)


agners:
There is only one user which triggers this warning, most likely this is a corrupted Supervisor container image.

```
InsecureRequestWarning: Unverified HTTPS request is being made to host 'o427061.ingest.sentry.io'. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/latest/advanced-usage.html#tls-warnings
```

agners:
> Good catch. Fix makes sense and tests are good although I guess for clarity I would've expected a test confirming the desired behavior rather then just implementation details? Is it possible to mock the situation they're running into in a test and confirm looping does not occur?

I haven't tried. This is most likely a corruption of some sort. Most likely the next corruption looks slightly different yet again :sweat_smile: . Yes the test are a bit synthetic, but it validates the Thread name using the Sentry SDK `BackgroundWorker` class, so any change to how this works the SDK should trigger a test error. This was a single user problem, I think what we have is good enough for now.

FWIW, Exceptions are already execluded via `capture_internal_exceptions`. The warning capture is really a Supervisor addition, hence we need this Sentry SDK exclusion on our end.

PR #6704, opened 2026-04-07
Let write-side OSError propagate during backup creation

agners:
## Proposed change

When a write-side `OSError` (e.g. ENOSPC, host down) occurs during backup creation, the outer tar file is left in a structurally corrupt state. Securetar's `create_tar` context manager uses a two-phase header write: it writes a placeholder header on enter and seeks back to rewrite it with the actual size on exit. If an `OSError` occurs mid-write, the inner tar entry has truncated data and a stale placeholder header.

Previously, `_folder_save` wrapped `OSError` as `BackupError`, which `store_folders` then caught and swallowed -- allowing the backup to continue writing to an already corrupt tar. Similarly, `_create_finalize` silently swallowed `OSError` when writing `backup.json`, and the `finally` block in `create()` could raise a secondary `OSError` from `_close_outer_tarfile` that replaced the original exception. This secondary exception is what was captured in Sentry as SUPERVISOR-B53 (18k events, 470 users), SUPERVISOR-1FAJ (5.5k events, 595 users), SUPERVISOR-BJ4 (3.7k events), SUPERVISOR-18KS, and SUPERVISOR-1HE6.

This PR introduces `BackupFatalError` (a `BackupError` subclass) for write-side I/O errors. Using a dedicated subclass rather than letting raw `OSError` propagate avoids the job decorator treating it as an unhandled exception (which would send extra events to Sentry and wrap it as `JobException`). Since `BackupFatalError` is a `HassioError`, the job decorator handles it cleanly, and `_do_backup` catches it via `except BackupError` to delete the incomplete backup file.

Changes:
- Add `BackupFatalError` exception for write-side I/O errors
- In `create()`, use except/else instead of finally so finalization is skipped on error. On the error path, close the tar suppressing errors) to release the file handle; the file is unlinked by the caller
- In `_create_finalize`, raise `BackupFatalError` on `OSError` instead of swallowing it
- In `_folder_save` and `store_supervisor_config`, wrap `OSError` as `BackupFatalError` instead of `BackupError`
- In `store_folders` and `store_addons`, re-raise `BackupFatalError` instead of swallowing it like regular `BackupError`

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/backups/backup.py
+++ b/supervisor/backups/backup.py
@@ -52,7 +52,7 @@ from ..coresys import CoreSys
 from ..exceptions import (
     AddonsError,
     BackupError,
-    BackupFatalError,
+    BackupFatalIOError,
     BackupFileExistError,
     BackupFileNotFoundError,
     BackupInvalidError,
@@ -605,7 +605,7 @@ class Backup(JobGroup):
         try:
             await self.sys_run_in_executor(_add_backup_json)
         except OSError as err:
-            raise BackupFatalError(
+            raise BackupFatalIOError(
                 f"Can't write backup metadata: {err!s}", _LOGGER.error
             ) from err
         except json.JSONDecodeError as err:
@@ -668,7 +668,7 @@ class Backup(JobGroup):
             try:
                 if start_task := await self._addon_save(addon):
                     start_tasks.append(start_task)
-            except BackupFatalError:
+            except BackupFatalIOError:
                 raise
             except BackupError as err:
                 self.sys_jobs.current.capture_error(err)
@@ -798,7 +798,7 @@ class Backup(JobGroup):
             if await self.sys_run_in_executor(_save):
                 self._data[ATTR_FOLDERS].append(name)
         except OSError as err:
-            raise BackupFatalError(
+            raise BackupFatalIOError(
                 f"Can't write tarfile: {err!s}", _LOGGER.error
             ) from err
         except (tarfile.TarError, AddFileError) as err:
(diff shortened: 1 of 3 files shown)


supervisor/backups/backup.py
```diff
@@ -505,10 +506,20 @@ def _close_outer_tarfile() -> int:
 
         try:
             yield
-        finally:
+        except Exception:
+            self._outer_secure_tarfile = None
+            # Close may fail (e.g. ENOSPC writing end-of-archive
+            # markers), but tarfile's finally ensures the file handle
+            # is released regardless. The file is unlinked by the caller.
+            with suppress(Exception):
```
mdegat01, line 514: The fact that this doesn't cause a linter error requiring an ignore comment seems like something we need to fix 😆

supervisor/backups/backup.py
```diff
@@ -505,10 +506,20 @@ def _close_outer_tarfile() -> int:
 
         try:
             yield
-        finally:
+        except Exception:
```
mdegat01, line 509: Wait we definitely have a `too-broad-except` rule or something like that, I've had to disable it before. Is something wrong with our linter? This should require a disable rule comment to pass ci...

supervisor/exceptions.py
```diff
@@ -1113,6 +1113,10 @@ class BackupFileExistError(BackupError):
     """Raise if the backup file already exists."""
 
 
+class BackupFatalError(BackupError):
```
mdegat01: Maybe `BackupFatalIOError`? The name seems a bit generic compared to the docstring, might get accidentally re-used for unrelated things in the future.

supervisor/backups/backup.py
agners: Hm, just checked, so it seems that when the pattern is:

```py
    except Exception:
        ...
        raise
```

The rule doesn't apply. Which kinda make sense.

PR #6633, opened 2026-03-13
Increase Supervisor update check interval to 12 hours

agners:
## Proposed change

Increase the Supervisor automatic update check interval from ~8 hours (29100s) to 12 hours (43200s). This slows down the rollout of new Supervisor versions to reduce pressure on the GitHub Container Registry, which has been hitting rate limits (`toomanyrequests`) during update waves (see [SUPERVISOR-1B8B](https://home-assistant-io.sentry.io/issues/SUPERVISOR-1B8B)).

## Type of change

- [ ] Dependency upgrade
- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (which adds functionality to the supervisor)
- [ ] Breaking change (fix/feature causing existing functionality to break)
- [x] Code quality improvements to existing code or addition of tests

## Additional information

- This PR fixes or closes issue: fixes #6634
- This PR is related to issue: [SUPERVISOR-1B8B](https://home-assistant-io.sentry.io/issues/SUPERVISOR-1B8B)
- Link to documentation pull request:
- Link to cli pull request:
- Link to client library pull request:

## Checklist

- [x] The code change is tested and works locally.
- [x] Local tests pass. **Your PR cannot be merged unless tests pass**
- [x] There is no commented out code in this PR.
- [x] I have followed the [development checklist][dev-checklist]
- [x] The code has been formatted using Ruff (`ruff format supervisor tests`)
- [ ] Tests have been added to verify that the new code works.

If API endpoints or add-on configuration are added/changed:

- [ ] Documentation added/updated for [developers.home-assistant.io][docs-repository]
- [ ] CLI updated (if necessary)
- [ ] [Client library][client-library-repository] updated (if necessary)

[dev-checklist]: https://developers.home-assistant.io/docs/en/development_checklist.html
[docs-repository]: https://github.com/home-assistant/developers.home-assistant
[cli-repository]: https://github.com/home-assistant/cli
[client-library-repository]: https://github.com/home-assistant-libs/python-supervisor-client/

PR #6705, opened 2026-04-07
Consolidate Supervisor auto-update into updater reload task

agners:
## Proposed change

The `_update_supervisor` task (24h schedule) and `_reload_updater` (~7.5h schedule) both triggered the Supervisor auto-update, meaning updates were effectively checked every 7.5h. This undermined the intent of #6633 and #6638 to reduce update frequency and container registry pressure.

Consolidate both into a single code path: `_reload_updater` fetches version data and, when a new Supervisor version is found, calls `_auto_update_supervisor()` (with the same job conditions as the old task). The reload interval is increased from ~7.5h to 24h.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR is related to issue: #6633, #6638
- Link to cli pull request:

## Checklist

### DIFF
--- a/tests/misc/test_tasks.py
+++ b/tests/misc/test_tasks.py
@@ -1,5 +1,6 @@
 """Test scheduled tasks."""
 
+import asyncio
 from collections.abc import AsyncGenerator
 from shutil import copy
 from unittest.mock import AsyncMock, Mock, PropertyMock, patch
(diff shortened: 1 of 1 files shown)


agners:
> So the code looks fine here. But given how much time and issues this little bit of code has cost us we might want to add a test case? Something like starting the scheduler with the update method mocked and a mock response for the version file showing a new version of supervisor. Then time travel 24 hours and confirm the update method was called exactly once.

Unfortunately it wasn't quite straight forward: `time_machine` doesn't affect `time.monotonic()` which is what asyncio's `call_later` uses.

But I came up with a fairly solid solution by manipulating `loop.time`.

PR #6707, opened 2026-04-07
Handle add-on filesystem errors gracefully and reduce Sentry noise

agners:
## Proposed change

Handle `OSError` (e.g. errno 74 / EBADMSG) in add-on metadata reads (`long_description`, `refresh_path_cache`) gracefully instead of letting them bubble up as unhandled exceptions. A new translatable `AddonFileReadError` is raised after calling `check_oserror()` to mark the system unhealthy, giving API consumers a proper error response.

Additionally, in `core.py` `setup()`, skip Sentry reporting when the resolution system has already handled the error (detected by checking if a new unhealthy reason was added during task execution). This avoids flooding Sentry with filesystem corruption errors that aren't actionable for developers -- the user is already notified via the resolution system. The log level is also lowered from `critical` (which triggers Sentry via `LoggingIntegration`) to `error` without stack trace in that case.

SUPERVISOR-BC6 alone accounts for 548K Sentry events from a single user with a corrupt filesystem.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR fixes or closes issue: fixes SUPERVISOR-BC6, SUPERVISOR-BZJ
- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/addons/addon.py
+++ b/supervisor/addons/addon.py
@@ -1682,8 +1682,9 @@ class App(AppModel):
         ]:
             await self._restart_after_problem(event.state)
 
-    def refresh_path_cache(self) -> Awaitable[None]:
+    async def refresh_path_cache(self) -> None:
         """Refresh cache of existing paths."""
         if self.is_detached or not self.app_store:
-            return super().refresh_path_cache()
-        return self.app_store.refresh_path_cache()
+            await super().refresh_path_cache()
+        else:
+            await self.app_store.refresh_path_cache()
--- a/supervisor/addons/model.py
+++ b/supervisor/addons/model.py
@@ -2,7 +2,7 @@
 
 from abc import ABC, abstractmethod
 from collections import defaultdict
-from collections.abc import Awaitable, Callable
+from collections.abc import Callable
 from contextlib import suppress
 from datetime import datetime
 import logging
@@ -91,6 +91,7 @@ from ..const import (
 from ..coresys import CoreSys
 from ..docker.const import Capabilities
 from ..exceptions import (
+    AppFileReadError,
     AppNotSupportedArchitectureError,
     AppNotSupportedError,
     AppNotSupportedHomeAssistantVersionError,
@@ -682,9 +683,15 @@ class AppModel(JobGroup, ABC):
             # Return data
             return readme.read_text(encoding="utf-8", errors="replace")
 
-        return await self.sys_run_in_executor(read_readme)
+        try:
+            return await self.sys_run_in_executor(read_readme)
+        except OSError as err:
+            self.sys_resolution.check_oserror(err)
+            raise AppFileReadError(
+                _LOGGER.error, app=self.slug, error=str(err)
+            ) from err
 
-    def refresh_path_cache(self) -> Awaitable[None]:
+    async def refresh_path_cache(self) -> None:
         """Refresh cache of existing paths."""
 
         def check_paths():
@@ -693,7 +700,13 @@ class AppModel(JobGroup, ABC):
             self._path_changelog_exists = self.path_changelog.exists()
             self._path_documentation_exists = self.path_documentation.exists()
 
-        return self.sys_run_in_executor(check_paths)
+        try:
+            await self.sys_run_in_executor(check_paths)
+        except OSError as err:
+            self.sys_resolution.check_oserror(err)
+            raise AppFileReadError(
+                _LOGGER.error, app=self.slug, error=str(err)
+            ) from err
 
     def validate_availability(self) -> None:
         """Validate if app is available for current system."""
--- a/supervisor/core.py
+++ b/supervisor/core.py
@@ -185,17 +185,32 @@ class Core(CoreSysAttributes):
 
         # Execute each load task in secure context
         for setup_task in setup_loads:
+            unhealthy_before = self.sys_resolution.unhealthy.copy()
             try:
                 await setup_task
             except Exception as err:  # pylint: disable=broad-except
-                _LOGGER.critical(
-                    "Fatal error happening on load Task %s: %s",
-                    setup_task,
-                    err,
-                    exc_info=True,
-                )
+                # If the error already caused a new unhealthy reason to
+                # be set (e.g. via check_oserror), it was handled by the
+                # resolution system and the user has been notified. Skip
+                # capturing to Sentry in that case.
+                capture_exception = self.sys_resolution.unhealthy == unhealthy_before
+
                 self.sys_resolution.add_unhealthy_reason(UnhealthyReason.SETUP)
-                await async_capture_exception(err)
+
+                if capture_exception:
+                    _LOGGER.critical(
+                        "Fatal error happening on load Task %s: %s",
+                        setup_task,
+                        err,
+                        exc_info=True,
+                    )
+                    await async_capture_exception(err)
+                else:
+                    _LOGGER.error(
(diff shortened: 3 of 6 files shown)


agners:
This PR got a bit out of hand: I simply wanted to introduce a API error for those read cases. But since the same code is also used during `setup()`, I had to also come up with a solution for this codepath.

I was considering simply handling the new `AddonFileReadError` exception in `setup()` , but this didn't seem too scale well (we'd have too keep a list of all the exception which we consider handled during `setup()` :thinking: .

The solution I chose now is also a bit hacky, so not sure if it's really better.

Simply consider all `HassioError` as "handled" doesn't work, we have quite some Sentry reports where we maybe need to inform the user in one form or another:

```
● OK, here's the consolidated list of Sentry issues coming from core.py:setup():

  From _adjust_system_datetime:

  ┌─────────────────┬───────────────────────────────────────────────────────────────────┬────────┬────────────┐
  │      Issue      │                               Error                               │ Events │   Status   │
  ├─────────────────┼───────────────────────────────────────────────────────────────────┼────────┼────────────┤
  │ SUPERVISOR-92R  │ DBusFatalError: Automatic time synchronization is enabled         │ 14,285 │ unresolved │
  ├─────────────────┼───────────────────────────────────────────────────────────────────┼────────┼────────────┤
  │ SUPERVISOR-W4W  │ DBusFatalError: Failed to set time zone: Permission denied        │ 333    │ unresolved │
  ├─────────────────┼───────────────────────────────────────────────────────────────────┼────────┼────────────┤
  │ SUPERVISOR-15YG │ DBusInterfaceMethodError: Invalid or not installed time zone      │ 77     │ unresolved │
  ├─────────────────┼───────────────────────────────────────────────────────────────────┼────────┼────────────┤
  │ SUPERVISOR-1GF9 │ DBusSystemdNoSuchUnit: Unit systemd-timesyncd.service not loaded. │ 165    │ unresolved │
  ├─────────────────┼───────────────────────────────────────────────────────────────────┼────────┼────────────┤
  │ SUPERVISOR-WD6  │ HostNotSupportedError: No timedate D-Bus connection available     │ 2,003  │ unresolved │
  └─────────────────┴───────────────────────────────────────────────────────────────────┴────────┴────────────┘

  From OSManager.load → data_disk:

  ┌────────────────┬────────────────────────────────────────────┬────────┬────────────┐
  │     Issue      │                   Error                    │ Events │   Status   │
  ├────────────────┼────────────────────────────────────────────┼────────┼────────────┤
  │ SUPERVISOR-S5W │ DBusNotConnectedError (via wrap_dbus)      │ 1,081  │ unresolved │
  ├────────────────┼────────────────────────────────────────────┼────────┼────────────┤
  │ SUPERVISOR-CPN │ DBusNotConnectedError (via connected_dbus) │ 18,155 │ unresolved │
  └────────────────┴────────────────────────────────────────────┴────────┴────────────┘

  From AddonManager.load → asyncio.gather:

  ┌──────────────────────────────┬──────────────────────────────────────────────────────────┬────────┬────────────┐
  │            Issue             │                          Error                           │ Events │   Status   │
  ├──────────────────────────────┼──────────────────────────────────────────────────────────┼────────┼────────────┤
  │ SUPERVISOR-BZJ               │ OSError: [Errno 74] Bad message (icon.png .exists())     │ 57     │ resolved   │
  ├──────────────────────────────┼──────────────────────────────────────────────────────────┼────────┼────────────┤
  │ SUPERVISOR-VAX               │ OSError: [Errno 74] Bad message (DOCS.md .exists())      │ 12     │ resolved   │
  ├──────────────────────────────┼──────────────────────────────────────────────────────────┼────────┼────────────┤
  │ SUPERVISOR-1BA8              │ OSError: [Errno 74] Bad message (translations .exists()) │ 1      │ resolved   │
  ├──────────────────────────────┼──────────────────────────────────────────────────────────┼────────┼────────────┤
  │ Multiple JobException issues │ Docker attach/install failures during addon.load()       │ varies │ unresolved │
  └──────────────────────────────┴──────────────────────────────────────────────────────────┴────────┴────────────┘
  From StoreManager.load → store/data.py:

  ┌──────────────────────────────┬─────────────────────────────────────────────────────────────┬────────┬────────┐
  │            Issue             │                            Error                            │ Events │ Status │
  ├──────────────────────────────┼─────────────────────────────────────────────────────────────┼────────┼────────┤
  │ SUPERVISOR-9X5               │ TypeError: expected string or bytes-like object, got 'dict' │ recent │ -      │
  └──────────────────────────────┴─────────────────────────────────────────────────────────────┴────────┴────────┘
```

Thoughts?

mdegat01:
I mean my first though reading the list is we probably shouldn't be relying on `setup` to report the `HassioError` type exceptions we want to know about. Like for those errors from `_adjust_system_datetime`, `setup` is not the only thing that calls that. If those errors are something we (the supervisor dev team) and the user needs to know about then they should be logged and reported from that method. Otherwise when they occur while making changes on a running Supervisor from the API neither of us will be properly informed.

So yea my take would be using setup for reporting should be a last resort. If `HassioErrors` are not being handled and reported properly in the places they are being raised then lets fix that. `setup` should just make sure the ones that weren't handled (non-`HassioError` type exceptions) have some last resort logging and capturing.

We could also require each of these load methods be jobs with the annotation and then even that handling can be dropped since the `Job` decorator takes care of that already.

agners:
> I mean my first though reading the list is we probably shouldn't be relying on `setup` to report the `HassioError` type exceptions we want to know about. Like for those errors from `_adjust_system_datetime`, `setup` is not the only thing that calls that. 

Since #6739 we do have a similar `HassioError` bubbling up to API -> report to sentry logic. So these D-Bus errors probably are now captured in two variants: One which bubble up to `setup()`, and the one which bubble up to `api_process()`.

> If those errors are something we (the supervisor dev team) and the user needs to know about then they should be logged and reported from that method. 

As for dev team, I don't think that sprinkling `async_capture_exception` everywhere is a good idea long term.

We can't handle every weird situation there is, this lead to unmaintainable code. If some of these Sentry issues have quite few reports, we probably can just ignore those, at least for now.

The ones which are high volume we need to dive into deeper: Is there an underlying issue/problem which needs addressing, then obviously that needs to be tackled. If it is just a rather common error where the user can/needs to take action, then we need a repair issue (the error handled by this PR is such a case). But just because we've informed the user, the exception still exists, and we need to decide what to do with it. In `setup()` the answer today is report to Sentry, log, but otherwise ignore and continue. Which I think is the best we can do.

But the question for this PR I'd like to answer is: Today, in `setup()` how should we detect the ones we consider handled? 

For errors which surface to API level we essentially have the same question. There we use the `APIError` class and inheritance to mark handled exception. I don't think we can simply apply that for `setup()`: An exception as a response to an API call is usually well handled, but for `setup()` the same error needs different handling.

Looking at it today, I think keeping a list of exception classes in `setup()` which we don't report to Sentry probably the cleaner and most straight forward way.

mdegat01 (approved):
Yep, looks good. Couple tests would be a nice addition if you wanted to:
1. Mock a bad message error raise on those reads and check that we get unhealthy (and probably mock that any other OSError just results in the exception if that test doesn't exist)
2. Test the scenario detailed in Core.setup for when to capture exceptions by just mocking all of the loads and calling it directly. I know `Core.setup` is a bit challenging to test though so up to you.

mdegat01 (approved):

Issue #6634, opened 2026-03-13
Can't run the update


Dave-Jones-1963:
### Describe the issue you are experiencing

Failed to perform the action update/install. Error updating Home Assistant Supervisor: Update of Supervisor failed: toomanyrequests: retry-after: 718.474µs, allowed: 44000/minute

### What type of installation are you running?

Home Assistant OS

### Which operating system are you running on?

Home Assistant Operating System

### Steps to reproduce the issue

1.click on update under settings!
2.
3.
...

### Anything in the Supervisor logs that might be useful for us?

```txt
[supervisor.host.apparmor] Adding/updating AppArmor profile: hassio-supervisor
2026-03-13 18:47:30.208 INFO (MainThread) [supervisor.supervisor] Update Supervisor to version 2026.03.0
2026-03-13 18:47:30.532 INFO (MainThread) [supervisor.docker.interface] Downloading docker image ghcr.io/home-assistant/aarch64-hassio-supervisor with tag 2026.03.0.
2026-03-13 18:47:34.280 ERROR (MainThread) [supervisor.docker.manager] toomanyrequests: retry-after: 718.474µs, allowed: 44000/minute
2026-03-13 18:47:34.281 CRITICAL (MainThread) [supervisor.supervisor] Update of Supervisor failed: toomanyrequests: retry-after: 718.474µs, allowed: 44000/minute
```

### System information

System information
Version	core-2026.3.1
Installation type	Home Assistant OS
Development	false
Supervisor	true
Docker	true
Container architecture	aarch64
User	root
Virtual environment	false
Python version	3.14.2
Operating system family	Linux
Operating system version	6.12.47-haos-raspi
CPU architecture	aarch64
Time zone	Europe/London
Configuration directory	/config
Home Assistant Community Store
GitHub API	ok
GitHub Content	ok
GitHub Web	ok
HACS Data	ok
GitHub API Calls Remaining	5000
Installed Version	2.0.5
Stage	running
Available Repositories	2902
Downloaded Repositories	9
Home Assistant Cloud
Logged In	false
Reach certificate server	ok
Reach authentication server	ok
Reach Home Assistant Cloud	ok
Home Assistant Supervisor
Host operating system	Home Assistant OS 17.1
Update channel	stable
Supervisor version	supervisor-2026.02.3
Agent version	1.8.1
Docker version	29.1.3
Disk total	116.6 GB
Disk used	21.5 GB
Nameservers	192.168.3.250, 192.168.3.251
[...]

Issue #6570, opened 2026-02-16
Blocked updates. Many reported on FB


tureka:
### Describe the issue you are experiencing

Can't update Addons or Supervisor. Many reports on Facebook about similar situation.

### What type of installation are you running?

Home Assistant OS

### Which operating system are you running on?

Home Assistant Operating System

### Steps to reproduce the issue

1. Don't know
2.
3.
...

### Anything in the Supervisor logs that might be useful for us?

```txt
[139759724312064] Error during service call to update.install: Error updating Let's Encrypt: 'AddonManager.update' blocked from execution, supervisor needs to be updated first
[139759724312064] Error during service call to update.install: Error updating Home Assistant Supervisor: Update of Supervisor failed: Your IP address has made too many requests to Docker Hub which activated a rate limit. For more details see https://www.home-assistant.io/more-info/dockerhub-rate-limit
```

## System Information

version | core-2026.2.2
-- | --
installation_type | Home Assistant OS
dev | false
hassio | true
docker | true
container_arch | amd64
user | root
virtualenv | false
python_version | 3.13.11
os_name | Linux
os_version | 6.12.67-haos
arch | x86_64
timezone | Europe/Warsaw
config_dir | /config

<details><summary>Home Assistant Community Store</summary>

GitHub API | ok
-- | --
GitHub Content | ok
GitHub Web | ok
HACS Data | ok
GitHub API Calls Remaining | 5000
Installed Version | 2.0.5
Stage | running
Available Repositories | 2806
Downloaded Repositories | 31

</details>

<details><summary>Home Assistant Cloud</summary>

logged_in | false
-- | --
can_reach_cert_server | ok
can_reach_cloud_auth | ok
can_reach_cloud | ok

</details>

<details><summary>Home Assistant Supervisor</summary>

host_os | Home Assistant OS 17.1
-- | --
update_channel | stable
supervisor_version | supervisor-2026.02.1
agent_version | 1.8.1
docker_version | 29.1.3
disk_total | 30.8 GB
disk_used | 22.6 GB
nameservers | fd19:11ba:fefe:fa00::251, 10.0.0.251
healthy | true
supported | true
host_connectivity | true
supervisor_connectivity | true
ntp_synchronized | true
virtualization | kvm
board | ova
supervisor_api | ok
version_api | ok
[...]

PR #6732, opened 2026-04-13
Detect container registry rate limits uniformly

agners:
## Proposed change

Container registry rate limits reach Supervisor in three distinct shapes:

1. **HTTP 429** from the daemon — currently recognised, but the resulting exception and resolution issue are both hardcoded to "Docker Hub". Since Supervisor/Core/plugin images all live on `ghcr.io` now, virtually every 429 we see in the field is actually a GHCR throttle that we mislabel. Issue [`SUPERVISOR-16BK`](https://home-assistant-io.sentry.io/issues/SUPERVISOR-16BK) (`>115k events`, `>93k users`) is exactly this — the image in the event context is `ghcr.io/home-assistant/amd64-hassio-supervisor:latest`, yet the user sees a "log into Docker Hub" suggestion.
2. **HTTP 500 with `toomanyrequests` in the body** — not currently recognised. Docker daemons before `28.3.0` wrap an upstream 429 into a 500 to the client. This was fixed upstream by moby/moby commit [`23fa0ae74a`](https://github.com/moby/moby/commit/23fa0ae74a) ("Cleanup http status error checks")
3. **JSON error event during a streaming pull** — not currently recognised. `POST /images/create` returns `200 OK` and streams progress/error events, so rate limits that land during layer download arrive as plain text inside the stream and have no HTTP status to key off of. Happens on all recent daemon versions. Issues [`SUPERVISOR-13FQ`](https://home-assistant-io.sentry.io/issues/SUPERVISOR-13FQ) (`>16k events`) and [`SUPERVISOR-13E0`](https://home-assistant-io.sentry.io/issues/SUPERVISOR-13E0) (`>8k events`) are examples.

Cases 2 and 3 propagate as plain `DockerError`, bypass the 429 detection in `docker/interface.py:install()` entirely, never produce a `DOCKER_RATELIMIT` resolution issue, and generate large amounts of Sentry noise. Case 1 is handled but routes every GHCR 429 through Docker-Hub-specific messaging and suggestions.

This PR addresses all three shapes and splits the registry-specific handling so `ghcr.io` failures produce a new `GITHUB_RATELIMIT` issue with appropriate guidance (no misleading Docker Hub login suggestion), while Docker Hub failures keep their existing behaviour.

Summary of the changes:

- New `DockerRegistryRateLimitExceeded` base exception with `DockerHubRateLimitExceeded` and a new `GithubContainerRegistryRateLimitExceeded` as subclasses. All extend `APITooManyRequests` so callers and future retry logic can key off a single type.
- New `GITHUB_RATELIMIT` `IssueType` (no `REGISTRY_LOGIN` suggestion, since GHCR authentication is different from Docker Hub).
- `PullLogEntry.exception` now maps stream errors containing `toomanyrequests` to `DockerRegistryRateLimitExceeded` (case 3).
- `docker/interface.py:install()` routes all three cases through a single `_registry_rate_limit_exception()` helper that picks the right resolution issue, suggestion and exception subclass based on the image's registry.
- `utils/sentry.py` filters `APITooManyRequests` (and anything wrapping it via `__cause__`) in both `capture_exception` and `async_capture_exception`. Single policy point, every caller benefits, no per-site changes needed.

Callers (`supervisor.update()`, plugin manager, HA core update) are intentionally **unchanged** — `UPDATE_FAILED` issues still get created alongside the registry-specific rate limit issue, giving users both the cause (rate limit) and the effect (update failed) in the resolution center.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR is related to issue: #6634, #6570
- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/docker/interface.py
+++ b/supervisor/docker/interface.py
@@ -34,17 +34,26 @@ from ..exceptions import (
     DockerJobError,
     DockerNotFound,
     DockerRegistryAuthError,
+    DockerRegistryRateLimitExceeded,
+    GithubContainerRegistryRateLimitExceeded,
 )
 from ..jobs.const import JOB_GROUP_DOCKER_INTERFACE, JobConcurrency
 from ..jobs.decorator import Job
 from ..jobs.job_group import JobGroup
 from ..resolution.const import ContextType, IssueType, SuggestionType
 from ..utils.sentry import async_capture_exception
-from .const import DOCKER_HUB, DOCKER_HUB_LEGACY, ContainerState, RestartPolicy
+from .const import (
+    DOCKER_HUB,
+    DOCKER_HUB_LEGACY,
+    GITHUB_CONTAINER_REGISTRY,
+    ContainerState,
+    RestartPolicy,
+)
 from .manager import CommandReturn, ExecReturn, PullLogEntry
 from .monitor import DockerContainerStateEvent
 from .pull_progress import ImagePullProgress
 from .stats import DockerStats
+from .utils import get_registry_from_image
 
 _LOGGER: logging.Logger = logging.getLogger(__name__)
 
@@ -222,6 +231,32 @@ class DockerInterface(JobGroup, ABC):
 
         return credentials, qualified_image
 
+    def _registry_rate_limit_exception(
+        self, image: str
+    ) -> DockerRegistryRateLimitExceeded:
+        """Create resolution issue and return typed exception for a rate limit.
+
+        The registry is derived from the image reference. Docker Hub and GHCR
+        each have their own issue type and exception subclass so users get
+        actionable guidance (e.g. registry login for Docker Hub). Unknown
+        registries fall back to a generic rate limit exception with no issue.
+        """
+        registry = get_registry_from_image(image)
+        if registry == GITHUB_CONTAINER_REGISTRY:
+            self.sys_resolution.create_issue(
+                IssueType.GITHUB_RATELIMIT,
+                ContextType.SYSTEM,
+            )
+            return GithubContainerRegistryRateLimitExceeded(_LOGGER.warning)
+        if registry is None or registry in (DOCKER_HUB, DOCKER_HUB_LEGACY):
+            self.sys_resolution.create_issue(
+                IssueType.DOCKER_RATELIMIT,
+                ContextType.SYSTEM,
+                suggestions=[SuggestionType.REGISTRY_LOGIN],
+            )
+            return DockerHubRateLimitExceeded(_LOGGER.warning)
+        return DockerRegistryRateLimitExceeded(_LOGGER.warning)
+
     @Job(
         name="docker_interface_install",
         on_condition=DockerJobError,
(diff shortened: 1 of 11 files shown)


agners:
Maybe the `GITHUB_RATELIMIT` issue type is not really all that useful since not actionable :thinking: . Maybe simply log (for the task case)/or raise errors to the caller if it happens on a Supervisor API request?

supervisor/docker/interface.py
```diff
@@ -326,14 +361,25 @@ async def process_pull_event(event: PullLogEntry) -> None:
...
+            raise self._registry_rate_limit_exception(image) from err
         except aiodocker.DockerError as err:
-            if err.status == HTTPStatus.TOO_MANY_REQUESTS:
-                self.sys_resolution.create_issue(
-                    IssueType.DOCKER_RATELIMIT,
-                    ContextType.SYSTEM,
-                    suggestions=[SuggestionType.REGISTRY_LOGIN],
-                )
-                raise DockerHubRateLimitExceeded(_LOGGER.error) from err
+            # Pre-28.3.0 daemons wrap registry rate limits as HTTP 500
```
mdegat01, line 370: So I just checked and we updated HAOS to 28.3.0 at 16.0. So 15.2 is the last version of HAOS that used a pre-28.3.0 version of docker, released April 14th, 2025. Is this even in our support window anymore? We should put an expected version of Supervisor this can be dropped on in the comment here if so.

supervisor/utils/sentry.py
```diff
@@ -59,11 +60,27 @@ def init_sentry(coresys: CoreSys) -> None:
...
+    return False
+
+
 def capture_exception(err: BaseException) -> None:
     """Capture an exception and send to sentry.
 
     Must be called in executor.
     """
+    if _is_rate_limit(err):
+        return
```
mdegat01, line 83: Can we move this to our existing filter method here:
https://github.com/home-assistant/supervisor/blob/a504d857454a58b0510fb0fd8354f8088a43c963/supervisor/misc/filter.py#L44-L50

Then we can do one `isinstance` check like `isinstance(exc_value, (AppConfigurationError, APITooManyRequests))` instead of multiple as they're kind of expensive. Or even it can't be merged due to the `err = err.__cause__` bit it still feels like our filtering out of exception noise should all be in one place. Or is the `err = err.__cause__` why it can't be merged into there?

mdegat01 (approved):
Small things. Going to approve so you don't have to wait for another review as I assume the fixes are small (or possibly rejected if the merge I suggested in the second one is impossible). If significant changes result I'll take another look 👍

supervisor/docker/interface.py
agners: Technically it's not in our support window anymore. But when it comes to Docker we always have been a bit more forgiving since it's such a central part. We also have a separate evaluation which currently has `MIN_SUPPORTED_DOCKER` at `24.0.0`. We could make an argument to bump that too, but I don't see a big need for that right now, so I'd prefer to just leave it at that support level for now.

supervisor/utils/sentry.py
agners: It does need to peak into `__cause__`, and with that potentially also multiple `isinstance` checks. But I think its still worth doing, centralizing these type of filtering make sense :+1:

agners:
> Maybe the `GITHUB_RATELIMIT` issue type is not really all that useful since not actionable 🤔 . Maybe simply log (for the task case)/or raise errors to the caller if it happens on a Supervisor API request?

I've removed the issue type, since we did not really have a use case for it.

supervisor/misc/filter.py
```diff
@@ -43,11 +46,21 @@ def sanitize_url(url: str) -> str:
...
+        err: BaseException | None = exc_value
+        while err is not None:
+            if isinstance(err, (AppConfigurationError, APITooManyRequests)):
+                _LOGGER.debug(
+                    "Skipping Sentry event for %s: %s",
+                    type(err).__name__,
+                    exc_value,
+                )
+                return None
+            err = err.__cause__
```
mdegat01, line 63: I was going to suggest we use our existing helper `check_exception_chain` but realizing its a bit different:
https://github.com/home-assistant/supervisor/blob/53f84ec15bb1e61cd9b6cd71b622aef251bfaeab/supervisor/utils/__init__.py#L60-L72

I'm not entirely sure what the difference is between `__cause__` vs `__context__` even after looking it up tbh, seems a bit subtle. Would rather not risk strange issues I suppose if this already works.

mdegat01 (approved):
Looks good to me 👍

supervisor/misc/filter.py
agners: Hm, looked into it, for the case we care here both would work. But I'd actually prefer `__cause__`, which is set when `raise B() from a` is used, much more explicit. I think I'd suggest to change `check_exception_chain` and use it then, but I'd rather prefer to do this in a follow up PR. Merging this as is.

agners: Refactoring `check_exception_chain` and use it for Sentry filter as well with #6757.

PR #6737, opened 2026-04-14
Always pass BUILD_FROM build arg for app builds

agners:
## Proposed change

#6694 made `build.yaml` optional for local app builds, but stopped passing the `BUILD_FROM` build arg when no build config file exists. This broke apps whose Dockerfiles use `ARG BUILD_FROM` without a default value (e.g. `FROM ${BUILD_FROM}`), as the variable resolves to empty and Docker rejects the build.

This PR restores the previous behavior: `BUILD_FROM` is always passed as a build arg, defaulting to the architecture-specific base image (`ghcr.io/home-assistant/{arch}-base:latest`) when no `build.yaml`/`build.json` is present.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/addons/build.py
+++ b/supervisor/addons/build.py
@@ -121,11 +121,11 @@ class AppBuild(CoreSysAttributes):
         return self.sys_arch.match([self.app.arch])
 
     @property
-    def base_image(self) -> str | None:
-        """Return base image for this app, or None to use Dockerfile default."""
-        # No build config (otherwise default is coerced when reading the config)
+    def base_image(self) -> str:
+        """Return base image for this app."""
+        # No build config - use default base image for the architecture
         if not self._build_config.get(ATTR_BUILD_FROM):
-            return None
+            return f"ghcr.io/home-assistant/{self.arch!s}-base:latest"
 
         # Single base image in build config
         if isinstance(self._build_config[ATTR_BUILD_FROM], str):
@@ -249,13 +249,11 @@ class AppBuild(CoreSysAttributes):
             build_cmd.extend(["--label", f"{key}={value}"])
 
         build_args = {
+            "BUILD_FROM": self.base_image,
             "BUILD_VERSION": version,
             "BUILD_ARCH": self.arch,
             **self.additional_args,
         }
-
-        if self.base_image is not None:
-            build_args["BUILD_FROM"] = self.base_image
 
         for key, value in build_args.items():
             build_cmd.extend(["--build-arg", f"{key}={value}"])
--- a/tests/addons/test_build.py
+++ b/tests/addons/test_build.py
@@ -23,6 +23,16 @@ def _is_build_arg_in_command(command: list[str], arg_name: str) -> bool:
     return f"--build-arg {arg_name}=" in " ".join(command)
 
 
+def _get_build_arg_value(command: list[str], arg_name: str) -> str | None:
+    """Get the value of a build arg from docker command."""
+    prefix = f"{arg_name}="
+    for i, part in enumerate(command):
+        if part == "--build-arg" and i + 1 < len(command):
+            if command[i + 1].startswith(prefix):
+                return command[i + 1][len(prefix) :]
+    return None
+
+
 def _is_label_in_command(
     command: list[str], label_name: str, label_value: str = ""
 ) -> bool:
@@ -284,12 +294,12 @@ async def test_no_build_file_no_deprecation_warning(
     assert "uses build.yaml which is deprecated" not in caplog.text
 
 
-async def test_no_build_yaml_base_image_none(
+async def test_no_build_yaml_base_image_default(
     coresys: CoreSys, install_app_ssh: App, tmp_path: Path
 ):
-    """Test base_image is None when no build file exists."""
+    """Test base_image returns default when no build file exists."""
     dockerfile = tmp_path / "Dockerfile"
-    dockerfile.write_text("ARG BUILD_FROM=ghcr.io/home-assistant/base:latest\n")
+    dockerfile.write_text("ARG BUILD_FROM\nFROM ${BUILD_FROM}\n")
 
     with patch.object(
         type(install_app_ssh),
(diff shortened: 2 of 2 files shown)


agners:
An example of such an app is https://github.com/sabeechen/hassio-google-drive-backup/blob/master/hassio-google-drive-backup/Dockerfile.

tests/addons/test_build.py
```diff
@@ -23,6 +23,16 @@ def _is_build_arg_in_command(command: list[str], arg_name: str) -> bool:
     return f"--build-arg {arg_name}=" in " ".join(command)
 
 
+def _get_build_arg_value(command: list[str], arg_name: str) -> str | None:
```
sairon, line 26: AI really likes this algorithm but I got rid of it because of its complexity, extending `_is_build_arg_in_command` with `arg_value` to do it the same way as `_is_label_in_command` does would make it consistent and it simpler.

supervisor/addons/build.py
```diff
@@ -121,11 +121,11 @@ def arch(self) -> CpuArch:
...
     @property
-    def base_image(self) -> str | None:
-        """Return base image for this app, or None to use Dockerfile default."""
-        # No build config (otherwise default is coerced when reading the config)
+    def base_image(self) -> str:
+        """Return base image for this app."""
+        # No build config - use default base image for the architecture
         if not self._build_config.get(ATTR_BUILD_FROM):
-            return None
+            return f"ghcr.io/home-assistant/{self.arch!s}-base:latest"
```
sairon, line 128: We can/should simply use `ghcr.io/home-assistant/{self.arch!s}-base:latest` now and let Docker use the correct platform from the manifest (which Claude doesn't know, so it makes in unnecessary complex).

tests/addons/test_build.py
```diff
@@ -330,7 +340,10 @@ async def test_no_build_yaml_no_build_from_arg(
             build.get_docker_args, AwesomeVersion("1.0.0"), "test-image:1.0.0", None
         )
 
-    assert not _is_build_arg_in_command(args["command"], "BUILD_FROM")
+    assert _is_build_arg_in_command(args["command"], "BUILD_FROM")
+    assert _get_build_arg_value(args["command"], "BUILD_FROM") == (
```
sairon, line 344: Could be simplified to a single `_is_build_arg_in_command("BUILD_FROM", "ghcr.io/home-assistant/amd64-base:latest")` with the above suggestion (plus the arch prefix should be removed too).

sairon (changes requested):
The change was intentional, in #6694 which was marked as a breaking change I stated:
> * Skip `BUILD_FROM` build arg when no build file exists, letting the Dockerfile's own `ARG BUILD_FROM=...` default take effect.

By re-introducing this build arg, we'll break builds for modernized apps with Dockerfiles which use this pattern:
```
ARG BUILD_FROM=ghcr.io/home-assistant/base:latest
FROM $BUILD_FROM
```

Meaning this PR will be a breaking change for those Dockerfiles that not only declare, but also define the `BUILD_FROM` argument, overriding the desired base image with plain base, leaving to even more cryptic errors instead of an obvious `base name ($BUILD_FROM) should not be blank`.

I must say, I don't like the implicit passing of `ghcr.io/home-assistant/base:latest` which may lead to breakages over time by itself and we should move away from it.

If we want to make possible building legacy apps which haven't been updated, we must take a more sophisticated approach, some random ideas:

1. Detect that `BUILD_FROM` isn't declared (probably only possible by iterating over the lines in a Dockerfile, which feels fragile).
2. Use some other heuristics, like the last change of the Git repo when building locally (not sure we can simply pass this, and feels hacky).
3. State that `BUILD_FROM` argument MUST NOT be used in modernized builds.

Either way, we shouldn't merge this PR as-is. Also, Claude is missing the fact we have multi-arch base images now, so it reintroduces unnecessary legacy cruft by using the arch prefixes.

agners:
> By re-introducing this build arg, we'll break builds for modernized apps with Dockerfiles which use this pattern:
> 
> ```
> ARG BUILD_FROM=ghcr.io/home-assistant/base:latest
> FROM $BUILD_FROM
> ```
> 
> Meaning this PR will be a breaking change for those Dockerfiles that not only declare, but also define the `BUILD_FROM` argument, overriding the desired base image with plain base, leaving to even more cryptic errors instead of an obvious `base name ($BUILD_FROM) should not be blank`.


Hm, I see, we between a rock and a hard place here :cry: . I must admit, I wasn't really aware of the fact that we had apps without `build.yaml`, and this breaking change. I just run into it locally when an app stopped building ([eWeLink Smart Home](https://github.com/CoolKit-Technologies/ha-addon) to be specific, that particular add-on is anyways archived).

> 
> I must say, I don't like the implicit passing of `ghcr.io/home-assistant/base:latest` which may lead to breakages over time by itself and we should move away from it.
> 
> If we want to make possible building legacy apps which haven't been updated, we must take a more sophisticated approach, some random ideas:
> 
>     1. Detect that `BUILD_FROM` isn't declared (probably only possible by iterating over the lines in a Dockerfile, which feels fragile).
> 
>     2. Use some other heuristics, like the last change of the Git repo when building locally (not sure we can simply pass this, and feels hacky).
> 
>     3. State that `BUILD_FROM` argument MUST NOT be used in modernized builds.
> 
> 
> Either way, we shouldn't merge this PR as-is. Also, Claude is missing the fact we have multi-arch base images now, so it reintroduces unnecessary legacy cruft by using the arch prefixes.

Other examples I've found like the rather popular Home Assistant Google Drive Backup do have an image tag as well, so it won't be built by Supervisor by default and therefor no problematic.

But yeah, I agree not passing `BUILD_FROM` and keeping this as a breaking change is cleaner. I'd not go down the road parsing the Dockerfile, this seems a slippery slope. We could maybe parse the error output on build failure and advice how to resolve, but then again the default error seems rather clear to me:

```
2026-04-16 14:44:12.608 ERROR (MainThread) [supervisor.addons.addon] Could not build image for app f4f71350_ewelink_smart_home_slug: Docker build failed for f4f71350/amd64-addon-ewelink_smart_home_slug:1.4.6 (exit code 1). Build output:
#0 building with "default" instance using docker driver



#1 [internal] load build definition from Dockerfile

#1 transferring dockerfile: 566B done

#1 WARN: InvalidDefaultArgInFrom: Default value for ARG ${BUILD_FROM} results in empty or invalid base image name (line 2)

#1 DONE 0.0s



 1 warning found (use docker --debug to expand):

 - InvalidDefaultArgInFrom: Default value for ARG ${BUILD_FROM} results in empty or invalid base image name (line 2)

Dockerfile:2

--------------------

   1 |     ARG BUILD_FROM

   2 | >>> FROM ${BUILD_FROM}

   3 |     

   4 |     ENV LANG=C.UTF-8

--------------------

ERROR: failed to build: failed to solve: base name (${BUILD_FROM}) should not be blank
```

Not sure what these double new lines are about, those seem unnecessary to me :thinking:

agners:
> Not sure what these double new lines are about, those seem unnecessary to me 🤔

Addressed with #6743.

agners:
There are a couple things in developer docs which need updating, tracked with https://github.com/home-assistant/developers.home-assistant/issues/3001.

PR #6739, opened 2026-04-15
Log unexpected errors in api_process wrappers

agners:
## Proposed change

The `api_process` and `api_process_raw` decorators silently swallowed any `HassioError` that bubbled up from endpoint handlers. The wrapped handler caught the exception, returned an error response, but never logged anything. When the exception also had no message (e.g. `DBusNotConnectedError` raised without arguments by `@dbus_connected`), `api_return_error` falls back to `"Unknown error, see Supervisor logs"` — telling the caller to check logs that contain literally nothing about the failure.

Concretely, Core's `analytics.save_preferences` hitting `POST /supervisor/options` with `diagnostics=true` on a host without OS-Agent on the bus (e.g. the Supervisor devcontainer) produces:

```
aiohasupervisor.exceptions.SupervisorBadRequestError: Unknown error, see Supervisor logs
```

…with no corresponding Supervisor log line to go look at.

This PR logs the caught `HassioError` with traceback in both `api_process` and `api_process_raw` before delegating to `api_return_error`, so the "see Supervisor logs" hint is actually actionable. The `APIError` branch is intentionally left alone — those carry explicit status codes and messages set by Supervisor code and are already visible in the response, so logging them would just be noise (e.g. validation errors).

This does not fix the underlying missing message on `DBusNotConnectedError` itself (that's a follow-up), but it makes any future occurrence of this pattern diagnosable from the logs alone.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/api/utils.py
+++ b/supervisor/api/utils.py
@@ -3,6 +3,7 @@
 import asyncio
 from collections.abc import Callable, Mapping
 import json
+import logging
 from typing import Any, cast
 
 from aiohttp import web
@@ -32,6 +33,8 @@ from ..utils import check_exception_chain, get_message_from_exception_chain
 from ..utils.json import json_dumps, json_loads as json_loads_util
 from ..utils.log_format import format_message
 from . import const
+
+_LOGGER: logging.Logger = logging.getLogger(__name__)
 
 
 def extract_supervisor_token(request: web.Request) -> str | None:
@@ -72,6 +75,7 @@ def api_process(method):
                 err, status=err.status, job_id=err.job_id, headers=err.headers
             )
         except HassioError as err:
+            _LOGGER.exception("Unexpected error during API call: %s", err)
             return api_return_error(err)
 
         if isinstance(answer, (dict, list)):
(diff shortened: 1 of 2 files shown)


mdegat01 (approved):
I'm fine with this. It feels like the flaw might be with the point at which the exception is raised though. HassioError means handled and if logging to inform the user is needed we should do that at the point we raise it right?

agners:
> I'm fine with this. It feels like the flaw might be with the point at which the exception is raised though. HassioError means handled and if logging to inform the user is needed we should do that at the point we raise it right?

"HassioError means handled" is maybe a bit overstatement. I means it is an exception created by us, but some of them are quite low level and come without any text (like the `DBusNotConnectedError` case showed).

IMHO, for the well handled exception cases we should have exception classes deriving from `APIError` and those should have user friendly strings. I guess we have quite some of them covered. What is left are `HassioError` derrived exception. Those are either cases where we just missed adding a good exception class (we should find those) or very rare exception we don't want to handle (e.g. happen due to very specific set of rare circumstances like corruption or weird system setup).

This PR is about `HassioError`. And I am tempted to report them to Sentry, but I fear the volume of such exception is maybe to high :thinking: .

mdegat01 (approved):
Ok I get it now. `HassioError` captured after `APIError` so its specifically to catch the non-`APIError` exceptions raised. This makes sense then.

As for sentry noise, I don't really know how much this will be. I think passing them to sentry and letting it aggregate them for us provides us the best way to track them down and improve handling so I'd say lets try it. If its clearly too noisy we'll drop it if necessary but I'd rather just quickly address the main problem points if possible with better logging and an APIError instead.

PR #6741, opened 2026-04-15
Add versioned v2 API with apps terminology

mdegat01:
## Proposed change

Introduce a versioned `/v2` API for the Supervisor that uses **apps** terminology consistently, while keeping the v1 API fully backward-compatible. Builds on the internal code rename from PR #6717.

The rebranding of addons → apps requires breaking API changes (path and field renames). Rather than making these in-place and breaking existing clients, a versioned v2 API allows both old and new clients to work during a transition period.

### Design

- **aiohttp sub-app** mounted at `/v2` — clean URL prefix with isolated routing and no code duplication
- **Feature flag gated** — `FeatureFlag.SUPERVISOR_V2_API` must be enabled; v2 routes are registered at `load()` time only when the flag is set
- **Thin v1 wrappers** — v1 endpoints delegate to v2 canonical methods, translating field names where needed (`apps` ↔ `addons`)
- **Strict per-version security regexes** — `_V1_PATTERNS` and `_V2_PATTERNS` match only their respective paths; no cross-version fallback
- **Shared business logic helpers** — `_rename_apps_to_addons_in_backups()`, `_process_location_in_body()`, `_all_store_apps_info()`, `_info_data()` eliminate duplication

### Path Mapping

| v1 | v2 |
|----|----|
| `/addons/{slug}/...` | `/v2/apps/{slug}/...` |
| `/store/addons/{slug}/...` | `/v2/store/apps/{slug}/...` |
| `/backups/...` (`addons` field) | `/v2/backups/...` (`apps` field) |
| `/store` (`addons` field) | `/v2/store` (`apps` field) |

### Constant Split

- `ATTR_APPS = "apps"` — v2 API responses/requests
- `ATTR_ADDONS = "addons"` — v1 API responses and backup file format (must never change — old backups would be unreadable)

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist


supervisor/api/middleware/security.py
```diff
@@ -30,13 +31,14 @@
...
 
-_CORE_FRONTEND_PATHS: Final = (
+_V1_FRONTEND_PATHS: Final = (
     r"|/app/.*\.(?:js|gz|json|map|woff2)"
     r"|/(store/)?addons/" + RE_SLUG + r"/(logo|icon)"
 )
 
-CORE_FRONTEND: Final = re.compile(
-    r"^(?:" + _CORE_FRONTEND_PATHS + r")$"
+_V2_FRONTEND_PATHS: Final = (
```
agners, line 39: Hm, I think we can drop frontend from v2 from the getgo.

supervisor/api/middleware/security.py
```diff
@@ -162,9 +77,194 @@
...
+            r"|/backups.*"
+            r")$"
+        ),
+        ROLE_MANAGER: re.compile(
+            r"^/v2(?:"
+            r"|/.+/info"
+            r"|/apps(?:/" + RE_SLUG + r"/(?!security).+|/reload)?"
+            r"|/audio/.+"
+            r"|/auth/cache"
+            r"|/available_updates"
```
agners, line 224: .. not sure if we should deprecate this one as well. From what I can see the Core isn't using it? 🤔

supervisor/api/middleware/security.py
```diff
@@ -162,9 +77,194 @@
...
+    api_bypass: re.Pattern[str]
+
+    # Paths that only Home Assistant Core may call
+    core_only: re.Pattern[str]
+
+    # Per-role allowed path patterns for installed apps
+    role_access: dict[str, re.Pattern[str]]
+
+    # Paths serving frontend assets (checked in core_proxy middleware)
+    core_frontend: re.Pattern[str]
```
agners: This is the Supervisor served frontend right? Maybe we should name this supervisor_frontend or something along these lines. Core always can be read as Home Assistant Core 😅

supervisor/api/backups.py
```diff
@@ -353,6 +397,34 @@ async def backup_partial(self, request: web.Request):
             job_id=job_id,
         )
 
+    @api_process
+    async def backup_partial(self, request: web.Request):
+        """Create a partial backup (v2: accepts "apps" key in request body)."""
+        body = await api_validate(SCHEMA_BACKUP_PARTIAL_V2, request)
```
agners: I'd name v2 without suffix (`SCHEMA_BACKUP_PARTIAL`) and suffix `SCHEMA_BACKUP_PARTIAL` with `_V1`. This way when we move to v3 it won't need changes 😅

agners (changes requested):
Looks good, looking forward to the new v2 API 😎

supervisor/api/middleware/security.py
mdegat01: Those are in the Manager role though, they aren't core only. The manager role is one that addons can declare they use so we'd need to check addons for usage of that.

However we can drop the `|/reload` bit. I just checked, that's there because there's a legacy route of `/addons/reload` that really maps to `/store/reload`. That path won't exist in v2. The others I expect to exist for now so I think we leave it as is unless we have a reason to deny apps access to these paths.

supervisor/api/middleware/security.py
mdegat01: Hm. So here's the actual paths in here:
```py
_V2_FRONTEND_PATHS: Final = (
    r"|/app/.*\.(?:js|gz|json|map|woff2)"
    r"|/v2/store/apps/" + RE_SLUG + r"/(logo|icon)"
)
```
`/app` is the app store panel. And yea I think you're right about that, that's gone now.

But what about the icons and logos? Are you sure frontend doesn't ask Supervisor for those in the new UI? Where else would they come from?

agners: It seems to load those through the API from what I can tell (e.g. `/api/hassio/addons/77f1785d_chip_tool/icon`). I think this can go :thinking: 

@bramkragten can you maybe confirm? Context: We are working on Supervisor v2 API, trying to get rid of legacy stuff as much as possible.

supervisor/api/middleware/security.py
agners: > Those are in the Manager role though, they aren't core only. The manager role is one that addons can declare they use so we'd need to check addons for usage of that.

Well, it's v2 no? We can make breaking changes :shrug: I mean just the fact that Core doesn't use this endpoint anymore tells me that we prefer alternative (better) ways to get this information.

mdegat01: Oh you're commenting specifically on `/available_updates`. Sorry its highlighting 4 lines in this UI and I figured you meant all 4 of them. So the fix in my first comment, while nice, wasn't even what you were referring to 😆 

Hm so `/available_updates`. Sure yea I suppose we could toss that. Core stopped using it when it moved to `update` entities since the coordinators just make API calls to `/supervisor/info`, `/os/info`, and `/addons/{slug}/info` on a schedule and then that info is in there. 

I can toss it out but the only thing to consider is I don't know what addons would need update info and why. But if they did its definitely more complex to use all those APIs then to just call `/available_updates` if info on what needs an update is all you need. Maybe we just wait for now and decide that in a follow-up PR? Since we have a lot of APIs to go through still.

supervisor/api/middleware/security.py
mdegat01: `/addons/77f1785d_chip_tool/icon`

@agners This API isn't planned to exist in v2 fyi. It was one of the ones in the legacy section for years and was really a redirect to `/store/addons/{slug}/icon`. For v2 I was only planning to make `/v2/store/apps/{slug}/icon`.

But I'm pretty sure whichever path is used here needs to be placed in the special "no auth required" section in order to work in the frontend properly.

mdegat01: This section of APIs I specifically told copilot to toss as super long standing legacy paths that we've thus far been unable to drop lol:
https://github.com/home-assistant/supervisor/blob/38ddb3df5466da2463f3656926a17f37466303b6/supervisor/api/__init__.py#L811-L825

supervisor/api/middleware/security.py
agners: I guess it's a bit a philosophical question which way to go at it: Remove everything and add what we know is going into v2, or start with everything and slowly remove what we no longer want 😅 

I think I'd preferred to remove what we know we don't want from the start. But it doesn't really matter.

I actually just realized that I mean to flag `/refresh_updates` below too, but somehow I did not save that comment. That one is actually even deprecated per docs. But I think I'd treat the two together. But ok, let's get discuss this in a follow up PR.

supervisor/api/middleware/security.py
agners: Hm, I see, I somehow assumed the Core hassio API proxy would add authentication unconditionally. But it seems we only add authentication if required, and this paths are explicitly excluded via hassio integrations `PATHS_NO_AUTH`.

But I wonder though, should we not make Core add authentication unconditionally? Then we could treat all endpoints equal from Supervisor perspective. Since Supervisor (v2 API) no longer serves its own frontend, we can argue it shouldn't have to accommodate frontend-specific constraints. That concern belongs at Core, which is the frontend-facing boundary. On the other hand, simply adding authentication to certain paths can also be tricky, what if the path filtering is broken? Then we allow unauthenticated access to Supervisor, which would be quite bad. So maybe maintaining the list twice is the more secure approach.

supervisor/api/backups.py
```diff
@@ -99,7 +100,17 @@ def _convert_local_location(item: str | None) -> str | None:
     }
 )
 
+# V1 schemas use "addons" as the request body key (legacy API contract).
 SCHEMA_RESTORE_PARTIAL = SCHEMA_RESTORE_FULL.extend(
```
agners: Same as with Backup below, I'd prefer if we suffix this with `_V1` and make `_V2` suffix less.

supervisor/api/middleware/security.py
mdegat01: I think icon and logo get special treatment because they are image assets. If you put image assets behind auth it makes caching more complicated I believe and we'd prefer clients cache them like they do other image assets. But we'd have to ask the frontend team for more info on why these APIs are special here.

supervisor/api/middleware/security.py
mdegat01: Oh! its deprecated for `/reload_updates`. Ok yea that one gets tossed then, that won't be in v2. But yea for the others I'd say lets just handle it in follow-up PRs.

mdegat01: So fun fact here - admin role is the only one that has ever been able to call the `/reload_updates` API. I guess whenever that change was made to rename that API it was only half finished since the deprecated one was the only one people could actually call from anywhere 🙈

agners (approved):
LGTM!

PR #6742, opened 2026-04-16
Add Unix socket support for Core communication with feature flag

agners:
## Proposed change

Reintroduce Unix socket support for Supervisor-to-Core communication, originally merged in #6590 and reverted in #6735. The key addition over the original PR is that the feature is now gated behind the `unix_socket_core_api` feature flag (introduced in #6719) and disabled by default.

When enabled and Core version supports it, Supervisor communicates with Core via a Unix socket at `/run/os/core.sock` instead of TCP, eliminating the need for access token authentication on that path (Core authenticates the peer by the socket connection itself).

### TCP path improvements (active by default, no feature flag needed)

The original PR also refactored the TCP communication path. Hence, the following improvements apply regardless of the feature flag:

- **Centralized WebSocket connection logic**: `APIProxy` and `HomeAssistantWebSocket` no longer implement their own auth handshakes. Both delegate to `api.connect_websocket()`, eliminating duplicate auth/retry logic.
- **`WSClient.connect_with_auth()` proper error handling**: The old version didn't close the websocket on failure or handle unexpected message types. It now closes the connection on any error and wraps unexpected exceptions in `HomeAssistantAPIError`.
- **`make_request()` early bail on stopped container**: Checks `is_running()` before attempting a request, giving a clear error instead of a cryptic connection failure.
- **`get_core_state()` response validation**: Now raises `HomeAssistantAPIError` if the response is `None` or not a dict (matching the validation `get_config()` already had).
- **"Connected to Core via TCP" log**: Logs transport info on first successful connection and re-logs after container restarts, improving observability.

**Note:** Once the feature flag is enabled, Unix socket support will only be used after the Core container got recreated. Hence this needs an `ha core rebuild` to take effect after enabling the feature flag!

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/homeassistant/websocket.py
+++ b/supervisor/homeassistant/websocket.py
@@ -3,9 +3,8 @@
 from __future__ import annotations
 
 import asyncio
-from contextlib import suppress
 import logging
-from typing import Any, TypeVar, cast
+from typing import Any, TypeVar
 
 import aiohttp
 from aiohttp.http_websocket import WSMsgType
@@ -45,14 +44,14 @@ class WSClient:
     ):
         """Initialise the WS client."""
         self.ha_version = ha_version
-        self._client = client
+        self.client = client
         self._message_id: int = 0
         self._futures: dict[int, asyncio.Future[T]] = {}  # type: ignore
 
     @property
     def connected(self) -> bool:
         """Return if we're currently connected."""
-        return self._client is not None and not self._client.closed
+        return self.client is not None and not self.client.closed
 
     async def close(self) -> None:
         """Close down the client."""
@@ -62,8 +61,8 @@ class WSClient:
                     HomeAssistantWSConnectionError("Connection was closed")
                 )
 
-        if not self._client.closed:
-            await self._client.close()
+        if not self.client.closed:
+            await self.client.close()
 
     async def async_send_command(self, message: dict[str, Any]) -> T:
         """Send a websocket message, and return the response."""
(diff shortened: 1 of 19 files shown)


supervisor/homeassistant/websocket.py
```diff
@@ -139,27 +138,105 @@ async def _receive_json(self) -> None:
...
 
-        return cls(AwesomeVersion(hello_message["ha_version"]), client)
+    @classmethod
+    async def connect_with_auth(
+        cls,
+        session: aiohttp.ClientSession,
+        url: str,
+        token: str,
+        *,
+        max_msg_size: int = 4 * 1024 * 1024,
```
mdegat01: Does this need to be an argument? I know it was but do we currently change the default anywhere? This feels like an anti-pattern like how ruff now flags arguments called something like `timeout` as something to set at the object level and not allow to be set by clients on a per call basis.

mdegat01 (approved):
Looks good. Largely the same code so can probably expediate it, especially since the bulk is now behind a feature flag. Had one small comment but seems ready to me.

If we start to build a bunch of these feature flags we may have to think of a documentation plan for them. For now though I might just recommend a comment somewhere to remind us that once this is turned on you must specifically call `ha core rebuild` to see it in action. A supervisor restart or even a host reboot won't turn it on alone since the container needs to be rebuilt with the new config.

supervisor/homeassistant/websocket.py
agners: We do use 64MB for the API proxy. I think it was driven by Node RED at one point. However, since we don't really allocate 64MB, and just use it as an upper boundary, we can just use the same default for our regular WebSocket connection too.

PR #6760, opened 2026-04-23
Refactor Supervisor network reattach path

agners:
## Proposed change

Clean up the reattach path in `DockerNetwork._create_supervisor_network()` that runs after the Supervisor network is (re)created.

On fresh startup, the Supervisor Docker network does not yet exist, so `_create_supervisor_network()` creates it and then re-attaches the known Supervisor-managed containers (Supervisor, Observer, CLI, DNS, Audio) to it. At that point the plugin containers legitimately do not exist yet, so `docker.containers.get()` returns 404. The callers wrapped these calls in `with suppress(DockerError):`, but `attach_container_by_name()` passed `_LOGGER.error` into the `DockerError` constructor, which logged the message at ERROR level before the exception was suppressed. This produced noisy, misleading ERROR lines during normal fresh startup:

```
ERROR [supervisor.docker.network] Can't find hassio_observer: 404 Client Error ... No such container: hassio_observer
ERROR [supervisor.docker.network] Can't find hassio_cli:      404 Client Error ... No such container: hassio_cli
ERROR [supervisor.docker.network] Can't find hassio_dns:      404 Client Error ... No such container: hassio_dns
ERROR [supervisor.docker.network] Can't find hassio_audio:    404 Client Error ... No such container: hassio_audio
```

Changes:

- `attach_container_by_name()` now raises `DockerNotFound` silently on 404 and `DockerError` without implicit logging on other Docker API errors.
- `_create_supervisor_network()` iterates all managed containers in a single loop using explicit `try/except`, replacing the three separate `suppress(DockerError)` blocks. Missing containers are logged at DEBUG; unexpected Docker errors at ERROR.
- Drop the `alias` argument on the reattach path. Docker adds the container name as an implicit network alias, and inter-container lookups go through `ExtraHosts` (`/etc/hosts`), not Docker DNS, so the explicit alias list was cosmetic and actually inconsistent with the first-create path in `DockerAPI._run()` (which uses the dashed hostname, e.g. `hassio-dns`, whereas the reattach path used the short slug `dns`).
- Consolidate `AUDIO_DOCKER_NAME`, `CLI_DOCKER_NAME`, `DNS_DOCKER_NAME` in `supervisor/const.py` alongside the existing `OBSERVER_DOCKER_NAME` and `SUPERVISOR_DOCKER_NAME` constants, and import them from there in the respective Docker wrapper modules.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/docker/network.py
+++ b/supervisor/docker/network.py
@@ -10,21 +10,18 @@ import aiodocker
 from aiodocker.networks import DockerNetwork as AiodockerNetwork
 
 from ..const import (
-    ATTR_AUDIO,
-    ATTR_CLI,
-    ATTR_DNS,
-    ATTR_OBSERVER,
-    ATTR_SUPERVISOR,
+    AUDIO_DOCKER_NAME,
+    CLI_DOCKER_NAME,
+    DNS_DOCKER_NAME,
     DOCKER_IPV4_NETWORK_MASK,
     DOCKER_IPV4_NETWORK_RANGE,
     DOCKER_IPV6_NETWORK_MASK,
     DOCKER_NETWORK,
     DOCKER_NETWORK_DRIVER,
-    DOCKER_PREFIX,
     OBSERVER_DOCKER_NAME,
     SUPERVISOR_DOCKER_NAME,
 )
-from ..exceptions import DockerError
+from ..exceptions import DockerError, DockerNotFound
 
 _LOGGER: logging.Logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


agners:
> * Drop the `alias` argument on the reattach path. Docker adds the container name as an implicit network alias, and inter-container lookups go through `ExtraHosts` (`/etc/hosts`), not Docker DNS, so the explicit alias list was cosmetic and actually inconsistent with the first-create path in `DockerAPI._run()` (which uses the dashed hostname, e.g. `hassio-dns`, whereas the reattach path used the short slug `dns`).

Maybe the more clean fix for this would be to read the hostname in `attach_container_by_name()` and do the same thing `DockerAPI._run()` is doing:

https://github.com/home-assistant/supervisor/blob/f7cce351e4b961ec0d80a5d77e2015b22e72e5f3/supervisor/docker/manager.py#L546

But this would need a `container.show()` to read the hostname. From what I can tell this alias is only really used for apps (if even).

supervisor/docker/network.py
```diff
@@ -218,24 +215,26 @@ async def _create_supervisor_network(
...
-                OBSERVER_DOCKER_NAME, [ATTR_OBSERVER], self.observer
-            )
-
-        for name, ip in (
-            (ATTR_CLI, self.cli),
-            (ATTR_DNS, self.dns),
-            (ATTR_AUDIO, self.audio),
+        # Reattach Supervisor-managed containers; missing ones are expected on fresh install.
+        for container_name, ipv4 in (
+            (SUPERVISOR_DOCKER_NAME, self.supervisor),
```
mdegat01: We might want to isolate this one for different handling. For example, a `DockerNotFound` exception may be acceptable for others but not for this one. That's where this code is running so no error is acceptable in this case.

supervisor/docker/network.py
```diff
@@ -218,24 +215,26 @@ async def _create_supervisor_network(
...
+                _LOGGER.debug(
+                    "No %s container to attach to Supervisor network",
+                    container_name,
+                )
+            except DockerError as err:
+                _LOGGER.error(
+                    "Can't attach %s to Supervisor network: %s",
+                    container_name,
+                    err,
                 )
```
mdegat01, line 250: Since we're cleaning this up, does continuing make sense in this case? What kind of errors result in this code path that won't just keep happening for every container we try? This is basically the "something is wrong with docker on this system" case right? Maybe alternatively we do something like this (just pseudo code):
```
try:
  attach supervisor by name
  Loop over others:
    try:
      attach i by name
    except DockerNotFound:
      debug log
except DockerError:
  error log
  capture sentry?
  schedule retry or pop repair? How does a user recover from this failure currently with the containers in the wrong network?
```

mdegat01 (commented):
Let me know what you think of my comment/suggestion. If we're seeing transient `DockerError` type exceptions in sentry then we can leave it as is instead. Though still curious what the recovery path is here

supervisor/docker/network.py
agners: Right, but what should we do when this one is missing? I guess at least print an error, but what else? :thinking:  Not much we can do :cold_sweat:

supervisor/docker/network.py
agners: Pop a repair is only really useful if we understand the problem and we can can give good advise to the user. Capture in Sentry goes down that road sprinkling it everywhere, which I am not sure if its a good idea 🤔 . Letting it bubble up will fail the whole coresys initialization, which is not a good idea as it could leave users completely stranded.

agners: I've came up with a good compromise: `_LOGGER.critical`, this is reported in Sentry, but avoids the explicit capture :sweat_smile:

mdegat01 (approved):
Yea that makes sense as a compromise to me. LGTM 👍

PR #6766, opened 2026-04-24
Improve Docker port-in-use detection and handling

agners:
## Proposed change

Triaging [`SUPERVISOR-1JWK`](https://home-assistant-io.sentry.io/issues/SUPERVISOR-1JWK) (`AppUnknownError` flooding Sentry from nginxproxymanager watchdog restarts) turned up that `RE_PORT_CONFLICT_ERROR` in `docker/manager.py` only matched one of the port-in-use message shapes Docker emits. The two variants produced by current moby — `Bind for <ip>:<port> failed: port is already allocated` from `portallocator.go` and `failed to bind host port <ip>:<port>/<proto>: address already in use` from `osallocator_linux.go` — fell through to `DockerAPIError`, got re-raised as the generic `AppUnknownError`, surfaced as HTTP 500 at the API, and (on the watchdog path) reached Sentry as an "unknown error."

Fixing that surfaced a few adjacent problems worth bundling:

- **Regex**: widen `RE_PORT_CONFLICT_ERROR` to match all known shapes (including an older form that embeds the container endpoint, still observed from older daemons and wrappers), anchored on the `failed to set up container networking` prefix and one of `address already in use` / `port is already allocated` as the suffix. Single capture group for the host port, so no call-site changes.
- **Diagnosability**: log the raw Docker message at `debug` level right before the conversion to `DockerContainerPortConflict`, so curious users can still see the exact upstream text (host IP, container endpoint, protocol) when investigating which process is holding the port.
- **Watchdog**: `_restart_after_problem` in `addons/addon.py` now catches `AppPortConflict` explicitly ahead of the generic `AppsError` handler: log a warning, break the retry loop, do *not* call `async_capture_exception`. A port conflict during watchdog restart is always an environment condition (another process grabbed the port while the add-on was down) — retrying cannot make it succeed and the Sentry capture is pure noise. This is what causes the SUPERVISOR-1JWK volume.
- **Cleanup**: with port conflicts now raised as typed `APIError` subclasses at the detection site, the `DockerAPIError` → `format_message()` rewrite fallback in `api_return_error` has no work left. Drop the fallback and delete `supervisor/utils/log_format.py` along with its tests — the module only ever handled port-conflict prose.

`test_app_start_port_conflict_error` and `test_observer_start_port_conflict` are parametrized across the three Docker message variants; a new `test_watchdog_port_conflict_does_not_retry` covers the watchdog behaviour (no retry, no Sentry capture, warning logged).

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR fixes or closes issue: fixes SUPERVISOR-1JWK
- Link to cli pull request:

## Checklist


mdegat01 (approved):
Hazards of parsing exception messages to find out what's really going on 🙈 

Good catch, hopefully there's no others out there but I suppose we'll keep updating as we see them in Sentry unless there becomes a better way to identify these without regexes.

PR #6767, opened 2026-04-24
Return proper API errors for mqtt/mysql service conflicts

agners:
## Proposed change

After #6739 added unexpected-error logging and Sentry capture to the `api_process` wrappers, two user-triggered service conflicts started surfacing on Sentry (SUPERVISOR-1JTQ and SUPERVISOR-1JWM) as if they were unexpected errors:

- `POST /services/{mqtt,mysql}` when another app already provides the service.
- `DELETE /services/{mqtt,mysql}` when no app currently provides it.

Both paths raised a generic `ServicesError`, which the API layer translated into an opaque HTTP 400 without an `error_key` or `extra_fields`, and which #6739 now also logs and captures via Sentry — so every legitimate conflict or mis-ordered request showed up as a bug.

This PR introduces two new-style API exceptions with translation keys and structured `extra_fields`:

- `ServiceAlreadyProvidedError` → 409 Conflict, `error_key: service_already_provided_error`, fields `{service, app}`.
- `ServiceNotProvidedError` → 404 Not Found, `error_key: service_not_provided_error`, fields `{service}`.

A shared `APIConflict` base class (status 409) is also added for future 409 responses. The `mqtt` and `mysql` service modules now raise these specific errors instead of the generic `ServicesError`, so the API returns structured, translatable responses and these expected user conflicts stop being captured as bugs.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR fixes or closes issue: fixes SUPERVISOR-1JTQ, fixes SUPERVISOR-1JWM
- This PR is related to issue: #6739
- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/services/modules/mqtt.py
+++ b/supervisor/services/modules/mqtt.py
@@ -69,7 +69,7 @@ class MQTTService(ServiceInterface):
         """Write the data into service object."""
         if self.enabled:
             raise ServiceAlreadyProvidedError(
-                _LOGGER.error,
+                _LOGGER.debug,
                 service=SERVICE_MQTT,
                 app=self._data[ATTR_APP],
             )
@@ -83,7 +83,7 @@ class MQTTService(ServiceInterface):
     async def del_service_data(self, app: App) -> None:
         """Remove the data from service object."""
         if not self.enabled:
-            raise ServiceNotProvidedError(_LOGGER.warning, service=SERVICE_MQTT)
+            raise ServiceNotProvidedError(_LOGGER.debug, service=SERVICE_MQTT)
 
         self._data.clear()
         await self.save()
--- a/supervisor/services/modules/mqtt.py
+++ b/supervisor/services/modules/mqtt.py
@@ -6,7 +6,7 @@ from typing import Any
 import voluptuous as vol
 
 from ...addons.addon import App
-from ...exceptions import ServicesError
+from ...exceptions import ServiceAlreadyProvidedError, ServiceNotProvidedError
 from ...validate import network_port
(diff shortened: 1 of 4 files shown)


sairon (approved):
LGTM. Definitely we don't want to have these captured, as the apps in their init scripts also pre-emptively unregister themselves on start: https://github.com/search?q=repo%3Ahome-assistant%2Faddons%20bashio%3A%3Aservices.delete&type=code

supervisor/services/modules/mqtt.py
```diff
@@ -68,9 +68,10 @@ def active(self) -> list[str]:
     async def set_service_data(self, app: App, data: dict[str, Any]) -> None:
         """Write the data into service object."""
         if self.enabled:
-            raise ServicesError(
-                f"There is already a MQTT service in use from {self._data[ATTR_APP]}",
-                _LOGGER.error,
+            raise ServiceAlreadyProvidedError(
+                _LOGGER.debug,
```
agners, line 72: As discussed separately, I've lowered this to debug. This avoids warnings like "The mqtt service is not currently provided by any app" which appear on regular Mosquitto app startup. The reasoning being, this are well handled errors from Supervisor's perspective. The client will be aware of the situation, and if clients care, they should handle/log it accordingly.

mdegat01: Yea makes sense. If you actually tried to start a second app that provided the same service I'd expect the second one to exit with error when it received this. So users should be aware already without us needing to do this.

mdegat01 (approved):

Issue #6745, opened 2026-04-16
Change add-ons to apps in Supervisor configuration


agners:
### Description

Change add-ons to apps in Supervisor configuration files (schema). This needs migration logic.

agners:
Fields in configuration files addressed by #6786.

Remaining is the `addon.json` configuration file name.

mdegat01:
There's a couple more. We also have an `addons` folder and an `addons_config` folder. This also made me realize a couple other challenges we have to think about:
1. The word `addon` appears as a mapping option in addon configuration. We'll have to deprecate that
2. Does renaming the `addon_config` folder may break existing containers with it mounted or will they be fine? Something to test
3. By default `all_addon_configs` is mounted at `addon_config` and local addons are mounted at `addons` in other addons. This will be a tough one to change

agners:
It seems that `/mnt/data/supervisor/discovery.json` still has `addon` as a key for its discovery information. This is probably tied to the API? 🤔  How should we deal with this?

PR #6769, opened 2026-04-24
Refactor API registration to support v1/v2 via shared methods

mdegat01:
## Summary

This PR adds clean v2 API support to the Supervisor REST API without duplicating route definitions, and expands the test suite to verify both versions.

### API Registration Refactor (`supervisor/api/__init__.py`, `supervisor/api/const.py`)

- Add `AppVersion` StrEnum (`V1`, `V2`) to `supervisor/api/const.py`
- Replace `self.v2_app` with a `versions` property (`dict[AppVersion, web.Application]`) computed dynamically so test fixtures that reassign `self.webapp` are automatically reflected in V1
- All `_register_*` methods now accept a required `app: web.Application` parameter; version-specific routes are gated with `if app is self.versions[AppVersion.V1/V2]:`
- `load()` loops over `enabled_versions` (V1 always, V2 when feature-flagged) and calls each registration method once per version — no duplication
- `_register_panel()` is called once outside the loop, always pinned to V1 — frontend static assets are V1-only
- Static resources are registered before `webapp.add_subapp()` to avoid registering into a frozen router
- Fold `_register_v2_apps` / `_register_v2_backups` / `_register_v2_store` into their respective unified methods

### Dual v1/v2 Test Coverage (`tests/api/`)

All 163 tests across 17 API modules that register identically on both versions now run against both via `api_client_with_prefix`.

**New fixtures in `tests/api/conftest.py`:**
- `api_client_v2` — TestClient with `SUPERVISOR_V2_API` feature flag enabled
- `api_client_with_prefix` — parametrizes over v1 (`""`) and v2 (`"/v2"`); use for APIs with identical behavior on both versions
- `core_api_client_with_root` — 3-way parametrize for Home Assistant Core endpoints: `/core` (v1 canonical), `/homeassistant` (v1 legacy alias), `/v2/core` (v2 canonical)
- `advanced_logs_tester` updated to use `api_client_with_prefix` so log-endpoint tests are auto-parametrized; accepts optional `v2_path_prefix` kwarg for paths that differ by version

**Tests intentionally kept V1-only:**
- `test_auth` / `test_discovery`: use indirect `api_client` parametrize for addon security context
- `test_panel`: frontend static assets are V1-only
- `test_jobs` (4 tests): inner `@Job`-decorated classes register names into the global `_JOB_NAMES` set — fixed by promoting them to module-level helpers (`_JobsTreeTestHelper` etc.) so registration only fires once at import time, enabling dual-version parametrization

### DIFF
--- a/supervisor/api/__init__.py
+++ b/supervisor/api/__init__.py
@@ -108,6 +108,11 @@ class RestAPI(CoreSysAttributes):
         enabled_versions = list(AppVersion) if v2_enabled else [AppVersion.V1]
         static_resource_configs: list[tuple[web.Application, StaticResourceConfig]] = []
 
+        # Panel is V1-only: the frontend assets are served from the root webapp only.
+        static_resource_configs.extend(
+            (self.versions[AppVersion.V1], config) for config in self._register_panel()
+        )
+
         for version in enabled_versions:
             app = self.versions[version]
             self._register_apps(app)
@@ -128,9 +133,6 @@ class RestAPI(CoreSysAttributes):
             self._register_network(app)
             self._register_observer(app)
             self._register_os(app)
-            static_resource_configs.extend(
-                (app, config) for config in self._register_panel()
-            )
             self._register_proxy(app)
             self._register_resolution(app)
             self._register_root(app)
--- a/supervisor/api/__init__.py
+++ b/supervisor/api/__init__.py
@@ -18,7 +18,7 @@ from .audio import APIAudio
 from .auth import APIAuth
 from .backups import APIBackups
 from .cli import APICli
-from .const import CONTENT_TYPE_TEXT
+from .const import CONTENT_TYPE_TEXT, AppVersion
 from .discovery import APIDiscovery
 from .dns import APICoreDNS
 from .docker import APIDocker
@@ -77,9 +77,9 @@ class RestAPI(CoreSysAttributes):
                 "max_field_size": MAX_LINE_SIZE,
             },
         )
-        # v2 sub-app: no middleware of its own — parent webapp's middleware
+        # V2 sub-app: no middleware of its own — the parent webapp's middleware
         # stack runs first for all requests including sub-app routes.
-        self.v2_app: web.Application = web.Application()
+        self._v2_app: web.Application = web.Application()
 
         # service stuff
         self._runner: web.AppRunner = web.AppRunner(self.webapp, shutdown_timeout=5)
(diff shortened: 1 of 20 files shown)


mdegat01:
As this PR is large I separated it into 3 commits to help with review:
- https://github.com/home-assistant/supervisor/pull/6769/changes/3b7ad2940b9d8bde792f960cf4c6e1324b7257b5 - The primary refactor. This changes the API loading process to register all shared paths on both v1 and v2 and then separately register v1 and v2 only endpoints. Main production code change
- https://github.com/home-assistant/supervisor/pull/6769/changes/f8242b2634860afb352241b9e31bc8ecb554203d - Mass parameterization of tests so that tests which test APIs that are dual registered with no changes other then path are exercised on both v1 and v2 with the same test.
    - This also includes one small production code change when I remembered the panel assets are v1 only. So rather then parameterizing those tests we should drop the static frontend assets registration from v2.
- https://github.com/home-assistant/supervisor/pull/6769/changes/35ff8c8cec30c5ee23e46d2e04538a64d58ab3d0 more test parameterization but not the same pattern as the other commit:
    - Parameterization of the homeassistant API tests to exercise all 3 possible paths since they are currently triple registered: legacy v1 `/homeassistant`, v1 `/core` and v2 `/core`
    - Adjusting the jobs API tests to support parameterization due to the issue mentioned at the top. Its still all really test parameterization but these tests could not be parameterized at all with the inline classes.

supervisor/api/__init__.py
```diff
@@ -89,72 +89,85 @@ def __init__(self, coresys: CoreSys):
...
+    @property
+    def versions(self) -> dict[AppVersion, web.Application]:
+        """Map of API version to its aiohttp application.
+
+        V1 is the root webapp (carries all middleware); V2 is a sub-app
+        mounted at /v2 — the parent middleware stack runs first for both.
+        Computed as a property so that reassigning self.webapp (e.g. in tests)
+        is automatically reflected.
+        """
+        return {AppVersion.V1: self.webapp, AppVersion.V2: self._v2_app}
```
agners, line 101: I wonder if it makes sense to create a base app, v1 and v2, where v1 just happens to live on the root? 🤔 This way it will be easy to move to v2 only at one point.

But we can also just do that when it comes to that.

supervisor/api/__init__.py
```diff
@@ -89,72 +89,85 @@ def __init__(self, coresys: CoreSys):
...
+            self._register_mounts(app)
+            self._register_multicast(app)
+            self._register_network(app)
+            self._register_observer(app)
+            self._register_os(app)
+            self._register_proxy(app)
+            self._register_resolution(app)
+            self._register_root(app)
+            self._register_security(app)
+            self._register_services(app)
```
agners, line 140: This now adds all the endpoints at once, which make sense since we largely want v2 to be pretty much v1.

That said, we still need to go through the endpoints to check for add-on usage. This one that has responses with `addon` as key for instance.

supervisor/api/__init__.py
```diff
@@ -568,137 +583,132 @@ def _register_proxy(self) -> None:
...
-                    partial(get_app_logs_v2, latest=True, no_colors=True),
-                ),
-                web.get("/apps/{app}/logs/boots/{bootid}", get_app_logs_v2),
-                web.get(
-                    "/apps/{app}/logs/boots/{bootid}/follow",
-                    partial(get_app_logs_v2, follow=True),
-                ),
-            ]
-        )
+        if app is self.versions[AppVersion.V1]:
```
agners, line 604: Not a big fan of these very long if blocks, not sure if we should just go with separate methods still to keep indention depth at bay... But more a stylistic thing.

agners (approved):
Quite big changes, especially in tests 😅 . But I guess it is sensible to test v2 too, so the changes make sense.

I leave it up to you if you want to implement/discuss the remarks, for me they are rather small nits and not deal breakers.

supervisor/api/__init__.py
mdegat01: We can if we want, once we had this property I didn't really see an advantage to making another subapp? Presumably there's some overhead and this makes it just as easy to turn off V1 someday. Or add new versions for that matter, just adjust active versions here for the loop.

supervisor/api/__init__.py
mdegat01: Interesting! Yep that one escaped before, I see it now. The validation is quite buried for services in that extension model we never really ended up using after `mqtt` and `mysql` 😓 

And yea that's fine, we'll give it the same treatment as the `backups` and `store` APIs where we register differently based on versions

supervisor/api/__init__.py
mdegat01: Yes so initially it was like that. There were a few problems:
1. Copilot pointed out we were initializing objects like `APIApps` multiple times, once per method, which was inefficient. Fixable via class properties but just adds more complexity
2. Apps would be pretty clearly cleaner as separate `_register_apps` and `_register_addons_v1` otherwise since there's really no overlap. But that only really applies to apps. For `store`, `backups` and now `services` we'd need something like `_register_store`, `_register_store_v2` and `_register_store_v1` to remove the if statements entirely. Since there are common paths, v2 only paths and v1 only paths. This was the part I didn't like, it was starting to get confusing to find everything split into 3. And this would be the pattern for any other modules with differences, only apps is cleanly split between v2 and v1.
3. In the end its all temporary as we'll set an enddate on v1 support. And then we cut these methods down to v2 stuff only.

That was my logic anyway 😆

mdegat01: Oh wait, you mean literally just move the block of code into another method called by this one. Slightly different then what I had. We can do that easily enough. I guess I didn't think one level of nesting was that bad and it kept each module together

supervisor/api/__init__.py
mdegat01: Oh wait. Actually the schema for the service `mysql` is this:
https://github.com/home-assistant/supervisor/blob/35ff8c8cec30c5ee23e46d2e04538a64d58ab3d0/supervisor/services/modules/mysql.py#L25-L33

Which does not include the `addon` field. That's only in `SCHEMA_CONFIG_MYSQL` which is used to validate the config stored with Supervisor, not the API postbody. So I think this API is safe from `addons`? Or is it somewhere else @agners

supervisor/api/__init__.py
agners: It's not bad per-se, I just feel it does not help readability. Anyways, let's not hold up this PR with stylistic things, we can address this later.

supervisor/api/__init__.py
agners: Yeah sorry, it is on the get side, it returns `addons`. Its in set_service_data, which writes `ATTR_APP` (which is `addon` still). From what I understand this is also saved information, so we needs to be addressed as part of #6745.

PR #6776, opened 2026-04-27
Return proper API errors for invalid hostnames

agners:
## Proposed change

`POST /host/options` with an invalid hostname (containing a space, an underscore, longer than 64 chars, etc.) lets the request travel all the way to systemd-hostnamed, which rejects it with `org.freedesktop.DBus.Error.InvalidArgs` and a message like `"Invalid static hostname 'bad name'"`. Supervisor mapped that D-Bus `INVALID_ARGS` error to `DBusInterfaceMethodError` together with `UNKNOWN_METHOD` and `INVALID_SIGNATURE` — three semantically very different conditions in one bucket. The result was a generic 400 with hostnamed's raw English text and, since #6739, a Sentry event and an exception traceback in the Supervisor log for what is really just user input validation.

This PR keeps the validation where it belongs (hostnamed's `hostname_is_valid()`) and only fixes how the rejection is surfaced:

- Split `ErrorType.INVALID_ARGS` out of `DBusInterfaceMethodError` into a new `DBusInvalidArgsError`. `UNKNOWN_METHOD` / `INVALID_SIGNATURE` continue to map to `DBusInterfaceMethodError`. The distinction matters: one means the call is broken (method missing or types wrong, a code/host-support problem), the other means the call is valid but a value was rejected (a client-input problem).
- Add `HostInvalidHostnameError(HostError, APIError)` with `error_key="host_invalid_hostname"` and `extra_fields={"hostname": ...}`, following the same `(SomethingError, APIError)` multi-inheritance pattern as `ObserverPortConflict` and `AppConfigurationInvalidError`. This gives clients a stable, machine-readable key and the offending hostname instead of systemd's prose.
- Catch `DBusInvalidArgsError` in `SystemControl.set_hostname` and re-raise `HostInvalidHostnameError`. `@api_process` then returns a clean 400 with the structured body and, because the error is now a modeled `APIError`, skips the unexpected-error logging and Sentry capture path added in #6739.

Tests cover the three layers: the D-Bus error mapping (`INVALID_ARGS` no longer collapses into `DBusInterfaceMethodError`), the host-control translation (a mocked hostnamed rejection becomes `HostInvalidHostnameError` with the right `error_key` / `extra_fields` / message), and the API surface (`POST /host/options` returns 400 with the structured body).

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR is related to issue: follow-up on #6739
- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/exceptions.py
+++ b/supervisor/exceptions.py
@@ -664,6 +664,23 @@ class HostLogError(HostError):
     """Internal error with host log."""
 
 
+class HostInvalidHostnameError(HostError, APIError):
+    """Hostname rejected by the host as semantically invalid."""
+
+    error_key = "host_invalid_hostname"
+    message_template = "Invalid hostname '{hostname}'"
+
+    def __init__(
+        self,
+        logger: Callable[..., None] | None = None,
+        *,
+        hostname: str,
+    ) -> None:
+        """Initialize exception."""
+        self.extra_fields = {"hostname": hostname}
+        super().__init__(None, logger)
+
+
 # Service / Discovery
 
 
@@ -743,6 +760,15 @@ class DBusInterfacePropertyError(DBusInterfaceError):
 
 class DBusInterfaceSignalError(DBusInterfaceError):
     """D-Bus signal not defined."""
+
+
+class DBusInvalidArgsError(DBusError):
+    """D-Bus argument value rejected by the service.
+
+    Distinct from DBusInterfaceMethodError: the method exists and the
+    argument types match the signature, but the service rejected an
+    argument's value as semantically invalid.
+    """
 
 
 class DBusParseError(DBusError):
--- a/supervisor/host/control.py
+++ b/supervisor/host/control.py
@@ -7,7 +7,11 @@ from awesomeversion import AwesomeVersion
 
 from ..const import HostFeature
 from ..coresys import CoreSysAttributes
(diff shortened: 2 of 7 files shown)


mdegat01 (approved):
So everything in this PR looks good and we should do it. But shouldn't we also modify this and add a regex?
https://github.com/home-assistant/supervisor/blob/ee1c6c48f29b349b5bb1d72e9db2d3e2542f32f0/supervisor/api/host.py#L77

This wouldn't be in lieu of all this other work. We're talking with a separate piece of software that has its own update schedule so even if we added better validation to our code it may be out of sync. But we could at least filter out the obvious stuff like whitespace without needing a dbus call right?

agners:
> This wouldn't be in lieu of all this other work. We're talking with a separate piece of software that has its own update schedule so even if we added better validation to our code it may be out of sync. But we could at least filter out the obvious stuff like whitespace without needing a dbus call right?

Yeah I did consider this, but then also I feel voluptuous validation errors are not that user friendly usually. Best UX would probably be to have frontend do a "live" validation (as you type, turn the input red or something along these lines). At this point we'd have 3 validation steps, which then seems also a bit overengineered :sweat_smile: .

Either way, we can add validation to Supervisor as a follow up PR if we decide it adds indeed value.

PR #6777, opened 2026-04-27
Treat JobException as a client-side API error

agners:
## Proposed change

Job condition guards (system not running, no free space, mounting not supported, etc.) and concurrency rejections ("Another job is running") raised by the `@Job` decorator are explicit precondition failures with descriptive messages — they are handled errors from the Supervisor's point of view, not unexpected internal failures. `JobException` previously inherited `HassioError` directly, so `api_process` caught the resulting `*JobError` mixins (`BackupJobError`, `AppsJobError`, `StoreJobError`, `HassOSJobError`, `SupervisorJobError`, `HomeAssistantJobError`, `MountJobError`, etc.) in its `HassioError` branch — which since #6739 logs them as unexpected and captures them to Sentry.

This change makes `JobException` inherit `APIError` so `api_process` surfaces these through its `APIError` branch with the original message intact and skips the unexpected-error path. Status defaults to `APIError`'s 400, so the API contract is unchanged. All `*JobError` mixins automatically pick up `APIError` via MRO; no per-class changes are needed. `JobNotFound` continues to be wrapped to `APINotFound` (404) at `supervisor/api/jobs.py:30`, unaffected.

`test_backup_immediate_errors` is extended to assert `async_capture_exception` is not called for the freeze and free-space condition guards. The assertion was confirmed to fail against the prior `JobException(HassioError)` definition (caught by the `except HassioError` branch, which logs as "Unexpected error" and captures to Sentry) and pass with the new inheritance.

This addresses a large share of `2026.04.1` Sentry noise — `BackupJobError`, `AppsJobError`, `HomeAssistantJobError`, `HassOSJobError`, `SupervisorJobError`, `StoreJobError` and bare `JobException` issues, all with culprit `supervisor.api.utils.api_process.<locals>.wrap_api`.

The MRO change pushes five plugin-specific job error mixins (`CliJobError`, `ObserverJobError`, `MulticastJobError`, `CoreDNSJobError`, `AudioJobError`) past pylint's `too-many-ancestors` threshold. Three other classes (`ResolutionNotFound` subclasses) already needed inline `# pylint: disable=too-many-ancestors` annotations for the same diamond-inheritance reason. Since the project's pylint config already disables every other `too-many-*` rule "for the sake of readability" — and Home Assistant Core disables `too-many-ancestors` globally with the same rationale ("it's too strict.") — the rule is added to the global disable list in `pyproject.toml` and the eight inline annotations are dropped.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -49,6 +49,7 @@ good-names = ["id", "i", "j", "k", "ex", "Run", "_", "fp", "T", "os"]
 # cyclic-import - doesn't test if both import on load
 # duplicate-code - unavoidable
 # locally-disabled - it spams too much
+# too-many-ancestors - it's too strict.
 # too-many-* - are not enforced for the sake of readability
 # too-few-* - same as too-many-*
 # unused-argument - generic callbacks and setup methods create a lot of warnings
@@ -61,6 +62,7 @@ disable = [
   "no-else-return",
   "not-context-manager",
   "too-few-public-methods",
+  "too-many-ancestors",
   "too-many-arguments",
   "too-many-branches",
   "too-many-instance-attributes",
--- a/supervisor/exceptions.py
+++ b/supervisor/exceptions.py
@@ -278,7 +278,7 @@ class CliUpdateError(CliError):
     """Error on update of a HA cli."""
 
 
-class CliJobError(CliError, PluginJobError):
+class CliJobError(CliError, PluginJobError):  # pylint: disable=too-many-ancestors
     """Raise on job error with cli plugin."""
 
 
@@ -293,7 +293,7 @@ class ObserverUpdateError(ObserverError):
     """Error on update of a Observer."""
 
 
-class ObserverJobError(ObserverError, PluginJobError):
+class ObserverJobError(ObserverError, PluginJobError):  # pylint: disable=too-many-ancestors
     """Raise on job error with observer plugin."""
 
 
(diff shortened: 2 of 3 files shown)


agners:
I've considered using `APIConflict` (HTTP code 409) instead of `APIError` (API code 400) here, but I'd rather keep this minimal and a non-breaking change.

If we treat them as user facing errors, we probably should turn those into translatable exceptions as well.

But for both points, I'd prefer to punt it further down, and just minimize the fallout of #6739, so we can release a Supervisor to stable again soon.

mdegat01 (approved):
Yea this makes sense. I thought about stripping out that pylint warning too but there was only a couple before.

On the topic of future changes in this space, the `JobException` model doesn't actually work particularly well with translatable errors. A `Job` can list multiple conditions but only one type of exception, can't really make a clear exception based on what failed 🤔

agners:
> On the topic of future changes in this space, the `JobException` model doesn't actually work particularly well with translatable errors. A `Job` can list multiple conditions but only one type of exception, can't really make a clear exception based on what failed 🤔

That is a general problem with Exception error handling. But it is also super common to just raise the first we encounter, and once that is resolved you'd just get the next error. I don't think it is a blocker really :shrug: .

PR #6803, opened 2026-05-05
Replace fixed-duration sleeps after bus events with gather

agners:
## Proposed change

Several tests use `await asyncio.sleep(...)` to "wait for the listener to run" after firing a bus event. The fixed duration is real wall-clock time and the wait can be indeterministic — if the handler chain happens to need slightly more time on a busy CI runner, the assertion races the handler.

`Bus.fire_event` returns the listener tasks since #6252; await them via a new `tests.common.fire_bus_event(coresys, event, data)` helper that bundles the `gather` so the call sites stay short. Touches `test_bus.py` (the bus tests were poking scheduling instead of verifying their assertions), `test_home_assistant_watchdog.py`, `test_plugin_base.py`, `addons/test_manager.py`, `docker/test_addon.py`, and `test_store_execute_reload.py`.

Other cleanups in the same spirit:

- `_fire_test_event` in `addons/test_addon.py` becomes `async def` and delegates to the helper, so its 17 call sites collapse to a single `await _fire_test_event(...)`.
- The two `test_store_execute_reload.py` sites that used the private `_update_connectivity()` helper are reworked to set the cached connectivity flag directly and fire the event themselves so they can gather the listener tasks the same way.
- The two `sleep(1)` post-pull drains in `docker/test_interface.py` collapse to `sleep(0)` (handler tasks are already gathered inside `pull_image`), saving ~2s.
- The `sleep(0.01)` waits inside `container_events()` task bodies (`api/test_addons.py`, `api/test_store.py`, `backups/test_manager.py`) are just one-yield-to-the-parent and become `sleep(0)`.

Switching to `gather` exposes a few latent test mocks that were silently swallowing TypeErrors as background-task failures before: `CGroup.add_devices_allowed` is `async def` but was patched as a plain `MagicMock` in `docker/test_addon.py` (now `new_callable=AsyncMock`), and the watchdog does `await (await self.start())` / `await (await self.restart())` because `App.start` / `App.restart` return `asyncio.Task` — the mocks in `addons/test_addon.py` (`test_app_watchdog`, `test_watchdog_on_stop`, `test_watchdog_during_attach`) needed `AsyncMock(return_value=<settled future>)` to mirror that shape rather than a plain `MagicMock`.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/tests/addons/test_addon.py
+++ b/tests/addons/test_addon.py
@@ -6,7 +6,7 @@ import errno
 from http import HTTPStatus
 from pathlib import Path, PurePath
 from typing import Any
-from unittest.mock import MagicMock, PropertyMock, call, patch
+from unittest.mock import AsyncMock, MagicMock, PropertyMock, call, patch
 
 import aiodocker
 from aiodocker.containers import DockerContainer
@@ -51,16 +51,18 @@ from tests.common import get_fixture_path, is_in_list
 from tests.const import TEST_ADDON_SLUG
 
 
-def _fire_test_event(coresys: CoreSys, name: str, state: ContainerState):
-    """Fire a test event."""
-    coresys.bus.fire_event(
-        BusEvent.DOCKER_CONTAINER_STATE_CHANGE,
-        DockerContainerStateEvent(
-            name=name,
-            state=state,
-            id="abc123",
-            time=1,
-        ),
+async def _fire_test_event(coresys: CoreSys, name: str, state: ContainerState) -> None:
+    """Fire a test event and await the listener tasks the bus spawned."""
+    await asyncio.gather(
+        *coresys.bus.fire_event(
+            BusEvent.DOCKER_CONTAINER_STATE_CHANGE,
+            DockerContainerStateEvent(
+                name=name,
+                state=state,
+                id="abc123",
+                time=1,
+            ),
+        )
     )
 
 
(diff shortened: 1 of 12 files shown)


tests/addons/test_addon.py
```diff
@@ -51,16 +51,18 @@
...
+    await asyncio.gather(
+        *coresys.bus.fire_event(
+            BusEvent.DOCKER_CONTAINER_STATE_CHANGE,
+            DockerContainerStateEvent(
+                name=name,
+                state=state,
+                id="abc123",
+                time=1,
+            ),
+        )
```
mdegat01: Could consider moving this to tests.common and making it a generic helper that accepts event, data and awaits the task. So we can use it in all the places shown.

mdegat01 (approved):
Good call. I really hated those sleeps, especially the ones that had to be > 0  to make the test work, much better 👍

agners:
Hm, there was a test issue in one of the runs:
```
ERROR tests/os/test_data_disk.py::test_datadisk_move_fail[unavailable drive by path] - blockbuster.blockbuster.BlockingError: Blocking call to io.TextIOWrapper.write
```

This seems an intermittent problem, nothing this PR touched.

PR #6824, opened 2026-05-08
Tighten mypy and fix findings under the new flags

agners:
## Proposed change

Tightens mypy in CI by enabling `warn_unreachable`, `warn_redundant_casts`, `warn_unused_ignores`, and the `exhaustive-match` and `possibly-undefined` error codes, then fixes every finding the new flags surfaced. Each fix is a separate commit so the rationale per finding is clear in `git log`.

The new flags caught a handful of real things — a never-reachable `_LOGGER.critical` fallback in the scheduler, an unbound `credentials` local on the docker pull error path, an unbound `repository_file` if `FILE_SUFFIX_CONFIGURATION` were ever empty, three `match` statements that fell through to a trailing `return`/`raise` instead of declaring a wildcard arm — plus stylistic noise (a stale `# type: ignore` for `eager_start` from before 3.14, three redundant `cast(int | None, …)` calls already covered by the source `TypedDict`).

The `[tool.mypy]` config follows the pattern Home Assistant Core uses: instead of `--ignore-missing-imports` (which globally suppresses import resolution errors), we set `disable_error_code = ["import-not-found", "import-untyped"]`. That keeps the five known untyped third-party deps quiet (`pyudev`, `log_rate_limit`, `pulsectl`, `cpe`, `atomicwrites`) and lets the CI invocation drop the CLI flag.

The scheduler dispatch was rewritten as `match task.interval: case int() | float(): ... case time(): ...`. With `exhaustive-match` enabled, broadening `_Task.interval` (e.g. `float | time | str`) now produces a build error pointing at this match, so future contributors can't silently expand the type without handling every case.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/docker/interface.py
+++ b/supervisor/docker/interface.py
@@ -332,16 +332,17 @@ class DockerInterface(JobGroup, ABC):
                 )
                 await async_capture_exception(err)
 
+        # Get credentials for private registries to pass to aiodocker.
+        # Done before registering the listener so a failure here does not
+        # leak a stale event listener.
+        credentials, pull_image_name = self._get_credentials(image)
+
         listener = self.sys_bus.register_event(
             BusEvent.DOCKER_IMAGE_PULL_UPDATE, process_pull_event
         )
 
         _LOGGER.info("Downloading docker image %s with tag %s.", image, version)
-        credentials: dict = {}
         try:
-            # Get credentials for private registries to pass to aiodocker
-            credentials, pull_image_name = self._get_credentials(image)
-
             # Pull new image, passing credentials to aiodocker
             docker_image = await self.sys_docker.pull_image(
                 current_job.uuid,
--- a/supervisor/docker/interface.py
+++ b/supervisor/docker/interface.py
@@ -337,6 +337,7 @@ class DockerInterface(JobGroup, ABC):
         )
 
         _LOGGER.info("Downloading docker image %s with tag %s.", image, version)
+        credentials: dict = {}
         try:
             # Get credentials for private registries to pass to aiodocker
             credentials, pull_image_name = self._get_credentials(image)
--- a/supervisor/host/configuration.py
+++ b/supervisor/host/configuration.py
@@ -267,7 +267,8 @@ class Interface:
                 return InterfaceMethod.AUTO
             case NMInterfaceMethod.MANUAL:
                 return InterfaceMethod.STATIC
-        return InterfaceMethod.DISABLED
+            case _:
+                return InterfaceMethod.DISABLED
 
     @staticmethod
     def _map_nm_addr_gen_mode(addr_gen_mode: int | None) -> InterfaceAddrGenMode:
@@ -333,7 +334,8 @@ class Interface:
                 return InterfaceType.WIRELESS
             case DeviceType.VLAN.value:
                 return InterfaceType.VLAN
-        raise ValueError(f"Invalid device type: {device_type}")
+            case _:
+                raise ValueError(f"Invalid device type: {device_type}")
 
     @staticmethod
     def _map_nm_wifi(inet: NetworkInterface) -> WifiConfig | None:
--- a/supervisor/misc/scheduler.py
+++ b/supervisor/misc/scheduler.py
@@ -73,19 +73,22 @@ class Scheduler(CoreSysAttributes):
 
     def _schedule_task(self, task: _Task) -> None:
         """Schedule a task on loop."""
-        if isinstance(task.interval, (int, float)):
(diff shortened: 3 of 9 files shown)


supervisor/docker/interface.py
```diff
@@ -337,6 +337,7 @@ async def process_pull_event(event: PullLogEntry) -> None:
         )
 
         _LOGGER.info("Downloading docker image %s with tag %s.", image, version)
+        credentials: dict = {}
```
mdegat01: Looking at the code for `self._get_credentials`, it appears to be using entirely our code. I don't see any way its possible for it to raise `DockerRegistryRateLimitExceeded` or `aiodocker.DockerError`. So rather then pre-defining the variable I think we can just call it outside of the `try` since that doesn't appear to be doing anything for it anyway. So something like this:

```py
        # Get credentials for private registries to pass to aiodocker
        credentials, pull_image_name = self._get_credentials(image)
        
        listener = self.sys_bus.register_event(
            BusEvent.DOCKER_IMAGE_PULL_UPDATE, process_pull_event
        )

        _LOGGER.info("Downloading docker image %s with tag %s.", image, version)
        try:
            # Pull new image, passing credentials to aiodocker
            ...
```

Moved to before the listener registration out of abundance of caution to so we don't risk leaving a stale event listener around.

supervisor/host/configuration.py
```diff
@@ -267,7 +267,8 @@ def _map_nm_method(method: str | None) -> InterfaceMethod:
                 return InterfaceMethod.AUTO
             case NMInterfaceMethod.MANUAL:
                 return InterfaceMethod.STATIC
-        return InterfaceMethod.DISABLED
+            case _:
+                return InterfaceMethod.DISABLED
```
mdegat01, line 271: What's the reasoning behind this rule? Is there a downside to doing it the way we were doing or is this just a very opinionated rule? It seems like the behavior would be identical here and I personally like the the former approach better. We generally prefer a shortcut approach in this code base like:
```py
if condition:
  # do stuff when condition is true
  return
# do stuff when condition is false
```
As opposed to:
```py
if condition:
  # do stuff when condition is true
else:
  # do stuff when condition is false
```
So adding what is effectively an `else` here when the others shortcut feels inconsistent to me. Unless there's a good reason for it, I'd probably prefer whatever rule this is stays disabled.

supervisor/misc/scheduler.py
```diff
@@ -73,26 +73,22 @@ async def _wrap_task():
...
-                "Unknown interval %s (type: %s) for scheduler %s",
-                task.interval,
-                type(task.interval),
-                task.id,
-            )
+        match task.interval:
+            case int() | float():
+                task.next = self.sys_call_later(task.interval, self._run_task, task)
+            case time():
+                today = datetime.combine(date.today(), task.interval)
```
mdegat01, line 80: This is neat! I didn't know you could do this now 👍

supervisor/store/repository.py
```diff
@@ -186,8 +186,7 @@ def validate_file() -> bool:
                 repository_file = Path(self._git.path / f"repository{filetype}")
                 if repository_file.exists():
                     break
-
-            if not repository_file.exists():
+            else:
```
mdegat01, line 189: Another neat trick! Going to have to remember these 👍

mdegat01 (approved):
My comments that aren't appreciations for python tricks I was unaware of are nitpicks so this can go in as is if you want. But I'm not a huge fan of preferring `case _:` over leveraging shortcircuiting and I think there's a better way to handle the `credentials` one.

supervisor/host/configuration.py
agners: I think the main difference to a if/else is that in if/else there can't be any other condition, but for match statement it could have more arms.

From what I understand, the `exhaustive-match` rule is about making sure that all possible cases are covered. Now if a default rule is used, then this ends up really a bit stylistic difference without adding much value by the type checker. But for an enum, it would flag again if someone extends the enum.

Actually, with that in mind I changed the match in mount.py: b8c00e8e4df165119bdc7355d31a8ce11fbbfa1f. This way if we extend the `MountUsage` enum mypy will flag and we need to make a decision in that location.

agners: Unfortunately we can't have it flag the trailing return if it were an open/default match, and have widening detection on matches where all enum types are matched. With that in mind, I prefer to have the rule enabled and just accept this style.

mdegat01: Ok I guess that's fine then

PR #6825, opened 2026-05-08
Run pytest in parallel with pytest-xdist

agners:
## Proposed change

Run the pytest suite in parallel using pytest-xdist. Most of the suite's wall-clock time is spent in fixture setup (D-Bus session, mock services, CoreSys construction) rather than test bodies, but each test pays this setup cost on its own — so distributing tests across CPU cores translates almost directly to a wall-clock reduction.

`pytest-xdist==3.8.0` is added to `requirements_tests.txt` and CI invokes pytest with `-n auto --dist=loadfile`. `loadfile` keeps all tests of a single file on the same worker, which preserves locality for module-level imports/fixtures and keeps pytest's progress output readable. This is the same pattern Home Assistant Core has been running for a long time in its `pytest-full` job (`--numprocesses auto --dist=loadfile`), so the configuration is already battle-tested in a sibling project.

GitHub Actions standard runners ship 4 vCPUs, so `-n auto` will pick 4 workers in CI. On an 8-core local machine the full local run drops from ~280s to ~88s (~3.2x); on the 4-vCPU CI runner expect roughly a ~2x reduction.

The session-scoped `dbus_session` fixture spawns its own dbus-daemon, so each xdist worker gets an independent bus — no shared D-Bus state across workers. `pytest-cov` already handles xdist worker coverage merging via the standard `.coverage.*` files, so no coverage configuration changes are needed.

A second commit fixes a flake exposed by xdist in `test_multiple_datadisk_add_remove_signals` and `test_disabled_datadisk_add_remove_signals`. Those tests fire UDisks2 `InterfacesAdded`/`InterfacesRemoved` signals and used `await asyncio.sleep(0.2)` to wait for supervisor's signal handler chain to update `coresys.resolution.issues`. The 200 ms cushion is effectively only 100 ms because `DataDisk._udisks2_interface_added` itself sleeps 100 ms internally to wait for `UDisks2._interfaces_added` (a sibling subscriber on the same signal) to refresh the block-device cache before the check runs. Under contended xdist scheduling the assertion races the handler. Replaced the four `sleep+assert` sites with a small `tests.common.wait_for(predicate)` polling helper (10 ms interval, 5 s deadline) so the tests stay fast on idle and robust under load. The product-code sleep inside the handler is still a real smell — handler ordering shouldn't depend on a fixed sleep — but is left for a separate fix.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -380,6 +380,8 @@ jobs:
           . venv/bin/activate
           pytest \
             -qq \
+            -n auto \
+            --dist=loadfile \
             --durations=10 \
             --cov supervisor \
             -o console_output_style=count \
--- a/requirements_tests.txt
+++ b/requirements_tests.txt
@@ -7,6 +7,7 @@ pytest-aiohttp==1.1.0
 pytest-asyncio==1.3.0
 pytest-cov==7.1.0
 pytest-timeout==2.4.0
+pytest-xdist==3.8.0
 pytest==9.0.3
 ruff==0.15.12
 time-machine==3.2.0
--- a/tests/common.py
+++ b/tests/common.py
@@ -1,7 +1,7 @@
 """Common test functions."""
 
 import asyncio
-from collections.abc import Sequence
+from collections.abc import Callable, Sequence
 from datetime import datetime
 from functools import partial
 from importlib import import_module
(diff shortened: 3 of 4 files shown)


mdegat01 (approved):
Well approved but there are test failures which is concerning. Perhaps we have some shared state issues since this shouldn't have caused anything to fail 😬 

Will be really nice to speed up tests. Particularly the API test suite which takes forever currently.

agners:
> Well approved but there are test failures which is concerning. Perhaps we have some shared state issues since this shouldn't have caused anything to fail 😬

Turned out this is a race condition: We do have sleeps in that test, and I guess the fact that we load the CPU more changes timings. I came up with a busy wait solution which should make the test resilient against different timings.


> Will be really nice to speed up tests. Particularly the API test suite which takes forever currently.

It won't automatically speed up local runs, you'll need to use `pytest -n auto` to use multiple CPUs.

mdegat01 (approved):
Nice one. Looks like an automation now 😆

PR #6786, opened 2026-04-29
Migrate persisted 'addon' field to 'app' in config files

mdegat01:
## Proposed change

Part of the ongoing addon→app terminology migration (following #6696, #6717, #6741).

Renames the `addon` key to `app` in persisted configuration files, with full backwards compatibility for existing installations.

### Config file migrations (backwards compatible)
- **`supervisor.json`**: `addons_custom_list` → `apps_custom_list`
- **`services.json`** (MQTT & MySQL): `addon` → `app` (the app slug that provides the service)
- **`discovery.json`**: `addon` → `app` in each discovery message

Each schema now has a pre-processor that transparently renames old keys on load, so existing config files on disk continue to work.

### Internal rename
- `Message.addon` → `Message.app` in the `Discovery` class
- `ATTR_APP` value changed from `"addon"` to `"app"` in both `const.py` and `services/const.py`
- Added `ATTR_ADDON = "addon"` and `ATTR_ADDONS_CUSTOM_LIST = "addons_custom_list"` constants for migration pre-processors and V1 API compat

### V1/V2 API compatibility
- `GET /services/{service}`: V1 returns `{"addon": ...}`, V2 returns `{"app": ...}`
- `GET /discovery` and `GET /discovery/{uuid}`: V1 returns `{"addon": ...}`, V2 returns `{"app": ...}`
- Implemented via dedicated `_v1` handler methods following the established backups/store pattern, registered with `AppVersion` guards in `_register_services()` and `_register_discovery()`

### Note on hassio_push/discovery
The internal push payload Supervisor sends to HA at `api/hassio_push/discovery/{uuid}` continues to use the `addon` key for Home Assistant compatibility.

### Type annotation
- Broadened `FileConfiguration` schema parameter type to accept `vol.All` validators in addition to `vol.Schema`

## Type of change

- [ ] Dependency upgrade
- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (which adds functionality to the supervisor)
- [ ] Breaking change (fix/feature causing existing functionality to break)
- [x] Code quality improvements to existing code or addition of tests

## Additional information

- This PR fixes or closes issue: fixes #
- This PR is related to issue: #6696, #6717, #6741
- Link to documentation pull request:
- Link to cli pull request:
- Link to client library pull request:

## Checklist

- [x] The code change is tested and works locally.
- [x] Local tests pass. **Your PR cannot be merged unless tests pass**
- [x] There is no commented out code in this PR.
[...]

PR #6837, opened 2026-05-13
Rename addon/addons to app/apps in filenames and imports

mdegat01:
## Proposed change

Continues the addon→app terminology migration started in #6786 and others.

Renames all source files, test files, fixture files, and directories that contained `addon`/`addons` in their names to use `app`/`apps` respectively, and updates all imports accordingly.

The resolution check slugs (`addon_pwned`, `deprecated_addon`, etc.) are part of the public API — exposed via `/resolution/info` and used to run checks by name. Each renamed check class now overrides the `slug` property to return the original filename-based slug, keeping the contract intact.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR fixes or closes issue: #6839
- This PR is related to issue: #6786
- Link to cli pull request:

## Checklist


mdegat01:
Huh. what happened to our bot? It didn't act on the checkbox 🤔

supervisor/resolution/validate.py
```diff
@@ -29,7 +44,9 @@ def get_valid_modules(folder, *, base=__file__) -> list[str]:
 
 SCHEMA_CHECKS_CONFIG = vol.Schema(
     {
-        vol.Required(check, default=SCHEMA_CHECK_CONFIG({})): SCHEMA_CHECK_CONFIG
+        vol.Required(
+            _get_check_slug(check), default=SCHEMA_CHECK_CONFIG({})
+        ): SCHEMA_CHECK_CONFIG
         for check in get_valid_modules("checks")
```
mdegat01, line 50: This bit feels a bit goofy but it is necessary here since this PR isn't changing the slugs of the checks. So we still accept the slugs with `addon` in them in the API and still write them into the file. I'll change that in a follow-up PR, this one is just about file and folder names.

Plus I need to consider V1 vs V2 API. V2 should only accept the new slugs. V1 presumably should continue to accept the old ones but perhaps allow both? Anyway, follow-up PR.

agners:
> Huh. what happened to our bot? It didn't act on the checkbox 🤔

Was that live? Did not notice so far :sweat_smile:

supervisor/resolution/validate.py
agners: > Plus I need to consider V1 vs V2 API. V2 should only accept the new slugs. V1 presumably should continue to accept the old ones but perhaps allow both? Anyway, follow-up PR.

V1 API: Whatever makes it easier for the Core transition. Maybe this is one of the cases where we anyways need to wait for the v2 client library/Core change, so probably we can just accept legacy on V1.

agners (changes requested):
There is still a fixture named `add-on.json` at `tests/fixtures/add-on.json`.

agners (approved):
LGTM

PR #6851, opened 2026-05-19
Coordinate graceful shutdown with Home Assistant OS

sairon:
## Proposed change

On a reboot or power off triggered from the API, enter the haos-pre-shutdown.target systemd target before stopping Home Assistant Core, so the OS can stop the console CLI cleanly instead of letting it restart-loop while the CLI plugin is torn down. Entering the target is best-effort and a no-op on OS releases that don't have it.

For host shutdowns not initiated by Supervisor, the SIGTERM handler now inspects the systemd manager state. When it is "stopping", Supervisor runs the managed Core.shutdown() to stop Core, apps and plugins gracefully before tearing itself down; on a plain Supervisor restart the state is "running", so only Supervisor stops as before. The manager state is exposed by every systemd version, so this works regardless of the OS release.

Add the SystemState D-Bus enum and Systemd.get_system_state() for the state lookup.

Refs home-assistant/operating-system#4642
Refs home-assistant/operating-system#4719

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/host/control.py
+++ b/supervisor/host/control.py
@@ -1,5 +1,6 @@
 """Power control for host."""
 
+import asyncio
 from datetime import datetime
 import logging
 
@@ -7,13 +8,19 @@ from awesomeversion import AwesomeVersion
 
 from ..const import HostFeature
 from ..coresys import CoreSysAttributes
+from ..dbus.const import StartUnitMode, UnitActiveState
 from ..exceptions import (
     DBusInvalidArgsError,
+    DBusSystemdNoSuchUnit,
+    HassioError,
     HostInvalidHostnameError,
     HostNotSupportedError,
 )
 
 _LOGGER: logging.Logger = logging.getLogger(__name__)
+
+HAOS_PRE_SHUTDOWN_TARGET = "haos-pre-shutdown.target"
+HAOS_PRE_SHUTDOWN_TIMEOUT = 10
 
 
 class SystemControl(CoreSysAttributes):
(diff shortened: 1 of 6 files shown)


mdegat01 (approved):
Looks good. LGTM 👍

supervisor/host/control.py
```diff
@@ -38,6 +45,28 @@ def _check_dbus(self, flag: HostFeature) -> None:
...
+        """Enter the HAOS pre-shutdown systemd phase if available."""
+        if not self.sys_dbus.systemd.is_connected:
+            return
+
+        try:
+            _LOGGER.info("Entering Home Assistant OS pre-shutdown phase")
+            await self.sys_dbus.systemd.start_unit(
+                HAOS_PRE_SHUTDOWN_TARGET, StartUnitMode.REPLACE
+            )
+            unit = await self.sys_dbus.systemd.get_unit(HAOS_PRE_SHUTDOWN_TARGET)
```
agners, line 58: Since we anyway have to get the unit, can't we do this before the log message above? That way we only get `Entering Home Assistant OS pre-shutdown phase` if we really enter that phase.

supervisor/host/control.py
```diff
@@ -38,6 +45,28 @@ def _check_dbus(self, flag: HostFeature) -> None:
...
+            async with asyncio.timeout(HAOS_PRE_SHUTDOWN_TIMEOUT):
+                await unit.wait_for_active_state({UnitActiveState.ACTIVE})
```
agners, line 60: Actually, since #6838 we have a nicer way via systemd Jobs to do this: Use `_run_systemd_job` to call `self.sys_dbus.systemd.start_unit`, which makes the waiting for the unit to become active unnecessary.

agners: We probably should get rid of `wait_for_active_state`, and cleanup the whole systemd service handling a bit 🤔 . But that is for a follow up. For this PR, I prefer if we use the Job based unit control from the get go.

agners (changes requested):

supervisor/host/control.py
sairon: ~Well, #6838 wasn't merged when I put up this PR :) But sure, I'll rebase and update it.~

Not relevant anymore with the changes requested in the OS PR.

sairon:
Replaced by home-assistant/operating-system#4736 and home-assistant/supervisor#6887.

PR #6856, opened 2026-05-20
Model client-state Apps* errors as APIError

agners:
## Proposed change

Follow-up on #6739. With `HassioError` now logged and captured by Sentry in `api_process`, a handful of `Apps*` exceptions raised from `AppManager.install/update/rebuild` and `AppModel._validate_availability` surfaced as "unexpected" 400s with a noisy log entry and a Sentry event, even though they are all client-state errors (user clicked install on an already-installed add-on, "no update available", local and store versions diverged, system architecture/machine/Home Assistant version incompatible, image-based app cannot be rebuilt, etc.). The dominant offender is SUPERVISOR-1JVV (`No update available for app core_mosquitto`, ~19k events / ~12k users on 2026.05.0); several siblings in the same `wrap_api` cluster show the same shape.

Same treatment as #6785 did for `BackupMountDownError`: map these through properly so the API returns clean, structured 400s.

- Add modeled `APIError` subclasses in `exceptions.py` for the previously raw raises in `apps/manager.py`: `AppAlreadyInstalledError`, `AppNotFoundError`, `AppNotInstalledError`, `AppNotInStoreError`, `AppNoUpdateAvailableError`, `AppRebuildVersionChangedError`, `AppRebuildImageBasedError`. Each gets a stable `error_key`, a `message_template`, and an `addon` extra field.
- Add `APIError` to `AppNotSupportedArchitectureError`, `AppNotSupportedMachineTypeError` and `AppNotSupportedHomeAssistantVersionError` so they behave the same as the rest of the `Apps*` `APIError`s instead of being treated as unexpected.
- Use the add-on's display name (`app.name` from the config) instead of the slug in `extra_fields` whenever an `App`/`AppModel` is available at the raise site, so users see e.g. "Mosquitto broker" rather than "core_mosquitto" in error messages. Slug remains the fallback when there is no app object (install of an unknown slug; update/rebuild of a slug that is not installed).
- Update raise sites in `apps/manager.py` and `apps/model.py` accordingly.

These are all runtime states users hit during normal interaction with the add-ons UI, not Supervisor bugs worth paging on.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR is related to issue: SUPERVISOR-1JVV and the broader `wrap_api` Sentry cluster introduced by #6739; follows the pattern set by #6785
- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/apps/manager.py
+++ b/supervisor/apps/manager.py
@@ -210,13 +210,11 @@ class AppManager(CoreSysAttributes):
         self.sys_jobs.current.reference = slug
 
         if slug in self.local:
-            raise AppAlreadyInstalledError(
-                _LOGGER.warning, addon=self.local[slug].name, slug=slug
-            )
+            raise AppAlreadyInstalledError(_LOGGER.warning, app=self.local[slug])
         store = self.store.get(slug)
 
         if not store:
-            raise AppNotFoundError(_LOGGER.error, slug=slug)
+            raise AppNotFoundError(_LOGGER.error, app=slug)
 
         store.validate_availability()
 
@@ -274,15 +272,15 @@ class AppManager(CoreSysAttributes):
         self.sys_jobs.current.reference = slug
 
         if slug not in self.local:
-            raise AppNotInstalledError(_LOGGER.error, slug=slug)
+            raise AppNotInstalledError(_LOGGER.error, app=slug)
         app = self.local[slug]
 
         if app.is_detached:
-            raise AppNotInStoreError(_LOGGER.error, addon=app.name, slug=slug)
+            raise AppNotInStoreError(_LOGGER.error, app=app)
         store = self.store[slug]
 
         if app.version == store.version:
-            raise AppNoUpdateAvailableError(_LOGGER.warning, addon=app.name, slug=slug)
+            raise AppNoUpdateAvailableError(_LOGGER.warning, app=app)
 
         # Check if available, Maybe something have changed
         store.validate_availability()
@@ -321,20 +319,18 @@ class AppManager(CoreSysAttributes):
         self.sys_jobs.current.reference = slug
 
         if slug not in self.local:
-            raise AppNotInstalledError(_LOGGER.error, slug=slug)
+            raise AppNotInstalledError(_LOGGER.error, app=slug)
         app = self.local[slug]
 
         if app.is_detached:
-            raise AppNotInStoreError(_LOGGER.error, addon=app.name, slug=slug)
+            raise AppNotInStoreError(_LOGGER.error, app=app)
         store = self.store[slug]
 
         # Check if a rebuild is possible now
         if app.version != store.version:
-            raise AppRebuildVersionChangedError(
-                _LOGGER.error, addon=app.name, slug=slug
-            )
+            raise AppRebuildVersionChangedError(_LOGGER.error, app=app)
         if not force and not app.need_build:
-            raise AppRebuildImageBasedError(_LOGGER.error, addon=app.name, slug=slug)
+            raise AppRebuildImageBasedError(_LOGGER.error, app=app)
 
         return await app.rebuild()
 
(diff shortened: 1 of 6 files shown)


supervisor/apps/manager.py
```diff
@@ -204,11 +210,11 @@ async def install(
         self.sys_jobs.current.reference = slug
 
         if slug in self.local:
-            raise AppsError(f"App {slug} is already installed", _LOGGER.warning)
+            raise AppAlreadyInstalledError(_LOGGER.warning, addon=self.local[slug].name)
```
mdegat01: I think we should use the slug not the name here. Or both if you prefer, but if so we should update all the other App errors to be consistent because we use slug in many of these right now (and even within this PR like for `AppNotInstalledError` below where we use `slug` but could get the `name`).

My reasoning is that we can look up name (and any other information about the app) from the slug later if the frontend wants. We cannot look up app info from the name if that's all we have.

mdegat01 (changes requested):
It looks like you're switching to using app name rather then slug in these errors. I get why but I think we should still return the slug for reasons below (both is fine too, just don't skip the slug imo).

Also we should be consistent across all these app errors with whichever we pick, name or slug. Exception of course for ones like `AppNotFoundError` where all we have is the input slug because its a nonexistent app.

supervisor/apps/manager.py
agners: I've noticed that the Frontend prefixes errors with the App name, which leads to a weird mix of App name and slug in the frontend (see screenshot https://github.com/home-assistant/epics/issues/50#issuecomment-4486256760 as an example).

Could we add slug as metadata always and format using the name? :thinking:

mdegat01: Yea that sounds totally reasonable to me. Want to update all of them like that? Or just do these and then another pr for the other unrelated errors?

agners: Ok, introduced a new `AppAPIError` exception class which takes the App model and extracts name/slug for all App exceptions. This way we have some consistency across App exceptions.

supervisor/exceptions.py
```diff
@@ -429,58 +502,32 @@ class AppNotSupportedError(HassioNotSupportedError):
...
-    def __init__(
-        self,
-        logger: Callable[..., None] | None = None,
-        *,
-        slug: str,
-        architectures: list[str],
-    ) -> None:
-        """Initialize exception."""
-        self.extra_fields = {"slug": slug, "architectures": ", ".join(architectures)}
-        super().__init__(None, logger)
```
mdegat01, line 447: These kind of became less well-defined as part of this. You sure you want to toss this? It doesn't have to set extra fields, we can keep the model for this specific subclass intact rather then making the caller dig into details to figure out what else they need to provide to make its message work and still let `AppAPIError` own `extra_fields` like this:
```py
    def __init__(
        self,
        logger: Callable[..., None] | None = None,
        *,
        app: Any,
        architectures: list[str],
    ) -> None:
        """Initialize exception."""
        super().__init__(logger, app=app, architectures=", ".join(architectures))
```
Similarly for the other 2 of this pattern.

supervisor/exceptions.py
```diff
@@ -361,6 +361,79 @@ class AppsError(HassioError):
...
+    def __init__(
+        self,
+        logger: Callable[..., None] | None = None,
+        *,
+        app: Any,
+        **extra_fields: Any,
+    ) -> None:
+        """Initialize exception."""
+        if isinstance(app, str):
+            self.extra_fields = {"slug": app, **extra_fields}
```
mdegat01: Do we need to force `AppNotFoundError` and `AppNotInstalledError` into this model? Since there is no `App` it feels like they are not instances of `AppAPIError` to me. They could just inherit from `AppsError` and `APIError` directly since they are in fact different from the others, there is no app context.

mdegat01 (approved):
I'm ok with it as is but I don't love the two things I called out below. I'd probably prefer we use `AppAPIError` specifically for app-related errors where there actually is an app for context only and let the `NotSupported` ones override the init with a tighter model.

supervisor/exceptions.py
agners: Ugh, yeah forcing it not into this model makes `AppAPIError` neater as well. Agreed, fixed!

PR #6858, opened 2026-05-20
docker: log when stored registry credentials are used for a pull

agners:
## Proposed change

Promote the "stored registry credentials are being used" log lines from DEBUG to INFO, in both the image pull path (`DockerInterface._get_credentials`) and the manifest fetcher (`RegistryManifestFetcher._get_auth_token`). The messages now also include the image (or repository) being accessed.

Background: stored registry credentials are applied globally to any pull whose image matches the configured registry hostname. This is fine while the credentials are valid, but our public images on `ghcr.io` (Core, Supervisor, plugins) can normally be pulled anonymously - so as soon as the stored credentials become invalid (token expired or revoked), every `ghcr.io` pull starts failing with 401/403 even though anonymous access would still work. The user has typically forgotten the credentials were ever added (often during a private add-on install), and the supervisor logs previously gave no hint that credentials were in play. See home-assistant/core#170281, where exactly this happened: an expired GitHub token blocked Core updates with a misleading `denied: denied` error.

With this change, a failing pull that used stored credentials leaves a visible breadcrumb in the logs pointing at the likely cause:

```
INFO  Using stored registry credentials for ghcr.io (user: …) to fetch manifest for home-assistant/green-homeassistant
WARN  Failed to fetch manifest for ghcr.io/home-assistant/green-homeassistant:2026.5.1 - 401
```

Usernames are not secret and are already routinely shared in support logs, so logging them at INFO is acceptable. Credentials are only used when registries are configured, so this does not add meaningful log noise for the typical installation.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR is related to issue: home-assistant/core#170281
- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/docker/interface.py
+++ b/supervisor/docker/interface.py
@@ -228,10 +228,11 @@ class DockerInterface(JobGroup, ABC):
             if registry in (DOCKER_HUB, DOCKER_HUB_LEGACY):
                 qualified_image = f"{DOCKER_HUB}/{image}"
 
-            _LOGGER.debug(
-                "Logging in to %s as %s",
+            _LOGGER.info(
+                "Using stored registry credentials for %s (user: %s) to pull %s",
                 registry,
                 stored[ATTR_USERNAME],
+                image,
             )
 
         return credentials, qualified_image
--- a/supervisor/docker/manifest.py
+++ b/supervisor/docker/manifest.py
@@ -172,7 +172,12 @@ class RegistryManifestFetcher:
             username, password = credentials
             if username and password:
                 auth = aiohttp.BasicAuth(username, password)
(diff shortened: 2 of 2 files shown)


mdegat01 (approved):
Good point, makes sense.

Not this PR but do we want to make this a repair? A repair flow would make this nice and easy to fix, prompt for updated credentials or a "revert to anonymous" option if the registry is one that supports that.

agners:
> Not this PR but do we want to make this a repair? A repair flow would make this nice and easy to fix, prompt for updated credentials or a "revert to anonymous" option if the registry is one that supports that.

Hm, yeah would definitely be more user friendly. I don't think the feature is used all that much, so it isn't high priority from my point of view.

PR #6862, opened 2026-05-20
apps: avoid blocking is_file() check in restore

agners:
## Proposed change

While restoring an app from a backup, `App.restore()` called `Path.is_file()` directly on the event loop to check whether the bundle contained an `image.tar`. Under blockbuster-based blocking-call detection this surfaced as an unhandled `BlockingError: Blocking call to os.stat`, originating from `os.path.isfile`.

Wrap the check in `self.sys_run_in_executor(...)` so the stat happens off the main thread. This matches the pattern already used for the same kind of check in `supervisor/api/backups.py`.

## Type of change

- [ ] Dependency upgrade
- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (which adds functionality to the supervisor)
- [ ] Breaking change (fix/feature causing existing functionality to break)
- [ ] Code quality improvements to existing code or addition of tests

## Additional information

- This PR fixes or closes issue: fixes #
- This PR is related to issue:
- Link to documentation pull request:
- Link to cli pull request:
- Link to client library pull request:

## Checklist

- [ ] The code change is tested and works locally.
- [ ] Local tests pass. **Your PR cannot be merged unless tests pass**
- [x] There is no commented out code in this PR.
- [ ] I have followed the [development checklist][dev-checklist]
- [x] The code has been formatted using Ruff (`ruff format supervisor tests`)
- [ ] Tests have been added to verify that the new code works.

If API endpoints or add-on configuration are added/changed:

- [ ] Documentation added/updated for [developers.home-assistant.io][docs-repository]
- [ ] [CLI][cli-repository] updated (if necessary)
- [ ] [Client library][client-library-repository] updated (if necessary)

[dev-checklist]: https://developers.home-assistant.io/docs/en/development_checklist.html
[docs-repository]: https://github.com/home-assistant/developers.home-assistant
[cli-repository]: https://github.com/home-assistant/cli
[client-library-repository]: https://github.com/home-assistant-libs/python-supervisor-client/

PR #6869, opened 2026-05-22
Enable blocking I/O detection in integration tests

mdegat01:
## Proposed change

The `run_supervisor` CI job runs integration tests against a live Supervisor instance to catch issues unit tests miss. We already use [blockbuster](https://github.com/PerchunPak/blockbuster) to detect blocking I/O in unit tests, but the integration tests have not had this enabled.

This adds a single line before Supervisor starts to write a `config.json` with `detect_blocking_io: true` into the data directory, so Supervisor activates blocking I/O detection at startup for all integration test runs.

## Type of change

## Additional information

## Checklist

### DIFF
--- a/.github/workflows/builder.yml
+++ b/.github/workflows/builder.yml
@@ -280,6 +280,7 @@ jobs:
       - name: Create the Supervisor
         run: |
           mkdir -p /tmp/supervisor/data
+          echo '{"detect_blocking_io": true}' > /tmp/supervisor/data/config.json
           docker create --name hassio_supervisor \
             --privileged \
             --security-opt seccomp=unconfined \
--- a/.github/workflows/builder.yml
+++ b/.github/workflows/builder.yml
@@ -191,7 +191,7 @@ jobs:
         with:
           arch: ${{ matrix.arch }}
           container-registry-password: ${{ secrets.GITHUB_TOKEN }}
-          cosign-base-identity: "https://github.com/home-assistant/docker-base/.*"
+          cosign-base-identity: 'https://github.com/home-assistant/docker-base/.*'
           cosign-base-verify: ghcr.io/home-assistant/base-python:3.14-alpine3.22
           image: ${{ matrix.image }}
           image-tags: |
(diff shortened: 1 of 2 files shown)


mdegat01:
We still should have a unit test for the issue caught by https://github.com/home-assistant/supervisor/pull/6862 but I realized we did have one test for it already: our "run_supervisor" workflow backs up and restores an app as part of its steps. But we don't have blockbuster enabled in these tests and we should so this fixes that.

mdegat01:
Clearly draft because I don't actually want to change the timeout to 30s, I just wanted it to fail faster temporarily 😆 

I'm not entirely sure what's wrong with backup/restore right now since I don't actually see a `BlockingError` anywhere. Going to let the other blocking error fix PRs go in to see if that fixes it somehow. If not I'll figure out whats wrong in the morning since its late here.

agners:
Hm, yeah the "Restore SSH app from backup" step fails, but the logs show a Docker 500 error :thinking: 

```
2026-05-21 21:11:25.870 ERROR (MainThread) [supervisor.docker.interface] Can't install homeassistant/amd64-addon-ssh:10.2.0: [500] Head "https://registry-1.docker.io/v2/homeassistant/amd64-addon-ssh/manifests/10.2.0": Get "https://auth.docker.io/token?scope=repository%3Ahomeassistant%2Famd64-addon-ssh%3Apull&service=registry.docker.io": net/http: request canceled (Client.Timeout exceeded while awaiting headers)
```

.github/workflows/builder.yml
```diff
@@ -308,7 +308,7 @@ jobs:
             sleep 1
           done
           echo "Waiting for Supervisor API at http://${SUPERVISOR}/supervisor/ping"
-          timeout=300
+          timeout=30
```
agners: Let's use a higher timeout, it might be slower depending on infrastructure load etc...

mdegat01: Fixed. Well actually just reverted this file entirely since we're going to use the dev system flag to set this config instead as we discussed. No changes are required to builder.yml anymore.

agners (approved):
Nice, LGTM!

PR #6877, opened 2026-05-25
fix(docker): restore add-on device access after USB re-enumeration

carlossg:
## Proposed change

Add-ons that get their serial device from the options schema (e.g. Z-Wave JS
`device:`) lose access whenever the device re-enumerates to a different minor
number (e.g. `ttyACM0` → `ttyACM1` after a HAOS reboot). The add-on then
crash-loops on `EACCES` until the user restarts it.

**Key prerequisite:** the user must have selected the device via a
`/dev/serial/by-id/…` symlink (the stable path shown in the add-on UI). On
re-enumeration the kernel assigns a new device node and a new major:minor pair,
but the by-id symlink stays the same. Supervisor must therefore match the
incoming hardware event via that stable by-id link and re-grant cgroup access
with the device's current major:minor.

Two independent bugs in `DockerApp` caused this:

1. **Listener never registered for options-based devices.** `_hw_listener` was
   only wired up when `addon.static_devices` (the manifest's `devices:` list)
   was non-empty. Add-ons that expose a device only via the options schema
   never registered the listener, so `add_devices_allowed` was never called
   when the device reappeared at a new minor.

2. **Options-based devices never matched in `_hardware_events`.** The handler
   compared the incoming `Device` only against `static_devices` (and only
   against `.path`/`.sysfs`, not `.links`). Devices configured via the options
   schema were never matched, so access was never re-granted even if the
   listener had been registered.

### Fix

Introduce `AppOptions.extract_device_paths` and `AppModel.option_device_paths`:
a cheap property that walks the raw schema and options to return the configured
device path strings as `set[Path]`, **without** running full options validation
(no hardware lookup, no pwnd hashing).

In `_hardware_events`, replace the old two-condition guard with a single
set-intersection:

```python
allowed_paths = set(self.app.static_devices) | self.app.option_device_paths
if not allowed_paths & {device.path, device.sysfs, *device.links}:
    return
```

This is strictly better:
- No per-event full validation.
- Matches by-id symlinks (via `device.links`) for both static and
  options-based devices.
- Uses the incoming device's current major:minor for the cgroup rule, which is
  the whole point of re-enumeration handling.

### Reproduction

1. Install Z-Wave JS, point its `device:` option at a `/dev/serial/by-id/…`
   symlink.
2. Replug the USB stick (or reboot HAOS) so it re-enumerates to a different
   minor.
3. Before fix: add-on crash-loops on `EACCES` reading the device.
   After fix: Supervisor re-grants cgroup access and the add-on recovers.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/docker/app.py
+++ b/supervisor/docker/app.py
@@ -618,8 +618,9 @@ class DockerApp(DockerInterface):
             _LOGGER.warning("Can't update DNS for %s", self.name)
             await async_capture_exception(err)
 
-        # Hardware Access
-        if self.app.static_devices:
+        # Hardware Access — register listener for both manifest static_devices and
+        # options-based devices (e.g. Z-Wave JS `device:` option).
+        if self.app.static_devices or self.app.devices:
             self._hw_listener = self.sys_bus.register_event(
                 BusEvent.HARDWARE_NEW_DEVICE, self._hardware_events
             )
@@ -887,9 +888,12 @@ class DockerApp(DockerInterface):
     )
     async def _hardware_events(self, device: Device) -> None:
         """Process Hardware events for adjust device access."""
-        if not any(
-            device_path in (device.path, device.sysfs)
-            for device_path in self.app.static_devices
+        if (
+            not any(
+                device_path in (device.path, device.sysfs)
+                for device_path in self.app.static_devices
+            )
+            and device not in self.app.devices
         ):
             return
 
--- a/supervisor/docker/app.py
+++ b/supervisor/docker/app.py
@@ -620,7 +620,7 @@ class DockerApp(DockerInterface):
 
         # Hardware Access — register listener for both manifest static_devices and
         # options-based devices (e.g. Z-Wave JS `device:` option).
-        if self.app.static_devices or self.app.devices:
+        if self.app.static_devices or self.app.option_device_paths:
             self._hw_listener = self.sys_bus.register_event(
                 BusEvent.HARDWARE_NEW_DEVICE, self._hardware_events
             )
@@ -888,13 +888,8 @@ class DockerApp(DockerInterface):
     )
     async def _hardware_events(self, device: Device) -> None:
         """Process Hardware events for adjust device access."""
-        if (
-            not any(
-                device_path in (device.path, device.sysfs)
-                for device_path in self.app.static_devices
-            )
-            and device not in self.app.devices
-        ):
+        allowed_paths = set(self.app.static_devices) | self.app.option_device_paths
+        if not allowed_paths & {device.path, device.sysfs, *device.links}:
             return
 
         try:
--- a/supervisor/docker/app.py
+++ b/supervisor/docker/app.py
@@ -892,6 +892,15 @@ class DockerApp(DockerInterface):
         if not allowed_paths & {device.path, device.sysfs, *device.links}:
             return
 
+        # Check hardware access policy before granting cgroup access
+        if not self.sys_hardware.policy.allowed_for_access(device):
+            _LOGGER.error(
+                "App %s tried to access blocked device %s via hardware event!",
+                self.app.name,
+                device.name,
+            )
+            return
+
         try:
             docker_container = await self.sys_docker.containers.get(self.name)
         except aiodocker.DockerError as err:
--- a/tests/docker/test_app.py
+++ b/tests/docker/test_app.py
@@ -28,6 +28,7 @@ from supervisor.docker.const import (
 from supervisor.docker.manager import DockerAPI
 from supervisor.exceptions import CoreDNSError, DockerNotFound
 from supervisor.hardware.data import Device
+from supervisor.host.const import HostFeature
 from supervisor.os.manager import OSManager
 from supervisor.plugins.dns import PluginDns
 from supervisor.resolution.const import ContextType, IssueType, SuggestionType
@@ -432,6 +433,11 @@ async def test_app_new_device(
         patch.object(App, "write_options"),
         patch.object(OSManager, "available", new=PropertyMock(return_value=is_os)),
         patch.object(
+            type(coresys.host),
+            "features",
+            new=PropertyMock(return_value=[HostFeature.OS_AGENT]),
+        ),
+        patch.object(
             CGroup, "add_devices_allowed", new_callable=AsyncMock
         ) as add_devices,
     ):
@@ -459,6 +465,11 @@ async def test_app_new_device_no_haos(
     with (
         patch.object(App, "write_options"),
         patch.object(OSManager, "available", new=PropertyMock(return_value=False)),
+        patch.object(
+            type(coresys.host),
+            "features",
+            new=PropertyMock(return_value=[HostFeature.OS_AGENT]),
+        ),
         patch.object(
             CGroup, "add_devices_allowed", new_callable=AsyncMock
         ) as add_devices,
@@ -619,6 +630,11 @@ async def test_app_options_device_hw_listener(
         # called exactly once — from the hardware event via the options listener.
         patch.object(OSManager, "available", new=PropertyMock(return_value=False)),
         patch.object(
+            type(coresys.host),
+            "features",
+            new=PropertyMock(return_value=[HostFeature.OS_AGENT]),
+        ),
+        patch.object(
             CGroup, "add_devices_allowed", new_callable=AsyncMock
         ) as add_devices,
     ):
@@ -656,6 +672,11 @@ async def test_app_options_device_policy_check(
     with (
         patch.object(App, "write_options"),
         patch.object(OSManager, "available", new=PropertyMock(return_value=True)),
+        patch.object(
+            type(coresys.host),
+            "features",
+            new=PropertyMock(return_value=[HostFeature.OS_AGENT]),
+        ),
         # Mock policy to block this device
         patch.object(
             coresys.hardware.policy,
--- a/tests/docker/test_app.py
+++ b/tests/docker/test_app.py
@@ -8,6 +8,7 @@ from typing import Any
 from unittest.mock import AsyncMock, MagicMock, Mock, PropertyMock, patch
 
 import aiodocker
+import attr
 import pytest
 
 from supervisor.apps import validate as vd
@@ -380,7 +381,7 @@ TEST_DEV_PATH = "/dev/ttyAMA0"
 TEST_SYSFS_PATH = "/sys/devices/platform/soc/ffe09000.usb/ff500000.usb/xhci-hcd.0.auto/usb1/1-1/1-1.1/1-1.1:1.0/tty/ttyACM0"
 TEST_HW_DEVICE = Device(
     name="ttyACM0",
-    path=Path("/dev/ttyAMA0"),
+    path=Path("/dev/ttyACM0"),
     sysfs=Path(
         "/sys/devices/platform/soc/ffe09000.usb/ff500000.usb/xhci-hcd.0.auto/usb1/1-1/1-1.1/1-1.1:1.0/tty/ttyACM0"
     ),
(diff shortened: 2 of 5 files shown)


agners:
> * Listener never registered for options-based devices. _hw_listener was only registered when addon.static_devices was non-empty. Add-ons that expose a device via the options schema (e.g. Z-Wave JS device:
>   option) never had the listener registered, so add_devices_allowed was never called when the device reappeared after a disconnect.

This is not true, the options validation adds it:

https://github.com/home-assistant/supervisor/blob/b54a01ed2210f946356712567ebe3564e6bc1fac/supervisor/apps/options.py#L171-L195

So it adds it to the resulting `self.devices`. The symlink themselfs don't need cgroups rules, since they are symlinks...

> 
> * by-id paths never matched in _hardware_events. The event handler compared only device.path and device.sysfs against static_devices. When static_devices (or the options path) contains a by-id symlink, the
>   match always failed because by-id paths live in device.links.

That is handled by the above code, by-id paths are resolved and the underlying device is added.

Did updating the permission not work for you? You should see `Added cgroup permissions ...` logs.

If its not working, what type of installation do you have? Can you share the logs?

carlossg:
@agners thanks for the review, I've checked a bit more and part of the PR is not needed
The issue happens when the device is re-enumerated (could be due to power issues), then zwave addon needds to be restarted to work again.
I've been running with a couple patches for a while to test before submitting and don't have logs, will revert them and see if I can reproduce

I'll post Claude analysis

> You're right that options.py populates options_schema.devices — but that's not where the bug is.
> 
> The device is added to the container correctly at start time: cgroups_rules already iterates app.devices (line 184), so Z-Wave JS gets the right cgroup rules baked in and works fine initially.
> 
> The problem is only on USB re-enumeration. When the stick disconnects and reconnects at a different minor number (e.g. ttyACM0 → ttyACM1), the baked-in rules are stale. _hw_listener exists specifically to
> call add_devices_allowed dynamically to fix this — but before this patch it was gated on:
> if self.app.static_devices:
> For Z-Wave JS, static_devices reads from the manifest's devices: field (model.py:357), which is empty — the device path comes entirely from the schema: options. So the listener was never registered and the
> stale rules were never updated after a reconnect.
> 
> You're correct about the by-id point though — options validation resolves symlinks to Device objects before adding them, so the device_all_paths expansion wasn't needed. I've removed that part and its test
> from the PR.

carlossg:
I upgraded core and it happened again. Attaching logs

[issue-zwave_js.log](https://github.com/user-attachments/files/28312834/issue-zwave_js.log)
[issue-supervisor.log](https://github.com/user-attachments/files/28312949/issue-supervisor.log)


```
Installation type: Home Assistant OS (HAOS)
Board:            Raspberry Pi 3 64-bit (rpi3-64)
HAOS version:     17.3
HA Core:          2026.5.4
Supervisor:       2026.05.1 (official, unpatched)
Z-Wave JS:        1.3.0 (core add-on)

Device configuration in Z-Wave JS options:
  "device": "/dev/serial/by-id/usb-0658_0200-if00"

Device state at time of failure:
  /dev/ttyACM1  (major 166, minor 1)
  /dev/serial/by-id/usb-0658_0200-if00 -> ../../ttyACM1

Previous device state (before 15:19 re-enumeration):
  /dev/ttyACM0  (major 166, minor 0)
  /dev/serial/by-id/usb-0658_0200-if00 -> ../../ttyACM0

Z-Wave stick: Sigma Designs USB Z-Wave controller (0658:0200)
  sysfs path: /sys/bus/usb/devices/1-1.4

Timeline:
  15:18:56  ttyACM0 removed (USB bus hang, kernel dwc_otg_hcd_urb_dequeue timeouts)
  15:19:01  USB device re-added at new address
  15:19:03  ttyACM1 appeared — minor changed from 0 to 1
  15:19:03  No "Added cgroup permissions" log from supervisor (bug)
  16:35:12  Z-Wave JS begins failing with "Operation not permitted"
```

```
  The chain of events that led to the Z-Wave failure:

  ┌──────────┬─────────────────────────────────────────────────────────────────────────┐
  │   Time   │                                  Event                                  │
  ├──────────┼─────────────────────────────────────────────────────────────────────────┤
  │ 15:11:52 │ Supervisor starts HA Core update to 2026.5.4                            │
  ├──────────┼─────────────────────────────────────────────────────────────────────────┤
  │ 15:11:55 │ Docker begins pulling new image                                         │
  ├──────────┼─────────────────────────────────────────────────────────────────────────┤
  │ 15:12:35 │ OOM killer fires — memory exhausted                                     │
  ├──────────┼─────────────────────────────────────────────────────────────────────────┤
  │ 15:12:42 │ Tailscale, nginx, SSH, ttyd killed                                      │
  ├──────────┼─────────────────────────────────────────────────────────────────────────┤
  │ 15:18:56 │ Z-Wave USB stick drops off (USB subsystem destabilised by OOM pressure) │
  ├──────────┼─────────────────────────────────────────────────────────────────────────┤
  │ 15:19:03 │ Stick reconnects as ttyACM1                                             │
  ├──────────┼─────────────────────────────────────────────────────────────────────────┤
  │ 15:19:03 │ No cgroup update — bug triggered                                        │
  ├──────────┼─────────────────────────────────────────────────────────────────────────┤
  │ 16:35+   │ Z-Wave JS failing Operation not permitted                               │
  └──────────┴─────────────────────────────────────────────────────────────────────────┘
```

agners (commented):
Can you update the PR description to accurately reflect what the current change is addressing?

tests/docker/test_app.py
```diff
@@ -572,3 +572,40 @@ async def test_ulimits_integration(coresys: CoreSys, install_app_ssh: App):
...
+    container: MagicMock,
+    docker: DockerAPI,
+):
+    """Test hw_listener is registered and fires for options-based devices.
+
+    Before the fix, _hw_listener was only registered when addon.static_devices
+    was non-empty. Add-ons like Z-Wave JS that configure the device via the
+    options schema (addon.devices) never had a listener, so cgroup permissions
+    were never updated after a USB re-enumeration.
+    """
```
agners, line 602: A full end-to-end test that sets a real device: option resolving to a by-id link (and ideally simulates a minor-number change on re-enumeration) would be nice.

supervisor/docker/app.py
```diff
@@ -887,9 +888,12 @@ async def stop(self, remove_container: bool = True) -> None:
...
         """Process Hardware events for adjust device access."""
-        if not any(
-            device_path in (device.path, device.sysfs)
-            for device_path in self.app.static_devices
+        if (
+            not any(
+                device_path in (device.path, device.sysfs)
+                for device_path in self.app.static_devices
+            )
+            and device not in self.app.devices
```
agners: Hm, this means on every hardware event we do a full validation, including pwnd hashing and whatnot.

The device extraction is just a side effect, but it's really not ideal doing this as part of the validation.

Now this is a bit bigger refactor, but I think its worth doing since this would introduce a lot of validation otherwise:

 The cleaner separation is two distinct concerns:                                                                                                                                                                                                              
                                                                                                                                                                                                                                                               
 1. What did the user configure? A set of device paths, depends only on schema + options.json. Changes only when options change.
 2. Does this event's device match one of them? Should be a cheap comparison.

And the key realization: in the event handler you don't need to resolve anything against live hardware. The incoming device already carries .path, .sysfs, and .links (which includes the by-id link). So you just intersect the configured paths against the device's own identifiers:
                                                                                                                                                                                                                                                               
 ```python                                                                                                                                                                                                                                                     
   async def _hardware_events(self, device: Device) -> None:                                                                                                                                                                                                   
       allowed_paths = set(self.app.static_devices) | self.app.option_device_paths                                                                                                                                                                             
       if not allowed_paths & {device.path, device.sysfs, *device.links}:                                                                                                                                                                                      
           return                                                                                                                                                                                                                                              
       ...                                                                                                                                                                                                                                                     
       permission = self.sys_hardware.policy.get_cgroups_rule(device)  # uses event's live major:minor                                                                                                                                                         
```

 This is strictly better because:
 - drops the per-event full validation entirely,                                                                                                                                                                                                               
 - also fixes by-id matching for both static_devices and options-based devices (which I think your earlier iteration tried to do too).
 - uses the incoming device's current major:minor for the cgroup rule, which is the whole point of re-enumeration handling.

**Review by agners — CHANGES_REQUESTED** [review]
Ok, the PR description is much better. But I think it misses one core point: The user needs to choose a by-id device, which can symlink to a different real device on reenumeration (it is stated in the reproduction steps, but I think this should be noted in the main description as well).

File: supervisor/docker/app.py [hunk 3]
```diff
@@ -887,9 +888,12 @@ async def stop(self, remove_container: bool = True) -> None:
...
     async def _hardware_events(self, device: Device) -> None:
         """Process Hardware events for adjust device access."""
-        if not any(
-            device_path in (device.path, device.sysfs)
-            for device_path in self.app.static_devices
+        if (
+            not any(
+                device_path in (device.path, device.sysfs)
+                for device_path in self.app.static_devices
+            )
```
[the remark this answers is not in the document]
agners: It is preexisting, but I think this one is worth implementing so it behaves the same as on app startup.

agners (changes requested):
@carlossg can you check the Copilot remark and the test failure?

I think other than that it looks quite good to me, thanks!

tests/docker/test_app.py
```diff
@@ -572,3 +572,50 @@ async def test_ulimits_integration(coresys: CoreSys, install_app_ssh: App):
...
+    install_app_ssh.persist["options"] = {"device": str(by_id_path)}
+
+    container.id = 123
+    docker._info = replace(docker.info, cgroup="1")  # pylint: disable=protected-access
+
+    # Re-enumerated device: same by-id symlink, different kernel node (ttyACM1).
+    reenumerated_device = replace(
+        TEST_HW_DEVICE,
+        name="ttyACM1",
+        path=Path("/dev/ttyACM1"),
```
agners, line 620: This doesn't affect the minor number testing. Not sure if the below `assert_called_once_with` below tests what this PR fixes since we add the same major/minor already on startup already.

supervisor/docker/app.py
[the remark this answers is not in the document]
carlossg: Added the policy check in _hardware_events at line 895. Now follows the same pattern as cgroups_rules on startup.

tests/docker/test_app.py
carlossg: Added comprehensive test coverage: (1) Fixed existing test to properly simulate minor number change (166:0 → 166:1), (2) Added new test_app_options_device_policy_check to verify policy enforcement for options-based devices.

tests/docker/test_app.py
carlossg: Fixed - the test now properly simulates re-enumeration with different minor numbers and verifies the new cgroup permission (c 166:1 rwm).

tests/docker/test_app.py
```diff
@@ -572,3 +584,117 @@ async def test_ulimits_integration(coresys: CoreSys, install_app_ssh: App):
...
+    install_app_ssh: App,
+    container: MagicMock,
+    docker: DockerAPI,
+):
+    """Test hardware policy blocks device access even for options-based devices.
+
+    Scenario: an add-on has a device option configured, but the device is blocked
+    by hardware policy (e.g., used by system). The hardware event handler must
+    respect the policy and not grant cgroup access.
+    """
```
agners, line 666: I think this needs the `option_device_paths` to actually work, otherwise the policy check we try to test isn't checked.

agners (changes requested):
A test checking `extract_device_paths` would be nice as well.

agners (approved):
I've allowed myself to tweak the tests a bit. Otherwise the code LGTM, thanks!

PR #6887, opened 2026-05-27
Coordinate graceful shutdown with Home Assistant OS

### LINKED ISSUE #4642: Improve systemd shutdown ordering for more graceful shutdowns
Currently when initiate a shutdown many services get stopped at once, which doesn't allow an orderly shutdown of Home Assistant Core and other containers.

The shutdown procedure should be coordinated with Supervisor.

See also https://github.com/home-assistant/supervisor/pull/6631#issuecomment-4068129770.

sairon:
## Proposed change

On newer HAOS (home-assistant/operating-system#4736), hassos-supervisor.service gets a long stop timeout (420 s docker stop + 450 s TimeoutStopSec) so Supervisor can handle the SIGTERM during host teardown and gracefully stop Core, apps and plugins via its existing shutdown handler. On older releases the timeout is too short, so reboot()/shutdown() keep stopping Core in-process before requesting the reboot/power off.

For host shutdowns not initiated by Supervisor (ACPI, power button, systemctl reboot), the SIGTERM handler now checks the systemd manager state. When it is "stopping", Supervisor runs Core.shutdown() to stop managed services gracefully; on a plain Supervisor restart the state is "running", so only Supervisor stops as before. The manager state is exposed by every systemd version, so this works regardless of the OS release.

Closes home-assistant/operating-system#4642

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR fixes or closes issue: fixes home-assistant/operating-system#4642
- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/__main__.py
+++ b/supervisor/__main__.py
@@ -13,6 +13,8 @@ zlib_fast.enable()
 
 # pylint: disable=wrong-import-position
 from supervisor import bootstrap  # noqa: E402
+from supervisor.dbus.const import SystemState  # noqa: E402
+from supervisor.exceptions import HassioError  # noqa: E402
 from supervisor.utils.blockbuster import BlockBusterManager  # noqa: E402
 from supervisor.utils.logging import activate_log_queue_handler  # noqa: E402
 
@@ -68,6 +70,32 @@ if __name__ == "__main__":
 
     # Create startup task that can be cancelled gracefully
     startup_task = loop.create_task(coresys.core.start())
+    shutdown_tasks: list[asyncio.Task] = []
+
+    async def host_is_shutting_down() -> bool:
+        """Return True if systemd is shutting the host down.
+
+        This relies only on the Systemd manager state, which is exposed by
+        every Systemd version, so it works regardless of the running HAOS
+        release (no dependency on OS-side units being present).
+        """
+        if not coresys.dbus.systemd.is_connected:
+            return False
+
+        try:
+            return await coresys.dbus.systemd.get_system_state() == SystemState.STOPPING
+        except HassioError as err:
+            _LOGGER.warning("Could not read systemd manager state: %s", err)
+            return False
+
+    async def stop_supervisor() -> None:
+        """Stop Supervisor, including managed services during host shutdown."""
+        try:
+            if await host_is_shutting_down():
+                _LOGGER.info("Host shutdown detected, shutting down managed services")
+                await coresys.core.shutdown()
+        finally:
+            await coresys.core.stop()
 
     def shutdown_handler() -> None:
         """Handle shutdown signals gracefully during startup."""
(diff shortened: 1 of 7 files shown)


supervisor/__main__.py
```diff
@@ -68,14 +70,43 @@ def run_os_startup_check_cleanup() -> None:
...
+            return False
+
+    async def stop_supervisor() -> None:
+        """Stop Supervisor, including managed services during host shutdown."""
+        try:
+            if await host_is_shutting_down():
+                _LOGGER.info("Host shutdown detected, shutting down managed services")
+                await coresys.core.shutdown()
+        finally:
+            await coresys.core.stop()
```
agners, line 98: So this means we basically attempt a coordinated `shutdown()` on `SIGTERM` on any OS version right?

Just thinking about the potential risks here, so this maybe makes it more likely that Supervisor gets killed hard because it attempts to wait for Core/Plug-ins/Apps to get shutdown gracefully 🤔

supervisor/__main__.py
```diff
@@ -68,14 +70,43 @@ def run_os_startup_check_cleanup() -> None:
...
         """Handle shutdown signals gracefully during startup."""
         if not startup_task.done():
             _LOGGER.warning("Supervisor startup interrupted by shutdown signal")
             startup_task.cancel()
 
-        coresys.create_task(coresys.core.stop())
+        if shutdown_tasks and not shutdown_tasks[0].done():
+            return
+
+        shutdown_tasks[:] = [coresys.create_task(stop_supervisor())]
```
agners, line 109: I think the Core state/asyncio.Event based race condition cleanup/re-entrant support in #6631 is a bit cleaner. But I can do it as a follow up to this, then we can discuss/hash out details.

supervisor/__main__.py
agners: So I thought a bit more about this, and what options we have. We could use `asyncio.timeout(...)` for the shutdown task and choose a timeout sensible depending on OS version (today maybe 20s, for HAOS 18 and newer maybe 420s). This would make sure that Supervisor gets around to call stop and (hopefully) manages to shutdown itself in time.

The question boils down to: do we value Supervisor itself exiting cleanly (run `stop()`, write `_update_last_boot`) more than giving more time to `shutdown()` to reach the last apps before Supervisor `SIGKILL`?

`_update_last_boot` seems not problematic, it is used for graceful Supervsior restarts. Our stop doesn't have all that important things to do, so maybe sacrifice Supervisor's graceful stop for the greater good is a reasonable tradeoff.

Let's go with what we have for now, we can do some testing and reconsider if we see this is problematic in practice.

agners (approved):

supervisor/__main__.py
sairon: Comparing what `shutdown()` and `stop()` does, I'd say attempting the coordinated shutdown and risking it gets killed there, while maybe skipping `stop` completely doesn't appear too harmful. But you might have better insights here.

(EDIT: I wrote this before I saw your comment.)

supervisor/__main__.py
sairon: Okay, follow-up works for me.

PR #6888, opened 2026-05-27
Fix off-by-one in unsupported os_version evaluation

sairon:
## Proposed change

Adjust check for unsupported OS version to match the boundary which is documented in [1], which says:

> Supervisor considers Home Assistant Operating System older than the last 4 major releases as unsupported.

Also adjust the wording in the message to match the statement from the docs.

This will make another major relase unsupported now, but it was the original intention anyway.

[1] https://www.home-assistant.io/more-info/unsupported/os_version/

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/tests/resolution/evaluation/test_evaluate_os_version.py
+++ b/tests/resolution/evaluation/test_evaluate_os_version.py
@@ -14,20 +14,33 @@ from supervisor.resolution.evaluations.os_version import EvaluateOSVersion
 @pytest.mark.parametrize(
     ("current", "latest", "expected"),
     [
-        ("10.0", "15.0", True),  # 5 major behind, should be unsupported
-        ("10.0", "14.0", True),  # 4 major behind, should be unsupported
-        ("14.0", "18.0", True),  # 4 major behind, should be unsupported
-        ("15.0", "18.0", False),  # 3 major behind, last supported
-        ("10.2", "11.0", False),  # 1 major behind, supported
+        ("10.0", "15.0", True),  # older than the last 4 major releases (15, 14, 13, 12)
+        ("10.0", "14.0", True),  # older than the last 4 major releases (14, 13, 12, 11)
+        ("14.0", "18.0", True),  # older than the last 4 major releases (18, 17, 16, 15)
+        ("15.0", "18.0", False),  # last of 4 supported major releases (18, 17, 16, 15)
+        (
+            "10.2",
+            "11.0",
+            False,
+            # not older than the last 4 major releases (11, 10, 9, 8)
+        ),
         ("10.4", "10.5", False),  # same major, supported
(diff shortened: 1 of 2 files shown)


tests/resolution/evaluation/test_evaluate_os_version.py
```diff
@@ -15,7 +15,9 @@
     ("current", "latest", "expected"),
     [
         ("10.0", "15.0", True),  # 5 major behind, should be unsupported
-        ("10.0", "14.0", False),  # 4 major behind, should be supported
+        ("10.0", "14.0", True),  # 4 major behind, should be unsupported
+        ("14.0", "18.0", True),  # 4 major behind, should be unsupported
+        ("15.0", "18.0", False),  # 3 major behind, last supported
```
agners: I find this comments a bit confusing actually, since in my reading 15 is 4 major versions behind 18 (18, 17, 16 and 15)...

agners (approved):
LGTM, nice find.

If you agree on the comments, then lets fix those before merging.

tests/resolution/evaluation/test_evaluate_os_version.py
sairon: Then we'll need to find different (less confusing) wording, because it's incorrect that 15 is 4 versions **behind** 18. How many major versions is 17 **behind** 18? It's 1. By extension, 15 is really just 3 majors behind it.

agners: Well, 18 is clearly a release by itself too, if its used as `latest_release`, no? So if you count the releases, it's quite clear to me that 14 is *more* 4 versions behind 18 (major release 18, 17, 16 and 15). In natural language, in my mind, its quite clear that we need to include 18 as a release as well, and hence just subtracting doesn't give you the answer :sweat_smile: 

Maybe we can use the same wording as in text plus list the releases to make it clear?

```py
        ("10.0", "14.0", True),  # older than the last 4 major releases (14, 13, 12, 11)
        ("14.0", "18.0", True),  # older than the last 4 major releases (18, 17, 16, 15)
        ("15.0", "18.0", False),  # not older than the last 4 major releases (18, 17, 16, 15)
```

sairon: It's obviously the word "behind" which makes it confusing here. Because otherwise, 18 would have to be 1 version behind itself to make 15 to be 4 versions behind 18. I'll reword it as you suggest.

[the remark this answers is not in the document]
agners: Doesn't seem necessary to me. If someone is on beta channel and more than 1 release behind it would be odd already :smile:

PR #6891, opened 2026-05-28
Make Core.shutdown idempotent and safe to call concurrently

agners:
## Proposed change

After #6887, `Core.shutdown()` runs from the SIGTERM signal handler during host shutdown in addition to the existing call sites (`host.control.reboot()`/`shutdown()`, backup restore). To avoid spawning a second `stop_supervisor()` task when a repeat SIGTERM arrives, `__main__.py` debounced the signal handler by stashing the in-flight task in a single-element list and bailing out on subsequent invocations. That workaround only covers the SIGTERM-vs-SIGTERM race; two API-initiated reboots, or a reboot API call racing with SIGTERM, would still re-enter the shutdown sequence.

This PR moves the idempotency into `Core.shutdown()` itself, where every caller benefits:

- A second call while a shutdown is in progress awaits the in-flight shutdown via an `asyncio.Event` rather than re-running the sequence.
- Calls while state is `STOPPING`/`CLOSE` return early — Supervisor is already going away, the work is moot.
- Calls while state is in `STARTING_STATES` (`INITIALIZE`/`STARTUP`/`SETUP`) return early too. There is nothing coherent to gracefully stop before startup completes, and on the SIGTERM-during-startup path the caller cancels `startup_task` first, so waiting for it to complete would deadlock (the `finally` in `Core.start()` cannot reach `set_state(RUNNING)` once awaits raise `CancelledError`).
- The sequence is wrapped in `try`/`finally` so the completion event is set even when an inner step raises.

With idempotency in `Core`, the closure-with-list workaround in `__main__.py` collapses to a plain `coresys.create_task(stop_supervisor())`: repeat SIGTERMs spawn extra tasks but each just observes the in-flight shutdown and waits.

Inspired by (a subset of) #6631.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR is related to issue: #6887
- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/core.py
+++ b/supervisor/core.py
@@ -43,6 +43,7 @@ class Core(CoreSysAttributes):
         self.coresys: CoreSys = coresys
         self._state: CoreState = CoreState.INITIALIZE
         self.exit_code: int = 0
+        self._shutdown_event: asyncio.Event = asyncio.Event()
 
     @property
     def state(self) -> CoreState:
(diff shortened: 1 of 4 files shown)


supervisor/core.py
```diff
@@ -364,28 +365,57 @@ async def stop(self) -> None:
...
+            _LOGGER.warning("Ignoring shutdown request, Supervisor is already stopping")
+            return
+
+        # Another shutdown is in progress, wait for it to complete
+        if self.state == CoreState.SHUTDOWN:
+            await self._shutdown_event.wait()
+            return
+
+        # Reset event for this shutdown cycle (supports repeated use, e.g. backup restore)
+        self._shutdown_event.clear()
```
mdegat01: The concept of repeated use doesn't really make sense here. Shutdown eventually leads to a stop in the python process which obviously clears the event. And there's no way to go back to `CoreState.RUNNING` from those closing states. So I'm not really sure when this would have any effect? At the very least the comment should probably be adjusted since its example isn't a real use case.

tests/test_core.py
```diff
@@ -233,3 +233,98 @@ async def test_setup_unhandled_exception_captured(
...
+    async def track_app_shutdown(startup):
+        nonlocal second_entered
+        second_entered = True
+        return await original_shutdown(startup)
+
+    with patch.object(coresys.apps, "shutdown", side_effect=track_app_shutdown):
+        await coresys.core.shutdown()
+
+    assert second_entered
+    assert coresys.core._shutdown_event.is_set()
```
mdegat01: I mean this is very much a fabricated use case, it can't ever happen in production. `CoreState` cannot get back to `RUNNING` from `SHUTDOWN`. Is there a plan to support some kind of live backup and restore for Supervisor I'm unaware of?

mdegat01 (changes requested):
The code is fine, works and definitely helpful.

My comments are just about a seemingly supported use case that confuses me. There is currently no way to return to `RUNNING` after `SHUTDOWN` without exiting and restarting the python process in production. Is there a future use case here we're looking to support I'm unaware of? If not we should probably remove the perceived support for that since its a bit confusing.

agners:
> My comments are just about a seemingly supported use case that confuses me. There is currently no way to return to `RUNNING` after `SHUTDOWN` without exiting and restarting the python process in production. Is there a future use case here we're looking to support I'm unaware of? If not we should probably remove the perceived support for that since its a bit confusing.

True, there is no way from `SHUTDOWN` to `RUNNING`, but the same shutdown functionality is used by in backup restore where we go from `RUNNING` to `FREEZE`, then call this shutdown function, then move back to `RUNNING`.

Have a look at the diagram at https://github.com/home-assistant/supervisor/pull/6631#issuecomment-4067986795. There were a few changes to it from #6887, but the multiple use cases `shutdown()` is being called still holds.

Maybe `shutdown` isn't a good term for this function. And the test definitely should replicate the real-world use case (backup restore via `FREEZE` state).

agners:
I've splitted `shutdown()` into `shutdown()` and `teardown_services()`. The latter is used for backup as well. Only `shutdown()` does now the state handling and reentrant checks.

mdegat01 (approved):
That diagram is great, I didn't realize we called `shutdown` during backup and restore, my bad. I like the separation here better, makes it more intuitive. LGTM 👍

PR #6904, opened 2026-06-02
Report progress during initial Core install

agners:
## Proposed change

On a fresh system, the initial Home Assistant Core install (`home_assistant_core_install`) reported no intermediate progress: its job jumped from 0% straight to 100% once the install finished. The download/extract progress is tracked on the internal `docker_interface_install` job, but `home_assistant_core_install` declared no `child_job_syncs`, so that progress never propagated up to the user-visible install job.

This mirrors what `home_assistant_core_update` already does and syncs the docker pull progress up to the install job, allocating it the full progress range. As with the update path, this treats the image pull as 100% of the task even though image cleanup and Home Assistant start are not accounted for, which is acceptable since the pull dominates the install time.

This is the Supervisor-side change for the landing page progress indicator added in home-assistant/frontend#52359, which surfaces this progress to the user during the initial Core install.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/homeassistant/core.py
+++ b/supervisor/homeassistant/core.py
@@ -19,7 +19,7 @@ from ..bus import EventListener
 from ..const import ATTR_HOMEASSISTANT, BusEvent, CoreState
 from ..coresys import CoreSys
 from ..docker.const import ContainerState
-from ..docker.homeassistant import DockerHomeAssistant
+from ..docker.homeassistant import HASS_DOCKER_NAME, DockerHomeAssistant
 from ..docker.monitor import DockerContainerStateEvent
 from ..docker.stats import DockerStats
 from ..exceptions import (
@@ -190,9 +190,15 @@ class HomeAssistantCore(JobGroup):
         # We assume for now the docker image pull is 100% of this task. But from
         # a user perspective that isn't true. Other steps that take time which
         # is not accounted for in progress include: image cleanup and Home
-        # Assistant start.
+        # Assistant start. The reference is scoped to the Home Assistant
+        # container so a Supervisor self-update done first (which also pulls an
+        # image via docker_interface_install) does not count towards this job.
         child_job_syncs=[
-            ChildJobSyncFilter("docker_interface_install", progress_allocation=1.0)
+            ChildJobSyncFilter(
+                "docker_interface_install",
+                reference=HASS_DOCKER_NAME,
+                progress_allocation=1.0,
+            )
         ],
     )
     async def install(self) -> None:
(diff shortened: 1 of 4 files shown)


supervisor/homeassistant/core.py
```diff
@@ -187,6 +187,13 @@ async def install_landingpage(self) -> None:
         name="home_assistant_core_install",
         on_condition=HomeAssistantJobError,
         concurrency=JobConcurrency.GROUP_REJECT,
+        # We assume for now the docker image pull is 100% of this task. But from
+        # a user perspective that isn't true. Other steps that take time which
+        # is not accounted for in progress include: image cleanup and Home
+        # Assistant start.
+        child_job_syncs=[
+            ChildJobSyncFilter("docker_interface_install", progress_allocation=1.0)
```
mdegat01: I'm not sure this feature works here as is. It'll work great in the happy path but you have a few error states that I think will cause trouble:

https://github.com/home-assistant/supervisor/blob/e2df1c094adfcb7f0a99a12e996855ad6c0569d0/supervisor/homeassistant/core.py#L242

This is inside the method and also eventually a call to `docker_interface_install`. If a Supervisor update is required then this job will move from 0 to 100% with its progress even though it actually has nothing to do with the task of installing Home Assistant Core. We probably need an enhancement here that would allow us to exclude this path from becoming progress of this job.

https://github.com/home-assistant/supervisor/blob/e2df1c094adfcb7f0a99a12e996855ad6c0569d0/supervisor/homeassistant/core.py#L219

The `docker_interface_install` we do actually want called via `self.instance.update` is done in a loop. On failure of that job this job does not fail, rather it tries again. I don't think this feature was designed with that use case in mind, it was more to sync a parent with a specific child. I'd make sure to add a test case for this scenario to see if it works the way you want it to.

mdegat01 (changes requested):
There's two non-happy path scenarios that I think will break this as is we should probably account for. Or at least have test cases validating they behave the way we want if they work fine.

I guess we could also add progress reporting to `install_landing_page` the same way but it would really just be for completeness sake. Nothing can be listening until at least the landing page is up and running.

supervisor/homeassistant/core.py
agners: Hm, yeah good point. For the Supervisor update, we can fix this by adding a filter for the sync (`reference="homeassistant"`).

Technically we add both jobs as sync, e.g. with 10%/90% allocation :thinking: . I think I'd rather keep it simpler for now. Mainly for point 2.

Hm, yeah, I was first thinking that Docker maybe also skipping layers if aborted half way during download. But this would also change the layer count/size, and resume from 0 again. But the job progress sync accumulates the progress, so we'll end up overflowing :see_no_evil: .

agners: Ok, so I now ended up implementing job sync with repeated jobs. Essentially, the percentage just goes back to what it was when the job first started. It added some complexity, I could see this come in handy in other places. Worst case we can rip it out again and just live with the overshooting behavior :shrug:

mdegat01 (approved):
Nice! I actually forgot we already had support for `reference` matching in the syncs 😆

PR #6909, opened 2026-06-03
Fix CIFS mount on kernel 6.1.x by checking kernel version for retrans parameter

### LINKED ISSUE #6896: CIFS mount fails on kernel 6.1.x: retrans parameter not supported
## The problem

Since Supervisor 2026.05.1 (which includes PR #6838), CIFS network mounts fail on HAOS devices running kernel 6.1.x with:

```
cifs: Unknown parameter 'retrans'
mount error(22): Invalid argument
```

The `retrans=0` option was added to CIFS mount options in `supervisor/mounts/mount.py` line 688 as part of PR #6838. However, the `retrans` mount option for CIFS was only introduced in Linux kernel 6.6. On kernels < 6.6, this parameter is unrecognized and causes the entire mount to fail with EINVAL.

## Environment

- **Supervisor version**: 2026.05.1
- **HAOS version**: 17.3
- **Kernel**: 6.1.115-haos
- **Board**: panther-x2 (also affects Green and other HAOS devices on kernel 6.1.x)

## Reproduction

1. Add a CIFS network mount via Supervisor (Settings → System → Storage → Add Network Storage)
2. Mount fails with `mount error(22): Invalid argument`
3. Kernel log shows: `cifs: Unknown parameter 'retrans'`

## Evidence

```
$ dmesg | grep cifs
[   86.267293] cifs: Unknown parameter 'retrans'
[   86.269438] cifs: Unknown parameter 'retrans'
[  863.183161] cifs: Unknown parameter 'retrans'
```

Manual mount without `retrans` works fine:
```
mount -t cifs "//server/share" /mnt/point -o vers=2.0,iocharset=utf8,sec=none
# succeeds
```

## Proposed fix

The `retrans` option should either:
1. Be removed from CIFS mount options (the `soft` + `echo_interval=10` already provides the timeout behavior), or
2. Be conditionally included only when the running kernel supports it (check kernel version >= 6.6)

For NFS, `retrans` has been available since kernel 2.x, so no change needed there.

## Additional context

- The `retrans` parameter was added in PR #6838 to bound per-operation timeout alongside `soft` + `echo_interval=10`
- However, the CIFS `retrans` option was only added in Linux 6.6 (see kernel commit adding retrans/retrans_inval to cifs)
- HAOS Green and panther-x2 boards ship with kernel 6.1.x, meaning ALL CIFS mounts on these devices are broken
- This is a regression - CIFS mounts worked before Supervisor 2026.05.1

C3H3-AI:
## Problem

PR #6838 added `retrans=0` parameter to CIFS mounts for better reliability. However, the `retrans` parameter for CIFS was only introduced in Linux kernel 6.6. This breaks CIFS mounts on:
- HAOS Green (kernel 6.1.x)
- panther-x2 (kernel 6.1.x)
- Any device running kernel < 6.6

Error from dmesg:
```
cifs: Unknown parameter 'retrans'
```

## Solution

Detect kernel version and only add `retrans=0` if kernel is 6.6 or newer.

## Changes

1. Add `_get_kernel_version()` helper to parse kernel version from `uname -r`
2. Add `_kernel_supports_cifs_retrans()` to check for kernel 6.6+
3. Wrap `retrans=0` addition with version check in `CIFSMount.options`

## Testing

- Kernel 6.1.x: Does not add retrans parameter - mount works
- Kernel 6.6+: Adds retrans=0 parameter - mount works with improved reliability

Fixes #6896

### DIFF
--- a/supervisor/mounts/mount.py
+++ b/supervisor/mounts/mount.py
@@ -7,6 +7,8 @@ from functools import cached_property
 import logging
 import os
 from pathlib import Path, PurePath
+import re
+import subprocess
 import time
 
 from dbus_fast import Variant
@@ -58,6 +60,35 @@ def _probe_network_mount(path: Path) -> bool:
     """
     os.statvfs(path)
     return path.stat().st_dev != path.parent.stat().st_dev
+
+
+def _get_kernel_version() -> tuple[int, int]:
+    """Get kernel major and minor version.
+
+    Returns (major, minor) tuple, e.g., (6, 1) for "6.1.115-haos".
+    Returns (0, 0) if version cannot be determined.
+    """
+    try:
+        output = subprocess.check_output(["uname", "-r"], text=True).strip()
+        match = re.match(r"(\d+)\.(\d+)", output)
+        if match:
+            return (int(match.group(1)), int(match.group(2)))
+    except (subprocess.SubprocessError, OSError):
+        pass
+    return (0, 0)
+
+
+def _kernel_supports_cifs_retrans() -> bool:
+    """Check if kernel supports CIFS retrans parameter.
+
+    The retrans parameter for CIFS mounts was introduced in Linux 6.6.
+    On older kernels (like 6.1.x used by HAOS Green/panther-x2),
+    using retrans causes mount failure with "Unknown parameter 'retrans'".
+
+    Returns True if kernel is 6.6 or newer.
+    """
+    major, minor = _get_kernel_version()
+    return major > 6 or (major == 6 and minor >= 6)
 
 
 # Three layered timeouts cooperate to keep the host alive when an NFS server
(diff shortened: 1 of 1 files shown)


sairon (commented):
There are no officially supported targets which use kernel 6.1 anymore. Latest supported HAOS version is 14.0 where all targets [use kernel 6.6](https://github.com/home-assistant/operating-system/blob/14.0/Documentation/kernel.md) already (even 13.0 uses 6.6.y everywhere).

Desmond-Dong:
@sairon I think it may be best to close this PR.

From what I understand, @C3H3-AI appears to be running a third-party Home Assistant OS distribution rather than the official Home Assistant OS. Because of that, it's difficult to determine whether the issue originates from Home Assistant itself or from changes introduced by the downstream distribution.

@C3H3-AI, since you're using a third-party distribution, I would recommend reaching out to its maintainer first. They will be in a better position to verify whether any distribution-specific modifications could be affecting this behavior.

At this point, I don't think there is enough information to conclude that this is an issue with the official Home Assistant project.

C3H3-AI:
Thanks @sairon for the clarification. Since kernel 6.1 is no longer used in any officially supported HAOS targets (14.0+ all use 6.6), this fix is no longer needed. Closing.

PR #6916, opened 2026-06-06
Improve port conflict detection and resolution across Core and apps

mdegat01:
## Proposed change

This PR now focuses on app-side port conflict handling and resolution framework
cleanup.
The previously attempted Home Assistant Core port conflict detection/handling was removed from this branch.

What is included:
- Keep app startup conflict handling based on definitive runtime Docker failures (address already in use).
- Create and maintain app port-conflict issues with per-port context via `reference_extra`.
- Add app-focused resolution behavior:
  - `IssueType.APP_PORT_CONFLICT`
  - `SuggestionType.CLEAR_PORT_CONFIG`
  - check: `app_port_conflict`
  - fixup: `app_clear_port_config`
- Remove pre-emptive app options port validation so option updates are not blocked by racy checks.
- Refactor resolution contracts to typed objects:
  - checks now use `approve_check(issue: Issue)`
  - fixups now use `process_fixup(suggestion: Suggestion)`
- Propagate `reference_extra` through resolution manager relationships and events.
- Update tests for the above behavior and typed signatures.

What is explicitly not included anymore:
- Core-side port conflict detection/handling logic.
- Core-specific conflict issue/check path `app_port_conflict_core`.
- Docker manager helper used only for the removed Core path.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to documentation pull request: home-assistant/developers.home-assistant#3212
- Link to cli pull request:
- Link to client library pull request: home-assistant-libs/python-supervisor-client#361

## Checklist

- [x] Tests have been added or updated to verify behavior.

- [x] Documentation added or updated for [developers.home-assistant.io][docs-repository]

[client-library-repository]: https://github.com/home-assistant-libs/python-supervisor-client


mdegat01:
Ok to facilitate review a bit, about half the changeset here is new tests.

Of the production code changeset, the reason every single check and fixup is in here is because I needed a new field in `Suggestion` and `Issue`. `reference` wasn't enough, I needed both the app and the particular port so I added `reference_extra`. This is why `needs-client-library` and `missing-documentation` are added since there's API changes due to this.

This also meant changing `CheckBase` and `FixupBase` so all descendents needed modifications. Only checks/app_port_conflict, checks/app_port_conflict_core, and fixups/app_clear_port_config need review of those. The others just had to update a signature but there's no runtime change.

Other changes outside of `resolution` do need review though as they are the core of this work. Sorry it ended up so big.

agners:
I did a bit of research how binding is behaving exactly:

```
  Bind outcome

  ┌─────┬─────────────────────────────┬───────────────────────┬────────────┬────────────┐
  │  #  │  Scenario (first → second)  │        Options        │    TCP     │    UDP     │
  ├─────┼─────────────────────────────┼───────────────────────┼────────────┼────────────┤
  │ 1   │ 0.0.0.0 → specific          │ none                  │ EADDRINUSE │ EADDRINUSE │
  ├─────┼─────────────────────────────┼───────────────────────┼────────────┼────────────┤
  │ 2   │ 0.0.0.0 → specific          │ SO_REUSEADDR on both  │ EADDRINUSE │ succeeds   │
  ├─────┼─────────────────────────────┼───────────────────────┼────────────┼────────────┤
  │ 3   │ specific → 0.0.0.0          │ none                  │ EADDRINUSE │ EADDRINUSE │
  ├─────┼─────────────────────────────┼───────────────────────┼────────────┼────────────┤
  │ 4   │ specific → 0.0.0.0          │ SO_REUSEADDR on both  │ EADDRINUSE │ succeeds   │
  ├─────┼─────────────────────────────┼───────────────────────┼────────────┼────────────┤
  │ 5   │ two distinct specific addrs │ none                  │ succeeds   │ succeeds   │
  ├─────┼─────────────────────────────┼───────────────────────┼────────────┼────────────┤
  │ 6   │ exact same addr+port        │ none                  │ EADDRINUSE │ EADDRINUSE │
  ├─────┼─────────────────────────────┼───────────────────────┼────────────┼────────────┤
  │ 7   │ exact same addr+port        │ SO_REUSEADDR on both  │ EADDRINUSE │ succeeds   │
  ├─────┼─────────────────────────────┼───────────────────────┼────────────┼────────────┤
  │ 8   │ exact same addr+port        │ SO_REUSEPORT on both  │ succeeds   │ succeeds   │
  ├─────┼─────────────────────────────┼───────────────────────┼────────────┼────────────┤
  │ 9   │ exact same addr+port        │ reuse opt on 2nd only │ EADDRINUSE │ EADDRINUSE │
  └─────┴─────────────────────────────┴───────────────────────┴────────────┴────────────┘

  Delivery semantics (UDP, for the rows that succeed)

  ┌─────────────────────────────────────┬────────────────────────┬─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
  │               Layout                │     Delivery rule      │                                                  Verified behavior                                                  │
  ├─────────────────────────────────────┼────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ wildcard + specific (rows 2, 4)     │ most-specific-match    │ datagram to the specific IP → specific socket; all other dest addrs → wildcard socket. Independent of bind order.   │
  ├─────────────────────────────────────┼────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ two distinct specific addrs (row 5) │ by destination address │ each socket gets only its own address's traffic                                                                     │
  ├─────────────────────────────────────┼────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ same addr, SO_REUSEADDR (row 7)     │ last-bind-wins         │ the most-recently-bound socket receives everything; earlier binds shadowed (0 packets). Follows bind order, not fd. │
  ├─────────────────────────────────────┼────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ same addr, SO_REUSEPORT (row 8)     │ load-balanced          │ per-flow hash across the group (measured ~108/99/93 of 300 flows). One flow always lands on one socket.             │
  └─────────────────────────────────────┴────────────────────────┴─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
```

When an non-host network app exposes ports, Docker is asked to bind to `""`, which is any (`0.0.0.0`/`::`). From the above table it becomes clear that reusing the same port with a specific bind is not possible in that scenario. 

Most servers, including Home Assistant Core, use `SO_REUSEADDR` to avoid issues reusing the same binding within `TIIME_WAIT` (60s).

supervisor/resolution/fixups/base.py
```diff
@@ -32,7 +33,10 @@ async def __call__(self, fixing_suggestion: Suggestion | None = None) -> None:
         # Process fixup
         _LOGGER.debug("Run fixup for %s/%s", self.suggestion, self.context)
         try:
-            await self.process_fixup(reference=fixing_suggestion.reference)
+            await self.process_fixup(
+                reference=fixing_suggestion.reference,
+                reference_extra=fixing_suggestion.reference_extra,
```
agners: Hm, should we maybe just pass the `Suggestion` object then? 🤔

supervisor/api/apps.py
```diff
@@ -315,23 +384,27 @@ async def options(self, request: web.Request) -> None:
...
+                raise AppConfigurationInvalidError(
+                    app=app.slug,
+                    validation_error=humanize_error(body[ATTR_OPTIONS], ex),
+                ) from None
+        if ATTR_BOOT in body and app.boot_config == AppBootConfig.MANUAL_ONLY:
+            raise AppBootConfigCannotChangeError(
+                app=app.slug, boot_config=app.boot_config.value
+            )
+        if (new_ports := body.get(ATTR_NETWORK)) is not None:
+            await self._validate_ports(app, new_ports)
```
agners: This preventive port conflicting management has a few flaws technically:

1. An app which uses host network doesn't necessarily bind the way we'd expect
2. At installation time we don't check the default port. So the user might end up with conflicts still
3. If the user wants to swap ports of two apps it gets annoying: Move to a unused port, switch the port in the second app, go back to the first app and switch port again
4. Apps can be stopped/disabled/unused... If we save configuration we still check port conflicts
5. Core itself might bind to other ports as well which would be not covered
6. OS binds to some ports as well
7. The port usage could change between when checked and when started (e.g. another app gets installed)
8. Potential use cases might be prevented (NGINX Proxy listen on public IP while Core listen on `172.30.32.1`)

Let's make this first step about improving the error detection/user notification when port conflicts happens. If we get an error from the operating system, that is a definitive proof that a conflicts indeed happened, and ideally needs action from user side. We can make it easier for the user to be aware of it (at least for the non-host network apps).

The main reason we'd like to have this conflict detection is so that the user doesn't get locked out accidentally. But there are still realistic ways how users can run into (like installing an app with the same default). I think in a second step I'd rather prefer to have a port management for Core specifically, where Supervisor actively reserves the port during startup until Core takes over or something along these lines.

supervisor/homeassistant/core.py
```diff
@@ -82,12 +115,18 @@ def __init__(self, coresys: CoreSys):
...
         self._watchdog_listener: EventListener | None = None
+        self._startup_abort: asyncio.Event | None = None
 
     @property
     def error_state(self) -> bool:
         """Return True if system is in error."""
         return self._error_state
 
+    @property
+    def is_starting(self) -> bool:
```
agners: This appears unused?

supervisor/docker/manager.py
```diff
@@ -285,6 +331,7 @@ def __init__(self, coresys: CoreSys):
         self._manifest_fetcher: RegistryManifestFetcher = RegistryManifestFetcher(
             coresys
         )
+        self._host_port_binding_cache: dict[DockerPortBinding, str] | None = None
```
agners: This too appears unused.

supervisor/apps/app.py
```diff
@@ -1257,11 +1285,9 @@ def _check_app_config_dir():
         try:
             await self.instance.run()
         except DockerContainerPortConflict as err:
-            raise AppPortConflict(
-                _LOGGER.error,
-                name=self.slug,
-                port=cast(dict[str, Any], err.extra_fields)["port"],
-            ) from err
+            port = cast(dict[str, Any], err.extra_fields)["port"]
+            self.create_port_conflict_issue(port)
```
agners, line 1307: Should we maybe also dismiss any existing port conflicts for this app if it succeeds starting?

agners (changes requested):
The repairs make a lot of sense, especially since often port conflicts only really appear on startup (since an app just installed does not start automatically).

I'd prefer if we don't do a preemptive port conflict check. It can be racy/still cause TOCTOU problems.

supervisor/resolution/fixups/base.py
mdegat01: We can. But itll make the PR a whole lot bigger. That'll break the contract of every fixup and every test about it. I was thinking that might make the most sense though.

supervisor/apps/app.py
mdegat01: That was included in the checks:
https://github.com/home-assistant/supervisor/blob/9961beb46205634be998a81ab8bf8c76a81e645c/supervisor/resolution/checks/app_port_conflict.py#L49-L51
https://github.com/home-assistant/supervisor/blob/9961beb46205634be998a81ab8bf8c76a81e645c/supervisor/resolution/checks/app_port_conflict_core.py#L52-L57

I can put it in here too if we want but I figured I'd just let the checks handle themselves.

supervisor/api/apps.py
mdegat01: Ok I completely tossed the bit in app options validation. And I removed all my code that attempted to add port bindings from metadata that docker didn't actually see. Its still doing the same logic for handling an options call from core and then handling conflicts detected on app startup with an issue. Let me know if this works for you or if you think it needs to be scaled back further.

supervisor/resolution/fixups/base.py
mdegat01: Ok I did this. Its done as a separate commit for easier review as its definitely a big refactor.

supervisor/apps/app.py
agners: The thing I am worried is that the user will see repair/issues even after the issue got resolved (by manually change ports and start the app again). From what I understand the next check would only happen later, no?

mdegat01:
Ok @agners as discussed I removed all the core port conflict detection work as its too flaky to be reliable with how port conflicts actually works. This PR is just about better handling of app port conflicts with a new repair now.

supervisor/exceptions.py
```diff
@@ -542,6 +542,24 @@ def __init__(
         super().__init__(None, logger)
 
 
+class AppNetworkPortInUseError(AppsError, APIError):
+    """Raise if a requested host port is already in use by another component."""
```
agners: This seems unused now.

supervisor/docker/manager.py
```diff
@@ -71,6 +71,52 @@
...
+        if self.ip is None or other.ip is None:
+            return False
+
+        # Different IP versions never conflict (assuming IPV6_V6ONLY=True)
+        if self.ip.version != other.ip.version:
+            return False
+
+        # They conflict if they're the same address, or either is the wildcard
+        wildcard = IPV4_WILDCARD if self.ip.version == 4 else IPV6_WILDCARD
+        return self.ip == wildcard or other.ip in {self.ip, wildcard}
```
agners: This seems also no longer used right now, so I guess we can remove it too and add it later if we are really using it?

agners (changes requested):
I like this PR as a first step 👍 .

agners (changes requested):
Another small nit :see_no_evil: .

supervisor/apps/app.py
mdegat01: I'm going to assume this was the nit in the review. I can change this. I guess it does provide a better experience and that's most important. But it does feel like we defeat the point of the issue/suggestion system a bit the more we work creation and resolution into other places instead of leaving them self-contained. Maybe we need to figure out how to tie them to events so they aren't poll only anymore.

mdegat01: Ok I found the real nit, more dead code. Hopefully its good now

agners (approved):

mdegat01:
PRs up for client library and documentation changes. Core changes have to wait for a bump in the client library. They're also going to be more complicated then normal because of that `reference_extra` field. Two things to do, possibly separate PRs:
1. Add `reference_extra` support to `SupervisorIssues` since its now part of the key that identifies issues uniquely
2. Add a repair for the new issue type

PR #6926, opened 2026-06-08
Return 404 for raspberrypi endpoints on boards without firmware

sairon:
## Proposed change

Return "not found" when board doesn't have the firmware update available. Any other error may indicate it's worth retrying later (as discussed in [1]), so 404 is more appropriate here.

[1] https://github.com/home-assistant/core/pull/172929#discussion_r3363706261

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/api/os.py
+++ b/supervisor/api/os.py
@@ -234,12 +234,13 @@ class APIOS(CoreSysAttributes):
         ):
             raise APINotFound(
                 f"OS Agent {RPI_FIRMWARE_MIN_OS_AGENT_VERSION} or newer required "
-                "for Raspberry Pi firmware management"
+                "for Raspberry Pi firmware management",
+                _LOGGER.debug,
             )
 
         if not self.sys_dbus.agent.board.has_rpi_firmware:
             raise APINotFound(
-                "Raspberry Pi firmware is not available on this board", _LOGGER.error
+                "Raspberry Pi firmware is not available on this board", _LOGGER.debug
             )
 
     @api_process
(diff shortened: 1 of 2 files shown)


supervisor/api/os.py
```diff
@@ -238,7 +238,7 @@ def _check_rpi_firmware_available(self) -> None:
             )
 
         if not self.sys_dbus.agent.board.has_rpi_firmware:
-            raise BoardInvalidError(
+            raise APINotFound(
```
agners, line 242: The error above does not pass `_LOGGER.error`, this one does, we should make this consistent.

Ideally we apply the new error scheme with translation etc. here too. Either create `BoardNotFoundError` which derives from `APINotFound`, or maybe make the existing `BoardInvalidError` derive from `APINotFound`? 🤔

agners (changes requested):

supervisor/api/os.py
agners: Especially the latter is technically a breaking change. But I suggest just do it, make `BoardInvalidError` a `APINotFound`, mark the PR as breaking change and be done with it. Optionally rename `BoardInvalidError` to `BoardNotFoundError`.

sairon: I'm a bit hesitant to touch `BoardInvalidError` here as it's used in other places for APIs that are already out in the wild. This only changes API that currently has WIP Core PR proposed. To make it consistent, I've changed both to `APINotFound` with debug log, which may be useful for development/troubleshooting, so PTAL.

agners: We changed quite some error handling in APIs lately, it worked out well so far 😅 .

But let's not hold back this fix, we can do the rework in a follow up. The tests are also written such that we can replace the exception here with a `BoardInvalidError` inheriting from `APINotFound` 👍 .

agners (approved):

PR #6927, opened 2026-06-08
Repair stale installed-app locations after addons->apps rename

### LINKED ISSUE #6917: Locally-built add-ons fail with "dockerfile is missing" after addons→apps path migration (#6865) — stored location not migrated
## Summary

The `addons/{core,data,local,git}` → `apps/{core,data,local,git}` directory/file migration introduced in #6865 renamed the on-disk directories and `addons.json` → `apps.json`, **but did not rewrite the absolute `location` path that is persisted per installed app inside that file.** As a result, locally-built add-ons keep pointing at the now-nonexistent `…/addons/…` source directory and fail to build with a misleading **"dockerfile is missing"** error on install/update/rebuild.

This affects anyone who had a locally-built (no `image:` in config) add-on installed before upgrading to a Supervisor that includes #6865 — including everyone who has **already** migrated.

## Symptom

```
INFO  (MainThread) [supervisor.docker.app] Updating image 0f1cc410/aarch64-addon-puppet:2.4.3 to 0f1cc410/aarch64-addon-puppet:2.5.0
ERROR (MainThread) [supervisor.apps.build] Cannot build app '0f1cc410_puppet' because dockerfile is missing. A repair using 'ha supervisor repair' will fix this if the cause is data corruption. Otherwise please report this to the app developer.
```

The add-on's repository is perfectly fine — the `Dockerfile` exists. The error is purely a stale stored path inside Supervisor.

## Root cause

1. #6865 (`355396aea`) migrates directories and the config file:
   - `_migrate_legacy_paths()` in `bootstrap.py` renames `addons/{core,data,local,git}` → `apps/{…}`.
   - `_migrate_addons_json()` in `apps/manager.py` renames `addons.json` → `apps.json`.
2. The installed app's source directory is persisted as an **absolute string** in the system data:
   - Set once by the store loader: `app[ATTR_LOCATION] = str(app_config.parent)` — `supervisor/store/data.py:212`
   - Stored as required in the system schema: `vol.Required(ATTR_LOCATION): str` — `supervisor/apps/validate.py:551`
   - Read back verbatim: `App.path_location` → `Path(self.data[ATTR_LOCATION])` — `supervisor/apps/model.py:611`, where `self.data` is the persisted snapshot `sys_apps.data.system[slug]` (`supervisor/apps/app.py:358–360`), **not** the store object.
3. The migration **never updates these stored strings**, so an installed add-on keeps `location = /data/addons/git/<hash>/<slug>` while the files now live at `/data/apps/git/<hash>/<slug>`.
4. On build, `AppBuild.is_valid()` checks `path_location.is_dir()` and `get_dockerfile().is_file()` against the stale path (`supervisor/apps/build.py:160–173`); the directory no longer exists → `AppBuildDockerfileMissingError` → *"dockerfile is missing"*.

Nothing self-heals it: store reloads only refresh `AppStore` objects, and the only writer of an installed app's `location` is `AppsData.update()`/`install()`/`restore()` — and `update()` runs **after** the failing build during `App.update()` (`supervisor/apps/app.py:1013` build, `:1029` data update), so an update can never get far enough to repair its own path.

## Reproduction

1. On a Supervisor predating #6865, install a locally-built add-on from a git store (e.g. `balloob/home-assistant-addons` → Puppet). `addons.json` records `location: /data/addons/git/<hash>/puppet`.
2. Upgrade to a Supervisor with #6865. Directories + `addons.json` are migrated; the stored `location` still says `/data/addons/git/…`.
3. Trigger a build (update the add-on, or rebuild). → "dockerfile is missing".

## Impact

- All installed locally-built add-ons (git stores and the local store) on any instance upgraded across #6865.
- **Already-migrated instances are affected too** — and a fix gated on "did we just rename the dir this boot" won't help them, because the rename already happened. The repair must run independently of the directory migration.

## Design question for the team

The minimal fix below repairs the stored strings on load. Before settling on an approach, it's worth deciding whether we'd rather:

- **(a)** Migrate the stored `location` strings in place (what the proposed branch does — smallest change, fixes already-migrated users), or
- **(b)** Stop persisting an absolute `location` for installed apps entirely and derive it from the store/repository + slug at read time (removes this whole class of bug, larger change), or
- **(c)** Reconcile an installed app's `location` from its `AppStore` on every store reload.

Given (a) needs to ship regardless to unbreak users who already upgraded, I've prepared it as the immediate fix and left (b)/(c) for discussion.

## Proposed minimal fix

Branch (no PR opened — for review/discussion): **`fix/migrate-stale-app-locations`**
https://github.com/home-assistant/supervisor/tree/fix/migrate-stale-app-locations
Commit: fac2d4046ef513b6ee05ddd1b9fc8c8e05348ed7

It adds a `_migrate_app_locations()` step in `AppManager.load_config()` that rewrites any system `location` starting with `<supervisor>/addons/` to `<supervisor>/apps/` after `read_data()`, persisting only if something changed. It runs unconditionally (not gated on the directory rename), so it also repairs instances that migrated on an earlier boot, and it's idempotent.

```python
def _migrate_app_locations(system: dict[str, Any], supervisor_path: Path) -> bool:
    """Rewrite stored app locations from legacy addons paths to apps paths."""
    legacy_prefix = f"{supervisor_path / 'addons'}/"
    new_prefix = f"{supervisor_path / 'apps'}/"
    changed = False
    for app in system.values():
        location = app.get(ATTR_LOCATION)
        if location and location.startswith(legacy_prefix):
            app[ATTR_LOCATION] = f"{new_prefix}{location[len(legacy_prefix):]}"
            changed = True
    return changed
```

Covered by a unit test in `tests/apps/test_manager.py` (rewrite + idempotency + leaves already-migrated paths untouched).

## Manual workaround (until a fix ships)

Edit `/data/apps.json` and fix the prefix for affected add-ons, then restart Supervisor:

```
system → "<repo>_<slug>" → "location":
  "/data/addons/git/<hash>/<slug>"   ❌
  "/data/apps/git/<hash>/<slug>"     ✅
```

(Local-store add-ons have `/data/addons/local/<slug>` → `/data/apps/local/<slug>`.)

agners:
## Proposed change

The `addons/{core,data,local,git}` → `apps/{...}` directory migration renamed the directories and `addons.json` → `apps.json`, but left the absolute source `location` persisted per installed app pointing at the old `addons` path. `App.path_location` reads that stored string directly, so for locally-built add-ons the build validity check resolves a now-nonexistent directory and the build fails with a misleading "dockerfile is missing" error on install, update, or rebuild.

This rewrites stale `location` values from the legacy `addons` prefix to the `apps` prefix when loading app config. The rewrite runs on every load and is idempotent, so it also repairs instances that migrated their directories on an earlier boot (before this fix existed), where no migration event is left to hook into. The rename and the location rewrite are each individually atomic (`Path.rename` and the atomicwrites-backed `save_data`), so they compose without ever persisting a partially migrated file.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR fixes or closes issue: fixes #6917
- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/apps/manager.py
+++ b/supervisor/apps/manager.py
@@ -4,12 +4,20 @@ import asyncio
 from collections.abc import Awaitable
 from contextlib import suppress
 import logging
-from typing import Self, Union
+from pathlib import Path
+from typing import Any, Self, Union
 
 from attr import evolve
 from securetar import SecureTarFile
 
-from ..const import FILE_HASSIO_ADDONS, FILE_HASSIO_APPS, AppBoot, AppStartup, AppState
+from ..const import (
+    ATTR_LOCATION,
+    FILE_HASSIO_ADDONS,
+    FILE_HASSIO_APPS,
+    AppBoot,
+    AppStartup,
+    AppState,
+)
 from ..coresys import CoreSys, CoreSysAttributes
 from ..exceptions import (
     AppAlreadyInstalledError,
@@ -47,6 +55,24 @@ def _migrate_addons_json() -> None:
             "Migrating %s to %s", FILE_HASSIO_ADDONS.name, FILE_HASSIO_APPS.name
         )
         FILE_HASSIO_ADDONS.rename(FILE_HASSIO_APPS)
+
+
+def _migrate_app_locations(system: dict[str, Any], supervisor_path: Path) -> bool:
+    """Rewrite stored app locations from legacy addons paths to apps paths.
+
+    The directory migration renamed addons/{core,data,local,git} to apps/{...}
+    but the absolute location persisted per installed app was left pointing at
+    the old addons path, which breaks local builds. Returns True if changed.
+    """
+    legacy_prefix = f"{supervisor_path / 'addons'}/"
+    new_prefix = f"{supervisor_path / 'apps'}/"
+    changed = False
+    for app in system.values():
+        location = app.get(ATTR_LOCATION)
+        if location and location.startswith(legacy_prefix):
+            app[ATTR_LOCATION] = f"{new_prefix}{location[len(legacy_prefix) :]}"
+            changed = True
+    return changed
 
 
 class AppManager(CoreSysAttributes):
@@ -98,6 +124,12 @@ class AppManager(CoreSysAttributes):
         """Load config in executor."""
         await self.sys_run_in_executor(_migrate_addons_json)
         await self.data.read_data()
+
+        if _migrate_app_locations(self.data.system, self.sys_config.path_supervisor):
+            _LOGGER.info(
+                "Repairing stale app source locations in %s", FILE_HASSIO_APPS.name
+            )
+            await self.data.save_data()
         return self
 
     async def load(self) -> None:
--- a/tests/apps/test_manager.py
+++ b/tests/apps/test_manager.py
@@ -11,9 +11,17 @@ from awesomeversion import AwesomeVersion
 import pytest
 
 from supervisor.apps.app import App
+from supervisor.apps.manager import _migrate_app_locations
 from supervisor.arch import CpuArchManager
 from supervisor.config import CoreConfig
-from supervisor.const import ATTR_INGRESS, AppBoot, AppStartup, AppState, BusEvent
+from supervisor.const import (
+    ATTR_INGRESS,
+    ATTR_LOCATION,
+    AppBoot,
+    AppStartup,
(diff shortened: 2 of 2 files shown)


mdegat01 (approved):
🙈 

Approved though there are ci failures

agners:
Closing in favor of #6929.

PR #6928, opened 2026-06-08
Fix CI race by waiting for Home Assistant Core running state

mdegat01:
## Proposed change

Update the `Run the Supervisor` workflow to wait for a **positive Home Assistant Core readiness signal** before backup/app actions.

The previous wait step inferred readiness by checking that no active `home_assistant_core_install` job was present. That can race when polling occurs before the install job appears, causing the workflow to continue while Core API/WebSocket is still unavailable.

This patch changes the wait logic to poll `ha core info` and require:

- `result == "ok"`
- `data.state == "running"`

It keeps install-job polling only for progress messaging.

Example failing run/job showing the issue:

- https://github.com/home-assistant/supervisor/actions/runs/27053089160?pr=6916
- https://github.com/home-assistant/supervisor/actions/runs/27053089160/job/79852289943?pr=6916

## Type of change

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/.github/workflows/builder.yml
+++ b/.github/workflows/builder.yml
@@ -333,32 +333,51 @@ jobs:
           exit 1
 
       # Wait for Core to be fully ready before backup/app actions.
-      # Relying only on install-job presence is racy because the job may not yet be
-      # visible when we first poll. Use system state from `ha info` as a positive signal.
+      # `home_assistant_core_install` includes both install + start of Core.
+      # We wait for that job to finish, and use a container/API fallback in case
+      # the job history rotates out quickly.
       - name: Wait for Core to be started
         run: |
-          echo "Waiting for Home Assistant Core to report running state..."
+          echo "Waiting for Home Assistant Core install/start job to finish..."
           timeout=300
           elapsed=0
+          install_job_seen=0
 
           while [ $elapsed -lt $timeout ]; do
-            system_info="$(docker exec hassio_cli ha info --no-progress --raw-json 2>/dev/null || true)"
-            core_result="$(echo "$system_info" | jq -r '.result // empty' 2>/dev/null)"
-            core_state="$(echo "$system_info" | jq -r '.data.state // empty' 2>/dev/null)"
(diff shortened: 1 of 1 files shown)


.github/workflows/builder.yml
```diff
@@ -332,32 +332,41 @@ jobs:
...
           timeout=300
           elapsed=0
 
           while [ $elapsed -lt $timeout ]; do
-            jobs=$(docker exec hassio_cli ha jobs info --no-progress --raw-json | jq -r '.data.jobs[] | select(.name == "home_assistant_core_install" and .done == false) | .name' 2>/dev/null)
-            if [ -z "$jobs" ]; then
-              echo "Home Assistant Core install/start complete (took ${elapsed}s)"
+            system_info="$(docker exec hassio_cli ha info --no-progress --raw-json 2>/dev/null || true)"
+            core_result="$(echo "$system_info" | jq -r '.result // empty' 2>/dev/null)"
+            core_state="$(echo "$system_info" | jq -r '.data.state // empty' 2>/dev/null)"
```
agners: That is Supervisor state no?

mdegat01: yea that's why I put this back in draft. This doesn't make sense right now. There really did seem to be a race condition in the workflow linked I think since a rebase and rerun fixed it with no other changes. But I don't think this solves it yet. I'll have to come back to this.

mdegat01:
Honestly I think I might just nix this. The more I look at this code the more I think there's not anything wrong here, we're correctly tracking this information. At least given what we have, the very strange part is that we are using the `home_assistant_core_install` job at all. I feel like `ha core info` should return some kind of basic `state` for Home Assistant core but it apparently doesn't right now. I feel like we should add that and then come back and adjust this to use it.

But the `home_assistant_core_install` job does complete only when core is both installed and started so I'm going to close this for now since I don't think there's an issue here. I'm not sure what happened in that run, if it happens again I'll dig more into it.

PR #6932, opened 2026-06-10
Don't offer a rebuild repair for a detached app

agners:
## Proposed change

Fixes a corner case where the resolution autofix loop crashes with an unhandled exception while trying to repair a detached, locally-built app.

The scenario: a locally-built app whose container image went missing for some external reason *and* which is also detached from its store (e.g. its local source folder was removed). There is no normal Supervisor command that produces this combination — `rebuild` removes the image before rebuilding, but it is rejected for detached apps, so it cannot leave a detached app without an image. The image loss here came from outside Supervisor's normal flow.

Such an app cannot be rebuilt: there is no source. But `App.load` still surfaced a `MISSING_IMAGE` repair with an `EXECUTE_REPAIR` suggestion for it, and the autofix loop then tried to rebuild it. That dead-ends in `App.path_location`, which raises for a detached app, crashing the autofix with an unhandled exception:

```
INFO (MainThread) [supervisor.resolution.fixup] Starting system autofix at state running
INFO (MainThread) [supervisor.resolution.fixups.app_execute_repair] Installing image for app local_deconz
ERROR (MainThread) [supervisor.jobs] Unhandled exception: App local_deconz has no source location: not available in the store (detached, or store accessed before it was loaded)
  ...
  File "supervisor/apps/build.py", line 89, in _read_build_config
    app.path_location, "build", FILE_SUFFIX_CONFIGURATION
  File "supervisor/apps/app.py", line 712, in path_location
    raise RuntimeError(...)
```

This PR:

- Doesn't create the rebuild repair in `App.load` when the app is detached. The detached-app check already surfaces a `DETACHED_ADDON_REMOVED` issue with an `EXECUTE_REMOVE` suggestion for these apps (the built-in `local` repository is always loaded), so the user is offered removal instead of a repair that can never succeed.
- Guards the repair fixup itself against a detached build app. With the `App.load` change this is only reachable via a race (the repair is created while the app is attached, then the app detaches before/between autofix runs); it now skips gracefully rather than throwing, mirroring the `is_detached` guards already in `AppManager.update`/`rebuild`.

Image-based detached apps are unaffected — they can still be re-pulled.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/apps/app.py
+++ b/supervisor/apps/app.py
@@ -306,7 +306,16 @@ class App(AppModel):
             await self.instance.check_image(self.version, default_image, self.arch)
         except DockerNotFound:
             _LOGGER.info("No %s app Docker image %s found", self.slug, self.image)
-            if self.need_build:
+            if self.need_build and self.is_detached:
+                # Image is gone and there is no store source to rebuild from.
+                # A rebuild repair would be futile; the detached-app check
+                # surfaces a remove suggestion instead.
+                _LOGGER.warning(
+                    "App %s is missing its image and is detached from its store; "
+                    "it cannot be rebuilt and should be removed",
+                    self.slug,
+                )
+            elif self.need_build:
                 # Don't run a local build during setup. Surface a repair so
                 # the resolution autofix loop can handle it off the critical
                 # path.
--- a/supervisor/resolution/fixups/app_execute_repair.py
+++ b/supervisor/resolution/fixups/app_execute_repair.py
@@ -49,6 +49,16 @@ class FixupAppExecuteRepair(FixupBase):
             )
             return
 
+        if app.is_detached and app.need_build:
+            # No store source to rebuild from; the repair cannot succeed. The
+            # detached-app check surfaces a remove suggestion instead.
+            _LOGGER.warning(
+                "Cannot repair app %s as it is detached from its store and "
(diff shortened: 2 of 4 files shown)


supervisor/apps/app.py
```diff
@@ -306,7 +306,16 @@ async def load(self) -> None:
...
-            if self.need_build:
+            if self.need_build and self.is_detached:
+                # Image is gone and there is no store source to rebuild from.
+                # A rebuild repair would be futile; the detached-app check
+                # surfaces a remove suggestion instead.
+                _LOGGER.warning(
+                    "App %s is missing its image and is detached from its store; "
+                    "it cannot be rebuilt and should be removed",
+                    self.slug,
+                )
```
mdegat01, line 317: Should this be its own repair suggesting removal or is it too niche for that?

mdegat01 (approved):
Looks good, seems like the error should be its own repair ideally but perhaps too niche for that effort.

supervisor/apps/app.py
agners: Yeah I think this is too niche. Besides my dev environment (where I clearly messed with Docker directly) we had a single user run into it on stable channel, but that instance had I/O errors anyways.

PR #6937, opened 2026-06-12
Add isolated physical network access for apps

agners:
## Proposed change

Implements the first phase of [isolated physical network access](https://github.com/home-assistant/architecture/discussions/1034): users can assign apps that use host networking an isolated endpoint on a physical network interface instead. The app then runs in bridge mode with its own IP and MAC address on the selected interface (Docker macvlan), providing full multicast/mDNS support and interface selection on multi-homed hosts while restoring the isolation that host networking gives up.

Key design points:

- **Dual attachment**: the container stays attached to the internal `hassio` network, so ingress, API proxy, DNS plug-in and inter-app communication work unchanged. The macvlan endpoint provides the physical network presence and wins the default route via `GwPriority`.
- **Shared networks per interface**: Docker allows only one macvlan network per parent interface, so networks (`hassio-macvlan-<iface>`) are a shared resource derived from the host interface configuration (subnet/gateway via NetworkManager), created on demand, recreated on drift and garbage collected when no app references them.
- **Static IPv4, stable MAC**: the user assigns a static IP (validated against subnet, host/gateway addresses and other apps); the MAC is pinned with the IP-derived scheme older engines used, since Docker 28 assigns random MACs per endpoint creation. Both are exposed in the app info API so users can set up router/firewall rules before the first start.
- **IPv6 via SLAAC**: external networks have IPv6 enabled (engine-allocated ULA satisfies the driver, real addressing comes from router advertisements). Endpoints set `accept_ra=2` and `accept_ra_rt_info_max_plen=64` so RFC 4191 routes announced by Thread border routers reach the container.
- **Graceful degradation**: if the endpoint cannot be set up at start (interface gone, subnet changed), the app falls back to host networking and a `network_isolation_failed` issue is raised; a scheduled resolution check detects drift. The fallback is blocked when the Docker gateway is unprotected, since that protection does not apply to isolated apps (their traffic bypasses the host network namespace, which also lifts the security rating penalty for host networking).

The feature requires Docker >= 28 (for `GwPriority` and per-endpoint sysctls) and is gated accordingly; existing installations and apps are unaffected unless a user opts in per app.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR is related to issue: https://github.com/home-assistant/architecture/discussions/1034
- Link to cli pull request:

## Checklist


agners:
This is AI assisted based on the design and discussion in https://github.com/home-assistant/architecture/discussions/1034. Currently it allows Apps which use host network to isolate.

I've tested this with the ESPHome and Matter App. The Matter App required a small tweak in the startup script to not fetch the primary interface from Supervisor (rather use the `ip route` based fallback). With this tweak I was able to communicate to Thread devices behind a BR running on a different host, which means the Matter app's isolated network processed IPv6 ND RIO correctly :tada: .

I think I'd prefer to make IPv4 completely optional. Currently the MAC address is derrived from the IPv4 address, which seems to be a design which Docker used previously. However, I think I'd rather prefer it to just assign a random MAC.

I am also not sure if we maybe should only support ipvlan. While macvlan is strictly more isolated, it seems not to work with WiFi really. If ipvlan gives all the isolation needs we have, maybe we can simply default to it instead :thinking:

PR #6391, opened 2025-12-03
Disable timeout for Docker image pull operations

agners:
## Proposed change

Disable timeout for Docker image pull operations to match docker-py behavior.

The aiodocker migration introduced a regression where image pulls could timeout during slow downloads. The session-level timeout (900s total) was being applied to pull operations, but docker-py explicitly sets `timeout=None` for pulls, allowing them to run indefinitely.

When aiodocker receives `timeout=None`, it converts it to `ClientTimeout(total=None)`, which aiohttp treats as "no timeout". This fixes `TimeoutError` exceptions that could occur during installation on systems with slow network connections or when pulling large images.

## Type of change

- [ ] Dependency upgrade
- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (which adds functionality to the supervisor)
- [ ] Breaking change (fix/feature causing existing functionality to break)
- [ ] Code quality improvements to existing code or addition of tests

## Additional information

- This PR fixes or closes issue: fixes #6390
- This PR is related to issue:
- Link to documentation pull request:
- Link to cli pull request:
- Link to client library pull request:

## Checklist

- [x] The code change is tested and works locally.
- [x] Local tests pass. **Your PR cannot be merged unless tests pass**
- [x] There is no commented out code in this PR.
- [x] I have followed the [development checklist][dev-checklist]
- [x] The code has been formatted using Ruff (`ruff format supervisor tests`)
- [x] Tests have been added to verify that the new code works.

If API endpoints or add-on configuration are added/changed:

- [ ] Documentation added/updated for [developers.home-assistant.io][docs-repository]
- [ ] CLI updated (if necessary)
- [ ] Client library updated (if necessary)

[dev-checklist]: https://developers.home-assistant.io/docs/en/development_checklist.html
[docs-repository]: https://github.com/home-assistant/developers.home-assistant
[cli-repository]: https://github.com/home-assistant/cli
[client-library-repository]: https://github.com/home-assistant-libs/python-supervisor-client/

PR #6955, opened 2026-06-17
Handle Docker status-check failures in Home Assistant API paths

mdegat01:
## Proposed change

This PR fixes unhandled Docker status-check failures that can bubble out of Home Assistant API communication paths.

Changes included:
- Add a dedicated `DockerTimeoutError` that inherits from `DockerError`.
- Raise `DockerTimeoutError` from `_get_container` when `TimeoutError` occurs while querying Docker container metadata.
- Standardize Core-running preflight checks in `HomeAssistantAPI` via a shared helper.
- Use that helper in both `make_request` and `connect_websocket`.
- Wrap `DockerError` from the running check as `HomeAssistantAPIError` to preserve existing fail-fast behavior and API contract.
- Add tests for timeout wrapping and DockerError wrapping paths.

The change to `_get_container` also addresses the separately linked sentry issue below about `App.start`. It will no longer bubble `asyncio.TimeoutError` as an unhandled exception and instead report a `DockerTimeoutError` that is handled the same as other errors communicating with docker in that codepath.

## Type of change

## Additional information

    - `make_request` codepath: https://home-assistant-io.sentry.io/issues/7241686140/events/e71e265dfe75468a9aaa7e4d852e30ba/, https://home-assistant-io.sentry.io/issues/7449775230/events/c6b273edaac442db8d6ba54d78d80bad/
    - `App.start` -> `is_running` codepath: https://home-assistant-io.sentry.io/issues/7241686140/events/8ecf59a6809c4b8c989872b265128102/ 
- Link to cli pull request:

Sentry impact context:
- ~165k reports in the latest release between the two events
- ~350k reports total in the past week
- `App.start` issue has ~1.5k reports in the past week, most from latest release

## Checklist

### DIFF
--- a/supervisor/docker/interface.py
+++ b/supervisor/docker/interface.py
@@ -35,6 +35,7 @@ from ..exceptions import (
     DockerNotFound,
     DockerRegistryAuthError,
     DockerRegistryRateLimitExceeded,
+    DockerTimeoutError,
     GithubContainerRegistryRateLimitExceeded,
 )
 from ..jobs.const import JOB_GROUP_DOCKER_INTERFACE, JobConcurrency
@@ -410,6 +411,10 @@ class DockerInterface(JobGroup, ABC):
         try:
             container = await self.sys_docker.containers.get(self.name)
             return await container.show()
+        except TimeoutError as err:
+            raise DockerTimeoutError(
+                f"Timeout occurred while getting container information for {self.name}: {err!s}"
+            ) from err
         except aiodocker.DockerError as err:
             if err.status == HTTPStatus.NOT_FOUND:
                 return None
--- a/supervisor/docker/interface.py
+++ b/supervisor/docker/interface.py
@@ -413,7 +413,7 @@ class DockerInterface(JobGroup, ABC):
             return await container.show()
         except TimeoutError as err:
             raise DockerTimeoutError(
-                f"Timeout occurred while getting container information for {self.name}: {err!s}"
+                f"Timeout occurred while getting container information for {self.name}"
             ) from err
         except aiodocker.DockerError as err:
             if err.status == HTTPStatus.NOT_FOUND:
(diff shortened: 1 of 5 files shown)


mdegat01:
I'm not sure if there's a separate sentry issue for `connect_websocket` or if there's just far fewer calls buried within the list of events in the same issue. The count of events seems to be mostly about `make_request` if so but they both suffer from the same issue.

We probably also need to follow this up with a similar wrapping for all other aiodocker interactions now. I thought it only raised things which inherit from `aiodocker.DockerError` but it appears `asyncio.TimeoutError` can pop out as well. That's unhandled pretty much everywhere right now 🙈

mdegat01:
Follow-up to comment above, if you search sentry for issues with the name `TimeoutError` you can see that yes we bubble this out of essentially every interaction with `aiodocker` at times. All are much much smaller then the one fixed by this PR though, this is the big one. I'll do a follow-up after this to address the other once the exception type is in to handle the others similarly

supervisor/docker/interface.py
```diff
@@ -410,6 +411,10 @@ async def _get_container(self) -> dict[str, Any] | None:
         try:
             container = await self.sys_docker.containers.get(self.name)
             return await container.show()
+        except TimeoutError as err:
+            raise DockerTimeoutError(
+                f"Timeout occurred while getting container information for {self.name}: {err!s}"
```
agners: Afaik `asycnio.TimeoutError` has an empty error string, so this end up to be:

```
Timeout occurred while getting container information for homeassistant:
```

Maybe better to just omit the error here 🤔

agners (approved):
Nice find. The aiodocker timeout handling had some fallout previously, e.g. #6391. Also timeout changed in the upstream 0.25.0 release (see #6448).

Not sure why the library does not wrap `asyncio.TimeoutError`, it seems to consider timeouts not something which need to be "vendorized" 🤔 . Anyways, I think for us it makes sense to separate timeouts related to Docker API from other timeouts, and generally handle them like other Docker releated exceptions. So I agree with the approach 👍 . One small nit, but otherwise LGTM.

PR #6963, opened 2026-06-19
Read store repository app folders in parallel

frenck:
## Proposed change

`StoreData.update()` reads each repository's app folder in sequence: the core repository, the local repository, then every custom git repository one after another. Each read does a recursive `**/config.*` glob and parses every add-on config for that repository, so the total time grows with the number of repositories. This runs on every store reload (startup, adding/removing a repository, and the periodic reload), and is more expensive with large or network-backed repositories.

The repositories are independent and each produces apps under its own repository-prefixed slug (`{repository}_{slug}`), so this reads all app folders concurrently with `asyncio.gather` instead of awaiting them one by one. The results are merged in the original order (core, local, then git repositories), so behavior is unchanged; only the wall-clock time drops toward the slowest single repository instead of the sum of all of them.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/store/data.py
+++ b/supervisor/store/data.py
@@ -1,5 +1,6 @@
 """Init file for Supervisor app data."""
 
+import asyncio
 from dataclasses import dataclass
 import errno
 import logging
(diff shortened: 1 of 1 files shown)


agners (commented):
> so behavior is unchanged; only the wall-clock time drops toward the slowest single repository instead of the sum of all of them.

Is that proven? 🤔 

Typically such recursive search in directory trees leads to lots of random I/O. I am quite sure that at least on spinning disks doing such operation in parallel actually kills performance. Probably modern SSD with multiple I/O queues can actually parallelize such operations, but I'd guess most cheaper SSD and eMMC/SD cards run with a single I/O queue, where running these in parallel most likely leads to no or negligible performance differences.

Unless measurements really show that this improves reload time on typical hardware in real world, I'd rather not parallelize workloads with lots of random I/O.

PR #6970, opened 2026-06-23
Refine Docker timeout handling and add regression tests

mdegat01:
## Proposed change

Align Docker timeout handling with aiodocker timeout semantics and improve timeout error mapping and regression coverage.

This PR:
- maps actionable timeout paths to `DockerTimeoutError`
- removes timeout handling around `run_command`/`container_logs` paths that use aiodocker long-running infinite-timeout behavior
- extracts container network setup in Docker manager `_run` into a helper for clarity
- updates container resolution evaluation so transient timeouts no longer create `CORRUPT_DOCKER`
- adds focused timeout regression tests across manager/interface/supervisor/resolution

### Timeout semantics table

This table denotes the `aiodocker` methods identified that we use which have no timeout by default, `TimeoutError` handling is not required for these. All others require `TimeoutError` handling.

| Call site | aiodocker method | Timeout semantics | TimeoutError handling in Supervisor | Notes |
| --- | --- | --- | --- | --- |
| `DockerAPI.run_command()` | `container.wait()` | Long-running resolver default (effectively infinite) | **No** | Removed timeout wrapping for this path. |
| `DockerAPI.run_command()` | `container.log(..., follow=False)` | Long-running resolver default (effectively infinite) | **No** | Removed timeout wrapping for this path. |
| `DockerAPI.container_logs()` | `container.log(...)` | Long-running resolver default (effectively infinite) | **No** | Removed timeout wrapping for this path. |
| `DockerAPI.run_command()` image pull fallback | `images.pull(..., timeout=None)` | Explicit infinite timeout | **No** | Timeout handling intentionally omitted per policy. |
| `DockerAPI.pull_image()` streaming pull | `images.pull(..., stream=True, timeout=None)` | Explicit infinite timeout | **No** | Timeout handling intentionally omitted per policy. |

Additional rule applied in this PR: awaited Docker calls that can actually time out in finite operations are mapped to `DockerTimeoutError`.

### Sentry

This change is a follow-up to #6955 to address the many other places `TimeoutError` can be raised that ended up as issues in Sentry. There is quite a lot in this case so rather then list them individually I'll just list the search I used. If you search for [`title is TimeoutError`](https://home-assistant-io.sentry.io/issues/?project=5370612&query=title%3ATimeoutError&referrer=issue-list&statsPeriod=14d) in the Supervisor project almost all of those have source `aiodocker` in the frames. Since the `TimeoutError` has no message the exact search makes them quite easy to find.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/docker/manager.py
+++ b/supervisor/docker/manager.py
@@ -697,7 +697,8 @@ class DockerAPI(CoreSysAttributes):
                 ) from err
             _LOGGER.info("Pulling image %s:%s", image, tag)
             try:
-                await self.images.pull(image, tag=tag, timeout=None)
+                # Timeout is disabled for pull operations by default, matching docker-py behavior.
+                await self.images.pull(image, tag=tag)
             except aiodocker.DockerError as pull_err:
                 raise DockerError(
                     f"Can't pull image {image}:{tag}: {pull_err}", _LOGGER.error
@@ -749,10 +750,9 @@ class DockerAPI(CoreSysAttributes):
         raises only if the get fails afterwards. Additionally it fires progress reports for the pull
         on the bus so listeners can use that to update status for users.
         """
-        # Use timeout=None to disable timeout for pull operations, matching docker-py behavior.
-        # aiodocker converts None to ClientTimeout(total=None) which disables the timeout.
+        # Timeout is disabled for pull operations by default, matching docker-py behavior.
         async for e in self.images.pull(
-            repository, tag=tag, platform=platform, auth=auth, stream=True, timeout=None
+            repository, tag=tag, platform=platform, auth=auth, stream=True
         ):
             entry = PullLogEntry.from_pull_log_dict(job_id, e)
             if entry.error:
--- a/supervisor/docker/manager.py
+++ b/supervisor/docker/manager.py
@@ -43,6 +43,7 @@ from ..exceptions import (
     DockerNoSpaceOnDevice,
     DockerNotFound,
     DockerRegistryRateLimitExceeded,
+    DockerTimeoutError,
 )
 from ..utils.common import FileConfiguration
 from ..validate import SCHEMA_DOCKER_CONFIG
@@ -288,7 +289,12 @@ class DockerAPI(CoreSysAttributes):
 
     async def post_init(self) -> Self:
         """Post init actions that must be done in event loop."""
-        self._info = await DockerInfo.new(await self.docker.system.info())
+        try:
+            self._info = await DockerInfo.new(await self.docker.system.info())
+        except TimeoutError as err:
+            raise DockerTimeoutError(
+                "Timeout while getting Docker information", _LOGGER.error
+            ) from err
         await self.config.read_data()
         self._network = await DockerNetwork(self.docker).post_init(
             self.config.enable_ipv6, self.config.mtu
@@ -457,6 +463,68 @@ class DockerAPI(CoreSysAttributes):
 
         return config
 
+    async def _setup_container_network(
+        self,
+        container: DockerContainer,
+        *,
+        name: str | None,
+        hostname: str | None,
+        network_mode: str | None,
+        networking_config: dict[str, Any] | None,
+        ipv4: IPv4Address | None,
+    ) -> None:
+        """Set up networking for container before start."""
+        if networking_config or network_mode not in ("host", None):
+            return
+
+        # Attach network
+        if not network_mode:
+            alias = [hostname] if hostname else None
+            try:
+                await self.network.attach_container(
+                    container.id, name, alias=alias, ipv4=ipv4
+                )
+            except DockerError:
+                _LOGGER.warning(
+                    "Can't attach %s to hassio-network!", name or container.id
+                )
+            else:
+                with suppress(DockerError):
+                    await self.network.detach_default_bridge(container.id, name)
+            return
+
+        try:
+            host_network = await self.docker.networks.get(DOCKER_NETWORK_HOST)
+            host_network_meta = await host_network.show()
+        except TimeoutError as err:
+            raise DockerTimeoutError(
+                "Timeout getting host network information from Docker",
+                _LOGGER.error,
+            ) from err
+        except aiodocker.DockerError as err:
+            raise DockerError(
+                f"Can't get host network information from Docker: {err!s}",
+                _LOGGER.error,
+            ) from err
+
+        # Check if container is register on host
+        # https://github.com/moby/moby/issues/23302
+        if name and name in (
+            val.get("Name") for val in host_network_meta.get("Containers", {}).values()
+        ):
+            try:
+                await host_network.disconnect({"Container": name, "Force": True})
+            except TimeoutError as err:
+                raise DockerTimeoutError(
+                    f"Timeout disconnecting container {name} from host network",
+                    _LOGGER.error,
+                ) from err
+            except aiodocker.DockerError as err:
+                if err.status != HTTPStatus.NOT_FOUND:
+                    raise DockerError(
+                        f"Can't disconnect container {name} from host network: {err!s}"
+                    ) from err
+
     async def _run(
         self,
         image: str,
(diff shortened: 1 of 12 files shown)


.github/skills/mypy-cache-fix/SKILL.md
agners, line 1: This should be a separate PR. Maybe by accident?

.github/copilot-instructions.md
```diff
@@ -3,6 +3,11 @@
 This repository contains the Home Assistant Supervisor, a Python 3 based container
 orchestration and management system for Home Assistant.
 
+## Pull Request Requirements
+
+When opening a pull request in this repository, always use the description format
+from `.github/PULL_REQUEST_TEMPLATE.md`.
+
```
agners, line 10: This too should be a separate PR.

agners (changes requested):

supervisor/docker/manager.py
```diff
@@ -457,6 +463,68 @@ def _create_container_config(
...
+        if not network_mode:
+            alias = [hostname] if hostname else None
+            try:
+                await self.network.attach_container(
+                    container.id, name, alias=alias, ipv4=ipv4
+                )
+            except DockerError:
+                _LOGGER.warning(
+                    "Can't attach %s to hassio-network!", name or container.id
+                )
```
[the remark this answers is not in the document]
mdegat01: These are exactly the kinds of bubbling unhandled errors this PR is seeking to avoid. We do not want that to bubble out of here, the handling was intended to swallow and log all errors caused from an interaction with docker at this point.

supervisor/docker/manager.py
```diff
@@ -662,12 +714,16 @@ async def run_command(
                 skip_cidfile=True,
                 **kwargs,
             )
+        except DockerError as err:
+            # _run() already logs the root cause when creating DockerError-derived exceptions.
+            # Re-raise here with command context without duplicating error-level logging.
+            raise DockerError(f"Can't execute command: {err}") from err
```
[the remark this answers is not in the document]
mdegat01: As with the other comment, this is intentional. Having `TimeoutError` bubble out of here was unintentional and problematic, this was intended to catch all errors caused by an interaction with Docker.

.github/skills/mypy-cache-fix/SKILL.md
mdegat01: Sure I can separate it. I just kept running into it repeatedly so I told copilot to make it a skill.

.github/copilot-instructions.md
mdegat01: Sure, was just another thing I kept repeatedly telling copilot to do 😆 . I'm fine with separating it.

.github/skills/mypy-cache-fix/SKILL.md
mdegat01: #6971

supervisor/docker/manager.py
```diff
@@ -288,7 +289,12 @@ def __init__(self, coresys: CoreSys):
 
     async def post_init(self) -> Self:
         """Post init actions that must be done in event loop."""
-        self._info = await DockerInfo.new(await self.docker.system.info())
+        try:
+            self._info = await DockerInfo.new(await self.docker.system.info())
+        except TimeoutError as err:
+            raise DockerTimeoutError(
+                "Timeout while getting Docker information", _LOGGER.error
+            ) from err
```
mdegat01, line 297: So an error raised here results in a catastrophic startup failure and a Supervisor crash. We aren't in the `run_forever` loop yet, we're in super early bootstrap. And there's no generic catch-alls, we'll just bail out.

But the thing is that's always been true. We have no error handling here for any docker errors and we can raise others from `DockerNetwork.post_init`. We also have no retry available for the early actions here so if we squashed and logged them there would be no recovery. So I think this is correct? The catch and re-raise here really just adds some context info to the failure since `TimeoutError` has none.

supervisor/docker/manager.py
```diff
@@ -457,6 +463,68 @@ def _create_container_config(
 
         return config
 
+    async def _setup_container_network(
```
mdegat01, line 466: For reference this exists because `_run` tripped the too complex pylint rule and it seemed like a logical chunk of code to separate.

supervisor/docker/manager.py
```diff
@@ -638,14 +686,18 @@ async def run_command(
...
+            ) from err
         except aiodocker.DockerError as err:
             if err.status != HTTPStatus.NOT_FOUND:
                 raise DockerError(
                     f"Can't inspect image {image}:{tag}: {err}", _LOGGER.error
                 ) from err
             _LOGGER.info("Pulling image %s:%s", image, tag)
             try:
-                await self.images.pull(image, tag=tag)
+                await self.images.pull(image, tag=tag, timeout=None)
```
mdegat01: I added `timeout=None` here to be consistent with this note:
https://github.com/home-assistant/supervisor/blob/32fad046e5c2eef779de8e5698c23dbcc58cde51/supervisor/docker/manager.py#L696-L700

But I'm not sure why either of these are like this. Looking at `aiodocker` code for `pull` it says it uses an infinite timeout by default. The AI pointed out that its logic does this:
```py
        if timeout is SENTINEL:
            # Inherit the parent client's timeout but override total and sock_read
            # because they don't make sense for long-running operations.
            return attrs.evolve(self._timeout, total=None, sock_read=None)
        elif timeout is None:
            # Infinite timeout
            return aiohttp.ClientTimeout()
```
So by providing `None` we ensure an entirely `None` `ClientTimeout` object instead of one with just those two fields set to `None`. Which makes sense except the default timeout we initialize `aiodocker` with only has a value for `total`, so it seems like adding `timeout=None` has no effect in either place to me:
https://github.com/home-assistant/supervisor/blob/32fad046e5c2eef779de8e5698c23dbcc58cde51/supervisor/docker/manager.py#L274-L279

.github/copilot-instructions.md
mdegat01: #6972

supervisor/docker/manager.py
agners, line 734: Technically this can throw Timeout error too.

supervisor/docker/manager.py
agners: Yeah this got introduced with #6391, but aiodocker 0.25.0 (merged with #6448) brought https://github.com/aio-libs/aiodocker/pull/990, which made this obsolete.

I think removing `timeout=None` in both places it is slightly better even, since that makes it use SENTINEL since it inherits the parent client's timeouts for connection establishment and only overrides the timeouts during read (which is the reason #6391 introduced None, since these are long running operations).

supervisor/docker/manager.py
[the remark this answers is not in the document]
agners: Timeout is still retained in the inner exception, so maybe it is fine as is.

But with this we'd retain the type 🤔 

```suggestion
        except DockerError as err:
            # _run() already logs the root cause when creating DockerError-derived exceptions.
            # Re-raise here with command context without duplicating error-level logging.
            raise type(err)(f"Can't execute command: {err}") from err
```

agners (changes requested):
This makes sense and should avoid a couple `asyncio.TimeoutError` reaching Sentry 👍. A couple smaller nits.

supervisor/docker/manager.py
mdegat01: Actually I don't think it can? Here's the code for this method in aiodocker v27.0.0 right now:
https://github.com/aio-libs/aiodocker/blob/c7aa432f847da092d9b7db910124f0ce806944da/aiodocker/containers.py#L407-L441

Specifically this bit in there:
```py
        # Use standard timeout resolution
        timeout_config = self.docker._resolve_long_running_timeout(timeout)
```
I double checked and am very sure that comment contradicts what's actually happening. Because when `_resolve_long_running_timeout` gets a `SENTINEL` object it returns the no timeout config.

Perhaps this is all a bug I'm not sure. But currently this code does not seem to raise `TimeoutError` unless we specifically add a timeout to the call.

mdegat01: Created a PR to fix the comment. If it is actually supposed to have finite timeout by default then I guess we'll learn when its reviewed: https://github.com/aio-libs/aiodocker/pull/1045

agners (approved):
LGTM!

PR #6976, opened 2026-06-24
Add unhealthy reasons for fsck failures and issues for other systemd failures

### LINKED ISSUE #6040: Inform user about failed fsck (create unhealthy status/repair)
### Description

When fsck fails, the user might not learn about it as the error is only printed to the logs and the system might still boot (especially if using data disk). Make sure there is a way of informing user about that so they can take action immediately.

As a further extension of this idea, maybe it might be good to somehow propagate status of all failed Systemd units?

### Additional context

Discovered through https://github.com/home-assistant/operating-system/issues/4182#issuecomment-3112946404

mdegat01:
## Proposed change

New check which marks the system as unhealthy when `fsck` fails on the OS or Data partition to denote there is a corrupted filesystem. Also looks for any other systemd unit failures. For any that aren't related to mount failures or our firewall service (since that handles itself) it creates issues out of them and reports a message to sentry so we can keep track of this.

The issues will only live in Supervisor, I don't plan to make generic systemd unit repairs. These will be sysadmin/developer level issues only visible in the CLI. If we identify other common and problematic repairs we'll add to this check to report new types of issues/unhealthy reasons that we do make into repairs.  It was also just a convenient way to keep track of which failures we'd already reported to sentry so we don't report the same failure repeatedly. 

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR fixes or closes issue: fixes #6040
- Link to documentation pull request: https://github.com/home-assistant/home-assistant.io/pull/46471
- Link to cli pull request:
- Link to client library pull request: https://github.com/home-assistant-libs/python-supervisor-client/pull/362
- Link to core pull request: https://github.com/home-assistant/core/pull/176928

## Checklist

### DIFF
--- a/supervisor/resolution/checks/systemd_unit_failure.py
+++ b/supervisor/resolution/checks/systemd_unit_failure.py
@@ -1,0 +1,157 @@
+"""Check for failed systemd units."""
+
+import asyncio
+from collections.abc import Awaitable
+from datetime import timedelta
+import logging
+
+from ...const import CoreState
+from ...coresys import CoreSys
+from ...host.firewall import FIREWALL_SERVICE
+from ...jobs.const import JobThrottle
+from ...jobs.decorator import Job
+from ...utils.sentry import async_capture_message
+from ..const import ContextType, IssueType, SuggestionType, UnhealthyReason
+from ..data import Issue
+from .base import CheckBase
+
+_LOGGER: logging.Logger = logging.getLogger(__name__)
+
+# Systemd fsck units for OS boot and overlay
+OS_FSCK_UNITS = {
+    "systemd-fsck@dev-disk-by\\x2dlabel-hassos\\x2dboot.service",
+    "systemd-fsck@dev-disk-by\\x2dlabel-hassos\\x2doverlay.service",
+}
+
+# Systemd fsck unit for data filesystem
+DATA_FSCK_UNIT = "systemd-fsck@dev-disk-by\\x2dlabel-hassos\\x2ddata.service"
+
+
+def setup(coresys: CoreSys) -> CheckBase:
+    """Check setup function."""
+    return CheckSystemdUnitFailure(coresys)
+
+
+class CheckSystemdUnitFailure(CheckBase):
+    """Check for failed systemd units and create repair suggestions."""
+
+    def __init__(self, coresys: CoreSys) -> None:
+        """Initialize systemd unit failure check."""
+        super().__init__(coresys)
+        self._failed_units_cache: list[tuple[str, ...]] = []
+
+    @Job(
+        name="resolution_check_systemd_unit_failure_list_units_filtered",
+        throttle=JobThrottle.THROTTLE,
+        throttle_period=timedelta(seconds=60),
+        internal=True,
+    )
+    async def _refresh_failed_units_cache(self) -> None:
+        """Refresh cached failed units at most once per throttle period."""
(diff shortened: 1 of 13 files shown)


supervisor/resolution/checks/systemd_unit_failure.py
```diff
@@ -0,0 +1,157 @@
...
+
+    @property
+    def context(self) -> ContextType:
+        """Return a ContextType enum."""
+        return ContextType.SYSTEM
+
+    @property
+    def states(self) -> list[CoreState]:
+        """Return a list of valid states when this check can run."""
+        return [CoreState.SETUP]
```
mdegat01, line 157: I realize this code contradicts this statement:
> It was also just a convenient way to keep track of which failures we'd already reported to sentry so we don't report the same failure repeatedly.

This check doesn't run actually repeatedly right now. It runs once during `SETUP` and that's it. I took this comment in the task:

> This check should be done before loading mounts (to avoid false positive from failing mount units).

To mean we intended to run it only once at startup so I set it up that way. I couldn't actually run it before mounts since resolution manager loads after mount manager but it was easy enough to filter them out. Plus couldn't they already be there on a Supervisor restart without a reboot?

However I wanted to confirm - is this actually what we want? Is it impossible for a systemd unit to fail during normal operation or can it only fail when started? And do we never create new ones during normal operation?

I kind of like the issues either way just as a way to inform developers and sysadmins of issues even when we haven't made proper UI for them yet via repairs. If it really is run once at supervisor start and never again though then I'll toss this line:
```py
if issue in self.sys_resolution.issues:
  continue
```
Because that'll never happen.

agners: > To mean we intended to run it only once at startup so I set it up that way. I couldn't actually run it before mounts since resolution manager loads after mount manager but it was easy enough to filter them out. Plus couldn't they already be there on a Supervisor restart without a reboot?

Right, so yeah filtering probably makes more sense. 

> However I wanted to confirm - is this actually what we want? Is it impossible for a systemd unit to fail during normal operation or can it only fail when started? And do we never create new ones during normal operation?

If a service process exits, the unit will fail at runtime. Most likely this is caused by either networking issue or resource constraints (out-of-memory).

The task is meant as a first step, the startup failures are most obvious and we can work on those first. Ultimately, a service can also fail multiple times, potentially in quick succession. Ideally we have a similar UX like Core has for duplicated error messages (show the count and when the last event was). But we need to draft first how we would exactly represent this (maybe multiple Supervisor issues, but group on Core/hassio side? 🤔 ).

supervisor/resolution/checks/systemd_unit_failure.py
```diff
@@ -0,0 +1,157 @@
...
+            )
+            if issue in self.sys_resolution.issues:
+                continue
+
+            capture_coroutines.append(
+                async_capture_message(
+                    f"Systemd unit failed: {unit_name}",
+                    level="error",
+                )
+            )
```
agners, line 123: I think I'd gone with `_LOGGER.critical` instead here. Yes it adds log, which at this point maybe we overwhelm users, but then we have lots of warnings in Supervisor logs too (from apps validation) 🤷 . It would avoid another interface.

On the other hand, `_LOGGER.critical` maps to Sentry level "fatal", which is probably wrong. So let's go with what we have here.

agners (approved):
LGTM!

PR #6989, opened 2026-06-30
Reject apps mapping a dynamic ingress port range port

agners:
## Proposed change

When an app uses dynamic ingress port selection (`ingress_port: 0`), `Ingress.get_dynamic_port()` picks a random port from the 62000-65500 range and hands it to the app to listen on for ingress. That port is reached only over the internal Docker network at the app's container IP (`api/ingress.py`); it is never published to the host.

If an app *also* maps a container port from that same range to the host via its `ports` config, the dynamically chosen ingress port could coincide with it. Docker would then publish that container port to the host, making the ingress endpoint reachable directly on the host and bypassing ingress authentication.

Rather than working around this at allocation time, this rejects such configs during validation: an app using dynamic ingress port selection must not map a port from the dynamic ingress port range itself. Validation runs on every config parse/update, so it also covers an app that introduces such a mapping in a later version (which an allocation-time check would miss, since the dynamic port is allocated once and persisted).

The container ports come exclusively from the app's `config.json` (users can only set host-port values for existing keys), so validating the app config fully covers this. The range bounds are extracted into `INGRESS_DYNAMIC_PORT_MIN`/`INGRESS_DYNAMIC_PORT_MAX` constants shared between the validator and the allocator.

This only affects apps that combine `ingress_port: 0` with a mapped port in the 62000-65500 range; none of the popular apps do that. Apps with a static ingress port are unaffected, and may still map a port in that range.

Alternative to #6983.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/apps/validate.py
+++ b/supervisor/apps/validate.py
@@ -97,6 +97,8 @@ from ..const import (
     ATTR_VIDEO,
     ATTR_WATCHDOG,
     ATTR_WEBUI,
+    INGRESS_DYNAMIC_PORT_MAX,
+    INGRESS_DYNAMIC_PORT_MIN,
     MACHINE_DEPRECATED,
     ROLE_ALL,
     ROLE_DEFAULT,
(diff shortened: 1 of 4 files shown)


mdegat01 (approved):
I still don't like that we let you list a fixed ingress port and put that port in ports but I guess since apps are doing that our hands are tied. This one looks good though, hopefully it prevents someone from having to deal with a confusing error in the future 👍

mdegat01:
Should we mention this limitation in app configuration documentation or not worth it?

agners:
> Should we mention this limitation in app configuration documentation or not worth it?

Yeah probably worth updating the developer docs indeed.

> I still don't like that we let you list a fixed ingress port and put that port in ports but I guess since apps are doing that our hands are tied. This one looks good though, hopefully it prevents someone from having to deal with a confusing error in the future 👍

Recommending not to use the same port for ingress and map it is probably almost more important. I try to come up with a PR with better recommendation wrt Ingress/port mapping.

PR #6992, opened 2026-07-01
Add app-based mapping options for app configs

### LINKED ISSUE #6745: Change add-ons to apps in Supervisor configuration

mdegat01:
## Proposed change

Add app-based mapping options for app configuration folders and local app folders. The new `local_apps`, `app_config`, and `all_app_configs` mappings are preferred, while the legacy `addons`, `addon_config`, and `all_addon_configs` values remain supported for compatibility. Validation now warns on legacy usage and incompatible legacy/new combinations, and Docker mounts prefer the app-based option when both are present.

This likely warrants a developer blog post describing the new mapping names and how to migrate from the legacy addon-based names.

## Type of change

## Additional information

- This PR fixes or closes issue: fixes [#6745](https://github.com/home-assistant/supervisor/issues/6745)
- Link to cli pull request:

## Checklist


supervisor/apps/const.py
```diff
@@ -19,11 +19,16 @@ class MappingType(StrEnum):
     DATA = "data"
     CONFIG = "config"
     SSL = "ssl"
-    ADDONS = "addons"
+    APPS = "apps"
```
agners: Should we maybe name this one a bit different anyways? Like `local_app_metadata` or something like this? 🤔

mdegat01: I thought about calling it `local_apps`. But then I realized I wanted to mount it to `/apps` to keep it simple and I liked that the name there matched the mount path so I left it.

If we want to pick a different name we can but `local_app_metadata` feels way too wordy (and its also not just metadata, its the full application including code). I also am not sure that `local_apps` is more intuitive for people? Local apps sounds like apps I've installed and that's not it either.

I'm open to a new name but the further we deviate from what we have and people are used to the more frustrating it is likely to be for people. So it seems like it should be worth it and significantly more intuitive.

supervisor/apps/validate.py
agners: This checks if the combination of `addon_config` and `homeassistant_config` is present. We probably should check the combination of `app_config` and `homeassistant_config` likewise? 🤔

supervisor/apps/validate.py
```diff
@@ -247,6 +247,22 @@ def _warn_app_config(config: dict[str, Any]):
...
         )
 
+    # Deprecated map types, deprecated as of 2026.07
+    _LEGACY_MAP_TYPES = {
+        MappingType.ADDONS: MappingType.APPS,
+        MappingType.ALL_ADDON_CONFIGS: MappingType.ALL_APP_CONFIGS,
+        MappingType.ADDON_CONFIG: MappingType.APP_CONFIG,
+    }
+    for volume in config.get(ATTR_MAP, []):
+        volume_type = volume.get(ATTR_TYPE) if isinstance(volume, dict) else None
```
agners: At this point map entries should always be dicts.

supervisor/apps/const.py
agners: We already have `/app_configs`. I think `/local_apps` is strictly better, it makes it clear that only the local apps are included. Just `apps` is also very generic, IMHO.

agners (changes requested):

sjbrenchley89 (commented):
A

supervisor/apps/validate.py
mdegat01: Yea good point. Also `ATTR_MAP` must exist and contain an array  and `ATTR_TYPE` must exist within each dictionary in the array at this point due to prior validation. Fixing this

supervisor/apps/validate.py
mdegat01: Actually this is checking if `CONFIG` and `ADDON_CONFIG` or `HOMEASSISTANT_CONFIG` is present. Because by default we map like this in the container:
1. `CONFIG` (now deprecated in favor of `HOMEASSISTANT_CONFIG`) -> `/config`
2. `HOMEASSISTANT_CONFIG` -> `/homeassistant`
3. `ADDON_CONFIG`/`APP_CONFIG` -> `/config`

These checks are for different reasons. `CONFIG` was deprecated by `HOMEASSISTANT_CONFIG`, there's no conflict there but we don't let you do use both since one deprecated the other.

`ADDON_CONFIG` is checked against `CONFIG` because by default there is a conflict, they both want `/config` and only one can win. `APP_CONFIG` should be added here too. Although technically there may not be a conflict because the user can change the default location for all of these to remove that. However this code does not acknowledge that so the warning there is accurate:
https://github.com/home-assistant/supervisor/blob/3f58d33d9f2be4b1e5742de22668db7269155a0c/supervisor/docker/app.py#L371-L407

And since its a legacy option anyway, I don't think we should change it to be honest. If they're using these other newer options they can also upgrade to `HOMEASSISTANT_CONFIG` (and then change that default as needed as well).

supervisor/apps/validate.py
```diff
@@ -394,6 +414,27 @@ def _migrate(config: dict[str, Any]):
...
+            (MappingType.ALL_APP_CONFIGS, MappingType.ALL_ADDON_CONFIGS),
+            (MappingType.APP_CONFIG, MappingType.ADDON_CONFIG),
+        ):
+            matching_volumes = [
+                volume
+                for volume in volumes
+                if volume
+                and volume[ATTR_TYPE] in {legacy_mapping_type, app_mapping_type}
+            ]
+            if len(matching_volumes) > 1:
```
agners: Actually, it seems 8e1d05d7f161c5d1bacc2fce08f390dfa384b6d2 weakened the test a bit too much, e.g. when the same volume is mapped twice. Claude suggests to use subset matching like so, which also seems a bit cleaner to me:

```suggestion
        volume_types = {volume[ATTR_TYPE] for volume in volumes}
        for app_mapping_type, legacy_mapping_type in (
            (MappingType.LOCAL_APPS, MappingType.ADDONS),
            (MappingType.ALL_APP_CONFIGS, MappingType.ALL_ADDON_CONFIGS),
            (MappingType.APP_CONFIG, MappingType.ADDON_CONFIG),
        ):
            if {app_mapping_type, legacy_mapping_type} <= volume_types:
```

supervisor/apps/validate.py
```diff
@@ -394,6 +414,27 @@ def _migrate(config: dict[str, Any]):
...
+                _LOGGER.warning(
+                    "App config using incompatible map options, '%s' and '%s'. Legacy option '%s' will be ignored. Please report this to the maintainer of %s",
```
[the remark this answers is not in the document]
agners: Right, this is what Claude found too, but it has the nicer suggestion. Applying that one.

tests/apps/test_config.py
```diff
@@ -278,6 +278,54 @@ def test_warn_advanced_deprecated(caplog: pytest.LogCaptureFixture):
...
+        ("addons", "local_apps"),
+        ("all_addon_configs", "all_app_configs"),
+        ("addon_config", "app_config"),
+    ],
+)
+def test_warn_incompatible_map_types_legacy_ignored(
+    app_mapping: str,
+    legacy_mapping: str,
+    caplog: pytest.LogCaptureFixture,
+):
```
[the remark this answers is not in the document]
agners: That is actually not true, it uses the name. But ordering it the same way seems still worthwhile.

agners (approved):
LGTM!

PR #7003, opened 2026-07-06
Run blocking hosts unlink in executor in CoreDNS.reset

agners:
## Proposed change

`CoreDNS.reset()` called `self.hosts.unlink()` directly, which performs blocking file I/O on the event loop. The sibling `write_hosts()` method in the same module already offloads its file write via `self.sys_run_in_executor()`. This change runs the unlink in the executor as well for consistency, keeping the existing `OSError` suppression (the exception still propagates out of the awaited call, so `suppress(OSError)` continues to catch it).

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/plugins/dns.py
+++ b/supervisor/plugins/dns.py
@@ -333,7 +333,7 @@ class PluginDns(PluginBase):
 
         # Resets hosts
         with suppress(OSError):
-            self.hosts.unlink()
+            await self.sys_run_in_executor(self.hosts.unlink)
         await self._init_hosts()
 
         # Reset loop protection


supervisor/plugins/dns.py
```diff
@@ -333,7 +333,7 @@ async def reset(self) -> None:
 
         # Resets hosts
         with suppress(OSError):
-            self.hosts.unlink()
+            await self.sys_run_in_executor(self.hosts.unlink)
```
mdegat01, line 336:
```py
try:
    await self.sys_run_in_executor(self.hosts.unlink)
except OSError as err:
```
Technically unrelated to this PR as its pre-existing. But since we're here, we should always be doing this check with OSErrors to warn users early on bad message errors.

mdegat01 (approved):
Approved because my comment is really just a separate bug/PR. It's tiny if you want to fold it in though.

supervisor/plugins/dns.py
agners: Addressed in #7007.

PR #7004, opened 2026-07-06
Run blocking tarfile unlink in executor on backup import

agners:
## Proposed change

In `BackupManager.import_backup()`, the cleanup after a failed `consolidate()` called the blocking `Path.unlink()` directly on the event loop. The same tarfile removal elsewhere in the module (in `_do_backup()`) already runs via `self.sys_run_in_executor()`. This change moves the `import_backup()` unlink to the executor as well so the blocking file I/O no longer runs on the event loop, making it consistent with the rest of the module.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/backups/manager.py
+++ b/supervisor/backups/manager.py
@@ -487,7 +487,7 @@ class BackupManager(FileConfiguration, JobGroup):
             try:
                 self._backups[backup.slug].consolidate(backup)
             except BackupInvalidError as err:
-                backup.tarfile.unlink()
+                await self.sys_run_in_executor(backup.tarfile.unlink)
                 raise BackupInvalidError(
                     f"Cannot import backup {backup.slug} due to: {err!s}", _LOGGER.error
                 ) from err


supervisor/backups/manager.py
```diff
@@ -487,7 +487,7 @@ async def import_backup(
             try:
                 self._backups[backup.slug].consolidate(backup)
             except BackupInvalidError as err:
-                backup.tarfile.unlink()
+                await self.sys_run_in_executor(backup.tarfile.unlink)
```
mdegat01, line 490: Same call out as in #7003 
```py
try:
    await self.sys_run_in_executor(backup.tarfile.unlink)
except OSError as err:
    if location in {LOCATION_CLOUD_BACKUP, None}:
```
Although in this case maybe we really should do this in a separate PR because I notice a couple things:
1. We have another `unlink` missing this boilerplate above as well:
https://github.com/home-assistant/supervisor/blob/912ec1da8a1825e624b7bc4d9c1b81e1e887fdfd/supervisor/backups/manager.py#L477
2. For some reason we're currently letting `OSError` exceptions bubble out of both of these locations unhandled to the `Job`? This doesn't seem like desired behavior and should probably have different handling.

mdegat01 (approved):
Approved. The comment is calling out a related issue but one that should probably just be a separate PR as it's not a simple this time.

Although test coverage would be very welcome. The codecov callout and the fact that this escaped to be found on production seems likely to be related 🙈

supervisor/backups/manager.py
agners: Yeah let's do this in a follow up PR :+1: .

agners: Addressed in #7007.

PR #7005, opened 2026-07-06
Migrate simple attrs classes to stdlib dataclasses

agners:
# Migrate simple attrs classes to stdlib dataclasses

## Proposed change

Several plain data-holder classes still used the attrs library while newer code in the project uses stdlib dataclasses. This change converts `EventListener` (bus), `Issue`/`Suggestion`/`HealthChanged`/`SupportedChanged` (resolution), `Device` (hardware), `Message` (discovery), `HostEntry` (DNS plugin), `ServiceInfo` (host services), `WhoamiData` and the scheduler's `_Task` to `@dataclass`, and switches the `attr.evolve`/`attr.asdict` call sites to `dataclasses.replace`/`dataclasses.asdict`. Field semantics map 1:1: `eq=False` becomes `compare=False`, `hash=False` and default factories carry over directly, so equality, hashing and copy behavior are unchanged (including the regenerated `uuid` on `replace()` of an `Issue`, matching `attr.evolve`).

Following review feedback, `slots=True` is now enabled on all converted classes (the attrs originals of `ServiceInfo`, `Message` and `HostEntry` were not slotted), and `HostEntry` is additionally frozen since instances are never mutated after creation. `Message` stays unfrozen on purpose: `Discovery.send()` updates the `config` field of an existing message when an app re-sends a discovery message with changed configuration.

The jobs module keeps using attrs for its validators and setter hooks, which have no stdlib dataclass equivalent, so the attrs dependency itself remains.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/discovery/__init__.py
+++ b/supervisor/discovery/__init__.py
@@ -28,7 +28,7 @@ CMD_NEW = "post"
 CMD_DEL = "delete"
 
 
-@dataclass
+@dataclass(slots=True)
 class Message:
     """Represent a single Discovery message."""
 
--- a/supervisor/discovery/__init__.py
+++ b/supervisor/discovery/__init__.py
@@ -2,11 +2,10 @@
 
 from __future__ import annotations
 
+from dataclasses import asdict, dataclass, field
 import logging
 from typing import TYPE_CHECKING, Any
 from uuid import uuid4
-
-import attr
 
 from ..const import (
     ATTR_ADDON,
@@ -29,14 +28,14 @@ CMD_NEW = "post"
 CMD_DEL = "delete"
 
 
-@attr.s
+@dataclass
 class Message:
     """Represent a single Discovery message."""
 
-    app: str = attr.ib()
-    service: str = attr.ib()
-    config: dict[str, Any] = attr.ib(eq=False)
-    uuid: str = attr.ib(factory=lambda: uuid4().hex, eq=False)
+    app: str
+    service: str
+    config: dict[str, Any] = field(compare=False)
+    uuid: str = field(default_factory=lambda: uuid4().hex, compare=False)
 
 
 class Discovery(CoreSysAttributes, FileConfiguration):
@@ -62,7 +61,7 @@ class Discovery(CoreSysAttributes, FileConfiguration):
         """Write discovery message into data file."""
         messages: list[dict[str, Any]] = []
         for message in self.list_messages:
-            messages.append(attr.asdict(message))
+            messages.append(asdict(message))
 
         self._data[ATTR_DISCOVERY].clear()
         self._data[ATTR_DISCOVERY].extend(messages)
@@ -121,7 +120,7 @@ class Discovery(CoreSysAttributes, FileConfiguration):
             _LOGGER.info("Discovery %s message ignore", message.uuid)
             return
 
-        data = attr.asdict(message)
+        data = asdict(message)
         data.pop(ATTR_CONFIG)
         # Home Assistant expects the legacy "addon" key in the push payload.
         data[ATTR_ADDON] = data.pop(ATTR_APP)
--- a/supervisor/host/services.py
+++ b/supervisor/host/services.py
@@ -101,7 +101,7 @@ class ServiceManager(CoreSysAttributes):
             _LOGGER.warning("Can't update host service information!")
 
 
-@dataclass(frozen=True)
+@dataclass(slots=True, frozen=True)
 class ServiceInfo:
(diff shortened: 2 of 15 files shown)


supervisor/host/services.py
```diff
@@ -102,13 +101,13 @@ async def update(self):
             _LOGGER.warning("Can't update host service information!")
 
 
-@attr.s(frozen=True)
+@dataclass(frozen=True)
```
mdegat01:
```suggestion
@dataclass(slots=True, frozen=True)
```
Any reason we can't add this here? Its generally more performant to use slotted dataclasses to my understanding

supervisor/discovery/__init__.py
```diff
@@ -29,14 +28,14 @@
 CMD_DEL = "delete"
 
 
-@attr.s
+@dataclass
```
mdegat01:
```suggestion
@dataclass(slots=True, frozen=True)
```
Unless there's something stopping us I think we should add slots and frozen to all of these as we migrate them. If we're worried about `frozen` creating a breakage in a gap in test coverage I guess we can skip it, it looks ok when I scanned just now though. But we should definitely add `slots` unless we have strong reason not to.

supervisor/misc/scheduler.py
```diff
@@ -2,28 +2,27 @@
...
-import attr
-
 from ..const import CoreState
 from ..coresys import CoreSys, CoreSysAttributes
 
 _LOGGER: logging.Logger = logging.getLogger(__name__)
 
 
-@attr.s(slots=True)
+@dataclass(slots=True)
```
mdegat01, line 16: Just noting that this one cannot be `frozen`, we do set fields of existing objects. `slots` is enough

supervisor/plugins/dns.py
```diff
@@ -55,12 +55,12 @@
 HOST_RESOLV: Path = Path("/etc/resolv.conf")
 
 
-@attr.s
+@dataclass
```
mdegat01:
```suggestion
@dataclass(slots=True, frozen=True)
```
Same comment as the others. Scanned and adding frozen here doesn't seem like an issue, slots definitely shouldn't be.

mdegat01 (changes requested):
Good one. I kept meaning to do this but I didn't like how I couldn't do all of them since `Job` was stuck with `attr` due to the hooks. But you're right, we should just do all the ones we can.

Small comments that we should upgrade all of these to `slots` in my opinion. And I think make a couple frozen but can skip that if you think its too risky. I scanned all the ones I suggested for usage first though and making them frozen doesn't appear like it would break anything.

supervisor/discovery/__init__.py
agners: If a discovery messsage is resent, we modify the existing message in `supervisor/discovery/__init__.py:90`. So this one we can't freeze.

supervisor/plugins/dns.py
```diff
@@ -55,12 +55,12 @@
 HOST_RESOLV: Path = Path("/etc/resolv.conf")
 
 
-@attr.s
+@dataclass(slots=True, frozen=True)
```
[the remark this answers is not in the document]
agners: This is intended behavior change according to review. Updated the PR description. Also `TypeError` was raised before too.

supervisor/host/services.py
```diff
@@ -102,13 +101,13 @@ async def update(self):
             _LOGGER.warning("Can't update host service information!")
 
 
-@attr.s(frozen=True)
+@dataclass(slots=True, frozen=True)
```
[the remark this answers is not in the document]
agners: This is intended behavior change according to review. Updated the PR description.

supervisor/discovery/__init__.py
```diff
@@ -29,14 +28,14 @@
 CMD_DEL = "delete"
 
 
-@attr.s
+@dataclass(slots=True)
```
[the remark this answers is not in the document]
agners: This is intended behavior change according to review. Updated the PR description.

supervisor/discovery/__init__.py
mdegat01: Hm. Ok I feel like we should treat that as a bug and replace these objects rather then modify them. I generally prefer immutable dataclasses to prevent unexpected niche behavior from cropping up (like if a process has a discover message, waits for another task and then finds it changed). But we can do that in a separate PR then.

mdegat01 (approved):
Looks good, LGTM

PR #6982, opened 2026-06-26
Make /os/update stop auto-rebooting after OS update

mdegat01:
## Proposed change

Changes OS update behavior so calling `/os/update` no longer automatically reboots the host after a successful update.

Instead, on successful update Supervisor now:
- Logs that the OS update completed and a reboot is required.
- Creates a `REBOOT_REQUIRED` issue with `EXECUTE_REBOOT` suggestion, matching the behavior used for Raspberry Pi firmware updates.

This is intentional and breaking for API consumers that previously relied on automatic reboot as part of `/os/update`.

Test updates in this PR:
- Updated OS manager success test to assert actual resolution issue/suggestion state instead of mocking `create_issue`.
- Added API tests for `/os/update`:
  - success path: no reboot occurs and reboot-required issue/suggestion are created
  - failure path: no reboot occurs and reboot-required issue/suggestion are not created

## Type of change

- [ ] Dependency upgrade
- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (which adds functionality to the supervisor)
- [x] Breaking change (fix/feature causing existing functionality to break)
- [ ] Code quality improvements to existing code or addition of tests

## Additional information

- This PR fixes or closes issue: fixes #6978
- This PR is related to issue:
- Link to documentation pull request: https://github.com/home-assistant/developers.home-assistant/pull/3214
- Link to cli pull request: https://github.com/home-assistant/cli/pull/662
- Link to client library pull request:

## Checklist

- [x] The code change is tested and works locally.
- [x] Local tests pass. **Your PR cannot be merged unless tests pass**
- [x] There is no commented out code in this PR.
- [x] I have followed the [development checklist][dev-checklist]
- [x] The code has been formatted using Ruff (`ruff format supervisor tests`)
- [x] Tests have been added to verify that the new code works.

If API endpoints or add-on configuration are added/changed:

- [ ] Documentation added/updated for [developers.home-assistant.io][docs-repository]
- [ ] [CLI][cli-repository] updated (if necessary)
- [ ] [Client library][client-library-repository] updated (if necessary)

[dev-checklist]: https://developers.home-assistant.io/docs/en/development_checklist.html
[docs-repository]: https://github.com/home-assistant/developers.home-assistant
[cli-repository]: https://github.com/home-assistant/cli
[client-library-repository]: https://github.com/home-assistant-libs/python-supervisor-client/

PR #7006, opened 2026-07-06
Track installed OS update pending reboot activation

agners:
## Proposed change

Since #6982 the OS update no longer reboots the host automatically, so the system keeps running the old version until the user reboots. In that window the update keeps being offered: `need_update` compares the running OS version, which only changes on reboot. Requesting the same update again re-downloads and reinstalls the full image. Worse, a Supervisor restart in that window drops the in-memory `REBOOT_REQUIRED` issue and cancels the pending update altogether, because `mark_healthy` marks the booted slot active on startup, reverting the primary boot slot set by the rauc install.

This PR tracks the installed version awaiting a reboot as `version_pending` and exposes it in `/os/info`:

- A successful install sets `version_pending`, and updating again to that version is rejected with a hint to reboot instead of re-downloading and reinstalling the image.
- `need_update` no longer reports `true` when the latest version is already installed, so `/os/info` (`update_available`) and `/available_updates` stop offering it.
- On load, the pending state is recovered from rauc by comparing the primary boot slot (via a new `GetPrimary` D-Bus wrapper) with the booted slot, re-creating the `REBOOT_REQUIRED` issue after a Supervisor restart as well.
- `mark_healthy` now skips marking the booted slot active while an update is pending, so a Supervisor restart no longer reverts the primary boot slot and cancels the installed update.

Note that the Core update entity derives update availability from `version` vs. `version_latest` itself, so a Core side change consuming `version_pending` is needed for the update entity to stop offering an already installed update.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR is related to issue: #6982
- Link to cli pull request:
- Link to client library pull request: https://github.com/home-assistant-libs/python-supervisor-client/pull/370

## Checklist


agners:
> Note that the Core update entity derives update availability from `version` vs. `version_latest` itself, so a Core side change consuming `version_pending` is needed for the update entity to stop offering an already installed update.

This means we would still have this unfortunate UX for older Core releases.

For now, let's make Supervisor "fake" `version` so that the update availability is correctly reflected even on older Core version. We can make Core consuming `version_pending` in future releases (and stop "fake" `version` depending on Core version).

supervisor/api/os.py
```diff
@@ -98,8 +99,14 @@ class APIOS(CoreSysAttributes):
     async def info(self, request: web.Request) -> dict[str, Any]:
         """Return OS information."""
         return {
-            ATTR_VERSION: self.sys_os.version,
+            # Report an installed update pending activation as the current
+            # version: the Core update entity compares version with
+            # version_latest, so it would offer the update again otherwise.
+            # Once Core consumes version_pending, this can be limited to Core
+            # versions predating that support.
+            ATTR_VERSION: self.sys_os.version_pending or self.sys_os.version,
```
mdegat01, line 107: Should we attach a core version gate to this or were you thinking to do that later?

mdegat01 (approved):
Looks good. One question about when you want to gate this fake version addition, now or later. Ready otherwise

supervisor/api/os.py
agners: I intended to do it later, when the change landed in Core.

PR #7008, opened 2026-07-07
Ignore wait-online unit failures in systemd unit failure check

agners:
## Proposed change

The systemd unit failure check introduced in #6976 creates a repair issue and reports to Sentry for any failed systemd unit. Wait-online oneshot units such as `NetworkManager-wait-online.service` end up in failed state whenever the network is not up in time at boot. Within the first week of the 2026.07.0 rollout, `NetworkManager-wait-online.service` alone is one of the most frequent Sentry reports of this check (72 users), with `systemd-networkd-wait-online.service` also appearing on supervised systems.

Home Assistant works offline, so a failed wait-online unit is transient and not actionable for the user. Skip units matching the `-wait-online.service` suffix in the check, the same way the firewall service is already skipped, so they no longer create repair issues or Sentry noise.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR is related to issue: #6976
- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/resolution/checks/systemd_unit_failure.py
+++ b/supervisor/resolution/checks/systemd_unit_failure.py
@@ -26,6 +26,10 @@ OS_FSCK_UNITS = {
 # Systemd fsck unit for data filesystem
 DATA_FSCK_UNIT = "systemd-fsck@dev-disk-by\\x2dlabel-hassos\\x2ddata.service"
 
+# Oneshot unit which ends up failed when the network is not up in time at
+# boot. Home Assistant works offline, so the failure is not actionable.
+NM_WAIT_ONLINE_SERVICE = "NetworkManager-wait-online.service"
+
 
 def setup(coresys: CoreSys) -> CheckBase:
     """Check setup function."""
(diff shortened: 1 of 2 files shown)


agners:
With #7009 in mind, I scoped this to NetworkManager only.

mdegat01 (approved):

supervisor/resolution/checks/systemd_unit_failure.py
```diff
@@ -92,6 +96,15 @@ async def run_check(self) -> None:
...
                 continue
 
+            # Skip NetworkManager-wait-online which fails when the network is
+            # slow to come up at boot, Home Assistant works offline
+            if unit_name == NM_WAIT_ONLINE_SERVICE:
+                _LOGGER.debug(
+                    "Ignoring failed wait-online unit '%s'",
+                    unit_name,
+                )
+                continue
```
mdegat01, line 106: Is this approach more readable then a `match-case`? Doesn't have to be decided now but it certainly meets the criteria for a `match-case` approach (many simple equality checks against one variable) so we could refactor to it. Could just wait until later to see if it keeps growing too.

agners: Claude says: A match-case is a poor fit here, for a Python-specific reason — in a match statement a bare name like case FIREWALL_SERVICE: is a capture pattern (it binds the variable, always matches), not an equality check.

So yeah, let's leave as is and we can still reconsider if the list grows.

PR #7009, opened 2026-07-07
Only report systemd unit failures to Sentry on Home Assistant OS

agners:
## Proposed change

The systemd unit failure check introduced in #6976 creates a repair issue and reports to Sentry for any failed systemd unit on the host. On supervised installations the first week of the 2026.07.0 rollout surfaced failed units like `gdm.service`, `tomcat8.service`, `isc-dhcp-server.service` or `snap.remmina.ssh-agent.service` — units entirely unrelated to Home Assistant, where there is nothing we can do with the report, it is pure noise in Sentry.

Only report failed units to Sentry when running on Home Assistant OS. The check itself still runs on supervised installations and keeps creating resolution issues, since a failed unit can be useful information for the admin of such a system.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR is related to issue: #6976
- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/resolution/checks/systemd_unit_failure.py
+++ b/supervisor/resolution/checks/systemd_unit_failure.py
@@ -63,6 +63,11 @@ class CheckSystemdUnitFailure(CheckBase):
 
     async def run_check(self) -> None:
         """Run check to list failed systemd units."""
+        # On a supervised installation the host units are not managed by us,
+        # only report failed units on Home Assistant OS
+        if not self.sys_os.available:
+            return
+
         # Get all failed units
         failed_units = await self._get_failed_units()
         capture_coroutines: list[Awaitable[None]] = []
--- a/supervisor/resolution/checks/systemd_unit_failure.py
+++ b/supervisor/resolution/checks/systemd_unit_failure.py
@@ -63,11 +63,6 @@ class CheckSystemdUnitFailure(CheckBase):
 
     async def run_check(self) -> None:
         """Run check to list failed systemd units."""
-        # On a supervised installation the host units are not managed by us,
-        # only report failed units on Home Assistant OS
-        if not self.sys_os.available:
-            return
-
         # Get all failed units
         failed_units = await self._get_failed_units()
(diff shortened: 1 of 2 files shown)


supervisor/resolution/checks/systemd_unit_failure.py
```diff
@@ -59,6 +59,11 @@ async def _get_failed_units(self) -> list[tuple[str, ...]]:
 
     async def run_check(self) -> None:
         """Run check to list failed systemd units."""
+        # On a supervised installation the host units are not managed by us,
+        # only report failed units on Home Assistant OS
+        if not self.sys_os.available:
+            return
```
mdegat01: Could also just add this to the check itself:
```py
@Job(name="check_systemd_unit_failure", conditions=[JobCondition.HAOS])
```
That being said, is this what we want? I think we don't want systemd unit failures from Supervised systems to get reported to sentry because there's nothing we can do with this information. But it might be useful to continue to create issues out of it so the person who can (the admin of this system) can learn that's occurring.

So perhaps instead of skipping the check altogether we just skip reporting to Sentry if its not HAOS? Though these issues are also only exposed in the CLI rather then the UI so idk if it actually makes the admins job any easier, we can skip it entirely if you think that'd just be noise for these users.

mdegat01 (approved):
Looks good. Suggested a couple alternate ideas if you want to take a look. It can also go in as is though.

supervisor/resolution/checks/systemd_unit_failure.py
agners: Yeah I was considering simply skipping the report to Sentry first. But then, the example which popped up (see PR description) are quite unrelated to HA. Not sure if users want HA get involved in those :thinking:

agners: Ok, let's skip just Sentry and report failures to the users. Maybe some even appreciate it. If not, we can still remove it entirely for Supervised users.

mdegat01 (approved):

Issue #6698, opened 2026-04-02
Change add-ons to apps in source


agners:
Refactor Supervisor code to use apps instead of add-ons.

For user facing things like configurations or backups, if we can make it such the user won't notice (backwards compatible) we can change field names etc. too.

agners:
Source code references are updated with #6717. Config files, backup metadata, and references in Docker API (container naming) are still open.

agners:
This is completed from Supervisor side.

PR #7014, opened 2026-07-08
Migrate addon job names to app and scope legacy API compatibility

mdegat01:
## Proposed change

This PR continues the addon->app migration for supervisor job names by renaming `@Job` names that still used `addon`/`addons` to `app`/`apps`.

Because job names are exposed over the API, this is intentionally treated as a breaking-change PR. To reduce compatibility risk while still moving forward:

- Keep a minimal compatibility mapping for API-exposed legacy names only.
- Retain legacy mapping for `app_manager_install`, `app_manager_uninstall`, and `app_manager_update`.
- Retain legacy mapping for backup jobs that use `cleanup=False`.
- Remove broader legacy aliases so those names now change as part of the breaking change.
- Mark `app_restart_after_problem`, `docker_app_*`, `tasks_*`, and `check_*` jobs as internal.

This keeps the compatibility surface focused while preserving the known legacy listener dependency for manager install/uninstall/update and long-lived backup jobs.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR is related to issue: #6698
- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/jobs/__init__.py
+++ b/supervisor/jobs/__init__.py
@@ -35,26 +35,16 @@ _CURRENT_JOB: ContextVar[str | None] = ContextVar("current_job", default=None)
 
 _LOGGER: logging.Logger = logging.getLogger(__name__)
 
-_LEGACY_JOB_NAMES: dict[str, str] = {
-    "app_manager_install": "addon_manager_install",
-    "app_manager_uninstall": "addon_manager_uninstall",
-    "app_manager_update": "addon_manager_update",
-    "backup_app_save": "backup_addon_save",
-    "backup_store_apps": "backup_store_addons",
-    "backup_app_restore": "backup_addon_restore",
-    "backup_restore_apps": "backup_restore_addons",
-    "backup_remove_delta_apps": "backup_remove_delta_addons",
-}
-
 
 def process_job_dict_for_legacy_compatibility(
     job_data: dict[str, Any],
 ) -> dict[str, Any]:
     """Map new job names to legacy names for API compatibility."""
-    if isinstance(name := job_data.get("name"), str) and (
-        legacy_name := _LEGACY_JOB_NAMES.get(name)
-    ):
-        return job_data | {"name": legacy_name}
+    # Home Assistant Core's hassio integration listens for this specific job name
+    # via the Supervisor websocket API. Core v2026.8 supports both names, so this
+    # compatibility can be removed when Core v2026.7 is no longer supported.
+    if job_data.get("name") == "app_manager_update":
+        return job_data | {"name": "addon_manager_update"}
     return job_data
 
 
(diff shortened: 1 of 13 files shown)


supervisor/jobs/__init__.py
```diff
@@ -35,6 +35,28 @@
...
+_LEGACY_JOB_NAMES: dict[str, str] = {
+    "app_manager_install": "addon_manager_install",
+    "app_manager_uninstall": "addon_manager_uninstall",
+    "app_manager_update": "addon_manager_update",
+    "backup_app_save": "backup_addon_save",
+    "backup_store_apps": "backup_store_addons",
+    "backup_app_restore": "backup_addon_restore",
+    "backup_restore_apps": "backup_restore_addons",
+    "backup_remove_delta_apps": "backup_remove_delta_addons",
+}
```
mdegat01: So the only known usage of this API is by core [here](https://github.com/home-assistant/core/blob/3e0aa0b81df4bc3eacca6d7a9ef0c3828ec02fb2/homeassistant/components/hassio/update.py#L222-L234). Well there's also 3 others in this file but this is the only one impacted by this since it subscribes to `addon_manager_update`.

I choose caution initially here. `app_manager_install` and `app_manager_uninstall` are also long-running docker interactions that seem more likely then others to have someone looking for progress on (well maybe just install really come to think of it, uninstall is just a stop and rm image). The `backup_` ones are all marked `clean=False` so presumably we expected someone to listen to them but I'm not sure anyone actually implemented that, I don't think we did in the backup overhaul. I also marked a lot of job as internal that I can't imagine anyone is listening for and are really just noise.

I can be more or less cautious here if we wish. The only one we truly need an answer for right now is `addon_manager_update`. The thing I'm not sure of though is how to handle deprecation here. We only return one name and its either old or new. I can change v2 API to use new names and v1 to use old names if we want but core is primarily listening for these updates over WSS...

So open to suggestions here to give a proper deprecation strategy that involves both working for a period. Else we just have to prepare for a release cutoff date or leave this indefinitely.

mdegat01: After discussion we're only going to handle backwards compatibility for `addon_manager_update` since job names were fully undocumented. So we handle the one known use case in core and that's it. Next version of core will handle either name and this can be removed when core 2026.7 is no longer supported.

mdegat01:
Needs core because we must update the addon update entity to handle tracking job updates to either `app_manager_update` or `addon_manager_update`.

agners:
> Needs core because we must update the addon update entity to handle tracking job updates to either `app_manager_update` or `addon_manager_update`.

Right, track both makes sure that Core is compatible to this backwards compatibility fix, and eventually allows to move to `app_manager_update`.

agners (approved):
LGTM, thanks!

PR #7018, opened 2026-07-09
Queue Home Assistant Core restart/stop behind a running start

agners:
## Proposed change

The `home_assistant_core` job group used `GROUP_REJECT` for `start`, `stop` and `restart`. Because `start` holds the group for the entire duration of `_block_till_run()` (polling Core's API until it reports `RUNNING`), any `stop`/`restart` requested *while Core is still starting* is rejected outright with:

```
Another job is running for job group home_assistant_core
```

This breaks legitimate early restarts. Concretely: Home Assistant Core's new UI-managed HTTP config applies a changed config as a "pending" trial and, when that config cannot bind (e.g. the port is already taken), reverts by restarting Core — which happens *during* Core startup. Supervisor refused that restart (see https://github.com/home-assistant/core/pull/176076):

```
aiohasupervisor.exceptions.SupervisorBadRequestError: Another job is running for job group home_assistant_core
```

This switches `stop` and `restart` to `GROUP_QUEUE`, so such a request **waits** for the in-progress `start` to finish and then runs, instead of failing. `start` keeps `GROUP_REJECT`. Reentrant same-group calls (e.g. `rebuild` → `start`) are unaffected: `JobGroup.acquire()` already short-circuits when the current job chain owns the lock.

### Note / follow-up

Now that `restart`/`stop` can be held open (queued) rather than returning immediately, callers using the client library's default 10s request timeout (`aiohasupervisor` `DEFAULT_TIMEOUT`) can hit a spurious `SupervisorTimeoutError` while the request is queued, even though the operation ultimately succeeds. These long-running lifecycle calls should use `timeout=None` (as `core/update` already does) in the [client library][client-library-repository]. That is a separate, independently-correct fix.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/homeassistant/core.py
+++ b/supervisor/homeassistant/core.py
@@ -493,7 +493,7 @@ class HomeAssistantCore(JobGroup):
     @Job(
         name="home_assistant_core_stop",
         on_condition=HomeAssistantJobError,
-        concurrency=JobConcurrency.GROUP_REJECT,
+        concurrency=JobConcurrency.GROUP_QUEUE,
     )
     async def stop(self, *, remove_container: bool = False) -> None:
         """Stop Home Assistant Docker."""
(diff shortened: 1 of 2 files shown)


agners:
Queuing semantics can also be problematic: What is the timeout? The user doesn't see that his request does not get processed (immediately). Let's stick with the current semantics here.

PR #7024, opened 2026-07-10
Rename Docker app container and builder names from addon_ to app_

### LINKED ISSUE #6744: Docker container names contain addon
### Description

Docker container names use addons still. There might be other strings related to Docker which might need migration.

### Additional context

_No response_

mdegat01:
## Proposed change

The Docker container name for each installed app (and its transient build container) was still derived as `addon_{slug}` / `addon_builder_{slug}`. This renames them to `app_{slug}` / `app_builder_{slug}` to match the broader addon→app rename.

The container naming pattern lives in `DockerApp.slug_to_name()` and `_build()` in `supervisor/docker/app.py`.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR fixes or closes issue: fixes #6744
- Link to cli pull request:

While direct container name interaction is undocumented and not recommended, it was technically possible for users to reference containers by name (e.g. `docker exec addon_local_ssh ...`). Those names now become `app_local_ssh`, etc. This is intentionally a breaking change as part of the addon→app migration.

## Checklist

### DIFF
--- a/supervisor/docker/app.py
+++ b/supervisor/docker/app.py
@@ -95,7 +95,7 @@ class DockerApp(DockerInterface):
     @staticmethod
     def slug_to_name(slug: str) -> str:
         """Convert slug to container name."""
-        return f"addon_{slug}"
+        return f"app_{slug}"
 
     @property
     def image(self) -> str | None:
@@ -719,7 +719,7 @@ class DockerApp(DockerInterface):
             f"{docker_version.major}.{docker_version.minor}.{docker_version.micro}-cli"
         )
 
-        builder_name = f"addon_builder_{self.app.slug}"
+        builder_name = f"app_builder_{self.app.slug}"
 
         # Remove dangling builder container if it exists by any chance
         # E.g. because of an abrupt host shutdown/reboot during a build
--- a/supervisor/docker/app.py
+++ b/supervisor/docker/app.py
@@ -9,6 +9,7 @@ from http import HTTPStatus
 from ipaddress import IPv4Address
 import logging
 from pathlib import Path
+import re
 import tempfile
 from typing import TYPE_CHECKING, Any, Literal, cast
 
@@ -164,6 +165,9 @@ class DockerApp(DockerInterface):
                 _LOGGER.error,
             ) from err
         except aiodocker.DockerError as err:
+            if err.status == HTTPStatus.NOT_FOUND:
+                # Legacy container removed between lookup and rename, nothing to migrate
+                return None
             raise DockerError(
                 f"Can't rename legacy app container {legacy_name} to {self.name}: {err!s}",
                 _LOGGER.error,
@@ -793,14 +797,17 @@ class DockerApp(DockerInterface):
             f"{docker_version.major}.{docker_version.minor}.{docker_version.micro}-cli"
         )
 
+        # app_builder is the new name used for the builder container but existing containers
+        # created by older versions of Supervisor might still use addon_builder
         builder_name = f"app_builder_{self.app.slug}"
-        legacy_builder_name = f"addon_builder_{self.app.slug}"
+        builder_name_pattern = rf"^(?:app|addon)_builder_{re.escape(self.app.slug)}$"
 
         # Remove dangling builder containers if they exist by any chance
         # E.g. because of an abrupt host shutdown/reboot during a build
         try:
             containers = await self.sys_docker.containers.list(
-                all=True, filters={"name": [builder_name, legacy_builder_name]}
+                all=True,
+                filters={"name": [builder_name_pattern]},
             )
             if containers:
                 await asyncio.gather(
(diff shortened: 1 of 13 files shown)


supervisor/docker/app.py
```diff
@@ -95,7 +95,7 @@ def __init__(self, coresys: CoreSys, app: App):
     @staticmethod
     def slug_to_name(slug: str) -> str:
         """Convert slug to container name."""
-        return f"addon_{slug}"
+        return f"app_{slug}"
```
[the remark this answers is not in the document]
agners: Yeah Copilot is right here. A Supervisor update happens while app containers are running, so they have the wrong name after updating Supervisor. As it stands now, Supervisor would start a second container, with conflicting resources (ports etc..). 

We either need to force stop/start all apps container, or remain compatible with `addon_` prefix until the next full system restart at least. I think I'd try thee latter approach: Make Supervisor accept `addon_`. Maybe we can also raise a reboot repair as long as we see `addon_` containers. Maybe we even need to mark the system unhealthy at one point if the user never reboots, so we can safely remove `addon_` prefix compatibility. But I'd wait with that for a year or so.

supervisor/docker/app.py
```diff
@@ -716,7 +716,7 @@ async def _build(self, version: AwesomeVersion, image: str | None = None) -> Non
             f"{docker_version.major}.{docker_version.minor}.{docker_version.micro}-cli"
         )
 
-        builder_name = f"addon_builder_{self.app.slug}"
+        builder_name = f"app_builder_{self.app.slug}"
```
agners, line 802: For the code below (dangling build container) we probably need to check `addon_` for a while as well so they get properly cleaned up.

supervisor/docker/app.py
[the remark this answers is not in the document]
agners: Actually, Claude suggests something else:

>  a one-time migration in DockerApp.attach() (or AppManager.load()): if app_{slug} isn't found, look up addon_{slug} and use Docker's container rename API — rename works on running containers, is atomic, and avoids restarting user workloads. Falling back to stop+remove of the legacy container before first start would also be acceptable but forces a restart of every add-on.

Seems easy enough. Maybe just raise a reboot required in case renaming fails 🤷 .

agners (changes requested):

supervisor/docker/app.py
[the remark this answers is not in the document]
mdegat01: Fixed in line with the proposal, makes sense. Skipped the reboot though, will that even help? If the rename fails then we won't find them to stop them so won't the dangling containers just start again?

supervisor/docker/app.py
```diff
@@ -719,23 +793,29 @@ async def _build(self, version: AwesomeVersion, image: str | None = None) -> Non
...
+        legacy_builder_name = f"addon_builder_{self.app.slug}"
 
-        # Remove dangling builder container if it exists by any chance
+        # Remove dangling builder containers if they exist by any chance
         # E.g. because of an abrupt host shutdown/reboot during a build
         try:
-            container = await self.sys_docker.containers.get(builder_name)
-            await container.delete(force=True, v=True)
+            containers = await self.sys_docker.containers.list(
+                all=True, filters={"name": [builder_name, legacy_builder_name]}
```
agners: Nit: This filters using unanchored regex, meaning we'll match substrings. Either anchor the matches or compare the list against `builder_name` and `legacy_builder_name` again below

supervisor/docker/app.py
```diff
@@ -95,13 +98,84 @@ def __init__(self, coresys: CoreSys, app: App):
...
+        # Raise on a rename fail. This means we found the legacy container and could not rename it
+        # so follow-up actions like attach will not work properly regardless
+        try:
+            await legacy_container.rename(self.name)
+        except TimeoutError as err:
+            raise DockerTimeoutError(
+                f"Timeout renaming legacy app container {legacy_name} to {self.name}",
+                _LOGGER.error,
+            ) from err
+        except aiodocker.DockerError as err:
```
agners, line 167: Theoretical issue, but in case the container got removed between the get above and here, Docker will raise 404. We can just return None in this case, so:



```suggestion
        except aiodocker.DockerError as err:
            if err.status == HTTPStatus.NOT_FOUND:
                # Legacy container removed between lookup and rename, nothing to migrate
                return None
```

agners (changes requested):
Other than the two minor nit, LGTM!

supervisor/docker/app.py
agners: Actually, double check what filters is doing here.

mdegat01: Yea it accepts a regex so we can bound it just match the ones we want

agners (approved):

PR #7039, opened 2026-07-13
Add API to manage SSH authorized keys on Home Assistant OS

agners:
## Proposed change

Add an API to manage root's SSH authorized keys on Home Assistant OS through the Supervisor, wrapping the `AddSSHAuthKey` and `ClearSSHAuthKeys` D-Bus methods of the OS Agent `io.hass.os` System object, which the Supervisor never used so far. This provides the backend for managing SSH access to the HAOS debug console (port 22222). The endpoints map 1:1 onto the OS Agent operations, so no request can partially succeed:

- `GET /os/ssh/authorized_keys` returns the configured keys (`{"keys": ["ssh-ed25519 AAAA…", …]}`).
- `POST /os/ssh/authorized_keys` with `{"key": "ssh-ed25519 AAAA…"}` appends a single key.
- `DELETE /os/ssh/authorized_keys` removes all keys.

The `GET` requires OS Agent 1.11.0 (which added the `ListSSHAuthKeys` D-Bus method, home-assistant/os-agent#281) and returns 404 on older releases, like the Raspberry Pi firmware endpoints do — this provides the auditability of what keys grant access to the box, including keys imported via USB or written by add-ons. Replacing the configured set means clear and re-add; an idempotent `PUT` to replace the whole set can be added later.

Proper key validation is left to OS Agent, which validates submitted keys since 1.10.0 (home-assistant/os-agent#273); a key it rejects surfaces as an error response including its validation message. The endpoints work with older OS Agent releases too, accepting that validation is lossy there: the Supervisor runs a basic sanity check on the submitted key — no control characters (one submitted key can never write more than one `authorized_keys` line) and at most 3000 bytes (dropbear — the consumer of the file on HAOS — ignores longer lines, which would leave a key that passes but never works). On releases before 1.10.0 the error OS Agent returns when clearing an already absent `authorized_keys` file (inverted error check) is treated as the empty state it reports. The endpoints are restricted to Home Assistant Core (`core_only`): add-on tokens of any role, the CLI plugin, and the observer are rejected, so host SSH key management is only possible from the Home Assistant UI and the host shell itself. This can be opened up later if needed.

Since dropbear on HAOS is gated by `ConditionFileNotEmpty` on the `authorized_keys` file, which systemd only evaluates at unit start, the service is started after a key is added and stopped after a clear — mirroring what the USB config import (`haos-config`) does, and ensuring a clear also terminates established SSH sessions (revocation). A running dropbear re-reads the file on every authentication attempt, and starting or stopping an already active/inactive unit is a no-op.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

Related OS Agent pull requests: home-assistant/os-agent#273 (key validation and hardening of the wrapped D-Bus methods, released in OS Agent 1.10.0; add and clear work with both older and newer releases) and the `ListSSHAuthKeys` addition released in OS Agent 1.11.0, which the `GET` endpoint requires.

## Checklist


supervisor/api/os.py
```diff
@@ -89,6 +92,77 @@
...
+        "sk-ssh-ed25519@openssh.com",
+        "sk-ecdsa-sha2-nistp256@openssh.com",
+    }
+)
+
+# dropbear, which consumes authorized_keys on Home Assistant OS, ignores
+# lines longer than 3000 bytes (and OS Agent rejects them)
+SSH_AUTH_KEY_MAX_LENGTH: Final = 3000
+
+RE_SSH_KEY_CONTROL_CHARS = re.compile(r"[\x00-\x1f\x7f]")
```
agners, line 111: Let's simplify this to just line length and control chars, but not check SSH type.

supervisor/os/manager.py
```diff
@@ -501,3 +513,55 @@ async def set_boot_slot(self, boot_name: str) -> None:
...
+        except DBusError as err:
+            # On affected OS Agent releases the missing-file error is the
+            # empty state clearing aims for, so treat it as success there.
+            if (
+                self.sys_dbus.agent.version >= CLEAR_SSH_AUTH_KEYS_FIXED_VERSION
+                or CLEAR_SSH_AUTH_KEYS_MISSING_FILE_ERROR not in str(err)
+            ):
+                raise HassOSError(
+                    f"Can't clear SSH authorized keys: {err!s}", _LOGGER.error
+                ) from err
```
mdegat01: Do we have to require users pass in every key every time? Can we offer a way to add to existing keys instead of replacing the full list each time? Since OSAgent offers a designated clear authorized keys API it seems unnecessary frustrating to make one API that does an upsert. Wouldn't it be easier to have a POST API that adds one or more keys and a DELETE API that clears the file?

supervisor/api/__init__.py
```diff
@@ -305,6 +305,7 @@ def _register_os(self, app: web.Application) -> None:
                 web.get("/os/datadisk/list", api_os.list_data),
                 web.post("/os/datadisk/wipe", api_os.wipe_data),
                 web.post("/os/boot-slot", api_os.set_boot_slot),
+                web.post("/os/ssh/authorized_keys", api_os.ssh_authorized_keys),
```
mdegat01: If we truly want this endpoint to replace the entire authorized keys file with this new set of keys and not add a key to the set we should use `PUT` here not `POST`imo. Per [mozilla guidelines](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Methods/PUT):

> The PUT HTTP method creates a new resource or replaces a representation of the target resource with the request [content](https://developer.mozilla.org/en-US/docs/Glossary/HTTP_Content).
> 
> The difference between PUT and [POST](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Methods/POST) is that PUT is [idempotent](https://developer.mozilla.org/en-US/docs/Glossary/Idempotent): calling it once is no different from calling it several times successively (there are no side effects).

As defined this is idempotent and replaces the resource (authorized key file in this case) so PUT is the better fit. But personally I would prefer this use `POST`, not be idempotent, and just append one or more keys to the existing file.

supervisor/os/manager.py
```diff
@@ -501,3 +513,55 @@ async def set_boot_slot(self, boot_name: str) -> None:
...
+        on_condition=HassOSJobError,
+        concurrency=JobConcurrency.REJECT,
+        internal=True,
+    )
+    async def set_ssh_authorized_keys(self, keys: list[str]) -> None:
+        """Replace root's SSH authorized keys on the host and start dropbear.
+
+        OS Agent validates each key since 1.10.0 and only offers clear and
+        append operations, so the replacement is not atomic: if an append is
+        rejected or fails, keys added before it remain in place.
```
mdegat01: This seems like a great reason to lay out the API like this:
1. `POST /os/ssh/authorized_keys` - accepts exactly one key. Calls the add ssh key in OS agent with it
2. `DELETE /os/ssh/authorized_keys` - accepts no arguments. Calls clear ssh auth keys in OS Agent

This makes the API easy to use, easy to translate to the CLI and ensures that no API call can partially succeed by having some OS Agent operations succeed before one fails. Since each API maps 1:1 with an OS Agent DBus API.

mdegat01 (changes requested):
I am not a fan of one upsert API like this. If we want to keep it then it should be PUT or PATCH not POST but I think we should change the API layout [per my comment below](https://github.com/home-assistant/supervisor/pull/7039/changes#r3582376731)

supervisor/os/manager.py
agners: Yeah agreed mirroring the OS Agent API makes more sense. Maybe we should have a get anyways at one point. I think I was concerned about unnecessary information leak when initially created the OS Agent implementation, but maybe that was a bit overly cautious.

supervisor/api/middleware/security.py
```diff
@@ -146,7 +146,7 @@ class _AppSecurityPatterns:
             r"|/multicast/.+"
             r"|/network/.+"
             r"|/observer/.+"
-            r"|/os/(?!datadisk/wipe).+"
+            r"|/os/(?!datadisk/wipe|ssh/authorized_keys).+"
```
mdegat01, line 154: Should we make this a core only endpoint rather then a manager one? Putting it in manager would give many apps the ability to call it. I know its hardly the only attack avenue if you consider the idea of a malicious app but it still just doesn't seem like capability we want to allow apps to do.

The downside would be the SSH app also can't call it which is probably the one app we'd prefer to allow. But as long as we make the proposed UI this seems like an acceptable situation to only allow host ssh key management from HA UI and the host shell itself.

supervisor/os/manager.py
```diff
@@ -501,3 +513,58 @@ async def set_boot_slot(self, boot_name: str) -> None:
...
+        except DBusError as err:
+            # On affected OS Agent releases the missing-file error is the
+            # empty state clearing aims for, so treat it as success there.
+            if (
+                self.sys_dbus.agent.version >= CLEAR_SSH_AUTH_KEYS_FIXED_VERSION
+                or CLEAR_SSH_AUTH_KEYS_MISSING_FILE_ERROR not in str(err)
+            ):
+                raise HassOSError(
+                    f"Can't clear SSH authorized keys: {err!s}", _LOGGER.error
+                ) from err
```
mdegat01, line 573: Shouldn't this stop dropbear service after a successful key clear? Since we've reset to initial state where the authorized key file is empty.

mdegat01 (changes requested):
I think I'd prefer we move these APIs to core only unless we really want to allow management from apps. And seems like we should stop the dropbear service after the keys are cleared.

supervisor/api/middleware/security.py
agners: Yeah I agree, let's make this Core only, at least for now. Opening up later is still possible. Core is what I am interested for now.

supervisor/os/manager.py
agners: Hm, right we start the service, so we should also stop on clear. Added that :+1:

mdegat01 (approved):
Good catch on the concurrency issue copilot. Looks good to me, LGTM 👍

PR #7042, opened 2026-07-14
Scope job names by API version and negotiate websocket event version

agners:
## Proposed change

Follow-up to #7014, which renamed externally visible job names from `addon_*` to `app_*` and shipped as a breaking change with a minimal legacy name mapping. This PR removes the need for the breaking change by scoping the job names to API surfaces:

- The v1 REST API (`/jobs`) maps all renamed jobs back to their legacy names, while the v2 REST API (`/v2/jobs`) exposes the new `app_*` names. The legacy name table covers every renamed job that is externally visible; internal jobs never surface through the API or websocket events.
- Websocket job events sent to Home Assistant Core use a per-connection negotiated event version: once the connection is established, Supervisor offers its supported versions via a new `supervisor/event_version` command and Core replies with the version it wants to receive. Version 1 sends job events with the legacy job names, version 2 with the new names. A Core without support for the command rejects it and Supervisor keeps the default version 1. Negotiation runs on every (re)connect, so a Core update or downgrade takes effect with the reconnection that follows it.

To make the per-connection decision possible, the legacy name mapping moves from job event creation time to send time: events queued during Supervisor startup are now mapped when they are actually sent, against the connection that carries them.

This allows Core to do the rename cleanly from one version to the next: switch to the v2 REST API and v2 websocket events in a single release, without a compatibility window where both name sets must be handled. Deprecating the legacy names is then tied to the v1 REST API and event version 1 as a whole, rather than to individual job names.

Core currently matches on the legacy names (`addon_manager_update` via websocket, `backup_store_addons`/`backup_addon_save` via the v1 REST API) and continues to work unchanged. Opting into version 2 requires a Core-side counterpart implementing the `supervisor/event_version` command in the hassio integration. The frontend consumes neither job events nor the jobs REST API today.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR is related to issue: #6698
- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/api/__init__.py
+++ b/supervisor/api/__init__.py
@@ -345,12 +345,20 @@ class RestAPI(CoreSysAttributes):
         api_jobs = APIJobs()
         api_jobs.coresys = self.coresys
 
+        # V1 keeps the legacy job names, V2 exposes the new ones.
+        if app is self.versions[AppVersion.V2]:
+            info_handler = api_jobs.info
+            job_info_handler = api_jobs.job_info
+        else:
+            info_handler = api_jobs.info_v1
+            job_info_handler = api_jobs.job_info_v1
+
         app.add_routes(
             [
-                web.get("/jobs/info", api_jobs.info),
+                web.get("/jobs/info", info_handler),
                 web.post("/jobs/options", api_jobs.options),
                 web.post("/jobs/reset", api_jobs.reset),
-                web.get("/jobs/{uuid}", api_jobs.job_info),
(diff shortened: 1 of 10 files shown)


agners:
Let's introduce a feature flag "WebSocket API v2" which makes Supervisor emit the new strings on WebSocket. This allows us to test that things work reliably while we go through all these places.

Once we completed all strings, and two years passed (the Supervisor -> Core backwards compatibility timeframe) we can make the feature flag the default (and remove the old strings), and finally remove all "or's" on Core side.

PR #7050, opened 2026-07-16
Fix Core API proxy stripping the multipart boundary from Content-Type

### LINKED ISSUE #7049: Core API proxy strips the multipart boundary from Content-Type, breaking all multipart requests from add-ons

Any `multipart/form-data` request an add-on sends to Home Assistant Core through the Supervisor Core API proxy (`http://supervisor/core/api/...`) fails: Core receives the request with `Content-Type: multipart/form-data` **without the `boundary` parameter**, so aiohttp cannot parse the body and raises:

```
ValueError: boundary missed for Content-Type: multipart/form-data
```

The same request sent directly to Core (`http://<core>:8123`, long-lived access token) works fine, which confirms the proxy is mutating the header.

This affects every multipart Core API endpoint when called from an add-on, e.g.:

- `POST /api/media_source/local_source/upload` (media library upload — what the media browser UI uses)
- `POST /api/webhook/<id>` with form-data payloads (e.g. push camera)

**Root cause**

The proxy rebuilds the forwarded request and uses aiohttp's `request.content_type` convenience property, which returns **only the media type and drops all parameters** — including `boundary`:

https://github.com/home-assistant/supervisor/blob/858630fb752a4dedcd2a3fb1027d846c88325c56/supervisor/api/proxy.py#L101

```python
content_type=request.content_type,   # aiohttp property: params (boundary) stripped
```

`make_request` then sets that already-stripped value verbatim on the outbound request:

https://github.com/home-assistant/supervisor/blob/858630fb752a4dedcd2a3fb1027d846c88325c56/supervisor/homeassistant/api.py#L286-L287

The body bytes are streamed unmodified (`data=request.content`), but the boundary string is declared **only** in the Content-Type header parameter — once it is gone, the multipart body is unparseable. JSON requests are unaffected (`application/json` has no required parameters), which is why this has stayed under the radar.

**Proposed fix (one line)**

```python
content_type=request.headers.get(hdrs.CONTENT_TYPE),
```

This forwards the original header value with its parameters. It does not interact with the `FORWARD_HEADERS` allow-list design (Authorization replacement is untouched); it only preserves the parameters of a header the proxy already chooses to forward — the same behavior as standard reverse proxies. Happy to submit a PR.

**My use case**

I maintain an add-on that runs an MCP (Model Context Protocol) server exposing Home Assistant to LLM agents. One of its tools uploads agent-generated images into the media library (`POST /api/media_source/local_source/upload`), whose returned `media_content_id` is then passed to services accepting a media selector (e.g. an e-paper integration's `upload_image`). That endpoint is the only supported upload path (multipart requires REST; there is no WS equivalent), and from an add-on the only route to Core is the Supervisor proxy — `SUPERVISOR_TOKEN` is not valid against Core directly. Because the proxy drops the boundary, the endpoint is unreachable from add-ons entirely. My current workaround is mapping `media` read-write and writing files into `/media` directly, which forfeits the read-only media mount and reimplements validation the Core endpoint already provides.

**History**

First reported over seven years ago against the wrong repo: home-assistant/core#19588 (Dec 2018, push-camera webhook via the hassio proxy, same failure) — closed stale without a fix. The stripping line is still present on current `main`.

Home Assistant OS

From inside any add-on container with `homeassistant_api: true`:

1. Create any small PNG file, e.g. `/tmp/test.png`
2. Run:

```sh
curl -sf -X POST http://supervisor/core/api/media_source/local_source/upload \
  -H "Authorization: Bearer ${SUPERVISOR_TOKEN}" \
  -F "media_content_id=media-source://media_source/local" \
  -F "file=@/tmp/test.png"
```

3. Request fails with HTTP 500; Home Assistant Core log shows `ValueError: boundary missed for Content-Type: multipart/form-data`
4. Control: the identical request sent directly to `http://<core>:8123/api/media_source/local_source/upload` with a long-lived access token succeeds and returns `{"media_content_id": "..."}`

```txt
Nothing relevant appears in the Supervisor logs — the proxy forwards the request without error.
The failure surfaces in the Home Assistant Core log:

ValueError: boundary missed for Content-Type: multipart/form-data

The behavior is deterministic and fully explained by the code path above.
```

### System information

```txt
Supervisor: 2026.07.3 (stable channel)
Home Assistant Core: 2026.7.2
Home Assistant OS: 18.1 (kernel 6.18.37-haos)
Arch: amd64 (KVM virtual machine)
```

### Additional information

The proxy code predates 2019 with this behavior unchanged; core#19588 documents the same failure mode from the webhook angle. I can submit the one-line PR (plus a test) if the fix direction is acceptable.

parkghost:
## Proposed change

The Core API proxy (`/core/api/...`) forwards the request's Content-Type via aiohttp's parsed `request.content_type` property, which returns only the media type and **drops all header parameters — including the multipart `boundary`**. Core then cannot parse any multipart body proxied from an add-on and rejects the request with:

```
ValueError: boundary missed for Content-Type: multipart/form-data
```

This makes every multipart Core API endpoint (e.g. `POST /api/media_source/local_source/upload`, form-data webhooks) unreachable from add-ons, since the Supervisor proxy is the only route available to `SUPERVISOR_TOKEN`. First reported from the webhook angle in home-assistant/core#19588 (2018, closed stale).

This PR forwards the raw `Content-Type` header instead, preserving its parameters — the same thing the `stream()` proxy path already does (`request.headers.get(CONTENT_TYPE, "")`). Behavior note: when the incoming request has no Content-Type header, the proxy previously forwarded aiohttp's `application/octet-stream` fallback; it now omits the header and lets the client default apply — net effect is unchanged.

A regression test is included: it POSTs a multipart body through the proxy and asserts the boundary parameter survives into `make_request` (fails before the fix with `'multipart/form-data' != 'multipart/form-data; boundary=...'`, passes after).

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR fixes or closes issue: fixes #7049
- This PR is related to issue: home-assistant/core#19588
- Link to cli pull request:

## Checklist

[dev-checklist]: https://developers.home-assistant.io/docs/development_checklist/

### DIFF
--- a/supervisor/api/proxy.py
+++ b/supervisor/api/proxy.py
@@ -98,7 +98,7 @@ class APIProxy(CoreSysAttributes):
                     for name, value in request.headers.items()
                     if name in FORWARD_HEADERS
                 },
-                content_type=request.content_type,
+                content_type=request.headers.get(CONTENT_TYPE),
                 data=request.content,
                 timeout=timeout,
                 params=request.query,
--- a/tests/api/test_proxy.py
+++ b/tests/api/test_proxy.py
@@ -361,6 +361,49 @@ async def test_api_proxy_delete_request(
         assert response.content_type == "application/json"
 
 
+async def test_api_proxy_multipart_content_type_preserved(
+    api_client: TestClient,
+    install_app_example: App,
+):
+    """Test multipart Content-Type keeps its boundary parameter when proxied."""
+    install_app_example.persist[ATTR_ACCESS_TOKEN] = "abc123"
+    install_app_example.data["homeassistant_api"] = True
+
+    with patch.object(HomeAssistantAPI, "make_request") as make_request:
+        # Mock the response from make_request
+        mock_response = AsyncMock()
+        mock_response.status = 200
+        mock_response.content_type = "application/json"
+        mock_response.read.return_value = b'{"result": "ok"}'
+        make_request.return_value.__aenter__.return_value = mock_response
+
+        boundary = "d1b1a3a3f0a94a5c9e3a"
+        body = (
+            f"--{boundary}\r\n"
+            'Content-Disposition: form-data; name="file"; filename="test.png"\r\n'
+            "Content-Type: image/png\r\n\r\n"
+            "fakepng\r\n"
+            f"--{boundary}--\r\n"
+        ).encode()
+
+        response = await api_client.post(
+            "/core/api/media_source/local_source/upload",
+            headers={
+                "Authorization": "Bearer abc123",
+                "Content-Type": f"multipart/form-data; boundary={boundary}",
+            },
+            data=body,
+        )
+
+        # The boundary parameter must survive the proxy — without it Core
+        # cannot parse the multipart body ("boundary missed for Content-Type").
+        assert (
+            make_request.call_args[1]["content_type"]
+            == f"multipart/form-data; boundary={boundary}"
+        )
+        assert response.status == 200
+
+
 async def test_api_proxy_mcp_headers_forwarded(
     api_client: TestClient,
     install_app_example: App,


agners (approved):
Nice catch, LGTM! Thank you!

PR #7059, opened 2026-07-17
Rename backup and restore job stages from addon(s) to app(s)

### LINKED ISSUE #7028: BackupJobStage and RestoreJobStage enum values
### Description

Several of our backup and restore job stages contain addon in the stage name. This is technically a breaking change to the API since the stage value is included in job updates over WS and `/jobs`. However since we decided blanket changing job names was acceptable in #7014 it should be even safer to do this one. There is no known usage of job stages currently in core or otherwise.

### Additional context

_No response_

mdegat01:
_Note: Description updated by @agners to reflect the backwards compatibility layer added in later commits._

## Proposed change

As part of the ongoing addon -> app refactor, this changes backup and restore job stage identifiers that still used addon/addons terminology to app/apps terminology.

This updates both enum members and emitted stage string values, then adjusts backup manager stage transitions and the related progress-event tests. It also fixes the `COPY_ADDITONAL_LOCATIONS` -> `COPY_ADDITIONAL_LOCATIONS` typo in the enum member name (string value unchanged).

Home Assistant Core parses these stage values in its `hassio` backup integration to report backup/restore progress, so the legacy stage names are preserved on REST v1 and WebSocket v1 output via the existing legacy compatibility mapping in the jobs module. The new `app` stage names are only emitted once a client opts into the v2 API via the `supervisor_websocket_v2_api` feature flag, so this is only a breaking change for clients opting into the v2 API.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR fixes or closes issue: fixes #7028
- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/backups/const.py
+++ b/supervisor/backups/const.py
@@ -27,24 +27,24 @@ class BackupType(StrEnum):
 class BackupJobStage(StrEnum):
     """Backup job stage enum."""
 
-    ADDON_REPOSITORIES = "addon_repositories"
-    ADDONS = "addons"
+    APP_REPOSITORIES = "app_repositories"
+    APPS = "apps"
     FINISHING_FILE = "finishing_file"
     FOLDERS = "folders"
     HOME_ASSISTANT = "home_assistant"
     SUPERVISOR_CONFIG = "supervisor_config"
     COPY_ADDITONAL_LOCATIONS = "copy_additional_locations"
-    AWAIT_ADDON_RESTARTS = "await_addon_restarts"
+    AWAIT_APP_RESTARTS = "await_app_restarts"
 
 
 class RestoreJobStage(StrEnum):
     """Restore job stage enum."""
 
-    ADDON_REPOSITORIES = "addon_repositories"
-    ADDONS = "addons"
-    AWAIT_ADDON_RESTARTS = "await_addon_restarts"
+    APP_REPOSITORIES = "app_repositories"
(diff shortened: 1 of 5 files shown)


supervisor/backups/const.py
```diff
@@ -27,24 +27,24 @@ class BackupType(StrEnum):
...
 
-    ADDON_REPOSITORIES = "addon_repositories"
-    ADDONS = "addons"
+    APP_REPOSITORIES = "app_repositories"
+    APPS = "apps"
     FINISHING_FILE = "finishing_file"
     FOLDERS = "folders"
     HOME_ASSISTANT = "home_assistant"
     SUPERVISOR_CONFIG = "supervisor_config"
     COPY_ADDITONAL_LOCATIONS = "copy_additional_locations"
```
agners: If we are at it, maybe we can fix this typo 😅 .

agners (changes requested):
> Although this is breaking change there is no known usage of this undocumented feature so no attempt at maintaining backwards compatibility is made here.

Unfortunately, that is not correct. The stages is convered into an enum on Core side:

https://github.com/home-assistant/core/blob/e14cab41770dcbf8a12f4b0641254d20e60fe18b/homeassistant/components/hassio/backup.py#L679

So unfortunately we need to make this in a backwards compatible manor too 😢 .

agners (approved):
LGTM!

PR #7061, opened 2026-07-18
Fix flaky test_adjust_system_datetime_if_time_behind

agners:
## Proposed change

`tests/test_core.py::test_adjust_system_datetime_if_time_behind` was flaky in CI, occasionally failing with `Failed: Timeout (>10.0s) from pytest-timeout`.

The test starts `_adjust_system_datetime()` as a background task and then emits the `systemd-timesyncd` `PropertiesChanged` signal that `wait_for_active_state()` blocks on. It used a fixed `await asyncio.sleep(0.1)` to give the task time to install its D-Bus signal subscription before emitting. On a busy CI runner (heavily parallel under pytest-xdist), 0.1 s was not always enough for the task to traverse `retrieve_whoami → set_timezone → stop_unit` and reach the `AddMatch` round-trip inside `wait_for_active_state()`. When the signal was emitted before the subscription was installed, it was broadcast to no subscriber and dropped; `wait_for_signal()` then blocked forever until pytest-timeout killed the test at 10 s.

This replaces the timing-based sleep with a deterministic handshake. The `SystemdUnit` D-Bus mock now exposes an `active_state_read` `asyncio.Event`, set whenever its `ActiveState` property is queried. Inside `wait_for_active_state()`, `ActiveState` is read only after the `PropertiesChanged` subscription has been installed, so awaiting that event before emitting guarantees the client is subscribed and the signal cannot be lost to a race. The event is a reusable synchronization primitive for any test that drives `wait_for_active_state()` this way.

Verified under full CPU load: the old sleep-based sync point timed out 10/10 runs, while the event-based version passed 30/30.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/tests/dbus_service_mocks/systemd_unit.py
+++ b/tests/dbus_service_mocks/systemd_unit.py
@@ -1,5 +1,7 @@
 """Mock of systemd unit dbus service."""
 
+import asyncio
+
 from dbus_fast.service import PropertyAccess, dbus_property
 
 from .base import DBusServiceMock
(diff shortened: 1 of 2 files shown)


mdegat01 (approved):
Makes sense. Wish we could generalize this somehow so you could attach an event to any property read like this if you needed that. But this fixes the issue in a simple way so all good as is

PR #7071, opened 2026-07-21
Reduce app config schema warning noise for regular users

agners:
## Proposed change

App config deprecation and misconfiguration advisories (e.g. "App config uses deprecated ..., please report this to the maintainer") were logged at warning level from the app config schema. Because that schema validates *every* app in *every* store repository on each store reload — not just installed apps — regular users saw a stream of "Please report this to the maintainer" warnings for apps they had never installed and could not fix. The wording is also backwards for that audience: the person reading Supervisor logs is almost never the app's maintainer.

This change routes those advisories through an injected logging callable so the store data layer can pick the level that matches who can act on the message:

- Apps from the **local** repository (authored by the user) keep logging at **warning** level.
- Any app validated while on the **dev** channel (a developer testing store apps) keeps logging at **warning** level.
- Everything else logs at **debug**, keeping the messages available for troubleshooting without spamming regular users' logs.

Installed apps, which are re-validated from `apps.json` on every boot, now also log these at debug level, since they were already migrated and their advisories logged when first read from the store.

The messages are also reworded to state the deprecation or misconfiguration as a fact rather than asking the reader to report it to the maintainer, since the remaining warning-level audiences (local authors and dev-channel testers) are the ones who can address it directly. The hard rejection for a dynamic ingress port conflict keeps its existing wording, as it aborts loading regardless of channel and is surfaced to the user directly.

No behavior other than log level and message wording changes: the migrations and validations themselves are unchanged.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/store/data.py
+++ b/supervisor/store/data.py
@@ -189,16 +189,23 @@ class StoreData(CoreSysAttributes):
         if not (app_config_list := await self._find_app_configs(path, repository)):
             return {}
 
-        # Deprecation/misconfiguration advisories are only actionable for a
-        # local app's author or a developer testing on the dev channel. Log them
-        # as warnings in those cases and keep them at debug level otherwise, so
-        # regular users don't see warnings for apps they cannot fix (including
-        # apps they haven't installed).
-        verbose = (
+        # App config deprecation/misconfiguration advisories are only actionable
+        # for someone who can fix the app, so only those cases log at warning
+        # level; everything else stays at debug so regular users don't see
+        # warnings for apps they cannot fix. Warn for the local repository
+        # (authored by the user), on the dev channel (a developer testing store
+        # apps), and for installed apps from custom repositories - the latter
+        # are often thin wrappers whose maintainer may have moved on, where the
+        # warning is the user's only heads-up before it breaks. Uninstalled apps
+        # and apps from the curated built-in stores (which get fixed via PRs)
+        # stay quiet.
+        repository_obj = self.sys_store.repositories.get(repository)
+        repository_is_builtin = bool(repository_obj and repository_obj.is_builtin)
+        always_verbose = (
             repository == REPOSITORY_LOCAL
             or self.sys_updater.channel == UpdateChannel.DEV
         )
-        schema = SCHEMA_APP_CONFIG if verbose else SCHEMA_APP_CONFIG_QUIET
+        installed_slugs = {app.slug for app in self.sys_apps.installed}
 
         def _process_apps_config() -> dict[str, dict[str, Any]]:
             apps: dict[str, dict[str, Any]] = {}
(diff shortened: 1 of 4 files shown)


supervisor/store/data.py
```diff
@@ -184,6 +189,17 @@ async def _read_apps_folder(
...
 
+        # Deprecation/misconfiguration advisories are only actionable for a
+        # local app's author or a developer testing on the dev channel. Log them
+        # as warnings in those cases and keep them at debug level otherwise, so
+        # regular users don't see warnings for apps they cannot fix (including
+        # apps they haven't installed).
+        verbose = (
+            repository == REPOSITORY_LOCAL
+            or self.sys_updater.channel == UpdateChannel.DEV
+        )
```
mdegat01: I think we need to add "or is installed" to this check.

The problem we have is there's a lot of apps out there which are simply repackaged versions of apps that are published as standalone docker containers. I can search on the forum for examples if you want but I can also pull 2 from my personal deployment: [signal messenger](https://github.com/haberda/signal-addon) and [caddy 2](https://github.com/einschmidt/app-caddy-2).

These apps are very clearly just as thin shells as possible around a working containerized application to allow it to be deployed as an app. Every time the wrapped application updates they immediately update with the new version, there are no other updates really. I think this is fairly common, I've definitely seen other apps like this.

In cases like this I expect no one will see these warnings. The actual developers of the service do not maintain or support the HA app. And the developers of the HA apps are effectively just users that were willing to put in some extra effort to make a service they wanted to use easy to deploy in their personal HA setup and then shared that with the world.

There's also app developers out there who have just moved on. They made something useful but they don't personally have a need for it anymore. So it sits available for people but unmaintained until it no longer works. This removes the small heads up we have for users in this situation.

We should absolutely silence these warnings for uninstalled store app. That is definitely just unwelcome noise that makes users gloss over all of them. But I think unless we want to build in a way to try and contact app developers directly for these issues we shouldn't suppress all warnings for installed apps. It seems very plausible that all users of some apps are normal users on production systems and this would just increase the likelihood of no one noticing an issue until it actually broke..

mdegat01 (changes requested):
I think suppressing the warning for installed apps is not a good idea for the reasons below. I am very much on board with this idea when it comes to uninstalled store pps and I like the dev check. But suppressing it for installed apps seems like a mistake to me.

supervisor/store/data.py
agners: On the other hand, we introduce config changes and do not apply it to our app repos timely! Either we make deprecation staged (e.g. show the message for three months just to dev channel users or something), so we even have a chance to silence warnings before they hit users, or we exclude the pre-installed app repositories... Either way is fine by me, but just adding `or is installed` still creates too much noise for regular users.

agners: I've implemented the second approach (exclude the pre-installed app store repositories) in 07c229c9c3d3866dfe5434481cd81c335117d8ad.

mdegat01: 🙈 

You make a good point. Maybe we need a `needs-apps` label to help us track when changes we make here require updates to our own apps.

But ok I like this approach. Logs from the built-in apps also qualifies as noise even if installed 👍

mdegat01 (approved):
LGTM 👍

Issue #7037, opened 2026-07-13
Media NFS mount reports active but Frigate writes to local ext4 after HAOS reboot/update


sean797:
### Describe the issue you are experiencing

After a HAOS update/reboot, Frigate records to the local VM disk instead of the configured NFS media mount. Supervisor reports the mount as `active`, but `/media/frigate` inside the addon is `ext4` (~50 GB VM disk), not `nfs4`.

This happened twice on the same system (2026-06-25 and 2026-07-10). On 2026-07-10, following a HAOS update, ~21.9 GB accumulated locally and the VM disk dropped to ~8% free before Frigate was stopped manually.

Related locked issues (same underlying failure): #5471, #4358, #4643.

### What type of installation are you running?

Home Assistant OS

### Which operating system are you running on?

Home Assistant Operating System (OVA / KVM)

### Network storage configuration

| Field | Value |
|-------|-------|
| Name | `frigate` |
| Usage | `media` |
| Type | NFS |
| Server | `192.168.0.180:/haos/frigate` |
| user_path | `/media/frigate` |
| Supervisor state | `active` |

Addon: **Frigate (Full Access)** (`ccab4aaf_frigate-fa`), `map: ["media:rw"]`, default paths under `/media/frigate`.

### Steps to reproduce

1. Configure NFS media mount named `frigate` (usage `media`) per [Frigate HA network storage docs](https://docs.frigate.video/guides/ha_network_storage/).
2. Run Frigate with recordings enabled.
3. Reboot HA or update HAOS (regression observed after HAOS 18.0 → 18.1 on **2026-07-10**).
4. Frigate starts (via delayed boot automation; see below).
5. Check storage inside Frigate (`/api/stats` → `service.storage`):
   - **Broken:** `/media/frigate/recordings` shows `mount_type: ext4`, total ~50 GB
   - **Expected:** `mount_type: nfs4`, total matches NAS capacity

### Observed failure mode

- `ha mounts info` shows `frigate` mount `state: active`.
- Host/SSH view of `/media/frigate` is local `ext4`, not NFS.
- Frigate writes recordings locally until the VM disk fills.
- No obvious Supervisor error in recent logs at time of failure (bind appears to fail silently or stale local directory blocks re-bind).

### Mitigation already tried (did not prevent recurrence)

- Frigate **Start on boot: disabled** (`boot: manual`)
[...]

PR #7013, opened 2026-07-08
Fix media/share mounts staying empty after mount reload fixup

agners:
## Proposed change

When a media/share network mount fails at runtime (e.g. the NFS server goes down), the reload fixup or the periodic mount reload escalates to stopping/restarting the data mount unit. systemd adds an implicit `RequiresMountsFor=` dependency from the bind mount unit to the data mount unit through its `What=` path (resulting in `Requires=` + `After=`, see `mount_add_mount_dependencies()` in systemd's `src/core/mount.c`), so the explicit stop/restart of the data mount unmounts the bind mount into media/share as well. Supervisor's `BoundMount` bookkeeping doesn't reflect that: after a successful reload, `MountManager.reload_mount()` only re-created the bind mount when it had been set up as an emergency fallback during load. The result is the sequence reported in #6938: the repair succeeds and the data mount is active again, but `/media/<name>` (resp. `/share/<name>`) stays an empty directory until the user re-sends the mount config via the API, which recreates all units.

Always re-create the bind mount after a successful reload instead. This also covers the case where the mount recovered on its own (probe passes, so the systemd reload is skipped) while the bind mount had already been torn down by an earlier failed restart cycle. `_bind_mount()` already handles unmounting a present bind mount and the emergency fallback, so the unconditional rebind is safe; and since `reload_mount()` only runs when the mount was failing (the periodic reload only calls it for probe failures, and the fixup only exists while a `MOUNT_FAILED` issue is present), the brief unmount/remount window doesn't disturb a healthy share.

## Type of change

- [ ] Dependency upgrade
- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (which adds functionality to the supervisor)
- [ ] Breaking change (fix/feature causing existing functionality to break)
- [ ] Code quality improvements to existing code or addition of tests

## Additional information

- This PR fixes or closes issue: fixes #6938
- This PR is related to issue: #6827
- Link to documentation pull request:
- Link to cli pull request:
- Link to client library pull request:

## Checklist

- [ ] The code change is tested and works locally.
- [x] Local tests pass. **Your PR cannot be merged unless tests pass**
- [x] There is no commented out code in this PR.
- [ ] I have followed the [development checklist][dev-checklist]
- [x] The code has been formatted using Ruff (`ruff format supervisor tests`)
- [x] Tests have been added to verify that the new code works.

If API endpoints or add-on configuration are added/changed:

- [ ] Documentation added/updated for [developers.home-assistant.io][docs-repository]
[...]

Issue #6938, opened 2026-06-13
Reload NFS mount - nothing happens


FredrikM97:
### Describe the issue you are experiencing

Hi! #6827 fixed a lot of the issues with  NFS mounts but I have noticed that using the reload on "network storage device failure" does not trigger the server to be considered online. In other words if I go into media, backup or share it is still empty even if home Assistant recognize it as repaired. Need to go into the shötorage and update the NFS mount (without changing anything) for it to properly load again.

Any way to a load a service call to reload it or make it a bit smarter to check if mount is available when an integration or app tries to access it? Alternatively make the reload trigger a check and not mark it as repaired until the mount responds?

### What type of installation are you running?

Home Assistant OS

### Which operating system are you running on?

Home Assistant Operating System

### Steps to reproduce the issue

1. Create NFS mount
2. Turn of server and validate media or backup no longer accessible
3. See the repair notification and repair
4. Check that media or backup files still not visible
5. Go into storage and select the mount and press update without changing anything
6. Now the media or backup files should be visible again 
...

### Anything in the Supervisor logs that might be useful for us?

```txt
Can't see anything
```

## System Information

version | core-2026.6.0
-- | --
installation_type | Home Assistant OS
dev | false
hassio | true
docker | true
container_arch | aarch64
user | root
virtualenv | false
python_version | 3.14.5
os_name | Linux
os_version | 6.12.75-haos-raspi
arch | aarch64
timezone | Europe/Stockholm
config_dir | /config

<details><summary>Home Assistant Community Store</summary>

GitHub API | ok
-- | --
GitHub Content | ok
GitHub Web | ok
HACS Data | ok
GitHub API Calls Remaining | 5000
Installed Version | 2.0.5
Stage | running
Available Repositories | 3124
Downloaded Repositories | 17

</details>

<details><summary>Home Assistant Cloud</summary>

logged_in | false
-- | --
can_reach_cert_server | ok
can_reach_cloud_auth | ok
can_reach_cloud | ok

</details>

<details><summary>Home Assistant Supervisor</summary>

host_os | Home Assistant OS 17.3
-- | --
update_channel | stable
supervisor_version | supervisor-2026.06.1
agent_version | 1.8.1
docker_version | 29.3.1
disk_total | 457.7 GB
disk_used | 33.9 GB
nameservers | 
healthy | true
supported | true
host_connectivity | true
supervisor_connectivity | true
ntp_synchronized | true
virtualization | 
board | rpi5-64
[...]

PR #7072, opened 2026-07-21
Check for local data before adding or updating a mount

agners:
## Proposed change

Adding or updating a mount whose target directory already contains local data only fails at the bind-mount step, via the non-empty check in `Mount.mount()`. This is a common situation: an add-on may have written into `/media/<name>` before network storage was set up (e.g. Frigate records to `/media/frigate` by default, and the mount created later per its network storage guide collides with the existing recordings — see the Frigate group of `MountInvalidError` events on Sentry), or data was written while no bind mount was in place after systemd tore it down together with a stopped/restarted data mount (#7013).

Failing that late leaves a half-created mount behind: the data mount unit is already mounted and the mount is registered in `MountManager._mounts`, but the config is never persisted, so the mount shows up in the API until the next Supervisor restart and then silently disappears while its transient unit lives on. Updating an existing mount in that state is worse: the working data mount is unmounted first (`remove_mount(retain_entry=True)`), only for the update to fail on the non-empty bind target afterwards.

This PR checks the mount's target directories (the data mount target, plus the media/share bind target where applicable) for local data upfront and fails the add/update with a clear message before anything is mounted or unmounted. Paths that are already mount points are skipped — they get unmounted before reuse — and paths that cannot be inspected (e.g. an unreachable network mount) are left to the authoritative mount-time check.

Should mounting still fail halfway (e.g. the bind mount unit fails to start), `create_mount()` now rolls back by unmounting the units it created instead of leaving the half-created mount in the list: registration in `_mounts` happens only after the data mount and bind mount both succeeded.

The target-directory errors are also made translatable following the established `error_key`/`message_template`/`extra_fields` pattern: `MountTargetNotEmptyError` (`mount_target_not_empty_error`) is raised by both the new upfront check and the mount-time non-empty check, and `MountTargetNotDirectoryError` (`mount_target_not_directory_error`) covers the target-is-not-a-directory case. Both remain subclasses of `MountInvalidError`, so existing exception handling is unaffected, and `extra_fields` carries `name` and `path` so the frontend can localize the message and display the user-facing path. Frontends without the translation strings fall back to the English message as before.

Also fixes the debug log statement in `create_mount()` which referenced `%s` without passing the mount name.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR is related to issue: #7037, #6938
- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/exceptions.py
+++ b/supervisor/exceptions.py
@@ -1454,6 +1454,37 @@ class MountInvalidError(MountError):
     """Raise on invalid mount attempt."""
 
 
+class MountTargetNotDirectoryError(MountInvalidError):
+    """Raise when a mount target exists but is not a directory."""
+
+    error_key = "mount_target_not_directory_error"
+    message_template = "Cannot mount {name} at {path} as it is not a directory"
+
+    def __init__(
+        self, logger: Callable[..., None] | None = None, *, name: str, path: str
+    ) -> None:
+        """Initialize exception."""
+        self.extra_fields = {"name": name, "path": path}
+        super().__init__(None, logger)
+
+
+class MountTargetNotEmptyError(MountInvalidError):
+    """Raise when a mount target directory contains existing data."""
+
+    error_key = "mount_target_not_empty_error"
+    message_template = (
+        "Cannot mount {name} because there is existing data at {path}. "
+        "Move it away first, then retry"
+    )
+
+    def __init__(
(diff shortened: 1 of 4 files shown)


mdegat01 (changes requested):
This looks good but I think copilot is right. Looks like you made two types of errors and one got skipped in its use case.

mdegat01 (approved):
Looks good, LGTM 👍

PR #7084, opened 2026-07-24
Fix CI race between Supervisor restart and backup restore

agners:
## Proposed change

The "Restore SSH app from backup" step in the build workflow occasionally fails with an empty CLI response, e.g. in https://github.com/home-assistant/supervisor/actions/runs/30083369091.

The restore request is in fact always sent to the old Supervisor instance: `ha supervisor restart` returns before the Supervisor actually shuts down, and the subsequent "Wait for Supervisor to come up" step polls `/supervisor/ping`, which the old instance still answers. The workflow therefore proceeds immediately without ever observing the restart. Whenever the restore (including waiting for the SSH add-on to start) does not complete within Core's 10 second stage 1 shutdown window, the API server is torn down mid-request and the CLI sees EOF, making the step flaky.

Fix this by recording the Supervisor container's `StartedAt` timestamp before issuing the restart and waiting until it changes, so the reused "Wait for Supervisor to come up" step (and everything after it) is guaranteed to talk to the new Supervisor instance.

## Type of change

- [ ] Dependency upgrade
- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (which adds functionality to the supervisor)
- [ ] Breaking change (fix/feature causing existing functionality to break)
- [x] Code quality improvements to existing code or addition of tests

## Additional information

- This PR fixes or closes issue: fixes #
- This PR is related to issue:
- Link to documentation pull request:
- Link to cli pull request:
- Link to client library pull request:

## Checklist

- [ ] The code change is tested and works locally.
- [ ] Local tests pass. **Your PR cannot be merged unless tests pass**
- [x] There is no commented out code in this PR.
- [ ] I have followed the [development checklist][dev-checklist]
- [ ] The code has been formatted using Ruff (`ruff format supervisor tests`)
- [ ] Tests have been added to verify that the new code works.

If API endpoints or add-on configuration are added/changed:

- [ ] Documentation added/updated for [developers.home-assistant.io][docs-repository]
- [ ] [CLI][cli-repository] updated (if necessary)
- [ ] [Client library][client-library-repository] updated (if necessary)

[dev-checklist]: https://developers.home-assistant.io/docs/en/development_checklist.html
[docs-repository]: https://github.com/home-assistant/developers.home-assistant
[cli-repository]: https://github.com/home-assistant/cli
[client-library-repository]: https://github.com/home-assistant-libs/python-supervisor-client/

PR #7085, opened 2026-07-24
Enter STOPPING state before Supervisor restart API response

agners:
## Proposed change

The Supervisor accepts new API requests in the window after responding to `/supervisor/restart`: `Supervisor.restart()` schedules `Core.stop()` as a task and returns immediately, so the state only changes to `stopping` after the restart response has been sent. A request accepted in that window keeps running while the Supervisor shuts down, and when `Core.stop()` tears down the API server after the 10 second stage 1 timeout the connection is dropped without a response — the client sees EOF instead of an error. This is what made the CI restore step flaky in https://github.com/home-assistant/supervisor/actions/runs/30083369091 (see #7084 for a CI-side fix attempt): the restore request landed on the old, dying instance and got killed mid-request.

Make `Supervisor.restart()` transition the core state to `STOPPING` before returning, via a new `Core.begin_stop()` which contains the transition part previously at the start of `Core.stop()`. The system validation middleware already rejects all requests outside of `STARTUP`/`RUNNING`/`FREEZE`, so any request arriving after the restart response now deterministically gets a clear "System is not ready with state: stopping" error from the old instance instead of possibly being accepted and killed. Since the state can now already be `STOPPING` when `Core.stop()` runs, its re-entry guard is changed from a state check to an explicit flag.

`Supervisor.update()` schedules the same stop task but is deliberately left unchanged: entering `STOPPING` before `update()` returns would suppress the final job progress event to Home Assistant (WebSocket messages are dropped in `CLOSING_STATES`), regressing update progress reporting. Closing that window needs the final job event to be ordered before the state transition and is left for a follow-up.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/core.py
+++ b/supervisor/core.py
@@ -343,7 +343,13 @@ class Core(CoreSysAttributes):
             return
         self._stop_initiated = True
 
-        await self.begin_stop()
+        try:
+            await self.begin_stop()
+        except BaseException:
+            # Nothing has been torn down yet at this point, so let a later
+            # stop() retry if the transition failed or was cancelled
+            self._stop_initiated = False
+            raise
 
         # Stage 1
         try:
--- a/supervisor/core.py
+++ b/supervisor/core.py
@@ -44,6 +44,7 @@ class Core(CoreSysAttributes):
         self._state: CoreState = CoreState.INITIALIZE
         self.exit_code: int = 0
         self._shutdown_event: asyncio.Event = asyncio.Event()
+        self._stop_initiated: bool = False
 
     @property
     def state(self) -> CoreState:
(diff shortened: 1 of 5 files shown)


agners:
This essentially fixes #7084 more elegant in my opinion.

Note there is another race condition: Once ping succeeds, the Supervisor can be in `STARTUP`. However, backup restore only works in `RUNNING`. But it seems this isn't that problematic, at least I haven't seen such failure. To fix this, we'd need to poll for state after the ping succeeds, before triggering the restore. Anyhow, if this ever comes up, the error in the restore attempt will quite clearly point at this race, making it trivial to fix then.

supervisor/core.py
```diff
@@ -315,17 +316,41 @@ async def start(self) -> None:
...
+            return
+
         # store new last boot / prevent time adjustments
         if self.state in (CoreState.RUNNING, CoreState.SHUTDOWN):
             await self._update_last_boot()
-        if self.state in (CoreState.STOPPING, CoreState.CLOSE):
-            return
 
         # don't process scheduler anymore
         await self.set_state(CoreState.STOPPING)
```
mdegat01, line 337: So at its core this is really a consequence of our blockbuster efforts right? Prior to that I believe this code was sync because we likely wrote the run state in the event loop:
https://github.com/home-assistant/supervisor/blob/f4ec256b94936ba98cbaad35e0ac87c032b5a1b8/supervisor/core.py#L79-L96

If so is this extra infrastructure a better answer then returning `set_state` to sync code? We can do that by having it return the task for writing the run state and telling callers they have to await it and handle errors or with a fire and forget task pattern. Although in this case we'd probably want to set up a waiter that listens for and handles errors properly with some kind of retry because we don't want disk out of sync with supervisor.

I'm fine with your approach if you want. I'm just now wondering if we have any other sneaky bugs created by making `set_state` async instead of sync and separately scheduled.

mdegat01 (approved):
Approved and good catch. But check my comment, I suggested an alternate proposal if you prefer it.

supervisor/core.py
agners: While related, this appears a bit orthogonal: Even back when `state` was a sync property setter, the stopping was a task and racy. So no, this is not a consequence of the blockbuster efforts, this needs rework either way.

Making `set_state` handing set state sync and returning a task for the write would make the `begin_stop` a bit more efficient, but doesn't fundamentally shape this PR. I'd prefer to do any set_state changes in a follow up, since it will affect more code then just the stopping.

PR #7087, opened 2026-07-24
Return a proper API error when adding a duplicate store repository

agners:
## Proposed change

Adding a store repository that is already in the store raises a plain `StoreError`. Since that is not an `APIError`, the `api_process` decorator logs it as "Unexpected error during API call" and reports it to Sentry, where it is one of the most frequent issues (SUPERVISOR-1JYE, ~8100 affected installations in the last 90 days). The events are ordinary client actions: add-on setup flows and documentation "add repository" links re-submitting a repository the user already has, plus third-party automation re-adding its repositories on every add-on start without checking the response.

Introduce `StoreRepositoryAlreadyAddedError`, which is both a `StoreError` and an `APIConflict` with an error key and message template, following the `ServiceAlreadyProvidedError` precedent. The request now fails with HTTP 409 Conflict (previously 400) and the response carries the structured error key and extra fields, and the error is no longer logged as unexpected or reported to Sentry.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR is related to issue: [SUPERVISOR-1JYE](https://home-assistant-io.sentry.io/issues/SUPERVISOR-1JYE) (auto-resolved by the commit message reference)
- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/exceptions.py
+++ b/supervisor/exceptions.py
@@ -1285,6 +1285,18 @@ class StoreAppNotFoundError(StoreError, APINotFound):
         super().__init__(None, logger)
 
 
+class StoreRepositoryAlreadyAddedError(StoreError, APIConflict):
+    """Raise when a repository is already added to the store."""
+
+    error_key = "store_repository_already_added_error"
+    message_template = "Can't add {url}, already in the store"
+
+    def __init__(self, logger: Callable[..., None] | None = None, *, url: str) -> None:
+        """Initialize exception."""
+        self.extra_fields = {"url": url}
+        super().__init__(None, logger)
+
+
 class StoreRepositoryLocalCannotReset(StoreError, APIError):
     """Raise if user requests a reset on the local app repository."""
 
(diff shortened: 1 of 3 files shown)


agners:
This is technically a breaking change since we change from 400 to 409, but the frontend appears to handle 409 just as well.

mdegat01 (approved):
I wasn't sure if 409 was actually a better fit then 400 to be worth a small breaking change but claude does agree. Although it also suggested just simply making it a no-op success instead if its idempotent, which it is in this case. Which seems fine and removes the error without a tiny breaking change. I'll leave it up to you, either is fine by me.

agners:
> Although it also suggested just simply making it a no-op success instead if its idempotent, which it is in this case. Which seems fine and removes the error without a tiny breaking change. I'll leave it up to you, either is fine by me.

Just accepting/no-op seems weird to me from a UX point of view: It feels a bit unexpected to add a repository, press safe, and the no new line is being added (since it already exists). I prefer the 409 approach.

PR #7089, opened 2026-07-24
Add repair to move local data blocking a mount target

agners:
## Proposed change

When an add-on writes into a media/share directory while its network mount is not in place — e.g. Frigate recording to `/media/frigate` before network storage was set up, or after systemd tore down the bind mount together with a stopped/restarted data mount (#7013) — the local data blocks re-creating the mount, since mounting over a non-empty directory is refused (#7072). Until now this failed silently at Supervisor startup: the bind mounts were created as fire-and-forget tasks, so the data mount reported `active` while `/media/<name>` stayed a local directory, and the only way out was removing the data manually over SSH/Samba and re-creating the mount via the API (see #7037).

This PR surfaces the condition as a new `mount_target_not_empty` issue with a new `move_local_data` suggestion (plus the existing `execute_remove`). The fixup moves the blocking data to a `<name>_local_recovery` folder in a user-accessible location instead of deleting it, then reloads the mount:

- **media/share mounts**: the recovery folder is created next to the mount in `/media` resp. `/share`, so users can inspect and clean it up via the media browser, the Samba shares or add-ons.
- **backup mounts**: their data mount directory (`/mnt/data/supervisor/mounts/<name>`) is not reachable for users, so the data is moved to local backup storage instead, where stranded backup files are visible and manageable.

The relocation is inherently safe against touching share content: paths that are mount points are skipped, and renaming a live mount point or moving data across filesystems fails (EBUSY/EXDEV) rather than misplacing data. Name collisions with an existing recovery folder get a numeric suffix.

Detection covers all paths where the condition can appear: bind mount failures during `load()` are now awaited and routed into resolution issues instead of being swallowed as fire-and-forget tasks (blocking local data raises the new issue; other bind failures create the existing `mount_failed` issue), the periodic mount reload picks it up through the same error routing, and the reload fixup raises it directly when re-creating the bind mount fails on local data. A successful mount reload dismisses a stale `mount_target_not_empty` issue.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR is related to issue: #7037, #7013, #7072
- Link to cli pull request:

## Checklist


agners:
Since we have already a repair for the failing mount, can't we just have it as part of that repair? Either we make a new repair "mount failed/data move needed" alongside "mount failed", but only raise one of the two.

Actually: We can have a single fixup with another suggestion, one which moves data and one which just retries (in case the user wants to resolve on its own).

supervisor/mounts/manager.py
```diff
@@ -175,12 +175,15 @@ async def reload(self) -> None:
...
-        """Await a list of tasks on mounts and turn each error into a failed mount issue."""
+        """Await a list of tasks on mounts and turn each error into a resolution issue."""
         errors = await asyncio.gather(*mount_tasks, return_exceptions=True)
 
         for i in range(len(errors)):  # pylint: disable=consider-using-enumerate
             if not (err := errors[i]):
                 continue
+            if isinstance(err, MountTargetNotEmptyError | MountTargetNotDirectoryError):
+                self._add_local_data_issue(mounts[i].name)
+                continue
```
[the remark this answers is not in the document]
agners: A file is very unusual, so I think we can accept that gap.

supervisor/mounts/manager.py
```diff
@@ -194,6 +196,24 @@ async def _mount_errors_to_issues(
...
+        Uses the same mount failed issue as other mount failures so at most
+        one issue exists per mount, with an additional suggestion to move the
+        blocking data aside. Reload stays available for users who prefer to
+        clear the data themselves. Adding is idempotent: an existing issue
+        just gains the extra suggestion.
+        """
+        self.sys_resolution.add_issue(
+            replace(mount.failed_issue),
+            suggestions=[
+                SuggestionType.MOVE_LOCAL_DATA,
```
agners, line 211: Unfortunately, this renders as an empty option which is quite confusing (since translations for the issue and the two existing suggestions already exist).

Best thing we can do is filter the suggestion based on Core version to avoid confusing entries on Core side 😢 .

agners: Solved by 8605c8b9316f16412255cf89c26f19f6eae938a1

supervisor/api/resolution.py
```diff
@@ -86,12 +97,12 @@ def _build_info_response(self) -> dict[str, Any]:
     @api_process
     async def info(self, request: web.Request) -> dict[str, Any]:
         """Return resolution information."""
-        return self._build_info_response()
+        return self._build_info_response(request)
```
mdegat01: Its fine to leave this since it may come up again. But technically we don't need this right now, no current version of core supports v2 anyway

mdegat01: Actually if we commit to putting the "supported suggestions for issue" concept in core soon we may never need this since any version of core that supports v2 will have its own suggestions filtering in place.

supervisor/mounts/manager.py
```diff
@@ -323,7 +343,103 @@ async def reload_mount(self, name: str) -> None:
...
+                    if path.is_dir() and not any(path.iterdir()):
+                        continue
+                except OSError:
+                    continue
+
+                target = recovery_base / f"{name}_local_recovery"
+                counter = 1
+                while target.exists():
+                    counter += 1
+                    target = recovery_base / f"{name}_local_recovery_{counter}"
```
mdegat01: So the odd part of doing this in the loop to me is that in the case of share or media it may make 2 of these to fix one issue. If there is something in both `mount.local_where` and something in `{share/media}/{mount.name}` then they will be moved to two separate `local_recovery` folders. I guess that's ok and maybe its too rare to matter? It feels like it would be nicer to me to move them both into one `local_recovery` folder to connect them as one fix.

supervisor/mounts/manager.py
```diff
@@ -323,7 +343,103 @@ async def reload_mount(self, name: str) -> None:
...
+                # Keep the path present for consumers even if the remount
+                # below fails: an empty directory instead of a missing one
+                path.mkdir()
+                moved.append((path, target))
+            return moved
+
+        try:
+            moved = await self.sys_run_in_executor(move_aside)
+        except OSError as err:
+            raise MountError(
```
mdegat01, line 429:
```suggestion
            self.sys_resolution.check_oserror(err)
            raise MountError(
```
Minor but as these are OSErrors raised from an interaction with the local data disk we should insert our OSError helper to look for bad message.

supervisor/resolution/module.py
```diff
@@ -198,6 +200,27 @@ def _make_issue_message(self, issue: Issue) -> dict[str, Any]:
...
+        return [
+            suggestion
+            for suggestion in suggestions
+            if (min_version := SUGGESTION_MIN_CORE_VERSION.get(suggestion.type)) is None
+            or (
+                version is not None
+                and version != LANDINGPAGE
+                and version_is_new_enough(version, min_version)
+            )
+        ]
```
mdegat01, line 223: If we want add core side filtering as discussed on the call then it may not be worth additional effort here. But for reference, the main reason I didn't add filtering on suggestions core side is because suggestions are really only sensible in the context of an issue.

For example, we use a suggestion with type `execute_start` and context `app` for issues of both type `boot_fail` and `app_port_conflict`. It's entirely possible that a version of core has a translation for this suggestion type for one issue but not for the other as translation keys are organized by issue type, not suggestion type. So this helper should perhaps accept both an Issue and a Suggestion and filter out suggestions when the combo does not match, rather then only on suggestion.

But we can also leave it as is and just commit to a core change. In which case we probably never need that capability here.

mdegat01 (changes requested):
Looks good. Only real thing that needs fixing is the addition of the oserror helper to look for bad message. The others are fine by me if you prefer them as is.

supervisor/resolution/module.py
agners: Past Core versions already pick all suggestions up, and render them without strings. We can't change old Core versions, they area already out and will be paired with new Supervisor versions. That is why we need the filter on Supervisor side.

agners: > But we can also leave it as is and just commit to a core change. In which case we probably never need that capability here.

Ah you meant for future issues/suggestion filtering. This would then work for new issues, so we don't need Supervisor filtering anymore 🤔 .

Yeah let's keep things simple here, at least until more sophisticated filtering is actually required.

agners: Core PR implementing this: https://github.com/home-assistant/core/pull/179540

mdegat01: sounds good to me 👍

supervisor/api/resolution.py
mdegat01: Since the mentioned core PR is now up should we drop this and add a v2 handler for "suggestions for issue" without it? No version of core currently supports v2 and any future version that does will have its own supported suggestions filtering.

mdegat01 (approved):
Looks good. Do you want to drop the filtering from v2 for the reasons I said or leave it? I'm fine either way.

PR #7092, opened 2026-07-27
Respect user-overridden Core image for install and update

agners:
## Proposed change

The `override_image` setting (set when a user configures a custom Core `image` via `/core/options`) was only honored by `load()`. The landingpage install, the initial Core install and Core updates always passed `image=self.sys_updater.image_homeassistant` to the Docker layer and wrote that image back to the Home Assistant config afterwards, discarding the user override on the next install or update.

This adds a `HomeAssistant.install_image` property that returns the user-overridden image when `override_image` is set, and the image from the update information otherwise. The landingpage install, Core install and Core update paths now use it for the image pull and persist that same image afterwards, so the override survives installs and updates. Version resolution is unchanged and still comes from the update information.

Behavior note: with an override set, a fresh install now pulls `<image>:landingpage` from the overridden image as well (previously the official landingpage image was used). An overridden image therefore needs a `landingpage` tag on the registry or a matching preinstalled local image for the fresh install flow. Since `override_image` is an explicit opt-in, the override is applied consistently to all image pulls.

As discussed in the PR, running a non-default Core image now marks the system as unsupported (`home_assistant_core_custom_image` reason). The evaluation compares the configured image against the default image for the machine rather than checking the `override_image` flag, so it also catches non-default images recorded through container adoption or manual configuration edits where the flag is not set.

Rollback note (raised in review): `update()` uses the currently configured install image for a rollback as well. If the image override is changed and a subsequent update fails, the rollback pulls the previous version on the newly configured image, not the image/version combination that was running before the update. Given this is a developer-focused feature, this behavior is accepted and documented here.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/homeassistant/core.py
+++ b/supervisor/homeassistant/core.py
@@ -221,6 +221,7 @@ class HomeAssistantCore(JobGroup):
                         _LOGGER.info("Home Assistant Core installation in progress")
 
         progress_task = self.sys_create_task(_periodic_progress_log())
+        install_image: str | None = None
         try:
             while True:
                 # read homeassistant tag and install it
@@ -262,10 +263,8 @@ class HomeAssistantCore(JobGroup):
                         )
 
                 try:
-                    await self.instance.update(
-                        to_version,
-                        image=self.sys_homeassistant.install_image,
-                    )
+                    install_image = self.sys_homeassistant.install_image
+                    await self.instance.update(to_version, image=install_image)
                     self.sys_homeassistant.version = self.instance.version or to_version
                     break
                 except DockerError, JobException:
@@ -283,7 +282,7 @@ class HomeAssistantCore(JobGroup):
             await progress_task
 
         _LOGGER.info("Home Assistant docker now installed")
-        self.sys_homeassistant.set_image(self.sys_homeassistant.install_image)
+        self.sys_homeassistant.set_image(install_image)
         await self.sys_homeassistant.save_data()
 
         # finishing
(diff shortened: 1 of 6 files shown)


agners:
The current override_image does not seem very useful to me, but then this is hacky too since it forces an alternative image source to create releases in lockstep with our release cadence :shrug: .

agners:
This essentially makes it easier to run your own fork. You'll have to make sure the release is in lockstep with Core (rather easy if you track beta/stable).

We probably want to know if a user runs a Core fork. So yes, let's do this, but mark the system unsupported (which we should have probably done anyways if it is set even without the updateable wrinkle).

supervisor/homeassistant/core.py
```diff
@@ -337,6 +334,7 @@ async def update(
             )
 
         old_image = self.sys_homeassistant.image
+        install_image = self.sys_homeassistant.install_image
```
mdegat01, line 337: So I just realized looking at this that rollback can get quite strange. If you set the custom image and then do an update and that update fails its going to try and rollback to the old version on the new image, not actually the version and image you were running before.

This feels like a beyond niche scenario for a developer only feature so I don't think we have to do anything about it. But perhaps we should note that in a comment or PR description or with wherever the doc lives for this feature.

mdegat01 (approved):
Looks good. Small comment on rollback but not one I think we need to address in the PR. Just noting odd niche behavior we could consider documenting as well, similar to the potentially unexpected landing page behavior when using this feature.

supervisor/homeassistant/core.py
agners: I've updated the PR description.

PR #7101, opened 2026-07-30
Add websocket addon-to-app compatibility shim

### LINKED ISSUE #7031: WSType and WSEvent enums around addons
### Description

When emitting messages over websockets we use `WSType` to set the overarching type of the message. One of those is `WSType.HASSIO_UPDATE_ADDON`. If the type is `supervisor/event` then the `WSEvent` enum is used to describe the type of event. One of those is `WSEvent.ADDON`.
https://github.com/home-assistant/supervisor/blob/6b0e1b7023c8058e6788a2d29a704ecd2188bfce/supervisor/homeassistant/const.py#L29-L48

As this is websockets we can't use V1 vs V2 to handle compatibility, these will require same handling as #7014 :
1. Update core to treat the new and old value as the same
2. Mark the value as deprecated in supervisor and to be removed when the last version of core that doesn't understand the two are the same is no longer supported.

### Additional context

_No response_

mdegat01:
## Proposed change

This PR updates websocket metadata and app update command terminology from `addon` to `app`, while keeping V1 clients compatible through a single send-time shim in `HomeAssistantWebSocket`.

The compatibility layer is applied at the shared websocket send point, so:
- **V1** continues to receive legacy addon metadata
- **V2** receives the new app terminology unchanged

This also adds a tests-only Ruff override for `PLR0917` so pytest-style test functions with many fixtures are not flagged as lint errors.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR fixes or closes issue: fixes #7031
- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/homeassistant/const.py
+++ b/supervisor/homeassistant/const.py
@@ -33,13 +33,13 @@ class WSType(StrEnum):
     SUPERVISOR_EVENT = "supervisor/event"
     BACKUP_START = "backup/start"
     BACKUP_END = "backup/end"
-    HASSIO_UPDATE_ADDON = "hassio/update/addon"
+    HASSIO_UPDATE_APP = "hassio/update/app"
 
 
 class WSEvent(StrEnum):
     """Websocket events."""
 
-    ADDON = "addon"
+    APP = "app"
     HEALTH_CHANGED = "health_changed"
     ISSUE_CHANGED = "issue_changed"
     ISSUE_REMOVED = "issue_removed"
--- a/supervisor/homeassistant/websocket.py
+++ b/supervisor/homeassistant/websocket.py
@@ -19,6 +19,7 @@ from ..const import (
     STARTING_STATES,
     BusEvent,
     CoreState,
+    FeatureFlag,
 )
 from ..coresys import CoreSys, CoreSysAttributes
 from ..exceptions import (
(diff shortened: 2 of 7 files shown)


supervisor/homeassistant/const.py
```diff
@@ -33,13 +33,13 @@ class WSType(StrEnum):
     SUPERVISOR_EVENT = "supervisor/event"
     BACKUP_START = "backup/start"
     BACKUP_END = "backup/end"
-    HASSIO_UPDATE_ADDON = "hassio/update/addon"
+    HASSIO_UPDATE_APP = "hassio/update/app"
```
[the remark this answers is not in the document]
mdegat01: No one imports these symbols, Supervisor is not published to any package repository it is only intended to be used as a standalone service. And these values are not accepted inbound, they are outbound value. The issue is talking about an old implementation plan, the feature flag is the new one.

supervisor/homeassistant/const.py
```diff
@@ -33,13 +33,13 @@ class WSType(StrEnum):
...
 
 
 class WSEvent(StrEnum):
     """Websocket events."""
 
-    ADDON = "addon"
+    APP = "app"
     HEALTH_CHANGED = "health_changed"
     ISSUE_CHANGED = "issue_changed"
```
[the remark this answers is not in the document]
mdegat01: These events are outbound only. This feature flag approach is compatible and is an improvement upon the plan presented in that task we worked out over the past several PRs.

supervisor/misc/tasks.py
```diff
@@ -145,7 +145,7 @@ async def _update_apps(self):
             # Ultimately auto updates should be handled by Home Assistant Core itself
             # through a update entity feature.
             message = {
-                ATTR_TYPE: WSType.HASSIO_UPDATE_ADDON,
+                ATTR_TYPE: WSType.HASSIO_UPDATE_APP,
                 "addon": app.slug,
```
agners: I guess this key needs updating too?

supervisor/homeassistant/websocket.py
```diff
@@ -250,6 +251,27 @@ def __init__(self, coresys: CoreSys):
...
+        self, message: dict[str, Any]
+    ) -> dict[str, Any]:
+        """Map new WS metadata back to legacy names for v1 compatibility."""
+        if self.sys_config.feature_flags.get(
+            FeatureFlag.SUPERVISOR_WEBSOCKET_V2_API, False
+        ):
+            return message
+
+        if message.get(ATTR_TYPE) == WSType.HASSIO_UPDATE_APP:
+            return message.copy() | {ATTR_TYPE: "hassio/update/addon"}
```
agners: Supposedly `|` already leads to a copy, so the other copy is no necessary.

```suggestion
            return message | {ATTR_TYPE: "hassio/update/addon"}
```

supervisor/homeassistant/websocket.py
```diff
@@ -250,6 +251,27 @@ def __init__(self, coresys: CoreSys):
...
+
+        if (
+            message.get(ATTR_TYPE) == WSType.SUPERVISOR_EVENT
+            and (data := message.get(ATTR_DATA))
+            and data.get(ATTR_EVENT) == WSEvent.APP
+        ):
+            return message.copy() | {ATTR_DATA: data.copy() | {ATTR_EVENT: "addon"}}
```
agners:
```suggestion
            return message | {ATTR_DATA: data.copy() | {ATTR_EVENT: "addon"}}
```

agners (changes requested):

supervisor/misc/tasks.py
mdegat01: Good catch, I didn't notice that one 👍

supervisor/homeassistant/websocket.py
mdegat01: Same thing with `data.copy()` then, fixed both

agners (approved):
LGTM!

PR #7123, opened 2026-08-12
Block supervisor/hassio WebSocket command types in the HA proxy

mdegat01:
## Proposed change

The Supervisor's Core WebSocket proxy (`/core/websocket`) accepts connections from any app with `homeassistant_api: true`. It authenticates the app with its own token but then opens the upstream connection to Core **as the Supervisor itself**. `_proxy_message` was forwarding frames verbatim with no content inspection.

An app could exploit this by sending a `supervisor/api` command (or other `supervisor/*`/`hassio/*` type) through the proxy. Core's `hassio` integration handles those commands by calling back into the Supervisor with Core's own token. That token bypasses all role checks in `security.py` — it is treated as the most privileged caller regardless of the app's declared `hassio_api` or `hassio_role`.

**Fix:** in `_proxy_message`, TEXT frames on the app-to-Core direction whose `type` field starts with `"supervisor/"` or `"hassio/"` are now rejected with a Core-shaped `result`/`unauthorized` response instead of being forwarded. The connection stays open so the app can continue making normal HA calls (e.g. `call_service`, `subscribe_events`). The Core-to-app direction is unfiltered. Fail-open on unparseable frames — Core's own validation already rejects those.

Tests cover: the exact `supervisor/api` attack vector, all denied command-type namespaces, continued usability of the connection after a rejection, and pass-through of non-JSON frames.

## Type of change

## Additional information

- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/api/proxy.py
+++ b/supervisor/api/proxy.py
@@ -2,7 +2,7 @@
 
 import asyncio
 from collections.abc import AsyncIterator
-from contextlib import asynccontextmanager, suppress
+from contextlib import asynccontextmanager
 import logging
 import re
 from typing import Final
@@ -44,12 +44,15 @@ CORE_API_DENY: Final = re.compile(r"^hassio(?:/|_)")
 DENIED_WS_TYPE_PREFIXES = ("supervisor/", "hassio/")
 
 
-def _denied_command_type(data: str) -> str | None:
+def _denied_command_type(data: str) -> tuple[str, str | None] | None:
     """Return the command type if it's one apps must never reach, else None."""
     try:
-        command_type = json_loads(data).get("type")
+        parsed = json_loads(data)
+        command_type = parsed.get("type")
         return (
-            command_type if command_type.startswith(DENIED_WS_TYPE_PREFIXES) else None
+            (command_type, parsed.get("id"))
+            if command_type and command_type.startswith(DENIED_WS_TYPE_PREFIXES)
+            else None
         )
     except ValueError, AttributeError:
         # ValueError: data wasn't valid JSON.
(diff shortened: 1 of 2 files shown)


agners:
> it is treated as the most privileged caller regardless of the app's declared `hassio_api` or `hassio_role`.

Technically, these are Supervisor API roles and are not related with the Core API roles.

I also think we should share the list of endpoints with the REST API filter (`CORE_API_DENY`), if possible. Or at least mention their relationship.

Finally, there are more admin end points on Core side, like `config/auth/create`, `config/auth/delete` or`auth/long_lived_access_token`, should we block these two?

Maybe we need to first take a step back and ask ourselfs what our exact intention is. The app flag `homeassistant_api` does not specify what Core credentials the flag gives, from [App configuration](https://developers.home-assistant.io/docs/apps/configuration) docs:

> This app can access the Home Assistant REST API proxy. Use `http://supervisor/core/api`.

supervisor/api/proxy.py
```diff
@@ -199,11 +217,40 @@ async def _proxy_message(
...
+        If filter_app_commands is set, TEXT messages whose command type is reserved
+        for Home Assistant's own frontend (see DENIED_WS_TYPE_PREFIXES) are rejected
+        instead of forwarded, to prevent apps from using this proxy to reach the
+        Supervisor API at full privilege through Home Assistant Core.
+        """
         while not source.closed and not target.closed:
             msg = await source.receive()
             match msg.type:
+                case WSMsgType.TEXT if filter_app_commands and (
+                    denied_type := _denied_command_type(msg.data)
```
agners: The return value here is really meant as a cutesy for this function. But we then need another `json_loads()` to get the `message_id`. Let's just simply return the message_id here too, which saves us the extra hassle below.

agners (changes requested):
> Core's `hassio` integration handles those commands by calling back into the Supervisor with Core's own token. That token bypasses all role checks in `security.py` — it is treated as the most privileged caller regardless of the app's declared `hassio_api` or `hassio_role`.

Oh right, this is about protecting the Supervisor API, through means of Core API proxy. With that in mind, the filter seems correct and complete.

Other than the minor restructuring suggestion the PR LGTM.

supervisor/api/proxy.py
mdegat01: Yea adjusted, return a tuple with both to remove the extra unnecessary full parse

agners (approved):
LGTM!

PR #7131, opened 2026-08-13
Reserve Core's host port before booting pre-Core apps

### LINKED ISSUE #6790: Reserve the Core port to prevent add-on conflicts on reboot
Prevent add-ons from binding to the port that Core is configured to use. If an add-on starts before Core and takes the port, Core will fail to start on reboot. Supervisor should be aware of Core's configured port and disallow conflicting add-ons from using it.

mdegat01:
## Proposed change

On a fresh system boot, Home Assistant Core and apps with `startup: system` or `startup: services` start in sequence before Core. Core runs with `--network host` and binds directly in the host network namespace. If an app claims Core's HTTP port (default 8123) before Core starts, Core fails to start entirely, leaving the user with no UI and no way to fix the situation without terminal access.

This change reserves Core's port in the host network namespace before any pre-Core apps are started, so that a conflicting app is the one that fails (surfacing the existing `APP_PORT_CONFLICT` issue/suggestion) rather than Core.

Because Supervisor runs on the `hassio` bridge network — not the host network namespace — a simple `socket.bind()` in Supervisor's own Python process is invisible to the host. The reservation is instead implemented by asking the host's systemd (over D-Bus, already a hard Supervisor dependency) to create a transient `.socket` unit (`hassio-port-reserve.socket`) listening on every configured bind address for Core's port (defaulting to both `0.0.0.0` and `::` if unset). Systemd binds the socket itself as part of activating the unit — no process or container is involved, and waiting for the unit to become active guarantees the bind has already happened before boot continues. The socket is paired atomically with a trivial oneshot `.service` unit (`hassio-port-reserve.service`, kept "active (exited)" via `RemainAfterExit`) so systemd always has a real activation target — without one, the first incoming connection attempt during boot would make systemd fail to activate a (nonexistent) service and tear the socket back down, silently releasing the reservation.

Both units are stopped right before Core is started so Core can claim the port itself. This release is attempted with retries and confirmed via the units' actual state rather than trusting a wait timeout, but it is still best-effort: Core is always started regardless of whether the release could be confirmed, since the reserved port only guards Core's own API/frontend bind — if Core can't claim it right away it just logs and keeps retrying the bind itself, which is far better than not starting Core at all (the single most critical piece of this orchestration).

On Supervisor restart (Core already running), no unit is created. If a previous Supervisor crash left a unit behind, it's detected and cleaned up (stopped, and reset if it ended up in a failed state) before a new one is created, since systemd refuses to redefine a transient unit that's still loaded under the same name.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR fixes or closes issue: fixes #6790
- Link to cli pull request:

### Design notes

- The reservation is best-effort: if systemd D-Bus isn't connected or the unit can't be created for any reason, reservation is skipped with a warning and boot continues unprotected, rather than blocking boot on a failed reservation.
- No container or subprocess is spun up, so there's no added boot latency and nothing visible in `docker ps` during boot.
- Waiting for the unit to reach the `active` state before continuing closes the race window a subprocess/container based approach would have — the bind is confirmed to have already happened by the time boot proceeds.
- Starting Home Assistant Core is never skipped because of this feature. The port reservation only protects Core's own bind; Core itself already handles and retries a failed bind, so the worst case of an unconfirmed release is a delayed UI, not a Supervisor that refuses to start Core.

## Checklist

- [x] The code change is tested and works locally
- [x] Local tests pass. Your PR cannot be merged unless tests pass
- [x] There is no commented out code in this PR
- [x] I have followed the [development checklist](https://developers.home-assistant.io/docs/supervisor/development)
- [x] Tests have been added to verify that the new code works

### DIFF
--- a/supervisor/core.py
+++ b/supervisor/core.py
@@ -7,8 +7,7 @@ from datetime import timedelta
 import logging
 from typing import Final, Self
 
-import aiodocker
-from aiodocker.containers import DockerContainer
+from dbus_fast import Variant
 
 from .const import (
     ATTR_STARTUP,
@@ -19,9 +18,10 @@ from .const import (
     CoreState,
 )
 from .coresys import CoreSys, CoreSysAttributes
-from .dbus.const import StopUnitMode, UnitActiveState
+from .dbus.const import StartUnitMode, StopUnitMode, UnitActiveState
 from .exceptions import (
     AppFileReadError,
+    DBusError,
     HassioError,
     HomeAssistantCrashError,
     HomeAssistantError,
@@ -37,8 +37,10 @@ from .utils.whoami import retrieve_whoami
 
 _LOGGER: logging.Logger = logging.getLogger(__name__)
 
-# Name of the short-lived container used to hold Core's host port during boot.
-_PORT_RESERVE_CONTAINER: Final = "hassio_port_reserve"
+# Name of the transient systemd unit used to hold Core's host port during boot.
+_PORT_RESERVE_UNIT: Final = "hassio-port-reserve.socket"
+_PORT_RESERVE_TIMEOUT: Final = 10
+_TERMINAL_STATES: Final = {UnitActiveState.INACTIVE, UnitActiveState.FAILED}
 
 
 class Core(CoreSysAttributes):
@@ -252,18 +254,19 @@ class Core(CoreSysAttributes):
                     await self.sys_supervisor.update()
                     return
 
-        # Start a temporary host-network container to hold Core's TCP port before
-        # booting any apps. Core runs with --network=host, so it competes in the
-        # host network namespace. Supervisor itself is on the hassio bridge and
-        # cannot bind there directly, so we use a container that can.
+        # Reserve Core's TCP port before booting any apps. Core runs with
+        # --network=host, so it competes with apps in the host network
+        # namespace. Supervisor itself is on the hassio bridge and cannot bind
+        # there directly, so a transient systemd socket unit is used instead
+        # (systemd performs the bind itself as part of activating the unit).
         # Released just before Core starts so Core can claim the port.
-        core_port_container: DockerContainer | None = None
+        core_port_reserved = False
         try:
             if not await self.sys_homeassistant.core.is_running():
                 hosts = self.sys_homeassistant.http_server_host
                 host = hosts[0] if hosts else "0.0.0.0"
                 port = self.sys_homeassistant.api_port
-                core_port_container = await self._reserve_core_port(host, port)
+                core_port_reserved = await self._reserve_core_port(host, port)
 
             # Start app mark as initialize
             await self.sys_apps.boot(AppStartup.INITIALIZE)
@@ -283,9 +286,9 @@ class Core(CoreSysAttributes):
             await self.sys_apps.boot(AppStartup.SERVICES)
 
             # Release the port reservation so Core can bind it
-            if core_port_container is not None:
-                await self._release_core_port(core_port_container)
-                core_port_container = None
+            if core_port_reserved:
+                await self._release_core_port()
+                core_port_reserved = False
 
             # run HomeAssistant
             if (
@@ -321,9 +324,9 @@ class Core(CoreSysAttributes):
             await self._update_last_boot()
 
         finally:
-            # Ensure the port reservation container is always removed
-            if core_port_container is not None:
-                await self._release_core_port(core_port_container)
+            # Ensure the port reservation is always released
+            if core_port_reserved:
+                await self._release_core_port()
 
             # Add core tasks into scheduler
             await self.sys_tasks.load()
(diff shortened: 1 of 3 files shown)


supervisor/core.py
```diff
@@ -444,6 +474,113 @@ async def _update_last_boot(self) -> None:
...
+        # Unlike a oneshot unit, our socket unit never exits on its own, so a
+        # unit left active by a crash would otherwise hold Core's port forever,
+        # and systemd refuses to redefine a transient unit that's still loaded
+        # (in any state) under the same name even with mode=replace.
+        await self._release_core_port()
+
+        properties: list[tuple[str, Variant]] = [
+            ("Description", Variant("s", "Home Assistant Core port reservation")),
+            ("Listen", Variant("a(ss)", [("Stream", f"{host}:{port}")])),
+        ]
```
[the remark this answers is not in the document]
mdegat01: Good catch, tied to a dummy service now to keep it alive.

supervisor/core.py
```diff
@@ -246,7 +254,20 @@ async def start(self) -> None:
...
+        # there directly, so a transient systemd socket unit is used instead
+        # (systemd performs the bind itself as part of activating the unit).
+        # Released just before Core starts so Core can claim the port.
+        core_port_reserved = False
         try:
+            if not await self.sys_homeassistant.core.is_running():
+                hosts = self.sys_homeassistant.http_server_host
+                host = hosts[0] if hosts else "0.0.0.0"
+                port = self.sys_homeassistant.api_port
+                core_port_reserved = await self._reserve_core_port(host, port)
```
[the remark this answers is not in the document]
mdegat01: Fixed by defaulting to ipv6 and ipv4 wildcard. Left the `http_server_host` bit in for now though cause I wanted to check with @agners on that. That was part of your PR, why didn't we persist that? I'd love to use it here to grab the exact binding in advance and not block more then necessary but it seems like it was deliberate from the comments. Can we change that?

supervisor/core.py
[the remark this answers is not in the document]
mdegat01: 👍

supervisor/core.py
```diff
@@ -444,6 +474,113 @@ async def _update_last_boot(self) -> None:
...
+                _PORT_RESERVE_UNIT, StartUnitMode.REPLACE, properties
+            )
+            unit = await self.sys_dbus.systemd.get_unit(_PORT_RESERVE_UNIT)
+            async with asyncio.timeout(_PORT_RESERVE_TIMEOUT):
+                state = await unit.wait_for_active_state(
+                    {UnitActiveState.ACTIVE, UnitActiveState.FAILED}
+                )
+            if state != UnitActiveState.ACTIVE:
+                raise DBusError(f"unit entered state {state}")
+        except (DBusError, TimeoutError) as err:
```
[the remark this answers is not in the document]
mdegat01: Changed all tehse over

supervisor/core.py
```diff
@@ -444,6 +474,113 @@ async def _update_last_boot(self) -> None:
...
+            with suppress(TimeoutError):
+                async with asyncio.timeout(_PORT_RESERVE_TIMEOUT):
+                    state = await unit.wait_for_active_state(_TERMINAL_STATES)
+
+            # Only a FAILED unit needs resetting -- systemd keeps those around
+            # until reset, but garbage collects a cleanly INACTIVE one on its
+            # own. A timeout above leaves state at FAILED so we still try.
+            if state != UnitActiveState.INACTIVE:
+                with suppress(DBusError):
+                    await self.sys_dbus.systemd.reset_failed_unit(_PORT_RESERVE_UNIT)
```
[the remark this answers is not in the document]
mdegat01: So partially addressed this. A short retry loop was added and the feedback handled. However we are always going to start core after. Not starting core would be a catastrophic failure here, users would effectively be left without a usable system at all.

The port in question here is core's API port. If core fails to bind it then it still starts but there's no frontend and no API. This is better then not starting it at all though since core is still running everything else, logging the failure and attempting to retry periodically if the service can be properly taken down later. This is preferrable to not starting at all

supervisor/core.py
```diff
@@ -444,6 +485,160 @@ async def _update_last_boot(self) -> None:
...
+        if not await self._release_core_port():
+            _LOGGER.warning(
+                "Cannot reserve Core port(s) %s:%d: a previous reservation "
+                "could not be confirmed released",
+                hosts,
+                port,
+            )
+            return False
+
+        service_properties: list[tuple[str, Variant]] = [
```
agners, line 521: Hm, maybe it is time to type this. We can't use dataclasses, but type aliases. But `start_transient_unit()` already uses bare types, so this PR does not introduce something new. So I'd do this in a follow up PR, something along these lines:

```python
type UnitProperty = tuple[str, Variant]
type UnitProperties = list[UnitProperty]
type AuxUnit = tuple[str, UnitProperties]
```

supervisor/core.py
```diff
@@ -444,6 +485,160 @@ async def _update_last_boot(self) -> None:
...
+
+        for _attempt in range(2):
+            # A stop error doesn't necessarily mean it didn't happen on the
+            # systemd side -- always re-check the real state below instead.
+            with suppress(HassioError):
+                await self.sys_dbus.systemd.stop_unit(unit_name, StopUnitMode.REPLACE)
+
+            with suppress(TimeoutError):
+                async with asyncio.timeout(_PORT_RESERVE_TIMEOUT):
+                    await unit.wait_for_active_state(_TERMINAL_STATES)
```
agners: This path can also raise `HassioError`, which we probably need to suppress too here.

supervisor/core.py
```diff
@@ -246,7 +267,16 @@ async def start(self) -> None:
                     await self.sys_supervisor.update()
                     return
 
+        # Reserve Core's TCP port before booting other apps, since Core runs
+        # with --network=host and competes with them for it. Released again
+        # just before Core starts so it can claim the port itself.
+        core_port_reserved = False
         try:
+            if not await self.sys_homeassistant.core.is_running():
+                hosts = self.sys_homeassistant.http_server_host or ["0.0.0.0", "::"]
```
agners, line 276: By default, Linux binds IPv6 dual-stack (default is `bindv6only=0`, which means IPv6 binds IPv4 as well). So passing both conflicts with itself.

The [Core passes both](https://github.com/home-assistant/core/blob/0f9e1bfaacaa7580e0ab02a981d4d491d609c9ee/homeassistant/components/http/server.py#L72-L73), but uses `loop.create_server()` which explicitly sets `IPV6_V6ONLY=1` on every `AF_INET6` socket.

So either we just pass IPv6 here:

```suggestion
                hosts = self.sys_homeassistant.http_server_host or ["::"]
```

Or, in my opinion cleaner/aligns better with Core, we set the systemd equivalent `BindIPv6Only=ipv6-only` (add `("BindIPv6Only", Variant("s", "ipv6-only")),` to `socket_properties`). This actually is the only correct way if `server_host` is in play.

agners (changes requested):

supervisor/core.py
mdegat01: Hm, its basically the same thing as the try except below I think? We're really just calling `unit.get_active_state()` inside `unit.wait_for_active_state`. We handle that by returning true so I'll put this inside the same `try` to handle `HassioError` the same here.

supervisor/core.py
mdegat01: nice, yea I like the second one better too

supervisor/core.py
mdegat01: yea this is pretty convoluted now. Ok ill follow-up with that

agners (approved):
LGTM!

PR #7137, opened 2026-08-14
Add local disk mounts to the mount API

MindFreeze:
> [!IMPORTANT]
> **This branch was rebuilt on top of #7167, not rebased.** The automount rework removed the bind-mount layer this feature rode on and made `Mount.from_dict` reject anything but CIFS/NFS, so `DiskMount`'s unit creation, mount location, and absent-device story were redesigned rather than merged. Earlier review comments are all still addressed (see the last reviewer note), but the diff is new work and deserves a fresh read rather than a delta skim.

## Proposed change

Add a `disk` mount type to the mount API so internal and USB drives can be used the same way as network storage, as discussed in home-assistant/feature-requests discussion [#1747](https://github.com/orgs/home-assistant/discussions/1747). A disk is picked from the new `GET /mounts/candidates` endpoint (UDisks2-backed, with guard rails that exclude system devices, HAOS partitions, the data disk and anything already mounted), resolved once to its filesystem UUID, and mounted by that UUID so it keeps working across reboots and device renames. Local disks follow the same usage restrictions (backup/media/share) and failure handling as network mounts.

Following #7167, a disk mount is a `.automount` + `.mount` pair like a network share. That is what replaces the retry the reconcile task used to do: a disk that was unplugged mounts itself again on the next access to its path, with no machinery of ours involved.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR is related to issue: https://github.com/orgs/home-assistant/discussions/1747
- Link to documentation pull request: https://github.com/home-assistant/developers.home-assistant/pull/3297, https://github.com/home-assistant/home-assistant.io/pull/47453
- Link to cli pull request: https://github.com/home-assistant/cli/pull/672
- Link to frontend pull request: https://github.com/home-assistant/frontend/pull/53643

Notes for reviewers:

- **Why `What=` is not the by-uuid path.** The unit mounts a link at `<data>/.mounts_devices/<name>` that points at `/dev/disk/by-uuid/<uuid>`. systemd derives a `Requires=` on the backing device unit whenever `What=` begins with `/dev/`, and while the disk is absent that dependency stalls **every** access for `DefaultDeviceTimeoutSec` — 90 s by default, per access, with no failure caching. Nothing settable on the unit shortens it, which I checked one by one: the unit's own `TimeoutUSec` does not cover the dependency wait; `JobTimeoutUSec` cannot be set on a transient unit; `JobRunningTimeoutUSec` is accepted but has no effect; `What=UUID=<uuid>` resolves to the same device unit; `AssertPathExists` and `ConditionPathExists` are evaluated only *after* dependencies are satisfied; and `DefaultDependencies=false` does not touch it, the device dependency being an automatic one. A `What=` outside `/dev` creates no device dependency at all, so `mount(8)` resolves the link itself and fails in milliseconds while the disk is away. There is a test pinning this specifically, because an innocent-looking simplification back to the by-uuid path silently restores a 90-second stall on every access to the mount point.
- **Why disks get an extra presence check.** A local filesystem answers `statvfs` from cache, so the probe that tells a network share's health cannot see that a disk was pulled from its port — it succeeds either way. Disk mounts therefore also ask UDisks2 whether the device is still present, and report inactive when it is not. An unavailable UDisks2 is deliberately not treated as a missing disk.
- **Supported filesystems**: ext4, ext3, ext2, vfat, exfat, ntfs, btrfs and f2fs — the set the HAOS kernel can mount. f2fs is included on @sairon's prompt, since HAOS 18.3 builds it on every board (home-assistant/operating-system#4977); on an older OS an f2fs mount fails with a plain mount error and starts working after the OS update — the same non-guarantee that already applies to Supervised installations, whose kernel is not ours to promise anything about. The systemd unit gets the kernel driver name (`ntfs` mounts with `ntfs3`, `ext2`/`ext3` with `ext4`), while the API always reports the filesystem as probed. `fs_type` is the override point for this; it raises rather than guess if read before the device is resolved.
- **Security model**: whether a device may be mounted (not a system disk, not in use, supported filesystem, ...) is checked when Supervisor resolves the device through UDisks2. To make sure that check always runs, the API ignores a caller-supplied `filesystem`, and a restored backup gets re-resolved and re-checked on the restoring host. Only `mounts.json` is trusted as-is, since editing it already requires root on the host.
- A write-protected device can only be added read-only; asking for it writable is an error, not a silent downgrade.
- A candidates entry may be posted straight back to `POST /mounts` with a name added. Both identifiers may be supplied together: the UUID drives resolution and a device supplied alongside is checked for agreement rather than silently ignored.
- Two backup edge cases: restoring a disk mount on a host without UDisks2 fails that mount with a clear error, and restoring a backup containing a disk mount on an *older* Supervisor skips restoring all mounts (the old schema rejects the unknown type — existing best-effort behavior).
- vfat/exfat/ntfs mounts have no ownership options (`uid`/`gid`/`umask`) yet, same as CIFS mounts today; possible follow-up.
- **All six items from the earlier review are still in place** after the rebuild, verified individually rather than assumed: the `isinstance` checks replaced by `mount.type == MountType.DISK` at both sites, the validate.py/api split with `uuid` required in the persisted schema, the scrubbed test serials, `type` in each candidates entry, and the postable-candidates round trip with the device/UUID agreement check.

Testing:

- Unit tests assert the exact transient unit properties for both units of the pair, and every behavior above has a test that fails when that behavior is reverted.
- The systemd and kernel behavior behind the two reviewer notes above was measured rather than assumed, on a Linux host using transient units against a loopback filesystem that could be attached and detached on demand: 90.1–90.2 s per access with an absent device and no failure caching, 0.0 s once `What=` moved outside `/dev`, a replugged device mounting itself on the next access in 0.1 s, and `statvfs` still succeeding on a mount whose device had been detached. Measured on systemd 259; HAOS runs 254.13, where all of these mechanisms are already present and no `system.conf` override changes the 90 s default.
- Mount activation on real hardware cannot be exercised in a container, since systemd only manages device units on a real host.
- Testing limitation: the devcontainer currently cannot exercise mounts of any type. UDisks2 inside it sees no host devices without a `/run/udev` passthrough, and `/data` is mounted without the `slave` propagation that `HostFeature.MOUNT` requires. Both may be deliberate, so this is noted as a limitation rather than fixed here.

## Checklist


Baxxy13:
Really nice new feature.
But wondering why f2fs is not included in the list of supported file systems.

MindFreeze:
f2fs is not currently supported by HA OS. Support can potentially be added to the OS but this is out of scope for this PR

supervisor/mounts/manager.py
```diff
@@ -387,6 +387,18 @@ async def restore_mount(self, mount: Mount) -> asyncio.Task:
...
+        # A backup is data from outside this host, so a disk mount out of one
+        # is not taken at its word. Forgetting the resolved filesystem makes
+        # activation resolve the device again, which is what runs the
+        # mountable-device guard — otherwise a crafted backup could name the
+        # UUID of a system disk and have it mounted unchecked. This is safe
+        # here and deliberately not done for mounts.json at startup: a restore
+        # runs on a live host where UDisks2 is available. If the disk is absent
+        # the activation fails, the entry is retained, and each later reload
+        # retries the resolution until it appears.
+        if isinstance(mount, DiskMount):
```
mdegat01:
```suggestion
        if mount.type == MountType.DISK:
```
Should be the same thing right? We try to avoid `isinstance` checks if there's an alternative since they tend to have a high performance cost.

supervisor/mounts/mount.py
```diff
@@ -814,3 +840,190 @@ def where(self) -> PurePath:
...
+        if "device" in self._data:
+            del self._data["device"]
+
+
+def disk_mount_uuids(mounts: list[Mount], *, exclude: str | None = None) -> set[str]:
+    """Return filesystem UUIDs already claimed by configured disk mounts."""
+    return {
+        mount.uuid
+        for mount in mounts
+        if isinstance(mount, DiskMount) and mount.uuid and mount.name != exclude
```
mdegat01:
```suggestion
        if mount.type == MountType.DISK and mount.uuid and mount.name != exclude
```
Same as above, avoiding `isinstance` if we can.

supervisor/mounts/validate.py
```diff
@@ -85,13 +102,47 @@ def usage_specific_validation(config: dict[str, Any]) -> dict[str, Any]:
     }
 )
 
+_SCHEMA_MOUNT_DISK_BASE = _SCHEMA_BASE_MOUNT_CONFIG.extend(
+    {
+        vol.Required(ATTR_TYPE): vol.All(MountType.DISK.value, vol.Coerce(MountType)),
+        vol.Exclusive(ATTR_DEVICE, "disk_device_identifier"): str,
+        vol.Exclusive(ATTR_UUID, "disk_device_identifier"): str,
```
mdegat01: Everything in here about requiring either UUID or device as identifier and about whether or not there can be a filesystem is for the API only right? The persisted config and backup will only ever have a UUID and will always have a filesystem right?

Assuming I'm right I think we should split this, its rather confusing right now. This file is where validation for the mounts.json config files go. The API layer borrows what it can use as is but API specific validation should go in `supervisor/api/mounts.py` instead to keep it close to its point of usage.

tests/api/test_mounts.py
```diff
@@ -920,3 +928,422 @@ async def test_mount_not_found(
...
+            "device": "/dev/sdc1",
+            "uuid": SDC1_UUID,
+            "label": "Backups",
+            "filesystem": "ext4",
+            "size": 2000397795328,
+            "read_only": False,
+            "drive": {
+                "vendor": "Seagate",
+                "model": "Expansion",
+                "serial": "NA9BQ8X4",
```
mdegat01: Just to check, if this is a real serial code feel free to scrub it into something super generic. I'm not sure if serial code of a disk constitutes a concerning reveal of PII but it feels like a better to be safe then sorry situation

supervisor/api/mounts.py
```diff
@@ -50,6 +70,68 @@ async def info(self, request: web.Request) -> dict[str, Any]:
...
+        """Return API representation of a mount candidate."""
+        return {
+            ATTR_DEVICE: block.device.as_posix(),
+            ATTR_UUID: block.id_uuid,
+            ATTR_LABEL: block.id_label,
+            ATTR_FILESYSTEM: block.id_type,
+            ATTR_SIZE: block.size,
+            ATTR_READ_ONLY: block.read_only,
+            ATTR_DRIVE: self._drive_to_dict(block),
+        }
```
mdegat01, line 154: Is there a world where we can potentially discover CIFS and NFS mounts (or other future types of mounts we might add support for)? I'm thinking perhaps we should add `ATTR_TYPE` to the fields in this output.

I realize every item would be the same currently (`type: disk`). But the API is `/mounts/candidates` so if we could somehow discover other types of mounts in the future they should probably be returned in the same list with a different type.

supervisor/api/mounts.py
```diff
@@ -50,6 +70,68 @@ async def info(self, request: web.Request) -> dict[str, Any]:
...
+
+            candidates.append(self._candidate_to_dict(block))
+
+        return {ATTR_CANDIDATES: candidates}
+
+    def _candidate_to_dict(self, block: UDisks2Block) -> dict[str, Any]:
+        """Return API representation of a mount candidate."""
+        return {
+            ATTR_DEVICE: block.device.as_posix(),
+            ATTR_UUID: block.id_uuid,
```
mdegat01, line 148: If you return both of these, doesn't that block the ability to pass this object directly into `POST /mounts`? `filesystem` is handled by `REMOVE_EXTRA` but this will run into the `vol.Exclusive` preventing two unique identifiers from coming in and return an error right? I think it would be good if the output of here can be used directly in a create call with just the addition of a name, perhaps with a test case to validate that. Would be nice for easy API usage.

mdegat01 (changes requested):
This looks quite nice! Just some smaller changes I think, nothing too significant. Will try to get it soon. I'll take a look at the related PRs next. Thanks for this!

supervisor/mounts/validate.py
MindFreeze: Hmm, one subtlety here is that when a mount is restored from a backup, it won't have a filesystem until the disk re-attaches. So that has to be optional

tests/api/test_mounts.py
MindFreeze: It was fake but I made it even more fake :)

supervisor/api/mounts.py
MindFreeze: Done. It can now be posted back directly with just name and usage added. When both device and uuid are present, resolution goes by uuid and the device must agree with it (mismatch is an error). The remaining fields are stripped by the schema.

MindFreeze:
Review comments are addressed but @agners wanted to delay this until mounts switch to `autofs`. Don't know if we should keep this in draft until then

mdegat01:
That makes sense. Ok I'll follow up with him

sairon:
@Baxxy13 @MindFreeze FTR, HAOS 18.3+ will get support for F2FS: https://github.com/home-assistant/operating-system/pull/4977

MindFreeze:
Updated with the autofs changes from #7167

felix-rub (approved):

Robbe-B:
@MindFreeze, 
I'm very happy to see someone put in the effort for this feature ! 🎉

PR #7145, opened 2026-08-17
Surface fixup failures to the caller applying a suggestion

agners:
## Proposed change

Applying a resolution suggestion whose fixup failed reported success to the caller: `FixupBase.__call__` swallowed `ResolutionFixupError`, so the API returned OK and the repair flow in Home Assistant completed — which removes the repair from the UI — while the issue persisted in Supervisor. Observed with the mount repairs (#7089 testing): the fixup logged a warning, the repair dialog claimed success, and the repair disappeared from Home Assistant while `ha resolution info` still showed the issue and suggestion.

Let the error propagate from the fixup instead, handling each caller appropriately:

- **User-applied suggestions** (`apply_suggestion` API): the failure now surfaces as an API error. Home Assistant's repair flow aborts with its existing "Could not apply the fix" message instead of completing, and the repair stays visible.
- **Autofix loop**: already catches and logs per-fixup errors — unchanged behavior.
- **Bus-event triggered fixups** (e.g. store reload on connectivity change): the event callback now catches and logs the failure the same way autofix does, instead of leaving an unhandled error in the bus event task.

`ResolutionFixupError` becomes an `APIError` with a translatable `resolution_fixup_error` error key and a default message, so it is reported as a client-visible error (HTTP 400 with message) rather than an unexpected server error with traceback and Sentry report. Fixup tests that exercise failing fixups now assert the raise; their existing assertions that the issue and suggestion remain are kept.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR is related to issue: #7089, home-assistant/core#179369
- Link to cli pull request:

Note: when this merges, the fixup-failure test added in #7089 (`test_fixup_failure_keeps_suggestion`) needs a `pytest.raises` added — whichever PR lands second takes the small adjustment.

## Checklist

### DIFF
--- a/supervisor/exceptions.py
+++ b/supervisor/exceptions.py
@@ -1212,16 +1212,7 @@ class ResolutionNotFound(ResolutionError):
     """Raise if suggestion/issue was not found."""
 
 
-class ResolutionFixupError(APIError):
-    """Raise if a fixup fails."""
-
-    error_key = "resolution_fixup_error"
-    message_template = (
-        "Applying the suggestion failed. Check Supervisor logs for details"
-    )
-
-
-class ResolutionFixupJobError(ResolutionFixupError, JobException):
+class ResolutionFixupJobError(ResolutionError, JobException):
     """Raise on job error."""
 
 
--- a/supervisor/exceptions.py
+++ b/supervisor/exceptions.py
@@ -1212,8 +1212,13 @@ class ResolutionNotFound(ResolutionError):
     """Raise if suggestion/issue was not found."""
 
 
-class ResolutionFixupError(HassioError):
+class ResolutionFixupError(APIError):
     """Raise if a fixup fails."""
+
+    error_key = "resolution_fixup_error"
+    message_template = (
+        "Applying the suggestion failed. Check Supervisor logs for details"
+    )
 
 
 class ResolutionFixupJobError(ResolutionFixupError, JobException):
--- a/supervisor/resolution/fixups/base.py
+++ b/supervisor/resolution/fixups/base.py
@@ -28,10 +28,11 @@ class FixupBase(ABC, CoreSysAttributes):
         if fixing_suggestion is None:
             return
 
-        # Process fixup. A failure propagates to the caller: the autofix
-        # loop logs and continues with the next fixup, while a user-applied
-        # suggestion surfaces the failure as an API error so the repair is
-        # not reported as successful (the issue and suggestion stay around).
+        # Process fixup. A failure propagates to the caller with its
+        # original, well-defined error: the autofix loop logs and continues
+        # with the next fixup, while a user-applied suggestion surfaces the
+        # failure as an API error so the repair is not reported as
+        # successful (the issue and suggestion stay around).
         _LOGGER.debug("Run fixup for %s/%s", self.suggestion, self.context)
         await self.process_fixup(fixing_suggestion)
 
--- a/supervisor/resolution/fixups/base.py
+++ b/supervisor/resolution/fixups/base.py
@@ -5,7 +5,6 @@ import logging
 
 from ...const import BusEvent
 from ...coresys import CoreSys, CoreSysAttributes
-from ...exceptions import ResolutionFixupError
 from ..const import ContextType, IssueType, SuggestionType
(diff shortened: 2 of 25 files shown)


supervisor/resolution/fixups/base.py
```diff
@@ -29,12 +28,12 @@ async def __call__(self, fixing_suggestion: Suggestion | None = None) -> None:
...
+        # Process fixup. A failure propagates to the caller: the autofix
+        # loop logs and continues with the next fixup, while a user-applied
+        # suggestion surfaces the failure as an API error so the repair is
+        # not reported as successful (the issue and suggestion stay around).
         _LOGGER.debug("Run fixup for %s/%s", self.suggestion, self.context)
-        try:
-            await self.process_fixup(fixing_suggestion)
-        except ResolutionFixupError:
-            return
+        await self.process_fixup(fixing_suggestion)
```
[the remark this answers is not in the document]
mdegat01: Yea this was my first thought as well reading it. We don't want these to surface to sentry so we should suppress them outside of here in that code path.

supervisor/exceptions.py
```diff
@@ -1212,9 +1212,14 @@ class ResolutionNotFound(ResolutionError):
...
 
 
-class ResolutionFixupError(HassioError):
+class ResolutionFixupError(APIError):
     """Raise if a fixup fails."""
 
+    error_key = "resolution_fixup_error"
+    message_template = (
+        "Applying the suggestion failed. Check Supervisor logs for details"
+    )
```
mdegat01: So this is the part I don't like. In many cases in these methods our fixups are calling are actually raising well-defined errors. We're then catching those errors, re-raising them as the ultra-generic `ResolutionFixupError` and pushing out this generic message instead the better defined one.

I want to suggest an alternate approach to consider. I think the entire reason `ResolutionFixupError` exists is because at the time we built all this the suggestion system and API was so hidden away the only way we actually expected them to run was via autofix. So we needed a way to differentiate between real issues with our code that needed to be reported to sentry and fix up failures due to environmental/config issues that just need to be reported to the user via logs. If we didn't expect anyone to actually use the apply suggestion API then the easiest way is to translate all environmental/config related errors into one suppressable error type. This makes the API quite difficult to use because a 200 response may be a failure but in this context that doesn't matter.

This doesn't apply at all anymore. The suggestion system is definitely not hidden away and the API is in regular use now. In fact I'd argue autofix is now the exception, the norm is applying suggestions via the API. So I think we should flip this around. Toss this exception type entirely, just let the original exception bubble. Autofix can catch and suppress all exceptions. Ones that extend from `HassioError` it can assume have been otherwise handled and reported to sentry if necessary. Ones that do not extend from `HassioError` it can continue to report to sentry.

mdegat01 (changes requested):
If we want to keep `ResolutionFixupError`, we need to fix the issue raised by Copilot to prevent all new sentry noise.

However see my second comment, in my opinion we should toss `ResolutionFixupError` entirely and just use the original exceptions as defined. In most cases they do a much better job of explaining what went wrong.

supervisor/exceptions.py
agners: Right, yeah I like that approach better too. Let's use the original exception. And if that one is not good (enough), improve it instead :+1: .

supervisor/resolution/module.py
```diff
@@ -136,7 +136,7 @@ def add_suggestion(self, suggestion: Suggestion) -> None:
                 async def event_callback(reference, fixup=fixup):
                     try:
                         await fixup(suggestion)
-                    except ResolutionFixupError as err:
+                    except HassioError as err:
```
mdegat01: This actually isn't the same as autofix because we don't also catch `Exception` and report to sentry from here. I know that's not new behavior from this PR so can leave it for follow-up but we probably should be consistent? Maybe a common helper we can use instead of this inline function?

mdegat01 (approved):
Love it! So much cleaner 😍 LGTM!

supervisor/resolution/module.py
agners: Hm, yeah this is indeed inconsistent. We add this particular duplication in this PR, so seems sensible to fix it as part of this PR as well.

PR #7146, opened 2026-08-17
Report storage usage per mount

MindFreeze:
## Proposed change

Parameterize the disk usage endpoint as `GET /host/disks/{disk}/usage` so it can also report storage usage for an active supervisor mount by name. `default` — the literal URL today's clients call — keeps its existing response byte for byte. Per-mount probes are shared between concurrent callers, run to the kernel's own bound rather than an artificial timeout, and are confined to an executor thread so a hung probe never blocks the rest of the API.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR is related to issue: https://github.com/orgs/home-assistant/discussions/1747
- Link to documentation pull request: https://github.com/home-assistant/developers.home-assistant/pull/3302
- Link to frontend pull request: https://github.com/home-assistant/frontend/pull/53683
- Link to cli pull request:

### Notes for reviewers

- Depth semantics are uniform across targets: level 1 is the system disk's labeled known-paths layer, which a mount does not have, so a mount breakdown starts at `max_depth=2`.
- Mounts default to totals only (`max_depth=0`). The directory walker recurses regardless of `max_depth` — the parameter only gates whether children appear in its output — so whenever it could not produce output (depth 0 and 1 for a mount) it is skipped outright rather than called; otherwise a request would walk the whole network mount to report a single number.
- When a breakdown is produced, whatever the walk cannot attribute is reported as an `other` child, keeping every node's children summing to its `used_bytes` — same invariant as the system disk's `system` entry.
- `default` wins over a mount literally named `default`, which the mount name pattern permits.
- The probe deliberately does not use `asyncio.shield` even though this file uses it elsewhere; the code comment at the call site explains why.

### Testing

End-to-end against a real CIFS mount (Samba in a container, guest share with known contents: 30 MiB `movies/`, 10 MiB `music/`, 5 MiB at the share root):

- Totals: `total_bytes` byte-identical to `df` on the server side; totals-only response carries no `children` key.
- Depth-2 breakdown: `movies` 31457280 and `music` 10485760 — exactly the sizes written into the share — plus `other` absorbing the rest; children sum exactly to `used_bytes`.
- Dead server: stopping Samba mid-flight returned a clean error in 10 s via the kernel's own `soft,echo_interval=10` bound (`Host is down`), while concurrent `/supervisor/ping` answered in 25–46 ms and the system disk usage in 52 ms during the hang.
- Inactive mount → 400 with no probe attempted; unknown mount → 404.
- `default` unchanged, including the v1 legacy addon-ID remap confirmed live through the parameterized route.
- Recovery: after the server returned, a mount reload restored usage reporting.

Unit tests: 69 pass in `tests/api/test_host.py` (46 pre-existing untouched, 23 new, v1+v2 parametrized); full suite green; 10 targeted mutations of the new behavior each produce a failing test.

## Checklist

### DIFF
--- a/supervisor/api/host.py
+++ b/supervisor/api/host.py
@@ -37,11 +37,17 @@ from ..const import (
     ATTR_TIMEZONE,
 )
 from ..coresys import CoreSysAttributes
+from ..dbus.const import UnitActiveState
 from ..exceptions import (
     APIDBMigrationInProgress,
     APIError,
     HostContainerLogEpochError,
     HostLogError,
+    MountNotFound,
+    MountUsageNotActiveError,
+    MountUsageNotMountedError,
+    MountUsageReadError,
+    MountUsageTimeoutError,
 )
 from ..host.const import (
     PARAM_BOOT_ID,
@@ -51,6 +57,7 @@ from ..host.const import (
     LogFormatter,
 )
 from ..host.logs import SYSTEMD_JOURNAL_GATEWAYD_LINES_MAX
+from ..mounts.mount import Mount
 from ..utils.systemd_journal import journal_logs_reader
 from .const import (
     ATTR_AGENT_VERSION,
@@ -59,6 +66,7 @@ from .const import (
     ATTR_BOOTS,
     ATTR_BROADCAST_LLMNR,
     ATTR_BROADCAST_MDNS,
+    ATTR_CHILDREN,
     ATTR_DT_SYNCHRONIZED,
     ATTR_DT_UTC,
     ATTR_FORCE,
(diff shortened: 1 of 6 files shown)


supervisor/api/host.py
```diff
@@ -412,6 +459,147 @@ async def _disk_usage_data(self, request: web.Request) -> dict[str, Any]:
             ],
         }
 
+    async def _mount_disk_usage_data(
+        self, request: web.Request, name: str
+    ) -> dict[str, Any]:
+        """Build disk usage response data for a supervisor mount."""
+        if name not in self.sys_mounts:
+            raise APINotFound(f"No mount exists with name {name}")
```
agners: In other places we started to create specific exception with a translation key to aid translation on client side. However, if we know already that we won't be using it on client side it is also not a hard requirement. I leave it up to you. But in any case, we'd make it consistent (create specific exceptions for all or none of the error cases).

MindFreeze: Done

agners (approved):
LGTM

PR #7154, opened 2026-08-19
Harden mount storage usage edge cases

MindFreeze:
## Proposed change

Follow-up to #7146, addressing three findings from the automated review that arrived after approval. Depths below 2 are normalized before keying the shared-probe registry, so concurrent callers varying only the depth share one probe; a breakdown whose walk raced a deletion (children summing past the parent) is dropped in favor of the consistent totals; and the v1 legacy addon-id remap is applied only to the system disk target, so a mount directory that happens to be named `apps_data` keeps its real name and the shared probe result is never mutated in place.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR is related to issue: #7146
- Link to cli pull request:

### Notes for reviewers

- One reviewed finding was deliberately not changed: a flat mount (files only at the root) at depth 2 still returns totals without an `other`-only breakdown — a single child covering the whole bar carries no information beyond the total.
- Each of the three fixes has a test that fails when the fix is reverted.

## Checklist

### DIFF
--- a/supervisor/api/host.py
+++ b/supervisor/api/host.py
@@ -475,6 +475,10 @@ class APIHost(CoreSysAttributes):
             raise MountUsageNotActiveError(name=name)
 
         max_depth = self._requested_max_depth(request, DISK_USAGE_MAX_DEPTH_MOUNT)
+        # All depths below 2 give totals only; normalize so concurrent callers
+        # share one probe regardless of the requested depth.
+        if max_depth < 2:
+            max_depth = DISK_USAGE_MAX_DEPTH_MOUNT
         return await self._mount_usage(mount, max_depth)
 
     async def _mount_usage(self, mount: Mount, max_depth: int) -> dict[str, Any]:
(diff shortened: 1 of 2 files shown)


supervisor/api/host.py
```diff
@@ -585,24 +589,35 @@ async def _mount_usage_data(self, mount: Mount, max_depth: int) -> dict[str, Any
...
             remainder = used - sum(child["used_bytes"] for child in children)
             if remainder > 0:
                 children = [
                     *children,
                     {"id": "other", "label": "Other", "used_bytes": remainder},
                 ]
+            elif remainder < 0:
+                # Walk raced a deletion; drop the breakdown rather than serve
+                # children summing past their parent.
+                children = []
```
mdegat01, line 611: I worry about silently dropping data here. The frontend will have no idea something went wrong here and will just carry on like it was a complete success. But users might notice and then think the feature is bugged.

How much of an edge case are we talking about here? Is this a theoretical issue the AI identified or something you actually noticed in practice? If its theoretical then maybe just a simple log will suffice, something that explains it and we can look for if a user does happen to notice and report an issue.

mdegat01 (changes requested):
One small comment because it seems like we're silently suppressing an error and that tends to bite us down the line. Good finds though

supervisor/api/host.py
MindFreeze: It was theoretical

MindFreeze: Added a log.
I think it's likely to happen on higher latency drives and we will probably never ask for this data for network mounts. So this leaves high latency local drives which should not be common.

mdegat01 (approved):
Looks good to me, thanks @MindFreeze !

PR #7161, opened 2026-08-22
Strip whitespace from repository URLs in store validation

thibaultmol:
## Proposed change

I noticed a bug where if you add a url in the 'add repository' screen on /config/apps/repositories and there were spaces before or after the url, it would error out without a clear error. Now they just get stripped

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

DISCLAIMER: GLM-5.2 was used to create the code for this PR. it's summary of what was changed:
> The fix is in validate_repository() in supervisor/store/validate.py — the repository URL is now stripped of leading/trailing whitespace before validation. This is the central validation function used by both the POST /store/repositories API endpoint and the store config file loader, so all entry points benefit from the fix.

## Checklist

### DIFF
--- a/supervisor/store/validate.py
+++ b/supervisor/store/validate.py
@@ -19,6 +19,8 @@ SCHEMA_REPOSITORY_CONFIG = vol.Schema(
 
 def validate_repository(repository: str) -> str:
     """Validate a valid repository."""
+    repository = repository.strip()
+
     if repository in BuiltinRepository:
         return repository
 
--- a/tests/api/test_store.py
+++ b/tests/api/test_store.py
@@ -129,6 +129,22 @@ async def test_api_store_add_repository_duplicate(
     assert result["message"] == f"Can't add {REPO_URL}, already in the store"
 
 
+async def test_api_store_add_repository_whitespace(
+    api_client: TestClient, coresys: CoreSys, supervisor_internet: AsyncMock
+) -> None:
+    """Test POST /store/repositories strips whitespace from the repository URL."""
+    with (
+        patch("supervisor.store.repository.RepositoryGit.load", return_value=None),
+        patch("supervisor.store.repository.RepositoryGit.validate", return_value=True),
+    ):
+        response = await api_client.post(
+            "/store/repositories", json={"repository": f"  {REPO_URL}  "}
+        )
+
+    assert response.status == 200
+    assert REPO_URL in coresys.store.repository_urls
+
+
 async def test_api_store_remove_repository(
     api_client: TestClient, coresys: CoreSys, test_repository: Repository
 ):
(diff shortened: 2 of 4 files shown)


thibaultmol:
just fyi, I added the email my git was configured with to my github account after the cla-bot mentioned it

supervisor/store/validate.py
```diff
@@ -19,6 +19,8 @@
 
 def validate_repository(repository: str) -> str:
     """Validate a valid repository."""
+    repository = repository.strip()
```
agners, line 22: This handles leading/trailing whitespace, but internal whitespace still slips through: `https://github.com/hassio-addons/repository #beta` passes validation — the url group of `RE_REPOSITORY` (`[^#]+` in `supervisor/validate.py`) matches the space before `#`, and `vol.Url()` tolerates it too — and then fails later during git clone with the same unclear error this PR is fixing. Suggest tightening the url group to `[^#\s]+` so whitespace that `.strip()` can't fix is rejected at validation time with a clear error.

tests/api/test_store.py
```diff
@@ -129,6 +129,22 @@ async def test_api_store_add_repository_duplicate(
     assert result["message"] == f"Can't add {REPO_URL}, already in the store"
 
 
+async def test_api_store_add_repository_whitespace(
```
agners: This is a near-verbatim copy of `test_api_store_add_repository` differing only in the payload. Parametrizing the existing test keeps the two from drifting apart:

```python
@pytest.mark.parametrize("repo_url", [REPO_URL, f"  {REPO_URL}  "])
async def test_api_store_add_repository(
    api_client: TestClient,
    coresys: CoreSys,
    supervisor_internet: AsyncMock,
    repo_url: str,
) -> None:
```

tests/store/test_validate.py
```diff
@@ -24,3 +24,27 @@ async def test_repository_validate(repo_list: list[str], valid: bool):
     else:
         with pytest.raises(Invalid):
             repositories(repo_list)
+
+
+@pytest.mark.parametrize(
```
agners, line 33: The three cases (spaces, `\n`, `\t`) all exercise the same single `str.strip()` call, so they can't fail independently — one mixed case gives the same coverage:

```python
(
    [" \t\nhttps://github.com/hassio-addons/repository \n\t"],
    ["https://github.com/hassio-addons/repository"],
),
```

agners (changes requested):
Thanks for the fix — stripping the URL is a good improvement. There's one whitespace case that still slips through validation and ends up with the same unclear git error this PR sets out to fix, plus two small test simplifications.

agners:
> Thanks for the fix — stripping the URL is a good improvement. There's one whitespace case that still slips through validation and ends up with the same unclear git error this PR sets out to fix, plus two small test simplifications.

Claude also mention that the unique constraint for repository entries can make the validation fail now (when previously URLs with spaces were saved). But this was only possible in Supervisor before 2022, so I don't think we need to handle it.

However, extending the regex and the test fixes seem worthwhile.

agners (approved):
LGTM, thanks!

PR #7166, opened 2026-08-25
Add endpoint to reset Docker storage

### LINKED ISSUE #6555: Implement command to reset Docker storage
### Description

There are quite some reports of Docker storage corruption. Likely this is caused by power outages during a container image write, or possible also just bit rot.

Today, Docker does not have a built-in feature to validate the authenticity of the container image on-disk. There is also no command which reliably rewrites a complete image: Deleting a container image is not reliable as some of the layers can be shared with other container images. This means that if corruption is in a shared layer, even a delete and re-pull of a particular image is not a guarantee for corruption to overcome.

Today, the only way to recover from such situations is to reinstall Home Assistant OS from scratch. This is a bit tedious and also requires a backup restore.

It would be nice to have a feature which simply allows to remove the complete Docker storage, forcing the system to re-download container images. 

### Additional context

_No response_

sairon:
## Proposed change

Corrupted Docker image layers cannot be always fixed by removing and re-pulling a single image, because layers are shared between images. The only (easy) way to recover is wiping all of Docker's storage which is currently cumbersome and requires OS shell access.

Home Assistant OS added a service that performs the Docker storage wipe on boot when a flag file exists in home-assistant/operating-system#4982, and home-assistant/os-agent#283 adds OS Agent DBus interface for creating that file.

This adds a new API method POST /docker/reset-storage which calls OS Agent's ScheduleDockerStorageReset and creates a reboot_required issue, following the same pattern used for the Docker storage driver migration.

This is expected to ship in HAOS 18.3, so the endpoint checks for that and returns 404 on older OS versions. Technically, it might be nicer to gate on OS Agent version, but from user perspective it's better to report the required OS version.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR fixes or closes issue: closes #6555
- Link to documentation pull request: https://github.com/home-assistant/developers.home-assistant/pull/3313
- Link to cli pull request: https://github.com/home-assistant/cli/pull/674
- Link to client library pull request: N/A (no `/docker/*` endpoints exposed there)

## Checklist


supervisor/api/docker.py
```diff
@@ -154,3 +154,34 @@ async def migrate_docker_storage_driver(self, request: web.Request) -> None:
...
             suggestions=[SuggestionType.EXECUTE_REBOOT],
         )
+
+    @api_process
+    async def reset_storage(self, request: web.Request) -> None:
+        """Schedule a Docker storage reset on next reboot."""
+        if (
+            not self.coresys.os.available
+            or not self.coresys.os.version
+            or self.coresys.os.version < AwesomeVersion("18.3.dev0")
```
agners: We usually use a constant for this, and I guess we'll need to have an actual dev build version.

supervisor/api/__init__.py
```diff
@@ -1071,6 +1071,7 @@ def _register_docker(self, app: web.Application) -> None:
                 web.get("/docker/registries", api_docker.registries),
                 web.post("/docker/registries", api_docker.create_registry),
                 web.delete("/docker/registries/{hostname}", api_docker.remove_registry),
+                web.post("/docker/reset-storage", api_docker.reset_storage),
```
agners, line 1077: `/docker` is allowed by Supervisor API manager role. Is that ok for this API as well? 🤔  I guess `/docker/migrate-storage-driver` is very similar and already enabled for the manager role 🤔 . For me manager is fine, just wanted to bring it up so this is a conscious decision.

agners (changes requested):
Other than the version thing, LGTM!

supervisor/api/__init__.py
sairon: Fair point. But if we want to restrict that, let's also restrict other "dangerous" endpoints in a follow-up.

supervisor/api/docker.py
sairon: Where exactly? It follows what's done a few lines above: https://github.com/home-assistant/supervisor/blob/69c5a538644eb2e8d4774293352535e3faf9e351/supervisor/api/docker.py#L140

This constraint will allow to also satisfy local dev builds, which makes testing a bit easier before dev jumps to 18.4.dev0/19.0.dev0.

sairon: Okay, here's another precedent confirming your preference: https://github.com/home-assistant/supervisor/blob/69c5a538644eb2e8d4774293352535e3faf9e351/supervisor/host/control.py#L22

Let's change it to const then and (for consistency) add a const for the above as well, so it would make a good example too.

agners (approved):
LGTM!

PR #6834, opened 2026-05-12
Fix host watchdog reset when an NFS mount's server is unreachable

agners:
## Proposed change

When a network mount's backing server becomes unreachable while the periodic mount reload runs, the host can be force-reset by the hardware watchdog within ~60 seconds. The sequence is:

1. `Mount.reload()` issues `ReloadUnit` over D-Bus. systemd spawns `mount -o remount` against the dead server; the remount child parks in NFS RPC wait.
2. After 30 s `Mount.reload()` gives up waiting and escalates by issuing `RestartUnit` on the same unit. systemd's stop phase SIGTERMs the in-flight remount child but never gets to spawn a regular `umount(8)` — the unit transitions back to `mounted` with the original kernel NFS mount still established.
3. The restart's start phase then runs `mount_enter_mounting()` on PID 1. Before forking `mount(8)`, that code does stat-family calls on the mount path (path canonicalisation, directory checks, empty-target warnings). On a still-live NFS dentry to a dead server those `fstatat()` calls route through the kernel NFS client and park in an RPC wait. **PID 1 itself** sits in D-state for the full NFS major timeout — not a worker thread, not a helper child.
4. During that window PID 1's main event loop doesn't run and the systemd watchdog ping doesn't fire. The iTCO chip resets the box.

The underlying PID 1 livelock is reported upstream as systemd/systemd#42050. The bug is reproducible with NFS; we have not been able to reproduce it with CIFS.

This PR removes the trigger from the Supervisor side:

- **Tighten the timeouts so the reload finishes before supervisor would escalate to restart.** Change the NFS `timeo` from `200` (20 s) to `100` (10 s) and make `retrans=2` (the default) explicit. With this the kernel RPC layer fails at roughly 30 s (one initial timeout plus two retransmit timeouts). Pass `TimeoutUSec=35s` as a unit property on each transient mount unit, and bump the supervisor's state-await from 30 s to 40 s. Three layered timeouts now cooperate to avoid the reload → restart escalation that triggers the bug:

    NFS RPC timeout   ~30 s
      < systemd unit `TimeoutSec`  35 s
      < supervisor state-await     40 s

[...]

Issue #6827, opened 2026-05-09
Supervisor crash when network mount (NFS) becomes unavailable during reload


Reinhard-M:
### Describe the issue you are experiencing

When a user-configured network mount (e.g. a NFS share on a device that can be powered off, like a SAT receiver or NAS) becomes temporarily unavailable, the Supervisor's new mount-reload/autofix logic introduced in **2026.05.0** triggers a crash of the Supervisor process itself, taking down Home Assistant Core along with it.

Prior to 2026.05.0 this scenario was handled gracefully — the mount would fail silently and HA continued running. After updating to 2026.05.0, a `MountActivationError` is raised inside the autofix path (`mount_execute_reload.py`) and the Supervisor crashes.

### What type of installation are you running?

Home Assistant OS

### Which operating system are you running on?

Home Assistant Operating System

### Steps to reproduce the issue

1. Configure a NFS network mount in **Settings → System → Storage** that points to a device that can be powered off (e.g. a SAT receiver with a USB hard drive shared via Samba, or a NAS with sleep/standby mode).
2. Power off or put the device into standby so the share becomes unreachable.
3. Wait for the Supervisor's automatic mount-reload interval (~15 minutes) to trigger.
4. **Result**: Supervisor crashes and restarts, taking down HA Core with it.

[supervisor_2026-05-09T08-02-43.432Z.log](https://github.com/user-attachments/files/27550873/supervisor_2026-05-09T08-02-43.432Z.log)

...

### Anything in the Supervisor logs that might be useful for us?

```txt
The Supervisor crashes and restarts. The following sequence is observable in the Supervisor log:

**Step 1 — Mount reload triggered, times out after 30 s:**

2026-05-09 02:29:35.896 INFO  [supervisor.mounts.manager] Reloading mount: VuDuo
2026-05-09 02:30:05.915 WARNING [supervisor.mounts.mount] Mount VuDuo still in state reloading after waiting for 30 seconds to complete
2026-05-09 02:30:05.915 INFO  [supervisor.mounts.mount] Mount VuDuo not correctly mounted after a reload. Trying a restart


**Step 2 — Supervisor crashes (no further log entries until restart ~49 s later):**

[gap — Supervisor process terminated]

2026-05-09 02:30:54.760 INFO  [supervisor.docker.supervisor] Attaching to Supervisor ghcr.io/home-assistant/aarch64-hassio-supervisor with version 2026.05.0
[...]

PR #7167, opened 2026-08-25
Use systemd automount units for network mounts

### LINKED ISSUE #6938: Reload NFS mount - nothing happens

agners:
## Proposed change

Pair every network mount with a systemd `.automount` unit and let the kernel's autofs machinery handle activation and recovery, instead of Supervisor orchestrating mount health itself. The `.automount` is the primary transient unit with the `.mount` as auxiliary (mirroring `systemd-mount --automount=yes`), created atomically via the `aux` parameter of `StartTransientUnit`. Media and share mounts now live directly at their container-facing path (`/media/<name>`, `/share/<name>`); the parent-directory rslave bind into containers surfaces the autofs trigger the same way it surfaced bind mounts.

This removes the entire class of mount failure modes we have been chasing (#6827, #6938, #7037, #7089):

- **PID 1 wedges and watchdog resets** (#6827): PID 1 path walks no longer cross dead NFS/CIFS lookups — autofs answers from kernel memory. The reload→restart escalation and its layered-timeout safety nets become unnecessary and are removed.
- **Bind mount teardown and local data pollution** (#6938, #7037): the inner bind layer is gone, and with it systemd's implicit `Requires=` teardown of the bind unit, the writable bare directory it exposed, and the `BoundMount` bookkeeping that could go stale. While a share is unreachable, the armed autofs trigger keeps the path a mount point — writes fail instead of landing on the local disk.
- **Server outages** self-heal: NFSv4 state manager and CIFS reconnection recover established mounts transparently, and the autofs trigger re-mounts torn-down mounts on the next access. The periodic task shrinks from remediation (reload→restart escalation) to pure reconciliation: re-arm dead triggers, refresh probe-based state, and sync the mount failed issue in both directions — so an issue raised during an outage is auto-dismissed after recovery and backup locations see fresh state.

What goes away: the inner bind-mount layer (`BindMount`, `BoundMount`, emergency read-only fallback directory), `Mount.reload()`/`Mount._restart()` escalation with the RELOADING safety net from #6834, and the periodic *remediation* of mounts (the 15-minute task remains, reduced to a probe-based reconcile that never reloads or restarts a healthy mount).

What changes shape:

- `NetworkMount.where` switches on usage: MEDIA → `path_extern_media/<name>`, SHARE → `path_extern_share/<name>`, BACKUP → `path_extern_mounts/<name>` (unchanged).
- `is_mounted()` no longer gates on systemd's ActiveState (the `.mount` is legitimately dormant until triggered); the statvfs probe is the source of truth — the statfs syscall walks with `LOOKUP_AUTOMOUNT`, so probing also arms/activates the mount. Reported mount `state` reflects reachability (ACTIVE/INACTIVE from the probe) rather than systemd unit state.
- `reload_mount()` becomes: re-arm the trigger if the `.automount` died (out-of-band unmount), probe, and — if an established mount is unreachable (permanently dead session the kernel cannot recover, e.g. a replaced server) — discard the session by stopping only the `.mount` unit while the `.automount` stays armed. systemd re-installs the autofs trigger over the path (the mechanism idle expiry uses; the automount's `Triggers=` reference keeps the transient `.mount` definition alive), so the path is never exposed as a writable directory, and the re-probe mounts fresh with a new session.
- The move-local-data repair (#7089) is carried over: with the bind layer gone there is only one target directory per mount, and a re-arm that fails on blocking local data raises the mount failed issue with the `move_local_data` suggestion.
- Backups skip network mount points when archiving media/share folders and unmount/re-arm nested mounts around folder restores.

Correctness constraints from a review of systemd's automount implementation (verified against v254.13, as shipped by HAOS) and the kernel autofs/VFS code — each of these was a silent failure mode in a naive implementation:

- The `.automount` must be the *primary* transient unit: aux units get no start job, so with the `.mount` as primary the trigger would never arm and the design silently degrades to eager mounting.
- `StartLimitIntervalUSec=0` on the `.mount`: with default start rate limiting, a fast-failing mount plus any polling consumer trips the limit within seconds and systemd then *detaches the autofs trigger entirely* (`AUTOMOUNT_FAILURE_MOUNT_START_LIMIT_HIT`) — the path silently becomes a plain writable local directory, recreating the #7037 pollution bug.
- No idle expiry (`TimeoutIdleUSec` stays 0): the kernel decides idle-expiry busyness via `may_umount_tree()`, which only sees the init namespace's mount instance — files held open by container processes are invisible, so expiry would unmount shares under actively writing add-ons.
- `LazyUnmount=true` (`MNT_DETACH`) on the `.mount`: unmount returns immediately even against a dead server; open file descriptors drain in their soft/softerr timeout regime instead of pinning the umount syscall.
- Warm upgrade: the previous design's bind unit occupies the exact unit name the network `.mount` uses now. `load()` strictly stops only that path-conflicting unit before arming (a failed stop leaves the path covered — safe and retryable), then stops the conflict-free legacy data mount best-effort *after* the automount is armed. Stopping the data mount first would not be safer: systemd unmounts the bind anyway as its dependent (the implicit `Requires=` from `RequiresMountsFor=`) before the data unmount can time out against an unreachable server, which would leave the path a plain writable directory with nothing armed.
- `Mount.unmount()` stops the `.automount` before the `.mount` so the trigger cannot re-fire the mount during cleanup.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR fixes or closes issue: fixes #6938
- This PR is related to issue: #6827, #7037, #7089, systemd/systemd#42050
- Link to cli pull request:

Notes for reviewers:

- Startup with the server offline: arming the trigger needs no server contact and always succeeds; the deliberate first probe fails within the usual ~35s budget and raises the mount failed issue. The armed trigger keeps the path a mount point throughout the outage, and the first access after the server returns mounts transparently — no reload, no repair needed. The reconcile then dismisses the issue within its 15-minute cycle. A Supervisor restart never adopts a dead trigger: `load()` re-arms a failed/stopped `.automount` and reports an unreachable adopted mount the same way a fresh mount would.
- Writable-window exposure during transitions (from an adversarial review pass): the reload escalation and the warm-upgrade teardown were restructured so the container-facing path is never left a plain writable directory — the escalation keeps the trigger armed throughout (session discard instead of unit re-creation), and the legacy teardown arms the automount before the best-effort stop of the old data mount. The one residual window is inherent and tiny: between the emptiness check and `StartTransientUnit` when first arming a path that is not yet covered (fresh mount, or re-arm after an out-of-band unmount). Writes landing before the check abort the re-arm into the `move_local_data` repair; the check-to-arm gap is two D-Bus calls.
- Issue lifecycle is probe-based, in both directions, because Supervisor is not in the access path. When a server goes offline under an established mount, the kernel keeps the mount active and no systemd unit state changes — there is no event to listen for (same in the old design; the systemd unit failure check keeps ignoring managed mount units). The mount failed issue is raised by the next probe: the 15-minute reconcile, a reload, or a Supervisor restart. Recovery is equally silent to Supervisor — access re-mounts a torn-down mount via the trigger, or the NFS/CIFS session reconnects under an attached one — so a repair can outlive the outage by up to one reconcile cycle before it is auto-dismissed (add-ons see the working share immediately). If that lag ever matters, a follow-up could subscribe to `PropertiesChanged` on the `.mount` unit's `ActiveState` for event-driven reconciliation.
- Mount `state` in the API now reports probe-based reachability instead of raw systemd unit state — a dormant-but-armed automount reports INACTIVE until first access or reload, ACTIVE after a successful probe.
- The pre-copy `is_mount()` health check in `_copy_to_location` no longer distinguishes a dead server: an armed trigger is always a mount point. The copy itself fails with a bounded ETIMEDOUT instead — slightly later error surfacing, same outcome.
- Local disk mounts (#7137) would need a chapter of their own on top of this: they currently rely on the bind layer and the periodic reload for absent-device retry. `Mount.from_dict` on this branch intentionally only accepts CIFS/NFS (user-visible mount configs were always restricted to these two by the API schema).

## Checklist


supervisor/docker/homeassistant.py
```diff
@@ -256,6 +256,10 @@ async def execute_command(self, command: list[str]) -> CommandReturn:
                     source=self.sys_config.path_extern_share.as_posix(),
                     target="/share",
                     read_only=False,
+                    # Receive mounts made after container start — without
+                    # this, accessing a not-yet-activated automount from
+                    # the container fails with ELOOP
+                    bind_options=MountBindOptions(propagation=PropagationMode.RSLAVE),
```
mdegat01, line 262: I was trying to figure out why this wasn't here already and realized this is in `execute_command` not the primary run path. I guess there's no harm but this method is only used for config check so I can't imagine share is involved much. And I think only the CLI uses runs config check this way, core can just run the referenced script directly it doesn't need Supervisor's help.

supervisor/mounts/manager.py
```diff
@@ -319,13 +260,58 @@ async def remove_mount(self, name: str, *, retain_entry: bool = False) -> Mount:
...
-        """Reload a mount to retry mounting with same config."""
+        """Probe a mount's health and surface the result.
+
+        With autofs handling re-activation transparently, a user-driven
+        "reload" reduces to: re-arm the automount trigger if it died
+        (out-of-band unmount, failure), then poke the mount via the
+        probe so the kernel triggers any pending activation, and update
+        the resolution issue accordingly. No reload/restart of a
+        healthy setup — the kernel repeats that work on its own when
+        something actually accesses the path.
```
mdegat01: This is less a comment for this PR and more something to consider adding to our agents file. I feel like our docstrings and comments are exploding in length because the AI likes to include details about what it changed and why like it would a commit message. There is definitely a point where extra verbosity is not helpful, we might want to guide the AI to focus on brevity and describing the situation as it exists now only in code. Generally avoid including history of changes in docstring and comments, leave that for commit messages and PR descriptions.

supervisor/mounts/mount.py
```diff
@@ -218,17 +250,205 @@ def __hash__(self) -> int:
...
+        exist and is adopted. If no `.automount` exists, any unit found
+        at the target path (or at the legacy data-mount location) is a
+        leftover from the eager-mount design — a warm upgrade — and is
+        torn down before the pair is set up fresh.
+        """
+        unit = await self._update_unit()
+        automount_state = await self._automount_state()
+
+        if automount_state is None:
+            await self._teardown_legacy_units()
```
mdegat01: If we recorded the version of Supervisor seen immediately after we determined it was a host reboot here:
https://github.com/home-assistant/supervisor/blob/69c5a538644eb2e8d4774293352535e3faf9e351/supervisor/core.py#L254-L257

We could use that information here. All this handling of legacy units is unnecessary if we know there has been a system reboot since the version of Supervisor including this change was installed. Probably 2026.8.0 at this point. We can check that here and skip all that extra handling.

supervisor/mounts/mount.py
```diff
@@ -218,17 +250,205 @@ def __hash__(self) -> int:
...
             await self.mount()
+            await self._cleanup_legacy_data_unit()
```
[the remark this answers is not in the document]
mdegat01: Seeing as its unreachable, we can probably live with this. Or maybe just suggest a host reboot via repair if the best effort clean up fails since that'll handle it anyway.

supervisor/mounts/mount.py
```diff
@@ -393,146 +647,117 @@ def ensure_empty_folder() -> None:
...
+                        ),
+                    )
+                    # A failed stop job is an error, not a success — treating
+                    # it as done is how a stale mount once survived a
+                    # "successful" cleanup (see #6938).
+                    if result != "done":
+                        raise MountError(
+                            f"Could not unmount {self.name} (systemd result: {result})",
+                            _LOGGER.error,
+                        )
```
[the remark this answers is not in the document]
mdegat01: Oof, ugly. But not wrong. There's no pretty way to handle this situation. Even if we put a finally to recreate the automount that can also fail and then we're just stuck with a writable directory. it'd be really nice if there was a way to get `stop_unit` to work like `start` and stop a primary and an aux as one atomic command the same way it starts it...

Without that we just have to decide on our level of comfort with error risk I guess. We can add recreation of the trigger and even retries on that but there's still always gonna be a code path that just throws up its hands and says "sorry, I couldn't fix it, its writable right now". A single retry on finally seems like a good enough best effort, we have the repair to fix local data in the way anyway.

supervisor/mounts/mount.py
```diff
@@ -393,146 +647,117 @@ def ensure_empty_folder() -> None:
...
+                "stop_unit",
+                self.sys_dbus.systemd.stop_unit(
+                    self.automount_unit_name, StopUnitMode.FAIL
+                ),
+            )
+            if result != "done":
+                raise MountError(
+                    f"Could not stop automount unit for {self.name} "
+                    f"(systemd result: {result})",
+                    _LOGGER.error,
```
mdegat01: All errors raised from unmount can make their way to the API via `DELETE /mounts/{name}`. If possible it would be good to raise translatable errors in as many of these places as possible to high-level describe what's wrong and mention that more details (like the systemd result) are in the log.

This also applies to any places added which raise generic `MountError` in `mount`. And existing ones if you or claude is feeling ambitious 😄

supervisor/mounts/mount.py
```diff
@@ -393,146 +647,117 @@ def ensure_empty_folder() -> None:
...
+            pass
         except DBusError as err:
-            _LOGGER.error(
-                "Could not reload mount %s due to: %s. Trying a restart", self.name, err
-            )
-            await self._restart()
-            self._dismiss_failed_issue()
-            return
+            raise MountError(
+                f"Could not unmount {self.name} due to: {err!s}", _LOGGER.error
```
mdegat01: Using `unmount` here feels confusing. Everywhere else unmount includes the automount trigger so it truly is gone. But here this is really just a kind of just the equivalent of a force stop and restart of a process. The automount isn't touched so at no point is your network mount disappearing from the user's perspective.

Maybe something like: `Could not stop {self.name} to force re-creation of the mount due to: {err!s}`. Maybe still too technical, I just wouldn't use the term unmount I don't think.

supervisor/mounts/mount.py
```diff
@@ -605,23 +832,37 @@ async def is_mounted(self) -> bool:
...
+                # The kernel returns ELOOP when a process loops on an
+                # autofs trigger whose activation cannot propagate into
+                # its mount namespace — a mount propagation
+                # misconfiguration, not a server problem.
+                _LOGGER.error(
+                    "Probe of mount %s failed with ELOOP — the automount "
+                    "cannot propagate into this mount namespace. Check "
+                    "mount propagation configuration",
+                    self.name,
+                )
```
mdegat01, line 873: If this happens that means the Supervisor container is misconfigured right? It has the wrong propagation mode? Is there any possible way this can occur in a supported setup?

These instructions aren't going to make any sense to an HAOS user and they won't have any way to follow them. So if its possible for it to occur on HAOS at all we probably should send it to sentry at minimum and perhaps suggest a host reboot? Just something they can actually follow that'll fix it.

If it can only occur on Supervised or when a user did a lot of tweaking at the OS level to get themselves into a bind then this is fine. If there's a sensible unsupported flag we can add it, otherwise this technical log is enough.

supervisor/bootstrap.py
```diff
@@ -255,13 +255,6 @@ def initialize_system(coresys: CoreSys) -> None:
         )
         config.path_mounts_credentials.mkdir(mode=0o600)
 
-    # Emergency folder
-    if not config.path_emergency.is_dir():
-        _LOGGER.debug(
-            "Creating Supervisor emergency folder at '%s'", config.path_emergency
-        )
-        config.path_emergency.mkdir()
-
```
mdegat01, line 264: Should we inject clean up for this folder somewhere if it exists? Otherwise it will just hang around forever until they do a restore on a new system. Maybe not a huge deal since its just a directory of directories but would be nice

mdegat01 (changes requested):
Few comments, main request for change is translatable errors in a few spots.

Whew though, this is some PR. Going to be great but we really gotta test this well before it goes out. Looks like a bit of a scary update for now trying to do this in an auto-update situation.

supervisor/mounts/mount.py
agners: Hm, yeah the legacy unit detection code path is run on every startup for each mount once :thinking: .

But I still prefer it over version detection: We haven't done that so far, and if it misfires for whatever reason it could get messy. I kinda like this "evidence based" cleanup.

supervisor/mounts/manager.py
agners: > Generally avoid including history of changes in docstring and comments, leave that for commit messages and PR descriptions.

Right, most of the time this is something for commit messages or PR description. I guess there are corner cases where a particular design might be tempting, but doesn't work for non-obvious reasons. In those cases I'd still prefer a comment in the code, to avoid anyone reverting to that tempting but non-working approach.

But in general I agree. FWIW, I've also observed similar behavior by Claude Code in OS commits.

supervisor/mounts/mount.py
agners: I think with the tweaked mount options unmounting should not fail typically. I guess we'll see how often this appears in practice. With that in mind, I am fine making this a bit technical now, if we see it often/if it confuses users, we can tweak then.

supervisor/mounts/mount.py
agners: Yes, from what I know this is most likely to happen on Supervised with an outdated Supervisor start script.

On HAOS we explicitly enable the network mount feature on 10.2 or newer, which enables the mount propagation feature (see https://github.com/home-assistant/operating-system/pull/2557).

So this essentially only can happen on unsupported systems, so I think this technical log is fine.

supervisor/mounts/mount.py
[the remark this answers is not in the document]
agners: If the stop fails, and automount creation works (and almost can't fail) this should be fine.

mdegat01 (approved):
Ok I think this looks good. Lets plan on keeping it in beta a little longer to make sure nothing comes up but I guess we can merge and get it out there soon.

One tiny note if you want to add it. Although it feels a little contradictory to my "comments should not be a git history" comment I do kind of like adding in "deprecated as of version X" comments to chunks of code with an expiration date as a small future todo reminder without needing to dig up the PR, compare to the version it was shipped in, etc. I'll leave it up to you if you want to add any of those around the code related to clean up and handling of legacy bind mount and emergency directory artifacts.

agners:
> Ok I think this looks good. Lets plan on keeping it in beta a little longer to make sure nothing comes up but I guess we can merge and get it out there soon.

Thanks a lot for the review! And yes I agree, this needs a bit longer beta phase.

> Although it feels a little contradictory to my "comments should not be a git history" comment I do kind of like adding in "deprecated as of version X" comments to chunks of code with an expiration date as a small future todo reminder without needing to dig up the PR, compare to the version it was shipped in, etc. I'll leave it up to you if you want to add any of those around the code related to clean up and handling of legacy bind mount and emergency directory artifacts.

I agree, if such a future date is somewhat clear/in the spec.

I think in this case it is not quite clear since technically we need to keep the legacy stuff forever if the user doesn't reboot :sweat_smile: .

I think we can just make a judgment call when we touch that code later if it's fine to remove then.

PR #7168, opened 2026-08-27
Add one-shot mode and docker-given totals to stats API

### LINKED ISSUE #7109: Container stats v2 API — design and implementation plan
# Container Stats v2 — Design & Implementation Plan

End-to-end plan for replacing the snapshot-based container stats API with a v2 API that has well-defined semantics, cheap sampling, and a live-update path for the frontend. Spans Supervisor, aiohasupervisor, Core (hassio integration), and frontend.

### Supervisor

- Fetch path: `DockerManager.container_stats()` (`supervisor/docker/manager.py:1008-1039`) does an inspect (`show()`) to check `running`, then aiodocker `stats(stream=False)` and takes `stats[-1]`. Every call is a fresh one-shot snapshot; Docker internally waits for two samples of its 1 s collector loop so `precpu_stats` is populated — measured ~1.01 s per call on this machine, vs ~44 ms for a `one-shot=1` request.
- Computation: `DockerStats` (`supervisor/docker/stats.py`). CPU: `(cpu_stats.total_usage - precpu_stats.total_usage) / (system_cpu_usage delta) * 100`. Since #2154 there is **no `online_cpus` multiply** — the value is "percent of total host capacity", not per-core (a container saturating 1 of 8 cores reads ~12.5). Undocumented.
- Memory: `usage - cache` with three cgroup-version fallbacks. Network/blkio are **cumulative counters**, not rates; blkio (`io_service_bytes_recursive`) is frequently empty on cgroup v2 → silently reports 0.
- Endpoints (all identical 8-key schema `cpu_percent, memory_usage, memory_limit, memory_percent, network_rx, network_tx, blk_read, blk_write`): `/addons/{app}/stats` (+ `/apps/{app}/stats` on v2), `/core/stats`, `/supervisor/stats`, `/dns|cli|observer|multicast|audio/stats`. No caching, no throttle, no `@Job`.
- Every managed container has exactly one long-lived `DockerInterface` instance that all stats consumers already funnel through: `App.instance = DockerApp(...)` (`supervisor/apps/app.py:151`), `HomeAssistantCore.instance = DockerHomeAssistant(...)` (`supervisor/homeassistant/core.py:82`), plugins likewise.
- v2 scaffolding already exists: `AppVersion` enum (`supervisor/api/const.py`), `FeatureFlag.SUPERVISOR_V2_API` (defaults **off**, `supervisor/api/__init__.py:108-111`), `/v2` sub-app + per-version `_register_*` branches in `supervisor/api/__init__.py`, v2 security patterns in `supervisor/api/middleware/security.py` (#6741, #6769, #6933).
- Push channel exists: `sys_homeassistant.websocket.supervisor_event(WSEvent.X, data)`; `WSEvent.JOB` (job progress) is the precedent. No stats push today.

### Docker stats API mechanics (verified empirically)

- **There is no bulk stats API in Docker.** Only per-container `GET /containers/{id}/stats`; even the `docker stats` CLI runs N parallel per-container requests. Any multi-container view is a fan-out on our side.
- `one-shot=1&stream=0` (API ≥1.41): single synchronous cgroup + `/proc/stat` read, **~44 ms** measured. `precpu_stats` is empty (by contract), but `cpu_stats.total_usage`, `system_cpu_usage`, `online_cpus`, `pids_stats` are populated, and the sample carries a nanosecond-precision `read` timestamp set by dockerd at collection time.
- `stream=0` without one-shot: dockerd subscribes the request to its 1 s collector loop and waits two ticks — **~1.01 s** measured (`read - preread ≈ 1.001 s`). aiodocker's `stats()` does not expose the `one-shot` param, so a small raw-query helper is needed.
- Because samples are self-timestamped and CPU counters are cumulative, **parallel fan-out staggering does not corrupt the math**: each container's percentage is computed from its own two `read`-timestamped samples. Only cross-container instant-coherence suffers (tens of ms), which nothing needs.
- The accuracy driver is **window length**, not simultaneity: `system_cpu_usage` derives from `/proc/stat` jiffies (~10 ms granularity), so a 200 ms window carries ~5 % quantization noise, 5 s ~0.2 %, 60 s negligible. Hence the minimum-sample-spacing guard in §3.1.

### Core (hassio integration)

- `HassioStatsDataUpdateCoordinator` (core #167080): 60 s interval, fetches stats **only for containers with at least one enabled stats entity** (all stats sensors are `entity_registry_enabled_default=False`), only for started add-ons. Each container is a separate GET via aiohasupervisor (`homeassistant.stats()`, `supervisor.stats()`, `addons.addon_stats(slug)`), gathered concurrently with per-container error isolation (a failing add-on logs a warning and stores `None` instead of failing the cycle).
- Only `cpu_percent` and `memory_percent` become sensors, though `ContainerStats` (aiohasupervisor `models/base.py`) carries all 8 fields.
- Supervisor-pushed events land via WS `supervisor/event` → `async_dispatcher_send(hass, EVENT_SUPERVISOR_EVENT, data)`; used for update/issue/job events, not stats.
- The `/api/hassio` HTTP proxy does **not** allow stats paths; frontend stats go through the admin WS `supervisor/api` passthrough.

### Frontend

- Single choke point: `fetchHassioStats(callWS, container)` in `src/data/hassio/common.ts` → WS `supervisor/api` proxy → `GET /{container}/stats`.
- Two consumers only:
  - App info page (`supervisor-app-info.ts`): CPU/RAM bars via `<supervisor-app-metric>`; metrics refetched when the `addon` object changes (5 s `setInterval` refreshes addon info, which cascades).
  - System information dialog (`dialog-system-information.ts`): core + supervisor stats, 10 s poll while dialog open.
- Precedent for live streaming: hardware dashboard uses Core WS subscription `hardware/subscribe_system_status` (push, ~5 s cadence, client keeps a 60-point rolling buffer for charts). Supervisor stats do NOT flow through `supervisor/subscribe` today.
- The frontend ships pinned inside Core, so frontend↔Core is version-locked.

### Known problems (prior art)

- #3275: "CPU and RAM usage not refreshing" — one-shot values stale/stuck.
- #2154 removed per-core normalization; semantics now differ from `docker stats` and are undocumented.
- Cumulative counters useless without client-side state; blkio 0 vs "no data" conflated on cgroup v2.
- Each stats call costs ~1 s inside dockerd (collector wait) plus an inspect round-trip.
- core #147390 → #167080 already moved Core to a dedicated, gated, 60 s stats coordinator — the consumer side is structured; the producer API is the weak link.

## 2. Goals

1. **Accurate, well-defined CPU**: average over a stated interval computed from dockerd's own `read` timestamps, not an instantaneous 1 s daemon-internal delta.
2. **Cheap**: ~44 ms one-shot sampling instead of the ~1 s collector wait; consumers sample exactly the containers they care about; zero cost when nobody is looking.
3. **Useful units**: rates (bytes/s) alongside cumulative counters; `null` for "unavailable" instead of 0.
4. **Live UI**: app page and dialogs update every few seconds while open — the existing frontend polling paths become cheap (~44 ms per sample) and accurate, no new plumbing.
5. **Non-breaking rollout**: v1 endpoints untouched; v2 additive behind the existing `SUPERVISOR_V2_API` feature flag.

### 3.1 Core mechanism: per-interface stats tracker (no background task)

A small stats tracker object per container, owned lazily by its `DockerInterface` instance (e.g. `self._stats_tracker` on `supervisor/docker/interface.py`, created on first stats call). All consumers of a container (Core coordinator, frontend poller via Core, add-on `self` endpoint) already converge on that one long-lived instance, so its state is automatically shared across consumers — no central registry, no name→state indexing, and uninstalling an add-on garbage-collects its tracker with the `App`/`DockerApp` objects.

Per tracker:

- **Ring buffer** of the last N (~12) timestamped raw cumulative samples, validated against **container id + started-at** — a container restart/recreate resets the kernel counters, so a mismatch drops the buffer. Negative deltas are guarded regardless.
- **On request** (`stats(window=...)`): take one `one-shot` sample now, push it into the ring, compute deltas against the buffered sample closest to `now - window` (default 60 s; oldest available if none old enough) and return averages over the **actual** elapsed interval, reported in the payload. Deltas use the samples' dockerd `read` timestamps and the `system_cpu_usage` delta as CPU denominator (robust against CPU hotplug; wall-clock × `online_cpus` would drift).
- **Minimum sample spacing (~1 s)**: if the newest buffered sample is younger than that, don't hit dockerd — compute from existing samples. Sub-second windows would amplify jiffy quantization (§1) anyway, so this doubles as an accuracy floor and a natural micro-cache when consumers overlap.
- **In-flight dedup**: at most one outstanding one-shot request per container; concurrent callers await the same future.
- **Cold start returns `null` deltas** (empty ring: first request ever, or after restart invalidation): the request still costs only one one-shot (~44 ms) and returns the point-in-time fields (state, memory, pids, cumulative counters) immediately, but `cpu_percent`, rates, and `interval` are `null` — there is no baseline yet, and we say so instead of hiding a ~1 s warm-up sleep in the request. This keeps latency predictable (always ~44 ms, never sometimes-1 s depending on invisible cache state) and makes the null rule uniform: disk rates can already be `null` (cgroup v2), so consumers must handle nullable delta fields anyway. A consumer that wants a quick value simply calls again after ≥1 s (the minimum-spacing guard is the only gate). The sample seeds the ring, so the second call gets real values.
- The one-shot mode also removes the pre-flight `show()` inspect: a 404/409 from the stats call itself tells us the container isn't running, which the API layer maps to `state` + `stats: null` (§3.2), not an error.

Stateless shared pieces: a `one-shot` raw-query helper on `DockerManager` (`docker._query_json("containers/{id}/stats", params={"stream": "0", "one-shot": "1"})`) and the pure parsing/delta math (successor of today's `DockerStats`).

Why request-driven per-interface state instead of the alternatives:

- **Background sampler**: rejected — sampling frequency should follow demand; a poller burns Docker calls when nobody looks. Request-driven gives 60 s averages to a 60 s poller and 5 s averages to an open page, automatically, with zero idle cost.
- **Central collector object keyed by container**: rejected — sharing is only ever per-container across consumers, which per-interface state already provides via object identity; a registry adds indexing and cleanup for nothing. (A central view was only needed by the bulk endpoint, which is dropped, see §3.2.)
- **Supervisor→Core push (`WSEvent.STATS` + lease/TTL)**: deferred — a ~44 ms endpoint is cheap enough to poll at 5 s; push can be layered on later without schema changes.
- **Streaming `docker stats`** (one long-lived stream per container): rejected — dockerd samples every second per stream, exactly the load we're avoiding.
- **Direct cgroupfs reads**: rejected for now — cheapest, but couples us to cgroup layout/mount availability inside the Supervisor container; the Docker API is the stable contract. Possible internal optimization later.

### 3.2 v2 API surface (Supervisor)

Per-container endpoints only, registered under the existing v2 branches in `supervisor/api/__init__.py`, gated by `SUPERVISOR_V2_API`. v1 endpoints stay exactly as they are.

- `GET /v2/apps/{app}/stats?window=60` (incl. `self` for add-on self-access, preserving the current addon-role security pattern)
- `GET /v2/core/stats?window=60`
- `GET /v2/supervisor/stats?window=60`
- `GET /v2/{plugin}/stats?window=60` (dns, audio, cli, observer, multicast)

**No bulk endpoint.** Considered and dropped: Docker has no bulk API, so a bulk endpoint only moves the per-container fan-out from Core into Supervisor without removing any Docker work; the round-trips it saves are local HTTP at 60 s cadence (~nothing); Core's coordinator already does per-container gather with per-container error isolation that a bulk handler would have to re-invent; cross-container timestamp coherence is a non-benefit (samples are self-timestamped, §1); and it would drag in a container-selection mini-language, an extra aiohasupervisor model, an extra security pattern, and duplicated tests. If a "give me everything" consumer ever materializes, a bulk endpoint composes trivially on top of the same per-interface trackers as an additive follow-up — it also would define the natural payload shape if Supervisor→Core push is ever added.

Response schema:

```json
{
  "state": "running",
  "timestamp": "2026-07-07T10:00:00.000Z",
  "interval": 58.7,
  "cpu_percent": 3.2,
  "cpu_cores": 8,
  "memory_usage": 123456789,
  "memory_limit": 2147483648,
  "memory_percent": 5.7,
  "network": { "rx_bytes": 1234, "tx_bytes": 5678, "rx_rate": 20.9, "tx_rate": 96.4 },
  "disk": { "read_bytes": 4096, "write_bytes": 8192, "read_rate": null, "write_rate": null },
  "pids": 12
}
```

Semantics, explicitly documented in the developer docs:

- `cpu_percent`: average CPU consumption over `interval` seconds, as percent of **total host capacity** (all cores = 100). `cpu_cores` (from the sample's `online_cpus`) lets consumers convert to per-core if desired. (Keeps the post-#2154 normalization the UI bars already assume, but now documented and averaged over a real window.)
- `interval`: actual seconds between the two samples used (from dockerd `read` timestamps) — consumers must not assume it equals the requested `window`. `null` on a cold tracker (no baseline sample yet), in which case all delta fields (`cpu_percent`, rates) are `null` too.
- Rates are bytes/second over `interval`; cumulative byte counters are also included so consumers can integrate.
- **`null` means "no data"** — never 0, and uniformly so: cold-start delta fields (above), and blkio on cgroup v2 with empty `io_service_bytes_recursive`. Every delta field is nullable by contract.
- Not-running containers: `state` + `stats`-fields `null` rather than a 404-style error, so consumers can poll through restarts without special casing.

Security: same per-container role patterns as the existing v1 stats endpoints, added to the v2 patterns in `security.py`.

### 3.3 Core (hassio integration)

1. **aiohasupervisor** (prerequisite release): add the v2 model (`ContainerStatsV2` with nested network/disk + nullable fields + state) and per-container v2 methods with a `window` argument — e.g. `client.addons.addon_stats_v2(slug, window=...)`, `client.homeassistant.stats_v2(...)`, `client.supervisor.stats_v2(...)` (naming TBD in that repo) — pointing at `/v2/...`.
2. **Stats coordinator**: keeps its shape (entity-subscription gating, concurrent gather, per-container error isolation) and simply switches to the v2 methods with `window=60`. Started-addon filtering becomes unnecessary since the response carries `state`. **No version gate and no fallback path**: Supervisor is always updated before Core (invariant — "old Supervisor + new Core" cannot exist), so once the aiohasupervisor bump lands, `/v2` stats are guaranteed present. The only sequencing requirement is the feature flag: `/v2` registration is gated on `FeatureFlag.SUPERVISOR_V2_API` defaulting to `False` (`supervisor/api/__init__.py:108-111`), so Core adoption must wait until the flag defaults to on in a shipped Supervisor release (recommended), or the v2 stats routes must be registered unconditionally outside the flag (only if stats shouldn't wait for general v2 availability). Cold-start responses carry `null` delta fields (§3.1); sensors map missing values to `unknown` for that first cycle, which is idiomatic. Optional nicety: when a response has `interval: null`, schedule a one-time quick re-fetch (~2 s) for that container instead of waiting a full 60 s cycle.
3. **New sensors** (all `entity_registry_enabled_default=False`, additive): network rx/tx rate, disk read/write rate, pids. They ride the same per-container calls the coordinator already makes.
4. **No new WS subscription command.** A `hassio/stats/subscribe` command (mirroring `hardware/subscribe_system_status`, shared poller, subscriber lifecycle) was considered and dropped: its main benefit — deduplicating sampling across concurrent viewers — is already provided server-side by the per-interface trackers (min spacing + in-flight dedup bound Docker sampling regardless of how many pollers exist upstream), so it would only deduplicate ~44 ms local HTTP calls from typically 1–2 open UIs, at the cost of a new Core WS command, subscriber lifecycle management, and a frontend migration. The frontend keeps polling through the existing admin `supervisor/api` WS passthrough, which handles `/v2/...` paths as-is. If Supervisor→Core push is ever added, a subscribe command is the natural conduit and can be introduced then, with the payload shape already defined by the v2 schema.

### 3.4 Frontend

The architecture stays as-is — polling through the existing `supervisor/api` WS passthrough — only retargeted to the v2 endpoints and schema:

1. `src/data/hassio/common.ts`: retarget `fetchHassioStats` to `GET /v2/{container}/stats?window=...` and replace `HassioStats` with the v2 type (nested network/disk, nullable delta fields, `state`).
2. **App info page** (`supervisor-app-info.ts`): poll its app's stats each 5 s tick (`window=5`) instead of only on addon-object change; bars go live with no architectural change. `<supervisor-app-metric>` unchanged (still feeds percent values, needs a null/loading state for cold-start); RAM tooltip keeps `memory_usage/limit`. Optional follow-up: client-side rolling buffer + sparkline like the hardware dashboard.
3. **System information dialog**: keep the 10 s `subscribePollingCollection`, retarget the two fetches (`core`, `supervisor`) to v2 with `window=10`.
4. **No version gate**: the frontend ships pinned inside Core, which requires the latest Supervisor — the only prerequisite is the same as Phase 3 (v2 routes reachable, i.e. the feature flag defaulting on). No dependency on any Core-side stats change.

## 4. Implementation phases

**Phase 1 — Supervisor (this repo)**
1. `DockerManager` one-shot raw-query helper (+ map 404/409 to not-running).
2. Per-interface stats tracker on `DockerInterface`: ring buffer, restart detection (container id + started-at), `read`-timestamp/window selection, minimum sample spacing, in-flight dedup, rate computation, `null` semantics incl. cold-start null deltas. Unit tests with synthetic sample sequences (restart reset, negative delta, empty blkio, first sample, overlapping consumers, sub-second re-request).
3. v2 per-container handlers (extend the existing stats handlers with the new schema + `window` param under the v2 branches); security middleware patterns; API tests (both feature-flag states).
4. Docs PR to developers.home-assistant (API reference + semantics).

**Phase 2 — aiohasupervisor**
5. `ContainerStatsV2` model + per-container v2 methods + tests; release; version bump in Core.

**Phase 3 — Core** (prerequisite: `SUPERVISOR_V2_API` defaults to on in a shipped Supervisor release — or stats routes registered outside the flag)
6. Coordinator switch to the v2 methods (`window=60`); drop the started-addon filtering; no fallback (Supervisor-before-Core invariant).
7. New rate/pids sensor descriptions (disabled by default).

**Phase 4 — Frontend** (same prerequisite as Phase 3; independent of the Core changes)
8. Retarget `fetchHassioStats` + type to v2; poll app stats on the app page's 5 s tick; retarget the system info dialog fetches.
9. (Optional follow-up) sparkline charts on the app page.

Each phase is independently shippable; v1 consumers keep working throughout.

## 5. Open decisions

1. **CPU normalization** — plan keeps host-capacity percent (documented, `cpu_cores` exposed). Alternative: per-core percent like `docker stats`. Recommendation: keep host-capacity; the UI bars are 0–100 gauges.
2. **Default `window`** — 60 s proposed (matches Core coordinator); frontend subscription uses 5 s.
3. **Serve v1 endpoints from the tracker too** — free accuracy/latency fix for legacy consumers, but changes v1's instantaneous semantics slightly. Recommendation: yes, as a separate small PR after Phase 1 settles.
4. **Feature-flag sequencing** — Phase 3 depends on `/v2` being reachable on every user's Supervisor. Recommendation: let stats ride the flag and start Phase 3 when `SUPERVISOR_V2_API` flips to default-on; register the stats routes unconditionally only if stats shouldn't wait for that.
5. **Host stats in the v2 API** — Core's `hardware` integration already streams host CPU/memory; adding host stats here would duplicate it. Recommendation: leave out.

mdegat01:
## Proposed change

Updates the `/stats` API for Home Assistant Core, Apps, Supervisor, and plugins (Audio, Cli, DNS, Multicast, Observer) to return docker-given totals and adds a one-shot mode. Also refactors `container_stats()` in `supervisor/docker/manager.py` with translatable errors and timeout handling, and introduces a consistent, component-flavored error translation pattern (`NotRunningError`/`StatsTimeoutError`/`UnknownError`) across all stats-producing components so API consumers get clear, translatable error messages instead of raw Docker errors.

Follow-up commit:

- Contains the reach into aiodocker's protected `_query` internals (needed to request one-shot stats until [aio-libs/aiodocker#1054](https://github.com/aio-libs/aiodocker/pull/1054) adds native support) to a single `DockerAPI._query_one_shot_stats()` helper, so it's easy to remove once that lands.
- Splits the stats response model between the v1 and v2 API contracts: v1 keeps the existing opt-in `one_shot` query string and its windowed `cpu_percent` field; v2 always requests one-shot stats and drops `cpu_percent` from its response, since it would never carry a meaningful (windowed) value in that mode. Both share the same `api_return_stats(stats, legacy=...)` helper, so the response model stays defined in one place. `DockerAPI.container_stats()` is unchanged aside from the internal refactor above; the legacy/v1 vs v2 distinction is handled entirely in the API layer.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR fixes or closes issue: fixes #7109
- Link to cli pull request:

## Checklist


supervisor/apps/app.py
```diff
@@ -1378,13 +1382,14 @@ def is_running(self) -> Awaitable[bool]:
...
 
-    async def stats(self) -> DockerStats:
+    async def stats(self, *, one_shot: bool = False) -> DockerStats:
         """Return stats of container."""
         try:
-            if not await self.is_running():
-                raise AppNotRunningError(_LOGGER.warning, app=self.slug)
-
-            return await self.instance.stats()
+            return await self.instance.stats(one_shot=one_shot)
```
agners, line 1388: I guess this will need https://github.com/aio-libs/aiodocker/pull/1054 to be merged/released first.

agners (changes requested):
Adding one shot as a query parameter avoids breaking the current API, which has both, upsides and downsides: Upside is we can deploy now, with v1, and transition case by case if we want to. Downside is It leaves the IMHO weird 1s CPU load snapshot in the API as a supported use case, along with the calculation which is always run (`_calc_cpu_percent`). I think I'd used the opportunity of v2 to get rid of that 1s CPU load snapshot entirely 😅 🤷 . But if you have other arguments/or feel strongly about deploying this in a backwards compatible way, I am also fine with it. Main aim here is to get the Core CPU load stats by default more accurate, and this enables that 👍 .

There are some linter errors.

supervisor/api/utils.py
```diff
@@ -99,6 +111,23 @@ def json_loads(data: Any) -> dict[str, Any]:
         raise APIError("Invalid json") from err
 
 
+def api_return_stats(stats: DockerStats) -> dict[str, Any]:
+    """Return the standard API response dict for a DockerStats object."""
+    return {
+        ATTR_CPU_PERCENT: stats.cpu_percent,
+        ATTR_CPU_USAGE: stats.cpu_usage,
+        ATTR_CPU_SYSTEM_USAGE: stats.cpu_system_usage,
+        ATTR_ONLINE_CPUS: stats.online_cpus,
```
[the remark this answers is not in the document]
agners: We've added fields in the past, so this is not really valid, it is a backwards compatible extension of the v1 (and v2) API.

supervisor/apps/app.py
mdegat01: oh nice! I forgot you already put that up, was gonna do that today. We don't have to wait for it tbh, can just add a task to change it over when that gets merged. It should be isolated to a one location change already, if not I'll update it to make it so 👍

supervisor/docker/manager.py
```diff
@@ -1003,38 +1007,71 @@ async def container_logs(self, name: str, tail: int = 100) -> list[str]:
...
+                if err.status == HTTPStatus.NOT_FOUND:
+                    raise DockerContainerNotFoundError(
+                        _LOGGER.warning, name=name
+                    ) from None
+                _LOGGER.error("Can't read stats from %s: %s", name, err)
+                raise DockerStatsUnknownError(name=name) from err
+
+            if not stats:
+                _LOGGER.error("Docker returned no stats for %s", name)
+                raise DockerStatsUnknownError(name=name)
```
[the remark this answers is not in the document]
mdegat01: Is this true for the one-shot call as well? Do you know @agners ? I'll test it tomorrow if not. I assumed since the one-shot call returned lifetime information for the container it would continue to return the lifetime information and that just wouldn't be changing while the container was stopped. I actually removed checks for the container being stopped in the one-shot path for this reason, figured it was a good way to reduce I/O.

If this is a valid issue for one-shot calls I'll address it (though maybe not here specifically, before this point). If this is only valid for windowed calls I'm going to say just leave it. When making a windowed call it still checks if the container is stopped first and immediately errors out, never getting to this point.

mdegat01:
Ok @agners I updated it per our discussion:
1. V1 gets a new one-shot option. Its an additive change so its compatible and then we can start using it immediately
2. V2 defaults to one-shot and currently does not support the windowed option. I figured when we get to making the client library and cli work for v2 we can decide then if we want to do one-shot only, handle the windowed calculations client side or add a `window=` option to the V2 API and let Supervisor handle it.

Basically leave the V2 decision for later and don't add something we may not need now.

supervisor/apps/app.py
mdegat01: Isolated the touch point to one tiny helper we can easily replace later. I'm ok with merging this now and coming back when that PR goes in. Or we can put it in draft for now and set a deadline or something where we merge anyway if the other PR still hasnt gone in.

agners (changes requested):
> Basically leave the V2 decision for later and don't add something we may not need now.

Nice, yeah I prefer this form.

supervisor/docker/stats.py
```diff
@@ -8,7 +8,10 @@ class DockerStats:
 
     def __init__(self, stats):
         """Initialize Docker stats."""
-        self._cpu = 0.0
+        self._cpu_percent: float | None = None
```
agners, line 11: Just a FYI, the client library for v1 (if we set one_shot parameter) won't like this. But I guess this is acceptable since we'll have to add the other fields as well anyways, and noone is using `one_shot` just yet.

supervisor/docker/stats.py
```diff
@@ -8,7 +8,10 @@ class DockerStats:
...
+        self._cpu_usage = 0
+        self._cpu_system_usage = 0
+        self._online_cpus = 0
```
agners, line 14: If a container is not running, all will be zero, which I guess is fine, but this particular field potentially leads to division by zero for calculations. We probably should point out that in particular online_cpu 0 can be used to detect if container is not running, and skip the whole calculation...

Unless we want to go some other way to detect not running containers (/in v2)?

agners (changes requested):
Whoops, clicked too early... Some findings.

supervisor/docker/stats.py
mdegat01: Yes, the schema change in a vacuum would break it. However we can get away with it here because:
1. Newer versions of the library with the `one_shot` option will have the widened typing for the field
2. Older versions of the library will not have any way to ask for a `one_shot` type response. And Supervisor always returns a `float` in this field when `one_shot=False`.

So I think the only way it could break on older versions of the library would be if someone made an HTTP call directly to Supervisor and included `one_shot` and then tried to create a `DockerStats` object out of the response with an old version of the library. I think we can live with that gap 😆

supervisor/docker/manager.py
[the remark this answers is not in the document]
mdegat01: Good catch — confirmed this was a real bug. Docker's `daemon.ContainerStats()` checks whether the container is running before branching into the `OneShot` path, so a stopped/restarting container in one-shot mode was returning a stats payload without `cpu_stats.online_cpus` instead of erroring out. Fixed in 40c764e by explicitly checking for that field and raising `DockerContainerNotRunningError` when it's absent, matching the windowed-mode behavior. Added test coverage for the empty-response case as well.

supervisor/docker/stats.py
mdegat01: Yes you're right. I looked it over with claude, I didn't realize it worked like this. What I've been referring to as lifetime counters isn't really that. It's really a cumulative counter for the time that container has been running, it gets wiped when the container is stopped. I didn't realize this.

Good news though, this is fixed and research around this identified a couple neat things:
1. The presence of the `online_cpus` field can be used to determine if a container you asked for stats on is stopped or not. So we don't have to first inspect the container to see if its running, we can just ask for stats and then use that to decide whether to raise `DockerContainerNotRunningError` or proceed, no all 0 responses and one less I/O call.
2. This also works in the windowed stats call. The windowed stats call returns immediately with an all 0s response if the container is stopped, it does not wait for the window to collect multiple samples. So we can skip the show/inspect I/O there too
3. It turns out that under the hood `containers.get` also does a show/inspect. So our logic of doing `container.get` followed by `show` was particularly wasteful. Claude found a way to construct a container obect without a show by doing `.containers(<name>)`. Which means both code paths are now down to a single I/O call that tells us all we need to know.

I realize the windowed stats path may be deprecated and gone in a bit but its used for now so performance improvements are welcome I think. Plus it makes the code a lot cleaner as you can see in 40c764e04 .

Point 3 also makes me think there are a lot of other places we are perhaps being wasteful with I/O. I'm going to follow up on that.

agners (approved):
Nice, LGTM!

PR #7172, opened 2026-08-28
Preserve discovery message uuids across app backup/restore

agners:
## Proposed change

Home Assistant links the config entry it creates from a discovery message to the uuid of that message. Several integrations make that uuid the unique ID of their config entry, most recently the Model Context Protocol integration in home-assistant/core#180378, so that uninstalling the app removes the entry and an app moving to another port updates the URL of the existing entry. Home Assistant also keys its rediscovery on it: removing such an entry re-triggers the discovery flow by looking the uuid up in Supervisor (home-assistant/core#127088).

The uuid identifies an installation of an app. It is minted when the app first announces a service and stays the same for as long as the data of the app does, since Supervisor matches a re-sent message on app and service and keeps the uuid. It only lives in `discovery.json` in Supervisor's own data directory though, which is not part of a backup. A restore brings the data of the app back, but not the uuid that identified it, so the app announces itself under a freshly generated one while the config entries restored with Home Assistant still reference the old one. The entry is then no longer removed together with the app, rediscovery finds nothing, and an app which comes back with a changed URL creates a second config entry next to the stale one.

This stores the discovery messages of an app in its backup metadata (`addon.json`) and restores them with their uuid, before the app can announce itself again. Backups written by older Supervisor versions carry no such metadata and restore exactly as before.

The uuid follows the data of the app. On restore this means:

- No live message for the service: the message of the backup is restored as is.
- Live message under the same uuid: same installation, only its config is taken from the backup, as the restore may bring back another version of the app which announces a different config.
- Live message under another uuid: it belongs to an installation which the restore just replaced. It is retracted as an uninstall would, so that Home Assistant drops its config entry, and the message of the backup takes its place. The app is then rediscovered under the uuid of the backup, which is what the config entries restored with Home Assistant reference.
- A message is skipped when the restored app no longer provides the service, or when its uuid is already taken by another message, which it would otherwise evict. A backup listing a service twice keeps the first message.

A restored message is not announced to Home Assistant during the restore. The service behind it is not up at that point, and it never comes up when the app is not started again. Only the app knows when its service is up and it sends its discovery message on every start, so that is where the push belongs. Such a send would normally be dropped as a duplicate, because the config the app sends is the config the backup restored, hence a restored message is marked as unannounced and pushed on the first send it gets.

### How the Core integrations deal with the discovery uuid

Every `async_step_hassio` in Core as of today. "New uuid" is a discovery for a service which already has a config entry, arriving under a uuid that entry does not know, which is what a restore without this PR produces. "Supervisor delete" is the delete Supervisor pushes on uninstall, and under this PR also for the message of a replaced installation; Home Assistant only removes entries whose unique ID equals the uuid.

| Integration | uuid stored as unique ID | Reaction to a new uuid | Entry removed on Supervisor delete | With the uuid preserved by this PR |
|---|---|---|---|---|
| otbr | yes | Not re-linked. The host fallback of home-assistant/core#179551 only matches entries without a unique ID, from the first version of the integration. When the existing entry is loaded, the router is found in use and the flow aborts, leaving the entry on the old uuid. When it is not loaded, which is the case while the app restarts, a second entry is created. | yes | Restored entries match the uuid the app announces. A replaced installation loses its entry through the delete and is rediscovered under the uuid of the backup. |
| wyoming | yes | Re-links the entry with the same host and port and adopts the new uuid. | yes | As otbr. |
| mcp | yes | Aborts on the matching URL, the entry keeps the old uuid. A changed URL creates a second entry. Shipped with home-assistant/core#180378. | yes | As otbr. Without this PR a restore onto a fresh system leaves the entry on a dead uuid, so it is no longer removed with the app and a changed URL duplicates it. |
| pyload, uptime_kuma | yes | Abort on the matching URL before looking at the uuid, the entry keeps the old uuid. A changed URL creates a second entry. | yes | As mcp. |
| adguard, matter, mealie, motioneye, mqtt, onewire | no, placeholder from `_async_handle_discovery_without_unique_id()` | Aborts whenever any entry of the integration exists. | never | Unaffected. |
| deconz, zwave_js, music_assistant | no, device identity (bridge ID, home ID, server ID) | Matches the device, updates host or URL. | never | Unaffected. |
| vlc_telnet | no, fixed `hassio` | Aborts, updates the config. | never | Unaffected. |
| esphome | no entry created | Records the dashboard location and aborts. | never | Unaffected. |

Independent of the unique ID, Home Assistant keeps the uuid of every discovery flow in the `discovery_keys` of the resulting entry and looks it up in Supervisor when the entry is removed, to offer the discovery again (home-assistant/core#127088). That lookup finds nothing for a uuid a restore did not preserve, for every integration in the table.

The retraction has a cost on a restore of a single app while Home Assistant keeps running and the app was reinstalled since the backup: the uuid keyed integrations lose their entry and offer the rediscovery again, so the user confirms the flow once more. That is the uninstall and reinstall Home Assistant would have seen if the app had been removed and set up again, which is what such a restore is.

Discovery data is still not cleaned up when an app update drops a service from its `discovery` config. That gap is independent of backups and left for a separate PR. Treating the uuid as a plain message handle in Home Assistant instead, keyed on app and service, is a larger change on the Core side and to be looked at separately as well.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR is related to issue: home-assistant/core#180378
- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/discovery/__init__.py
+++ b/supervisor/discovery/__init__.py
@@ -46,6 +46,10 @@ class Discovery(CoreSysAttributes, FileConfiguration):
         super().__init__(FILE_HASSIO_DISCOVERY, SCHEMA_DISCOVERY_CONFIG)
         self.coresys: CoreSys = coresys
         self.message_obj: dict[str, Message] = {}
+        # Messages restored from a backup which Home Assistant was not told
+        # about yet. Only the app knows when its service is actually up, so the
+        # push waits for the app to send the message itself (see send()).
+        self._unannounced: set[str] = set()
 
     async def load(self) -> None:
         """Load exists discovery message into storage."""
@@ -76,6 +80,10 @@ class Discovery(CoreSysAttributes, FileConfiguration):
         """Return list of available discovery messages."""
         return list(self.message_obj.values())
 
+    def messages_for_app(self, slug: str) -> list[Message]:
+        """Return list of discovery messages sent by an app."""
+        return [message for message in self.list_messages if message.app == slug]
+
     async def send(self, app: App, service: str, config: dict[str, Any]) -> Message:
         """Send a discovery message to Home Assistant."""
         # Create message
@@ -88,6 +96,10 @@ class Discovery(CoreSysAttributes, FileConfiguration):
             if exists_msg.config != config:
                 message = exists_msg
                 message.config = config
+            elif exists_msg.uuid in self._unannounced:
+                # Restored from a backup, this is the first time the app
+                # announces it, so Home Assistant still needs to hear about it
+                message = exists_msg
             else:
                 _LOGGER.debug("Duplicate discovery message from %s", app.slug)
                 return exists_msg
@@ -97,6 +109,7 @@ class Discovery(CoreSysAttributes, FileConfiguration):
             "Sending discovery to Home Assistant %s from %s", service, app.slug
         )
         self.message_obj[message.uuid] = message
+        self._unannounced.discard(message.uuid)
         await self.save()
 
         self.sys_create_task(self._push_discovery(message, CMD_NEW))
@@ -105,6 +118,7 @@ class Discovery(CoreSysAttributes, FileConfiguration):
     async def remove(self, message: Message) -> None:
         """Remove a discovery message from Home Assistant."""
         self.message_obj.pop(message.uuid, None)
+        self._unannounced.discard(message.uuid)
         await self.save()
 
         _LOGGER.info(
@@ -113,6 +127,47 @@ class Discovery(CoreSysAttributes, FileConfiguration):
             message.app,
         )
         self.sys_create_task(self._push_discovery(message, CMD_DEL))
+
+    async def restore_app_messages(self, app: App, messages: list[Message]) -> None:
+        """Restore discovery messages of an app from a backup.
+
+        Home Assistant ties a config entry to the uuid of the discovery message
+        it came from, so the uuid has to survive a backup/restore to keep that
+        link intact. Messages are only restored for services without a live
+        message: an existing one already carries the uuid Home Assistant knows,
+        restoring on top of it would break the link instead of preserving it.
+
+        A restored message is not pushed to Home Assistant here. The service
+        behind it is not up at this point, and it may never come up if the app
+        is not started again. Instead it is marked as unannounced so that the
+        push happens once the app sends the message itself.
+        """
+        live_services = {message.service for message in self.messages_for_app(app.slug)}
+        restored = False
+
+        for message in messages:
+            if message.service in live_services:
+                continue
+            if message.service not in app.discovery:
+                _LOGGER.info(
+                    "Skipping discovery message for service %s, app %s does not provide it anymore",
+                    message.service,
+                    app.slug,
+                )
+                continue
+
+            self.message_obj[message.uuid] = message
+            self._unannounced.add(message.uuid)
+            restored = True
+            _LOGGER.info(
+                "Restored discovery %s for service %s from %s",
+                message.uuid,
+                message.service,
+                app.slug,
+            )
+
+        if restored:
+            await self.save()
 
     async def _push_discovery(self, message: Message, command: str) -> None:
         """Send a discovery request."""
--- a/supervisor/discovery/__init__.py
+++ b/supervisor/discovery/__init__.py
@@ -142,11 +142,15 @@ class Discovery(CoreSysAttributes, FileConfiguration):
         is not started again. Instead it is marked as unannounced so that the
         push happens once the app sends the message itself.
         """
-        live_services = {message.service for message in self.messages_for_app(app.slug)}
+        # Services which already have a message, grown while restoring so that
+        # a backup listing a service twice does not end up with two messages
+        known_services = {
+            message.service for message in self.messages_for_app(app.slug)
+        }
         restored = False
 
         for message in messages:
-            if message.service in live_services:
+            if message.service in known_services:
                 continue
             if message.service not in app.discovery:
                 _LOGGER.info(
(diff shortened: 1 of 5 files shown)


supervisor/discovery/__init__.py
```diff
@@ -114,6 +128,47 @@ async def remove(self, message: Message) -> None:
...
+            if message.service not in app.discovery:
+                _LOGGER.info(
+                    "Skipping discovery message for service %s, app %s does not provide it anymore",
+                    message.service,
+                    app.slug,
+                )
+                continue
+
+            self.message_obj[message.uuid] = message
+            self._unannounced.add(message.uuid)
```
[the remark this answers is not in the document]
agners: Good catch on both, fixed in ace2ec2cf.

Messages are keyed by uuid across all apps, so an already taken uuid is now skipped with a warning instead of evicting the message it collides with, which could have belonged to another app.

The set of services which already have a message is now grown while restoring, so a backup listing the same service twice keeps only the first message. That was the worse of the two: both messages pass validation and both get served to Home Assistant once the app is started, while only one of them can ever be updated or removed again, as send() and remove() match on app and service.

Both need a malformed backup to trigger rather than normal operation, but the input is a tar file and the store is global and persisted, so they are worth guarding. Tests added for each.

supervisor/discovery/__init__.py
```diff
@@ -114,6 +128,62 @@ async def remove(self, message: Message) -> None:
...
+        # Services which already have a message, grown while restoring so that
+        # a backup listing a service twice does not end up with two messages
+        known_services = {
+            message.service for message in self.messages_for_app(app.slug)
+        }
+        restored = False
+
+        for message in messages:
+            if message.service in known_services:
+                continue
```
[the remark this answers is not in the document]
agners: Deliberate, the live message wins in every mode.

The case is real: a full restore does not reinstall an app which is already installed, remove_delta_apps only uninstalls apps missing from the backup, so the live message and its uuid survive and a restored config entry of Home Assistant can be left stale.

Propagating whether Home Assistant is restored would only cover part of it though. An app which is not part of the restore keeps its message no matter what, as restore_app_messages is never called for it, so the same mismatch stays reachable. That would fix one variant and leave its neighbour behaving differently.

So Supervisor decides instead: its live message is the one the app announces, and a stale entry is the accepted cost. The integration keeps working, the entry is just no longer removed together with the app. Documented in 6288bcdce, and the PR description now spells out the rule and its cost.

agners:
Maybe the better fix is to actually create a "app installation instance" a UUID which we reuse in discovery? This would then naturally be backed up with the app backup.

But then if an app has multiple discovery services they would have the same UUID, so this probably won't work either. Unless we key it by discovery service, but have it in the app data instead of the discovery data (`discovery.json`).

supervisor/discovery/__init__.py
```diff
@@ -114,6 +128,96 @@ async def remove(self, message: Message) -> None:
...
+            if (exists_msg := known.get(message.service)) is not None:
+                if exists_msg.uuid != message.uuid:
+                    _LOGGER.info(
+                        "Keeping discovery %s for service %s of app %s, it replaces %s of the backup",
+                        exists_msg.uuid,
+                        message.service,
+                        app.slug,
+                        message.uuid,
+                    )
+                    continue
```
mdegat01, line 181: This section right here feels like the better answer to be honest. You're trying to identify duplicate messages here to know when one needs to be suppressed. To do that you're ignoring UUID and instead considering messages to be duplicate when the values in app and service are identical.

Instead of trying to fix this in Supervisor perhaps a better fix would be in core? Do not store the UUID with the config entry anymore. Instead use the combination of app + service as the unique identifier for the discovery message tied to the config entry. If core needs the UUID for some Supervisor interaction, it can obtain the current one via lookup with the two fields it has.

This fully solves the problem you're trying to tackle here. And it fixes it for existing backups as well, not just for backups made from this point forward. And puts the control back in the hands of apps. Supervisor remains just a messenger from apps to core and doesn't attempt to figure out exactly what the app is or isn't providing anymore after a backup restore.

It's a bit of a hassle to do this now of course since we'd have to update existing config entries with the new identifier. Unless there's no way to do that? I can't say I've tried it but I feel like there must be... 🤔

mdegat01 (changes requested):
Not sure if this is best approach. Take a look at my comment for an alternative way to tackle this issue and let me know what you think.

supervisor/discovery/__init__.py
agners: Hm, reading the code, this actually was not what I intended to do: My intention was to preserve UUID from the backup, essentially making sure the UUID stays the same across a installation lifecycle (as long as `/data` is the same). So this would mean on restore, two things should happen: If the app is already installed, and the UUID is the same, then simply accept the backup config and go on. If the app is already installed, but the local UUID is different then the backup one, then the backup wins since we replace the `/data` as well. Supervisor should send a delete as if the app got uninstalled, and then allow a new discovery to be sent out once the app sends its discovery (with the backup UUID).

I've updated the PR accordingly, so we have this approach.

> Instead of trying to fix this in Supervisor perhaps a better fix would be in core? Do not store the UUID with the config entry anymore. Instead use the combination of app + service as the unique identifier for the discovery message tied to the config entry. If core needs the UUID for some Supervisor interaction, it can obtain the current one via lookup with the two fields it has.

Yeah, I think looking back on how discovery works, this was the intended behavior. For some reason Core started to read into UUID more than it represents :cry: .

The shift actually happend with https://github.com/home-assistant/core/pull/91970 (April 2023) "Include UUID in hassio discovery data to support multiple addons providing the same service", with the TODO "update integrations to set the config entry unique_id to the addon UUID". That is the moment Core started reading identity into it. slug had already been in HassioServiceInfo since https://github.com/home-assistant/core/pull/80094 (Oct 2022), so (slug, service) would have solved the same problem.

The question really becomes where to go from here.
a) Accept discovery UUID as "installation identity", and implement the consequences of it. Essentially what this PR attempts to do (e.g. make sure it stays stable accross backup restore).
b) Make it clear that discovery UUID is not meant to be used as an "installation identity", and adjust all Core integrations accordingly. Maybe remove UUID from the discovery data again (revert https://github.com/home-assistant/core/pull/91970) to avoid Core integrations reading too much into it.

agners: Let's have this discussion/decision in a task: https://github.com/home-assistant/supervisor/issues/7193

PR #7181, opened 2026-08-31
Add `/time` API for setting NTP servers on HAOS

### LINKED ISSUE #6278: Support setting NTP server
### Description

Since https://github.com/home-assistant/os-agent/pull/207 OS Agent supports configuring NTP servers. This requires Supervisor backend support.

Supervised requires systemd-timesyncd, so technically this should also work on Supervised.

This also needs CLI support, e.g. `ha network ntp`  to set global NTP server. Or maybe even a dedicated `ha ntp`.

### Additional context

_No response_

sairon:
## Proposed change

Add API for customizing NTP servers, using fixed DBus methods available since OS 18.3. The API is limited to HAOS only with OS Agent that sets the values in a daemon drop-in config. Changing the values triggers a systemd-timesyncd unit restart.

The API currently exposes only the configuration overrides, not the currently applied value (like `timedatectl show-timesync --property=ServerName` does). If we want to add the effective NTP server, another field should be introduced in the API for this.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR fixes or closes issue: fixes #6278
- Link to documentation pull request: https://github.com/home-assistant/developers.home-assistant/pull/3321
- Link to cli pull request: https://github.com/home-assistant/cli/pull/675
- Link to client library pull request: https://github.com/home-assistant-libs/python-supervisor-client/pull/403

## Checklist


supervisor/api/time.py
```diff
@@ -0,0 +1,60 @@
...
+import voluptuous as vol
+
+from ..const import ATTR_SERVERS
+from ..coresys import CoreSysAttributes
+from ..exceptions import APINotFound
+from ..host.const import HostFeature
+from .const import ATTR_FALLBACK_SERVERS
+from .utils import api_process, api_validate
+
+RE_NTP_SERVER = re.compile(r"^[^\s#]+\Z")
```
agners: This allows null characters, which would break D-Bus connection. Maybe better to just allow list instead:

```suggestion
RE_NTP_SERVER = re.compile(r"^[a-zA-Z0-9._:%-]+\Z")
```

supervisor/host/manager.py
```diff
@@ -136,6 +136,13 @@ def supported_features(self) -> list[HostFeature]:
         ):
             features.append(HostFeature.MOUNT)
 
+        if (
+            self.sys_os.available
+            and self.sys_dbus.agent.timesyncd.is_connected
+            and self.sys_dbus.agent.version >= NTP_MIN_OS_AGENT_VERSION
```
agners, line 143: Claude says this version compare could fail due to None on left side. In other places we have `self.sys_dbus.agent.is_connected` to prevent that, maybe we can do this here too.

agners:
As raised in our discussion, maybe we should consider if `/ntp` is a good fit. I had Claude take a stab at it, this is what I got:


# Where to anchor the NTP settings API (PR #7181)

Brainstorm on the API namespace for NTP/time configuration, prompted by review of
https://github.com/home-assistant/supervisor/pull/7181.

## What the codebase already says about "where time lives"

The time domain **already lives under `/host` today** — more than one might remember:

- `GET /host/info` returns `timezone`, `dt_utc`, `dt_synchronized`, `use_ntp`
  (`supervisor/api/host.py:179-183`). So read-side NTP state is already a `/host`
  concern.
- `supervisor/host/control.py` already has `set_datetime()` (used by `core.py`'s
  boot-time clock repair) and `set_timezone()` — and PR #7181 itself adds
  `set_ntp_servers()` **into that same file**. The manager-layer code is anchored in
  host; only the API is anchored at top level. That mismatch is itself a signal.
- The `system_enable_ntp` fixup toggles NTP via `timedate1` — host plumbing again.

Meanwhile `/dns` is a **plugin namespace** — it configures the CoreDNS plugin
container, sitting next to `/audio`, `/cli`, `/multicast`, `/observer`. A top-level
`/ntp` visually joins that family and falsely suggests an NTP plugin exists. So `/dns`
is precedent *against* top-level `/ntp`, not for it.

## Why /os and /host are separate at all

The split is by **abstraction contract**, not by physical machine:

- **`/host`** = "any systemd Linux host", spoken through standard freedesktop D-Bus
  APIs: `systemd1`, `hostname1`, `timedate1`, `resolved1`, `logind`, UDisks2,
  NetworkManager. Historically this worked on Supervised (Debian) installs too.
- **`/os`** = "Home Assistant OS the product": RAUC updates, boot slots, data disk,
  Raspberry Pi firmware, and OS Agent config like swap. Gated on `os_available`,
  meaningless on Supervised.

Swap sits under `/os/config/swap` because swap-on-the-data-partition genuinely *is* an
OS-image implementation detail. NTP servers are not — "which time servers does my host
use" is a host question on any Linux box. It's only the **write mechanism** (OS Agent
authoring a timesyncd drop-in, because Supervisor can't write the host's `/etc`) that
is HAOS-specific. Anchor by concept and it's `/host`; anchor by mechanism and it's
`/os`. Supervisor's own precedent (hostname, timezone, datetime) anchors by concept.

## Two different "timesyncd" D-Bus surfaces — and the "currently used server" point

This is the architecturally interesting bit. There are two independent interfaces:

1. **`io.hass.os` → `Config.Timesyncd`** (OS Agent, what the PR uses): *writes
   config*. HAOS-only, version-gated.
2. **`org.freedesktop.timesync1`** (systemd-timesyncd itself): *runtime status*,
   read-only. Generic systemd, **no OS Agent needed** — Supervisor could wrap it
   exactly like it wraps `timedate1`. It exposes `ServerName` and `ServerAddress` (the
   currently selected server — which answers "what's actually in use" directly,
   whether it came from config, DHCP, or fallback), plus `PollIntervalUSec`,
   `Frequency`, and `NTPMessage` (stratum, leap, root delay/dispersion, offset, delay,
   jitter, packet count — everything `timedatectl timesync-status` prints).

So the natural shape is: **config** (agent-gated, HAOS ≥ 18.3) and **status**
(available on anything running timesyncd) as two capabilities in one namespace with
per-field availability — another reason not to name the namespace after the config
mechanism.

On "learned through DHCP": HAOS has honored DHCP-provided NTP servers since 2022 — a
NetworkManager dispatcher script writes DHCP option 42 into
`/run/systemd/timesyncd.conf.d/10-ntp.conf` and restarts timesyncd
(https://github.com/home-assistant/operating-system/pull/1798). Since HAOS 18.3, the
whole timesyncd configuration is layered as drop-ins
(https://github.com/home-assistant/operating-system/pull/4988 — the mechanism the
OS Agent API builds on), from lowest to highest priority:

* `10-ntp.conf` (in `/run`): NTP servers from DHCP
* `20-custom.conf`: `timesyncd.conf` imported from the CONFIG partition
* `50-os-agent.conf`: NTP servers set through the OS Agent D-Bus API

Because a later drop-in's `NTP=` replaces an earlier one's, setting servers through
the new API **silently overrides DHCP-learned servers** (and any CONFIG-partition
customization). That makes a status surface showing the effective server *and its
source* (`dhcp` / `custom` / `os-agent` / `fallback`) more valuable, not less —
`timesync1.ServerName` reports the effective server truthfully in every case, though
attributing it to a layer would mean inspecting which drop-ins exist.

## The full future feature space

**timedatectl / timedate1** offers: timezone get/set + `ListTimezones`, NTP on/off
(`SetNTP` — the existing fixup), `NTPSynchronized`, manual `SetTime`, `LocalRTC`
(RTC-in-local-time flag) + `RTCTimeUSec` (read the RTC directly). **timesyncd config**
additionally has `RootDistanceMaxSec`, `PollIntervalMinSec/MaxSec`,
`ConnectionRetrySec`, `SaveIntervalSec`.

**chrony** (if HAOS ever outgrew SNTP) adds a much bigger space: multiple weighted
sources with per-source options (iburst, prefer, trust, minpoll/maxpoll), **NTS**
(authenticated NTP, RFC 8915), acting as an **NTP server to the LAN** (very
Home-Assistant-shaped: serve time to IoT devices on flaky internet), GPS/PPS reference
clocks, hardware timestamping/PTP, makestep policy, offline/online source control for
intermittent links, `chronyc tracking`/`sources`/`sourcestats` telemetry, and real
**RTC drift tracking** (`rtcfile`, `trimrtc` — systemd only does the kernel's crude
11-minute sync).

**RTC** as its own sub-domain: presence (`/sys/class/rtc/rtc0`), RTC time vs system
time, hctosys source, wake alarms, and health — several RTC drivers report the
voltage-low/data-loss flag (`RTC_VL_READ` ioctl, e.g. NXP PCF85xx parts common on
SBCs), and the Pi 5's onboard RTC exposes battery-charging configuration. "RTC battery
dead → clock resets every power cycle → TLS/automations break until NTP sync" is a
real support pattern that a `rtc` info block plus a resolution-center issue could
catch.

None of that is "NTP". All of it is "time". That's the strongest argument that the
namespace should be domain-named, not protocol-named.

## Options on the table

| Anchor | For | Against |
|---|---|---|
| `/ntp` (PR as-is) | Matches merged docs/CLI PRs; short | Reads as a plugin namespace next to `/dns`; protocol-named, so timezone/RTC/sync-status never fit; splits time across `/ntp` + `/host/info` |
| `/os/config/ntp` | Mirrors the swap precedent, honest about mechanism | Wrong layer conceptually; strands the generic status half (timesync1) in an HAOS-only namespace |
| `/host/time` or `/host/timedate` | Where the code already lives; where `timezone`/`use_ntp`/`dt_synchronized` are already reported; room for `ntp` config, `sync` status, `timezone`, `rtc` sub-objects; per-capability feature flags fit the existing `HostFeature` pattern | Docs/CLI/client PRs need re-touching |
| `/timedate` top-level | systemd-familiar name | Every other top-level is a *component* (core, os, host, store, plugins); a config-domain top-level breaks that taxonomy |

## Recommendation

**`/host/time`** (or `/host/timedate` for the systemd echo), shaped roughly as
`GET /host/time/info` returning `{timezone, ntp_enabled, synchronized, servers,
fallback_servers, active_server, active_source, dt_utc, rtc: {...}}` and `POST /host/time/options`
accepting the writable subset, each field guarded by its own feature/version gate.
It's the only anchor where the PR's config write, the "currently used server" status,
the existing timezone/datetime plumbing, and every future item (NTS, LAN serving, RTC
health) all land in one coherent place — and it matches where the PR already put its
own manager code.

The pragmatic counterweight: the `/ntp` shape apparently already went through docs,
CLI, and client-library reviews, so moving costs coordination. But this is exactly the
moment it's still cheap — nothing has shipped in a Supervisor release yet.

---

Claude pushes a bit back on top-level mostly for structural consistency (components/plug-ins). I don't see this as a major factor, but yeah it is still a bit a departure from the current scheme.

IMHO, going with `time` or `timedate` makes a lot of sense, especially with future features in mind (timezone, RTC information, and to a lesser extend NTS does not fit well into `ntp`). I think put it underneath `/host` for consistency also doesn't seem to bad, since we anyway move/rename. Personally, I think I'd go for `/host/timedate`.

supervisor/api/time.py
```diff
@@ -0,0 +1,60 @@
...
+
+    @api_process
+    async def info(self, request: web.Request) -> dict[str, Any]:
+        """Return NTP server configuration."""
+        self._check_available()
+
+        return {
+            ATTR_SERVERS: self.sys_dbus.agent.timesyncd.ntp_servers,
+            ATTR_FALLBACK_SERVERS: self.sys_dbus.agent.timesyncd.fallback_ntp_servers,
+        }
```
agners, line 49: With the current server configuration in mind, maybe it would be better to have the configuration in a separate field (e.g. `config`) or even endpoint? 🤔

agners (changes requested):
Some smaller remarks. Let's consider what features might go into this endpoint, and with that the API endpoint location/name settled first.

supervisor/host/manager.py
sairon: I think it's purely hypothetical - we also do `if self.sys_dbus.agent.version ...` at other places and I don't see issues with that in Sentry. `self.sys_os.available` pretty much ensures OS Agent will be there, if not, we've got bigger problems.

sairon:
> IMHO, going with `time` or `timedate` makes a lot of sense, especially with future features in mind (timezone, RTC information, and to a lesser extend NTS does not fit well into `ntp`). I think put it underneath `/host` for consistency also doesn't seem to bad, since we anyway move/rename. Personally, I think I'd go for `/host/timedate`.

I had this PR ready for over a year, so I just finished what I implemented back then, so it wasn't a fresh memory for me to tell why I did this like this, but finally I recalled where the current nesting comes from by looking at the original linked task.

There we considered `ha network ntp` or `ha ntp` CLI. Then decided it doesn't really fit under `network` so it went directly to `ha ntp`. The API mirrors the CLI structure, so if we change where the API endpoint goes to, we probably shoud relocate it in CLI as well for consistency. And  that's where it doesn't really feel right to me - although we have some other time-related stuff in `info` there, we don't nest other features next to it. Putting `ntp` here feels a bit unprecedented:
```
Available Commands:
  disks       Manage host disk operations
  info        Provides information on the host system
  logs        View the log output of the host systemd journal logs
  options     Allow to set options on host system
  reboot      Reboots the host machine
  reload      Reload information from the host machine
  shutdown    Shutdown the host machine
```

To follow the existing idioms, it would mean to put it under `/host/options` (which currently only takes `hostname`) and there I'm also not sure if it's the way we want to go...

sairon:
After some more thought and private chat, let's move this to `/time`. There were some good points about the narrow scope of `/ntp` terminology, but the suggestion to put it under host breaks pre-existing concepts. Some randomly ordered things considered when deciding for this placement:
- timedate is a systemd artifact - timedated exposes org.freedesktop.timedate1 but Supervisor consistently doesn't name endpoints after the D-Bus service behind them (network, logs, disks, services...)
- time sounds less awkward than timedate, and it's where people would be looking for the time-related stuff; the original CLI PR even used it as an alias
- the nesting under host doesn't have any precedent in endpoints exposing config - we have `ha os swap info|options` (yet not `ha os options`) or `ha host disk` (but that has no `info|options`)
- time is not just a host thing, it's system-wide

supervisor/api/time.py
```diff
@@ -0,0 +1,62 @@
...
+
+class APITime(CoreSysAttributes):
+    """Handle REST API for time and date configuration."""
+
+    def _check_available(self) -> None:
+        """Check if OS Agent Timesyncd configuration is available."""
+        if HostFeature.NTP not in self.sys_host.features:
+            raise APINotFound(
+                "Home Assistant OS 18.3 or newer required for NTP settings"
+            )
```
mdegat01, line 37: What status code to return here feels like such a mystery 😆 If you search our exception types for "NotSupported" you'll find a pattern or at least a similar idea. What to tell the client when the API is known but they can't use it due to how the system is set up.

It's especially confusing with this one (and similarly with `AppNotSupportedHomeAssistantVersionError`) where its due to a version of something that could be updated. Since most likely the client here is someone that could update the system but at the same time "update the system" is not a typical API response since normally it isn't something a client could do.

404 Not Found seems not quite right to me since typically those come with no explanation, "does not exist" is assumed to be self-explanatory. But those other places I mentioned all use 400 Client Error and that doesn't seem any better, there's nothing the client can change on retry that will suddenly work.

I guess I'm fine with 404 but then we should probably go back and review the others for consistency. "API exists but isn't supported here due to: \<x\>" seems like a common pattern so we should try to have consistency in status code and response.

mdegat01 (approved):
Looks good to me.

I should mention that community apps includes the NTP app [chrony](https://github.com/hassio-addons/app-chrony). Could be a nice follow-up to update it to leverage this new feature somehow. I'm not sure if having it automatically set itself as the NTP server on the system is ideal but perhaps it could include a config option to do that? Or if it could create a repair or discovery pop-up that asks the user if they want to use it as their NTP server that'd be really cool but I'm not entirely sure how to do that 😆

supervisor/api/time.py
sairon: > 404 Not Found seems not quite right to me since typically those come with no explanation, "does not exist" is assumed to be self-explanatory. But those other places I mentioned all use 400 Client Error and that doesn't seem any better, there's nothing the client can change on retry that will suddenly work.

That's right. And I am quite sure we had this discussion in the past and that we reached the same conclusion - 404 is meh but it is semantically still better than 400, exactly because of that "client can't retry with a different request" argument. Also, for newly added endpoints, clients will get 404 "naturally" on older Supervisor versions.

supervisor/host/control.py
```diff
@@ -130,3 +139,35 @@ async def set_timezone(self, timezone: str) -> None:
...
+            await self.sys_dbus.agent.timesyncd.set_fallback_ntp_servers(
+                fallback_servers
+            )
+            restart_required = True
+
+        if restart_required:
+            await self.sys_dbus.agent.timesyncd.update()
+            await self.sys_dbus.systemd.restart_unit(
+                TIMESYNCD_UNIT, StartUnitMode.REPLACE
+            )
```
agners: `restart_required` is derived from the OS Agent property cache, but the `Set` calls above mutate that same cache — both Timesyncd properties are `prop.EmitTrue`, so `PropertiesChanged` updates the proxy before the call even returns. The operation that creates the obligation to restart therefore destroys the record of it.

Concretely: `set_ntp_servers` succeeds, `set_fallback_ntp_servers` raises `DBusError`, and the exception propagates before the restart. The drop-in now holds the new primary servers but timesyncd still runs the old ones. On an identical retry both comparisons are equal, `restart_required` stays `False`, and the API returns 200 as a no-op — the written config stays inactive until an unrelated reboot, with no way to force it through the API. Same shape if `restart_unit` itself fails.

Since the `Set` calls are idempotent, dropping the diff makes every retry self-healing. `timesyncd.update()` also becomes unnecessary, as nothing reads the cache afterwards and the proxy self-syncs via `PropertiesChanged`:

```suggestion
        if servers is None and fallback_servers is None:
            return

        if servers is not None:
            _LOGGER.info("Setting NTP servers: %s", servers)
            await self.sys_dbus.agent.timesyncd.set_ntp_servers(servers)

        if fallback_servers is not None:
            _LOGGER.info("Setting fallback NTP servers: %s", fallback_servers)
            await self.sys_dbus.agent.timesyncd.set_fallback_ntp_servers(
                fallback_servers
            )

        await self.sys_dbus.systemd.restart_unit(TIMESYNCD_UNIT, StartUnitMode.REPLACE)
```

(The docstring above would want to lose the "when changed".)

The cost is a timesyncd restart on a no-change POST. If that's unwelcome, the alternative is to diff against the running daemon rather than the written config — `org.freedesktop.timesync1` exposes `SystemNTPServers`, so a stale daemon after a failed restart shows a real difference. That needs checking against the real interface (fallback servers may not be exposed separately, and the merged view includes the DHCP drop-in), but it's plumbing that would be useful anyway for reporting the currently active server.

supervisor/host/manager.py
```diff
@@ -136,6 +136,13 @@ def supported_features(self) -> list[HostFeature]:
...
+        ):
+            features.append(HostFeature.NTP)
```
agners, line 145: This gate and `SystemControl._check_dbus(HostFeature.NTP)` disagree: `_check_dbus` additionally requires `self.sys_dbus.systemd.is_connected`, which the timesyncd restart needs.

With systemd's D-Bus disconnected on an otherwise eligible host, `/host/info` advertises `ntp` and `GET /time/info` returns 200, but `POST /time/options` passes the `APINotFound` gate and then raises `HostNotSupportedError` → 400 "No ntp D-Bus connection available", a hard error on a capability the API just advertised.

```suggestion
        if (
            self.sys_os.available
            and self.sys_dbus.agent.timesyncd.is_connected
            and self.sys_dbus.systemd.is_connected
            and self.sys_dbus.agent.version >= NTP_MIN_OS_AGENT_VERSION
        ):
            features.append(HostFeature.NTP)
```

I looked at whether a shared helper would be better than duplicating the predicate, and I don't think it is: `supported_features` is `lru_cache`d (cleared in `reload()`) while `_check_dbus` is evaluated live at call time, and that difference is deliberate — the write path wants the current connection state, not a cached snapshot. A helper would have to pick one. The existing `HOSTNAME`/`TIMEDATE` features duplicate their checks the same way, so matching that seems better than introducing a new mechanism for one feature.

agners (approved):
LGTM!

PR #7186, opened 2026-09-01
Treat containers with corrupt storage metadata as missing

### LINKED ISSUE #7182: DockerAPIError: Docker API error occurred while getting container information: [500] RWLayer of container <id> is unexpectedly nil
Sentry Issue: [SUPERVISOR-1MF0](https://home-assistant-io.sentry.io/issues/7507464266/?referrer=github_integration)

```
DockerError: RWLayer of container 93933b4a7140cdab98f752b51a48711aaddbac34b9a1715e4663a0d107f7f0f5 is unexpectedly nil
(1 additional frame(s) were not displayed)
...
  File "aiodocker/containers.py", line 116, in get
    data = await self.docker._query_json(
  File "aiodocker/docker.py", line 556, in _query_json
    async with self._query(
  File "contextlib.py", line 214, in __aenter__
    return await anext(self.gen)
  File "aiodocker/docker.py", line 448, in _query
    yield await self._do_query(
  File "aiodocker/docker.py", line 529, in _do_query
    raise DockerError(response.status, data["message"])

DockerAPIError: Docker API error occurred while getting container information: [500] RWLayer of container 93933b4a7140cdab98f752b51a48711aaddbac34b9a1715e4663a0d107f7f0f5 is unexpectedly nil
(10 additional frame(s) were not displayed)
...
  File "supervisor/jobs/decorator.py", line 307, in wrapper
    raise err
  File "supervisor/jobs/decorator.py", line 299, in wrapper
    return await method(obj, *args, **kwargs)
  File "supervisor/apps/app.py", line 1442, in begin_backup
    if not await self.is_running():
  File "supervisor/docker/interface.py", line 432, in is_running
    (container_metadata := await self._get_container())
  File "supervisor/docker/interface.py", line 425, in _get_container
    raise DockerAPIError(
```

agners:
## Proposed change

Since Docker 29.4, containers whose RW layer fails to load while the daemon restores its state at startup — for example after an unclean shutdown corrupted the storage metadata — are kept registered so they can still be removed, instead of being dropped (https://github.com/moby/moby/pull/51724). Every other operation on such a container fails with a 500 `RWLayer of container <id> is unexpectedly nil` error, including the inspect at the start of nearly every Supervisor container operation. A single corrupt container record therefore permanently breaks starting, restarting, rebuilding and backing up of the affected add-on, plugin or Home Assistant Core: the start path errors out on the initial state check, before it reaches the cleanup that would remove the broken record. When the Home Assistant Core container is the affected one, add-on authentication requests fail as well, since they check whether Core is running.

Docker versions before 29.4 dropped such containers during daemon startup, so Supervisor saw them as missing and recreated them, healing the installation as a side effect. This change restores that behavior explicitly: a Docker API call failing with this error now reports the container as missing (`DockerNotFound`, or the not-found semantics of the state checks). The recovery paths then recreate the container just as they did before — `run()` sees the container as not running and proceeds to its cleanup, where `stop_container()`'s force delete (which no longer inspects first since #7175) removes the broken record before the new container is created. A warning naming the affected container is logged whenever this happens.

With the containerd image store the inspect of such a container succeeds (the RW layer check in Docker's inspect path applies to graph drivers only) and the error surfaces only from the actual start or restart call. Those are handled as well: the broken record is removed there and the container reported as missing, so the next start takes the recreate path. Sentry shows this shape in the field (SUPERVISOR-1B6Q).

Detection is deliberately narrow: only Docker's deterministic nil-RW-layer message is matched, in all its per-call-site phrasings (container id or name, word order — "unexpectedly nil" appears nowhere else in Docker). The sibling corruption class where the RW layer loads but its on-disk directories are gone (`stat …: no such file or directory`, `open …/lower: not a directory`, `error creating overlay mount …`; e.g. SUPERVISOR-1BGE, and the dominant shape in SUPERVISOR-1B6Q with ~900 affected users since February) keeps its current behavior: removal is verified safe for a nil RW layer (Docker skips releasing it on delete), but not for a loaded-but-broken one. Extending the same recovery to that class is a candidate follow-up once `docker rm -f` is verified to succeed there.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR fixes or closes issue: fixes #7182
- This PR is related to issue: https://github.com/home-assistant/operating-system/issues/4875
- Link to cli pull request:

## Checklist


supervisor/docker/interface.py
```diff
@@ -423,6 +428,20 @@ async def _get_container(self) -> dict[str, Any] | None:
...
+                # can recreate any of its containers, recovery paths then
+                # recreate this one, with stop_container's force delete
+                # removing the broken record without inspecting it first.
+                _LOGGER.warning(
+                    "Container %s storage metadata is corrupt, "
+                    "treating the container as missing: %s",
+                    self.name,
+                    err,
+                )
+                return None
```
mdegat01, line 444: Instead of acting like the container is missing, why wouldn't we just remove the container right here so it is actually missing? This comment applies to every one of these if blocks that doesn't include a "remove container" step actually. Shouldn't we just immediately remove the container as soon as we detect this situation? What's the value in keeping it around to let something else remove it later? We can't fix it so in the end something will have to remove it, might as well do it right now.

Or even what if we changed the helper from `is_corrupt_container_error(err) -> bool` to `check_corrupt_container_error(err, container) -> Coroutine[None]` and centralized this checking and handling in the helper?

agners: Went with removal, but at the call sites that can do it safely rather than in every block that detects the error — done in 14121bd9e.

`start_container()`/`restart_container()` now force-remove the broken record when the corrupt error comes out of the actual start/restart call, and raise `DockerNotFound` so callers treat the container as missing. With the containerd image store this is the only place the error surfaces at all (the RW layer check in moby's inspect path is graph-driver specific, so the inspect of a corrupt container succeeds there), and removing here is what makes the next start take the recreate path instead of failing into the same wall forever. The pre-run cleanup in `run()` already removes the record without inspecting since #7175, so every recreate flow disposes of it moments after detection.

The read-side checks (`_get_container()`, `container_is_initialized()`, logs) detect and report the container as missing but deliberately don't remove it: they run outside the job locks that serialize a container's lifecycle, and removal has to go by name. Names are only unique at a given instant — the heal itself deletes the broken record and then recreates the container under the same name — so a delete-by-name issued from an unlocked path after a stale inspect could land on the freshly recreated container rather than the broken one. Start, restart and stop all hold the job lock, so no such handover can happen under them. I also looked at making unlocked removal safe by parsing the container id out of the error message and deleting by id, but relying on Docker's message text for targeting (rather than just detection, where there's no alternative) felt too fragile to be worth it.

In practice nothing useful is left behind: every flow that wants the container running reaches a locked removal right after detection (boot heal, user start, rebuild). The only case where the record outlives detection is when nothing tries to run the container — there it's a zombie entry in `docker ps -a` until the next start or stop, and the warning log makes it diagnosable.

mdegat01: @agners something seems off with 14121bd9e  or else I'm misunderstanding your comment. Your comment above seems imply a behavior change in that commit to remove the container in safe places. But there's no actual change in production code in that commit, only changes to a comment, a docstring and test cases. Did something get left out of the changeset to commit and pushed early?

agners: Hm, yeah I tried to autofix in all cases, but Claude was really not all that happy, also due to potential race condition with container name/deletion (I think it is rather theoretical). In the end I conceded and allowed to paddle back a bit, this is where Claude got confused with rebases/commit messages.

Anyhow, I've went through the commits again, and let Claude reword. Also note that this aligns with how things worked in Docker before:


  ```
┌────────────────────────────────┬─────────────────────────────────────────────────────────────────────────────────────────────┬─────────────────────────────────────────────────────────┐
  │                                │                                 Docker < 29.4 (accidental)                                  │                  Docker ≥ 29.4 + #7186                  │
  ├────────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────┤
  │ Read paths (is_running, state) │ 404 → missing                                                                               │ corrupt 500 → missing (identical semantics)             │
  ├────────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────┤
  │ Recreate flow                  │ create new under same name                                                                  │ stop's force-delete removes the record, then create new │
  ├────────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────┤
  │ The broken record              │ left registered-but-invisible? No — never registered; containers/<cid> dir orphaned forever │ actually removed via docker rm equivalent               │
  ├────────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────┤
  │ Visibility                     │ none                                                                                        │ warning log naming the container                        │
  ├────────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────┤
  │ Layer dirs                     │ orphaned                                                                                    │ orphaned (same moby gap)                                │
  └────────────────────────────────┴─────────────────────────────────────────────────────────────────────────────────────────────┴─────────────────────────────────────────────────────────┘
````


And below Claude's verbose exploitation on why.

---

You read it right, sorry for the confusion — that commit's message was misleading, I've reworded it (now f09bf16c1: "Document why corrupt container removal stays at the job-locked sites"). Its actual change is only a comment, a docstring and tests.

The removal itself was already in place when you reviewed: 3e4f28891 factors `stop_container`'s force-delete into `_remove_container_record()` and calls it from the `start()`/`restart()` error handlers when the failure is the corrupt-container error, raising `DockerNotFound` so the next start takes the recreate path. Together with the pre-run cleanup in `run()` (which removes without inspecting since #7175), every place that can act on the record does so, following where Docker raises the error on each backend.

For the remaining blocks — `_get_container()`, `container_is_initialized()`, the `containers.get()` blocks in start/restart/logs — I deliberately decided against removing right there: those run outside the job locks that serialize a container's lifecycle, and removal has to go by name. Names are only unique at a given instant — the heal itself deletes the broken record and then recreates the container under the same name — so a delete-by-name issued from an unlocked path after a stale inspect could land on the freshly recreated container rather than the broken one. Start, restart and stop hold the job lock, so no such handover can happen under them. I did try your suggestion in full (2f7bcd4ea, since force-pushed away): removing from every detection site by parsing the container id out of the error message, which makes the delete safe to target — but relying on Docker's message text for targeting rather than just detection felt too fragile, so I backed it out. What remained of that round is the docstring/comment explaining the lock constraint and tests asserting the read paths don't remove.

In practice every flow that wants the container running reaches a locked removal right after detection; the record only outlives detection when nothing tries to run the container, where it's a zombie in `docker ps -a` until the next start or stop, with the warning log making it diagnosable. Happy to revisit if you'd still rather remove in the read paths.

Comment (mdegat01) [reply supervisor/docker/interface.py:444]: Ok I guess I get it. By consistently treating it identically to `NOT_FOUND` we really don't have to remove it in these other places, it should heal itself automatically. I still think the concerns around the name are pretty theoretical because we manage the system and force consistent names for components but its fine as it is.

File: supervisor/docker/manager.py [hunk 2]
```diff
@@ -884,6 +884,14 @@ async def container_is_initialized(
         except aiodocker.DockerError as err:
             if err.status == HTTPStatus.NOT_FOUND:
                 return False
+            if is_corrupt_container_error(err):
+                _LOGGER.warning(
```
mdegat01, line 888:
```suggestion
                _LOGGER.info
```
Definitely a nit but since we are treating this as not an issue we should probably log it as such. This is self-healing so users don't really need to know about it, warning makes it sound like something is wrong. Info or debug seems better I think? Applies through the changeset if you agree.

mdegat01 (approved):
Ok I'm on the same page. Minor nit that the log level may be too aggressive as it isn't something users have to worry about it but good with it as is if you want 👍

supervisor/docker/manager.py
agners: It is still a corruption of some kind which lead to this. If this happens only once, nobody should really bother, but I think if this happens regularly, it could hint at storage/file system issues. I'd prefer if it stands out a bit, so I'll leave it as a warning.

PR #7188, opened 2026-09-01
Report mount storage usage from the probe, not cached state

MindFreeze:
## Proposed change

The per-mount storage usage endpoint gated on the mount's cached state, refusing anything not currently `active`. Since #7167 made mounts autofs-triggered, that state is only as fresh as the last reconcile probe — 15 minutes apart — so a share that came back stayed unreportable for up to that long even though a probe would have found it healthy. This drops the gate and lets the probe decide: its `statvfs` walks with `LOOKUP_AUTOMOUNT`, so it activates a dormant trigger and reports what it finds.

**To reproduce:** ask for usage on a healthy mount whose last probe failed, and you get `400` until the next reconcile. The parametrized test added here is the repro — it covers a cached state of `INACTIVE`, `FAILED` and unknown, all of which return `400` before this change and real usage after it.

**One caveat:** a usage request is no longer strictly read-only with respect to system state. The probe can now cause the mount it measures. Callers still share one probe rather than stacking executor threads.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR is related to issue: #7146, #7167
- Link to documentation pull request: https://github.com/home-assistant/developers.home-assistant/pull/3325
- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/api/host.py
+++ b/supervisor/api/host.py
@@ -37,14 +37,12 @@ from ..const import (
     ATTR_TIMEZONE,
 )
 from ..coresys import CoreSysAttributes
-from ..dbus.const import UnitActiveState
 from ..exceptions import (
     APIDBMigrationInProgress,
     APIError,
     HostContainerLogEpochError,
     HostLogError,
     MountNotFound,
-    MountUsageNotActiveError,
     MountUsageNotMountedError,
     MountUsageReadError,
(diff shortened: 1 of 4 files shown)


mdegat01 (approved):
Good catch, thanks @MindFreeze !

Issue #7110, opened 2026-08-04
Network API v2: separate runtime state from configuration


agners:
# Network API v2 — Design

Status: draft (research phase)
Scope: Supervisor REST API `/network/*`, host network manager, NM D-Bus layer.

## 1. Problem statement

The v1 network API uses a single data structure for two distinct things:

1. **Observed state** — what NetworkManager reports right now (active addresses, gateway, DNS from DHCP lease, connectivity, wifi signal).
2. **Desired configuration** — what the user asked for (method, static addresses, nameservers, route metric, wifi credentials).

Internally the split *exists* (`IpConfig` vs `IpSetting` in `supervisor/host/configuration.py`), but it is fused back together at every boundary:

- `interface_struct()` / `ip4config_struct()` (`supervisor/api/network.py:112`) interleave both in one JSON object: `method`/`route_metric` come from the *setting*, `address`/`gateway`/`nameservers`/`ready` from *runtime state* — as indistinguishable siblings.
- The PUT handler accepts (roughly) the same shape back and rebuilds a full `IpSetting` from it.
- Most fundamentally: settings are only read **through the active connection** (`NetworkInterface.settings` = `self.connection.settings if self.connection else None`, `supervisor/dbus/network/interface.py:102`). NM's actual model — persistent connection profiles that exist independently of activation — is collapsed into "one interface = one active connection = one settings blob".

### Concrete problems caused by the conflation

| # | Problem | Where |
|---|---------|-------|
| P1 | GET interleaves state and settings; configured static addresses are invisible when interface is down; DHCP lease indistinguishable from config | `api/network.py:112-135` |
| P2 | GET→modify→PUT round-trips bake runtime state into persistent config (DHCP lease becomes static address; DHCP DNS gets pinned) | confirmed in frontend, see §3 |
| P3 | Missing `method` in PUT body defaults to `static` (`api/network.py:247`) — `{"ipv4": {"nameservers": [...]}}` silently rewrites the interface to static with an empty address list; `create_vlan` defaults the same field to `auto` | `api/network.py:247,333` |
| P4 | Per-family full replace but cross-family merge in a single PUT — three different write semantics in one endpoint | `api/network.py:244-280` |
[...]

PR #7192, opened 2026-09-02
Add Network API v2: state/config split with non-destructive disable

mdegat01:
## Proposed change

First chunk of the Network API v2 work described in #7110: a `GET /v2/network/info`, `GET /v2/network/interfaces/{name}`, and `PUT /v2/network/interfaces/{name}/config` that split each interface's *observed state* from its *desired config*, resolved independent of whether the interface is currently active. v1 stays byte-identical and frozen throughout - this is purely additive to the shared internals.

Compared to v1:

- Config is resolved independent of activation: `GET /v2/network/interfaces/{name}` still returns full `config` even when the interface is down/disconnected, by enumerating stored NetworkManager connection profiles instead of only reading the active connection.
- `PUT .../config` requires `method` per IP family (no silent default to `static`), is a full-document replace of `config` rather than a partial per-field patch, and validates cross-field contradictions (static with empty addresses, gateway of the wrong family, `psk` without `auth: wpa-psk`) with a 400 instead of silently accepting them.
- `enabled: false` is non-destructive in v2: it deactivates the connection and clears `autoconnect` instead of deleting the stored profile, and re-enabling reuses that same profile (also avoids v1's "duplicate profile on re-enable" side effect for v2 callers). v1 keeps its destructive delete-on-disable behavior, needed for the frontend's Wi-Fi "forget" flow.
- Response shape uses a hard `state`/`config` split (replacing v1's fused flat `ipv4`/`ipv6`/`wifi` objects), plural `addresses`, adds `path` and `psk_set`, drops the `default` interface alias, and `config` is `null` when no NetworkManager profile matches at all instead of v1's always present dummy config.

Also, while touching the `apply_changes()` error paths for the non-destructive disable work, cleaned up its error handling for all callers (v1 and v2):

- Split the old `_apply_disable()` (which branched on a `destructive_disable` argument) into two dedicated helpers, `_apply_delete()` and `_apply_deactivate()`, so log messages and error handling are specific to each path instead of sharing generic wording.
- Added an `error_key` to every error `apply_changes()` can raise, so API callers can translate them instead of matching on message text.
- Corrected the HTTP status codes for these errors to reflect who is actually responsible: bad client-supplied connection settings stay `400` (activation failed/timed out), a request to update an interface that does not exist or is disabled is `404`, and DBus-call failures that aren't caused by anything the caller sent become `500` (`APIUnknownSupervisorError`) instead of the previous blanket `400`.
- Stopped interpolating raw DBus exception text into client-facing error messages (that unbounded text can't be translated). The full error is still logged at ERROR level exactly as before; the raised exception now only carries bounded fields such as the interface name.

This does not complete the Network API v2 work tracked in #7110 - more PRs will follow.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR is related to issue: #7110
- Link to cli pull request:

## Checklist


supervisor/dbus/network/setting/__init__.py
```diff
@@ -334,3 +337,24 @@ async def reload(self):
...
+    id/parent) and device-type compatibility are intentionally not part of
+    this helper, they stay call-site specific.
+    """
+    if setting.match and setting.match.path:
+        return setting.match.path == [path]
+
+    return (
+        setting.connection is not None
+        and setting.connection.interface_name == interface_name
+    )
```
mdegat01, line 372: Note that I did not include "R6 — explicit ownership. Supervisor manages exactly one profile per device, tracked by UUID " from #7110 in this PR. Since that involves tracking a UUID across restarts I left it for a follow-up and used a similar idea to `equals_dbus_interface` for now.

supervisor/api/network.py
```diff
@@ -109,6 +116,90 @@
...
+    def validator(config: dict[str, Any]) -> dict[str, Any]:
+        addresses = config.get(ATTR_ADDRESSES) or []
+
+        if config[ATTR_METHOD] == InterfaceMethod.STATIC and not addresses:
+            raise vol.Invalid(
+                f"{schema_key}: at least one address is required when method is static"
+            )
+
+        if config.get(ATTR_GATEWAY) and not addresses:
+            raise vol.Invalid(f"{schema_key}: gateway requires at least one address")
```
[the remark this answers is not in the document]
mdegat01: Hm fair point. Validation should reject this currently as this setup cannot be configured via Supervisor and isn't supported. This is a situation that would require an externally managed profile per #7110

supervisor/host/network.py
```diff
@@ -299,133 +359,255 @@ async def _apply_settings_in_place(
...
-            try:
-                await inet.settings.delete()
-            except DBusError as err:
-                raise HostNetworkError(
-                    f"Can't disable interface {interface.name}: {err}", _LOGGER.error
-                ) from err
+            if destructive_disable:
+                await self._apply_delete(interface, existing_settings)
+            else:
+                await self._apply_deactivate(inet, interface, existing_settings)
```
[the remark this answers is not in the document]
mdegat01: Looking closely at #7110 it appears a PUT of `config: null` is intended as a way to create the Supervisor managed profile for the interface so we can then store its UUID. And then I suppose it would wipe existing config once that profile was already there? However the use case above "creating a disabled and otherwise blank profile via our API" doesn't really seem like a feature our users need as we are not intending to mimic nmcli with our API.

I'm going to say the fix here for now is to make this an exception to round-tripping support. Validation will not allow this use case. We will revisit in a future PR when we get to `config_source` and the different types of profiles.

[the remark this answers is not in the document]
mdegat01: Ok I realized I misunderstood what was going on here. Validation already blocks a PUT with just a body of `null` so `config: null` cannot be round-tripped. Copilot was calling out a specific scenario that was being prevented:
1. GET returns `config: null` (currently only happens if the interface has no profile and is simply using NM defaults, there are other possible reasons for this in #7110 though)
2. Client does a PUT with a fully configured profile but `enabled: false`
3. The API was allowing this and returning success but under the hood this was being dropped, it wouldn't create a brand new but disabled profile

Validation now also blocks this specific scenario and returns a 400. We could allow it but to be honest I think it should just stay blocked. As mentioned above we're not intending to be an NM interface. The work around disabled profiles was just so that users could make changes in the UI without losing their prior work, not to draft complicated profiles and iterate on them. Requiring anything new start active seems fine for our use case.

supervisor/api/network.py
```diff
@@ -175,6 +278,115 @@ def interface_struct(interface: Interface) -> dict[str, Any]:
...
+    """Return a dict with information of an interface for the v2 API."""
+    interface = resolved.interface
+    return {
+        ATTR_NAME: interface.name,
+        ATTR_TYPE: interface.type,
+        ATTR_MAC: interface.mac,
+        ATTR_PATH: interface.path,
+        ATTR_PRIMARY: interface.primary,
+        ATTR_STATE: interface_state_struct(interface),
+        ATTR_CONFIG: interface_config_struct(resolved),
```
agners, line 386: Exactly, the essence of this PR: These two lines 🤩 😄 !

supervisor/api/__init__.py
```diff
@@ -278,27 +278,46 @@ def _register_network(self, app: web.Application) -> None:
...
+                    web.post("/network/reload", api_network.reload),
+                    web.get("/network/interfaces/{name}", api_network.interface_info),
+                    web.put(
+                        "/network/interfaces/{name}/config",
+                        api_network.update_config,
+                    ),
+                    web.get(
+                        "/network/interfaces/{interface}/accesspoints",
+                        api_network.scan_accesspoints,
+                    ),
```
agners, line 318: There is no `create_vlan` endpoint, is this a conscious omission from v2? FWIW, I think we can't drop it ultimately, but since v2 is still shaping I am also fine (re)adding it in a follow up PR.

supervisor/api/network.py
```diff
@@ -224,11 +436,123 @@ async def info(self, _: web.Request) -> dict[str, Any]:
...
         interface = self._get_interface(request.match_info[ATTR_INTERFACE])
 
-        return interface_struct(interface)
+        return interface_struct_v1(interface)
+
+    async def _get_resolved_interface(self, name: str) -> ResolvedInterface:
+        """Get a resolved interface (state+config) by name for the v2 API.
+
+        Unlike `_get_interface()`, this does not support the `default` alias
+        (dropped in v2) and excludes VLAN interfaces (not shown in v2 for now).
```
agners, line 449: Ok, I guess that resolves the VLAN discussion 😄 

But let's mention that in the PR description. And don't forget the follow up PR.

supervisor/api/network.py
```diff
@@ -109,6 +116,102 @@
 )
 
 
+def _validate_ip_config_v2(schema_key: str):
```
agners, line 119: We kinda have a mix of v1 and v2 postfixes. I think in other places we decided to make v2 the new default (so no postfix), and postfix everything which is old/v1. I think this makes sense here too, ultimately we want this to be the new normal.

agners (changes requested):
Generally looks like a very sensible first step. I have some AI findings in a follow up review.

supervisor/dbus/network/setting/generate.py
```diff
@@ -304,5 +307,19 @@ def get_connection_from_interface(
...
                 )
             conn[CONF_ATTR_802_WIRELESS_SECURITY] = wireless_security
+        else:
+            # No security requested: explicitly clear any previously stored
+            # 802-11-wireless-security section instead of leaving it
+            # untouched. `NetworkSetting.update()` only merges sections that
+            # are present, so an old WPA/WEP section (and its PSK) would
+            # otherwise survive a switch to open auth. An empty dict here
+            # tells it to remove the section instead of merging it.
+            conn[CONF_ATTR_802_WIRELESS_SECURITY] = {}
```
agners, line 317: **Blocking**

The empty `802-11-wireless-security: {}` is a sentinel that only `NetworkSetting.update()` understands (its `_merge_settings_attribute` pops the section on an empty dict). But `get_connection_from_interface()` is also the payload builder for `_apply_create()` in `host/network.py`, which hands the dict to `AddAndActivateConnection` unchanged. NetworkManager receives an empty `a{sv}` for the section and instantiates an `NMSettingWirelessSecurity` with every property at its default, so `key-mgmt` is NULL. `nm_device_complete_connection()` then runs `nm_connection_normalize()` and the verify in `nm-setting-wireless-security.c` fails:

```c
if (!priv->key_mgmt) {
    g_set_error_literal(error, NM_CONNECTION_ERROR,
                        NM_CONNECTION_ERROR_MISSING_PROPERTY, _("property is missing"));
    g_prefix_error(error, "%s.%s: ", "802-11-wireless-security", "key-mgmt");
    return FALSE;
}
```

I checked every `_normalize_*` function in `nm-connection.c` on NM main: none removes a wireless-security setting, so nothing rescues it before verify. Result: creating any open-auth wifi profile (v1 `POST /network/interface/wlan0/update` with `auth: open`, or v2 PUT on a wlan interface with no stored profile) now fails with `HostNetworkCreateConfigError` where base succeeded, because base simply omitted the section.

The mock `AddAndActivateConnection` only checks that `802-11-wireless` has an `ssid`, and `test_generate.py` asserts the sentinel as expected output, so tests encode this rather than catch it.

Suggestion: emit the clear only for the update path (e.g. a flag on `get_connection_from_interface`, or strip `802-11-wireless-security == {}` plus the empty `security` reference right before `add_and_activate_connection` / `add_connection`).

(The `802-11-wireless.security = ""` half is harmless on current NM, that property has no from-dbus handler and is ignored.)

supervisor/host/network.py
```diff
@@ -299,133 +359,255 @@ async def _apply_settings_in_place(
...
+            uuid=existing_settings.connection.uuid,
+            autoconnect=False,
+        )
+
+        try:
+            if inet.connection:
+                await self.sys_dbus.network.deactivate_connection(
+                    inet.connection.object_path
+                )
+            await existing_settings.update(settings)
```
agners, line 473: **Blocking**

Ordering issue: `deactivate_connection()` runs first, then `existing_settings.update(settings)` to clear autoconnect. When the interface is active, `existing_settings` is `inet.settings`, i.e. `inet.connection.settings`, the `NetworkSetting` owned by the very `NetworkConnection` being deactivated.

On deactivation NM emits `PropertiesChanged` on the Device with `ActiveConnection = /`. `NetworkInterface.update()` (dbus/network/interface.py:160) sets `self.connection = None`, the setter shuts down the old `NetworkConnection` (interface.py:97-98), and `NetworkConnection.shutdown()` shuts down its `settings` (connection.py:180-181). Both are `dbus_connected`-guarded, and `dbus_connected` returns `noop()` once `is_shutdown` is set (dbus/utils.py:15-16).

So depending on when the signal lands relative to `update()`:
- before: `update()` is a silent no-op, PUT returns 200, autoconnect stays true, GET still shows `config.enabled: true`, and the interface comes back on reboot.
- during one of `update()`'s awaits: `connected_dbus` raises `DBusNotConnectedError`, which is a `HostNotSupportedError`, not a `DBusError`, so it escapes the `except DBusError` here and becomes a generic 500.

The mock `DeactivateConnection` (tests/dbus_service_mocks/network_manager.py:270) emits no property change, so the test never sees either.

Fix: write `autoconnect=false` first (the settings object is still alive), then deactivate. Alternatively re-resolve the profile by object path after deactivation.

supervisor/host/network.py
```diff
@@ -299,133 +359,255 @@ async def _apply_settings_in_place(
...
+
+        try:
+            settings_changed = await existing_settings.update(settings)
+            if await self._apply_settings_in_place(
+                inet, interface, settings, settings_changed, update_only=update_only
+            ):
+                return None
+
+            _LOGGER.info("Activating connection for interface %s", interface.name)
+            activated = await self.sys_dbus.network.activate_connection(
```
agners, line 394: **Blocking**

The v2 headline case is editing a profile whose device has no active connection. In that case `_apply_update()` persists via `existing_settings.update()`, then `_apply_settings_in_place()` returns False at the "requires an active connection" check (line 330), so we unconditionally `activate_connection()` and `apply_changes()` awaits `_wait_for_activation()` (line 571) on the HTTP request.

For an ethernet device without carrier, NM accepts a user-requested activation and the AC sits in ACTIVATING waiting for carrier. `_wait_for_activation` times out after `CONNECTION_ACTIVATION_TIMEOUT` (60 s) and raises `HostNetworkActivationTimeoutError` with "Timed out waiting for connection to activate, check connection settings". So `PUT .../eth0/config` changing only `mdns` on an unplugged eth0 blocks a minute, returns a 400 that blames the settings, and a subsequent GET shows the new config was applied. Same shape for wifi with a wrong PSK: persisted, then 400.

No test covers PUT on a down interface with a real activation cycle; the mock `ActiveConnection` is ACTIVATED from the start. The design doc (§4.2) had the activation wait off the request path (background job).

Suggestion: if there's no active connection, persist and return (optionally kick activation without awaiting it), or move the wait into a job and return the job id. At minimum don't map "no carrier yet" to "check connection settings".

supervisor/api/network.py
```diff
@@ -224,11 +436,123 @@ async def info(self, _: web.Request) -> dict[str, Any]:
...
+            # support, so reject this combination as an explicit exception
+            # to round-tripping for now; revisit with `config_source` (#7110).
+            raise APIError(
+                f"Interface {interface.name} has no existing configuration; "
+                "it cannot be replaced with a disabled one"
+            )
+
+        if interface.type == InterfaceType.WIRELESS and body.get(ATTR_WIFI) is None:
+            raise APIError(
+                f"Interface {interface.name} is wireless and requires a wifi configuration"
```
agners, line 511: **Should Fix**

`_map_nm_wifi()` (host/configuration.py:377-383) returns None for any `key-mgmt` other than `none`/`wpa-psk` (sae/WPA3, wpa-eap, owe). With `interface.wifi` None:

- `interface_state_struct` (line 309-314) reports `state.wifi: null` even though the device is connected with a signal, because `active_ssid`/`signal` are read off `interface.wifi` rather than `inet.wireless.active`. That's the state/config split leaking: an unsupported *config* parse hides observed *state*.
- `interface_config_struct` reports `config.wifi: null`.
- PUTting that GET body back (e.g. to change only `mdns`) hits this check and gets a 400. The only way through is to supply open/wep/wpa-psk, which overwrites a working WPA3 security section.

Design R1 says GET → PUT must round-trip, R4 says unsupported auth is reported and only refused on write. Suggest sourcing `state.wifi` from `inet.wireless.active` independent of the config parser, and accepting `wifi: null` on PUT when the stored profile's auth is unsupported (leaving `802-11-wireless-security` untouched).

supervisor/api/network.py
```diff
@@ -109,6 +116,102 @@
...
+            raise vol.Invalid(
+                f"{schema_key}: addresses and gateway are only supported when method is static"
+            )
+
+        return config
+
+    return validator
+
+
+def _validate_wifi_config_v2(config: dict[str, Any]) -> dict[str, Any]:
```
agners, line 150: **Should Fix**

The validator rejects psk-without-wpa-psk but never requires psk when `auth` is `wpa-psk`. On an interface with no stored profile, `PUT {wifi: {auth: "wpa-psk", ssid: "X"}}` passes validation; `generate.py` emits `key-mgmt: wpa-psk` and, since `interface.wifi.psk` is falsy (line 310), no `psk`. `AddAndActivateConnection` accepts it (missing secrets are legal at verify), the profile is persisted, activation fails with NO_SECRETS → `HostNetworkActivationFailedError` 400, no rollback. GET then reports `psk_set: true` (line 367 infers it from `auth`), so a client never re-prompts for the secret.

Omitting `psk` is intentional only when an existing wpa-psk profile is being kept. So the missing rule is: require `psk` when `auth == wpa-psk` unless `resolved.has_profile and resolved.interface.wifi.auth == WPA_PSK`. `update_config()` has `resolved` in hand, so that's the place.

Note `test_api_network_update_config_v2_wifi_psk_set_round_trip` (tests/api/test_network_v2.py:376) says it echoes `psk_set` on an existing WPA profile, but the mock settings have no `802-11-wireless-security` section, so it actually exercises this gap and asserts 200. Seeding the mock with a wpa-psk section would make it test what the docstring says.

supervisor/dbus/network/__init__.py
```diff
@@ -130,6 +137,44 @@ async def add_and_activate_connection(
...
+        Callers that don't support VLAN devices yet (currently the v2 API) are
+        responsible for filtering those out themselves - this is a temporary
+        restriction expected to be lifted in a subsequent PR, so it's kept out
+        of this shared lookup to avoid spreading it across layers.
+        """
+        for object_path in await self.settings.list_connections():
+            setting = NetworkSetting(object_path)
+            try:
+                await setting.connect(self.connected_dbus.bus)
+            except DBusError:
```
agners, line 162: **Should Fix**

`find_connection_settings()` only catches `DBusError` around `setting.connect()`. A profile removed between `ListConnections` and the Introspect/GetSettings inside `connect()` (another client running `nmcli con delete`, or a v1 destructive disable) comes back as UnknownObject/UnknownMethod, which `DBus.from_dbus_error` (utils/dbus.py:82-104) maps to `DBusObjectError` / `DBusInterfaceMethodError`. Those derive from `HassioNotSupportedError`, not `DBusError` (exceptions.py:1114-1122), so they escape the loop, escape `get_with_config()`, and `GET /v2/network/info` fails as a whole with a 500 and a Sentry event instead of skipping the vanished profile.

`NetworkManager.update()` a few lines down (line 262) already catches `(DBusFatalError, DBusInterfaceError)` for the analogous vanished-device race. Suggest `except DBusError, DBusInterfaceError, DBusObjectError:` here (or `HassioError`, since the intent is "skip anything we can't read").

---

## Posting

Inline comments go via the GitHub review API with `path` + `line` + `side: RIGHT`. Do NOT post until confirmed. Command skeleton:

```bash
gh api repos/home-assistant/supervisor/pulls/7192/reviews \
  --method POST \
  --input review.json
```

where `review.json` has `event: "REQUEST_CHANGES"`, `body: <main comment>`, `comments: [{path, line, side: "RIGHT", body}, ...]` using the six entries above.

agners (changes requested):
Nice work on the state/config split, the overall shape matches the design. Went through the write paths with a focus on what NetworkManager actually receives and how the D-Bus objects are torn down. Three things I consider blocking (open-auth create is broken, disable can silently not stick, PUT on a down interface blocks for 60 s and then reports an error after persisting), plus three smaller ones inline. Naming feedback I've left separately.

supervisor/dbus/network/__init__.py
mdegat01: Btw claude was *very* unhappy with this "Posting" bit at the bottom. Said the reviewer was very suspicious for injecting a potential prompt-injection attack and directly contradicts the repos AI policy on autonomous actions. I assume this got in here by mistake, probably a copy and paste error lol

mdegat01:
@agners good finds. All addressed, PR re-opened and ready for review again

felix-rub (approved):

Issue #7189, opened 2026-09-01
Core started via API during app boot fails to bind its port (held by Supervisor's own reservation)


agners:
Starting Core through the API while Supervisor is still booting apps produces a Core instance that cannot bind its own port. Core comes up without the `http` component, never answers `/api/`, and Supervisor waits the full startup timeout before failing.

## What causes it

Since https://github.com/home-assistant/supervisor/pull/7131 Supervisor reserves Core's host port with a transient systemd socket unit (`hassio-port-reserve.socket`) before booting apps, so host-network apps cannot claim it. The reservation is taken in `Core.start()` before the `initialize` phase and released only after the `services` phase, immediately before Supervisor starts Core itself:

```python
core_port_reserved = await self._reserve_core_port(hosts, port)
await self.sys_apps.boot(AppStartup.INITIALIZE)
...
await self.sys_apps.boot(AppStartup.SERVICES)
if core_port_reserved:
    core_port_reserved = not await self._release_core_port()
# run HomeAssistant
```

That ordering is correct for Supervisor's own start of Core, but the reservation is known only to `Core.start()` in `supervisor/core.py`. `HomeAssistantCore.start()` and `HomeAssistantCore.restart()` do not consult or release it, so any caller that reaches them during the reservation window starts a Core that finds its port occupied by Supervisor's own reservation unit:

- `POST /core/start` and `POST /core/restart` (`supervisor/api/homeassistant.py`) — how this was hit in practice
- backup restore (`supervisor/backups/manager.py`)
- the Core container watchdog (`homeassistant/core.py`), which is registered before the reservation is taken

The window lasts for the whole app boot, i.e. as long as the `initialize`, `system` and `services` phases take. Each phase blocks until every app in it reports `started`, up to `STARTUP_TIMEOUT` (120 s) per app, so on a system with apps that are slow to report healthy the window can be minutes rather than seconds. That is also the situation in which a user is most likely to press "Start" on a Core that looks stuck.

## Observed

`POST /core/start` was issued 32 s into a boot, while the `services` phase was still waiting on app healthchecks. Core started, failed to bind 8123, and Supervisor released the reservation about 26 ms later. Supervisor's own boot flow then found the container already running and logged `Skipping start of Home Assistant`, so it never noticed Core was broken; the API call's `_block_till_run()` spun for ten minutes and raised `HomeAssistantStartupTimeout`.

Supervisor:

```
[...]

Issue #6762, opened 2026-04-23
Supervisor potentially outdated on first Home Assistant Core installation


agners:
### Describe the issue you are experiencing

Through user reports, a pattern emerged where Home Assistant installation types got missdetected as "Home Assistant Container" instead of "Home Assistant OS".

Home Assistant Core logs show the following issue:
```
2026-04-14 17:07:33.881 ERROR (MainThread) [homeassistant.setup] Error during setup of component hassio: Field "duplicate_log_file" of type bool is missing in HomeAssistantInfo instance
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/setup.py", line 425, in _async_setup_component
    result = await task
             ^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/components/hassio/__init__.py", line 678, in async_setup
    await update_info_task
  File "/usr/src/homeassistant/homeassistant/components/hassio/__init__.py", line 607, in update_info_data
    await asyncio.gather(
    ^^^^^^^^^^^^^^^^^^^^^
    ...<8 lines>...
    ),
    ^
  File "/usr/local/lib/python3.14/site-packages/aiohasupervisor/homeassistant.py", line 21, in info
    return HomeAssistantInfo.from_dict(result.data)
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^
  File "<string>", line 122, in __mashumaro_from_dict__
mashumaro.exceptions.MissingField: Field "duplicate_log_file" of type bool is missing in HomeAssistantInfo instance
```

The reason for this error is that a recent version of Home Assistant Core (2026.4.2 in this case) connected to an old version of Supervisor (2025.11.1). This fundamentally breaks an invariant in our system: We expect Home Assistant Supervisor always to be newer than the Home Assistant Core. This is enforced by the `JobCondition.SUPERVISOR_UPDATED`, on Home Assistant Core installation.

However, it appears that this is not always checked, specifically on first installation of Home Assistant Core. While normally the system updates Home Assistant Supervisor on first start, this might fail for some reason. However, it does not seem to hold back subsequent Core installation.

In a test locally this is reproducible by plugging in Ethernet ~1 min after startup.

 

### What type of installation are you running?

Home Assistant OS

### Which operating system are you running on?

Home Assistant Operating System

### Steps to reproduce the issue

1. Install HAOS 16.3 on a Home Assistant Green
2. Wait for 60s to plug-in Ethernet
[...]

PR #7194, opened 2026-09-03
Block start/restart/rebuild/update APIs while Supervisor is not fully running

mdegat01:
## Proposed change

Supervisor boots apps and Home Assistant Core in a specific, ordered sequence (`initialize` → `system` → `services`, then Core itself), and replays a similarly ordered sequence while restoring a backup (during which Supervisor is in the `FREEZE` state). Calling the API to start, restart, rebuild or update something while one of those sequences is still in progress can race with it: the boot sequence has *not yet* released state it holds for its own scheduled start (e.g. the port reservation behind #7189), it can find the container already started/updated and skip its own scheduled step, or the two calls can outright collide on job concurrency (`JobConcurrency.GROUP_REJECT` on the same `JobGroup` rejects the second caller immediately rather than queuing it).

This PR adds a `require_running_system` decorator that rejects the affected API calls with a clean 503 error unless `CoreState.RUNNING` (i.e. Supervisor has fully finished booting and is not currently restoring a backup), and applies it to:

- **Apps**: `start`, `restart`, `rebuild` (this is what #7189 asked for), and `update`
- **Home Assistant Core**: `start`, `restart`, `rebuild`, `update`
- **Plugins**: `restart`/`update` for Audio, DNS and Multicast, and `update` for the CLI and Observer plugins (they have no restart route)

This only affects calls coming in through the REST API - it does not change anything about the internal calls Supervisor itself makes while executing its own boot/restore sequences.

### Why more than just Core start/restart

The issue's reproduction was `POST /core/start`/`restart`, but the same race is not specific to Core:

- **Apps** boot in the same ordered phases as Core and can likewise be started/rebuilt out of order, or reject a scheduled boot outright via `JobConcurrency.GROUP_REJECT` if an API call to the same app is in flight when Supervisor's boot flow reaches it.
- **`update`** is included alongside `start`/`restart`/`rebuild` because app/Core update jobs share the same `JobGroup` and `GROUP_REJECT` concurrency as their other lifecycle jobs, so an in-progress update can cause a concurrent, Supervisor-scheduled boot-time start to fail immediately rather than queue.
- **Plugins** (Audio, DNS, Multicast, CLI, Observer) are, if anything, more exposed than Apps/Core: they are not `JobGroup`s and their `start()`/`restart()` methods have no `@Job` protection at all, so a racing API call has no concurrency guard to even reject it - it goes straight to Docker.

### Backup restore

Restoring a backup also drives Home Assistant Core, apps and plugins through an ordered restart/start sequence, while Supervisor is in the `FREEZE` core state rather than `RUNNING`. `require_running_system`'s single `JobCondition.RUNNING` check naturally also blocks calls during `FREEZE`, so this closes the same race during backup restore with no extra logic.

### Error handling

`require_running_system` runs `Job.check_conditions()` directly rather than wrapping the handler in `@Job(...)`, since the API view classes are a separate layer from the `CoreSysAttributes` business-logic classes (`App`, `HomeAssistantCore`, ...) that already have their own `@Job`-tracked methods; adding a second `@Job` here would just create a duplicate, misleading job entry for the same logical operation.

The resulting `JobConditionException` (meant for job/log consumers, and naming the internal method, e.g. `'App.start' blocked from execution, ...`) is caught and converted into a new, translatable `APISystemNotReadyError` (`error_key = "system_not_ready_error"`) so API callers get a clean, actionable message instead.

The error is a 503 (via a new `APIServiceUnavailable` base class, which `APIDBMigrationInProgress` now shares) rather than a 400: aiohasupervisor maps 400 to `SupervisorBadRequestError`, which Core treats as a definitive answer in places (e.g. "no update available" in https://github.com/home-assistant/core/pull/171129), whereas 503 surfaces as `SupervisorServiceUnavailableError` and signals a transient condition callers should retry.

## Type of change

## Additional information

- This PR is related to issue: #7189

## Checklist

### DIFF
--- a/supervisor/exceptions.py
+++ b/supervisor/exceptions.py
@@ -132,7 +132,7 @@ class APIUnknownSupervisorError(APIError):
 
 
 class APISystemNotReadyError(APIError):
-    """Raise when an API call is rejected because Supervisor has not fully started.
+    """Raise when an API call is rejected because Supervisor isn't in a state that allows it.
 
     Used by require_running_system to reject start/restart/rebuild/update
     calls made while Supervisor's boot or backup-restore sequences are still
@@ -142,8 +142,8 @@ class APISystemNotReadyError(APIError):
 
     error_key = "system_not_ready_error"
     message_template = (
-        "Supervisor has not fully started yet, please try again once startup "
-        "has completed"
+        "Supervisor is currently starting up or restoring a backup, please "
+        "try again once that has completed"
     )
 
     def __init__(self, logger: Callable[..., None] | None = None) -> None:
--- a/supervisor/exceptions.py
+++ b/supervisor/exceptions.py
@@ -142,8 +142,7 @@ class APISystemNotReadyError(APIError):
 
     error_key = "system_not_ready_error"
     message_template = (
-        "Supervisor is currently starting up or restoring a backup, please "
-        "try again once that has completed"
+        "Supervisor is not ready to perform this operation, please try again later"
     )
 
     def __init__(self, logger: Callable[..., None] | None = None) -> None:
--- a/supervisor/exceptions.py
+++ b/supervisor/exceptions.py
@@ -131,6 +131,26 @@ class APIUnknownSupervisorError(APIError):
         super().__init__(None, logger, job_id=job_id)
 
 
+class APISystemNotReadyError(APIError):
+    """Raise when an API call is rejected because Supervisor has not fully started.
+
+    Used by require_running_system to reject start/restart/rebuild/update
+    calls made while Supervisor's boot or backup-restore sequences are still
+    in progress (see #7189), instead of letting the internal
+    JobConditionException (meant for job/log consumers) bubble up as-is.
+    """
+
+    error_key = "system_not_ready_error"
+    message_template = (
+        "Supervisor has not fully started yet, please try again once startup "
+        "has completed"
+    )
+
+    def __init__(self, logger: Callable[..., None] | None = None) -> None:
+        """Raise & log."""
+        super().__init__(None, logger)
+
+
 # JobManager
 
 
--- a/supervisor/exceptions.py
+++ b/supervisor/exceptions.py
@@ -103,14 +103,18 @@ class APIInternalServerError(APIError):
     status = 500
 
 
+class APIServiceUnavailable(APIError):
+    """API service unavailable error."""
+
+    status = 503
+
+
 class APIAppNotInstalled(APIError):
     """Not installed app requested at apps API."""
 
 
-class APIDBMigrationInProgress(APIError):
+class APIDBMigrationInProgress(APIServiceUnavailable):
     """Service is unavailable due to an offline DB migration is in progress."""
-
-    status = 503
 
 
 class APIUnknownSupervisorError(APIError):
@@ -131,7 +135,7 @@ class APIUnknownSupervisorError(APIError):
         super().__init__(None, logger, job_id=job_id)
 
 
-class APISystemNotReadyError(APIError):
+class APISystemNotReadyError(APIServiceUnavailable):
     """Raise when an API call is rejected because Supervisor isn't in a state that allows it.
 
     Used by require_running_system to reject start/restart/rebuild/update
--- a/supervisor/api/apps.py
+++ b/supervisor/api/apps.py
@@ -108,7 +108,13 @@ from ..exceptions import (
 )
 from ..validate import docker_ports
 from .const import ATTR_BOOT_CONFIG, ATTR_REMOVE_CONFIG, ATTR_SIGNED
-from .utils import api_process, api_return_stats, api_validate, json_loads
+from .utils import (
+    api_process,
+    api_return_stats,
+    api_validate,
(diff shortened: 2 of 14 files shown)


supervisor/exceptions.py
```diff
@@ -131,6 +131,25 @@ def __init__(
...
+class APISystemNotReadyError(APIError):
+    """Raise when an API call is rejected because Supervisor isn't in a state that allows it.
+
+    Used by require_running_system to reject start/restart/rebuild/update
+    calls made while Supervisor's boot or backup-restore sequences are still
+    in progress (see #7189), instead of letting the internal
+    JobConditionException (meant for job/log consumers) bubble up as-is.
+    """
+
+    error_key = "system_not_ready_error"
```
agners, line 147: Suggest `status = 503` here, matching `APIDBMigrationInProgress` above. aiohasupervisor maps 400 to `SupervisorBadRequestError`, which Core treats as a definitive client-side answer in several places, including "no update available" in https://github.com/home-assistant/core/pull/171129. A transient not-ready state should surface as `SupervisorServiceUnavailableError` so callers retry instead of concluding. Details in the review summary.

agners (commented):
Ok, I had a hunch that this could be problematic with things like https://github.com/home-assistant/core/pull/171129 in Core, so I feed it into Claude, and it went to town with it :sweat_smile: 

But I think the findings are real. AI generated stuff below

---

The guard itself looks correct: `@require_running_system` sits inside `@api_process` so the error is rendered properly, `Job.check_conditions()` works on the API views since they are `CoreSysAttributes`, and the new `test_require_running_system.py` covers every gated route across `SETUP`/`STARTUP`/`FREEZE`. A few points, one of which I think should change before merge.

### Status code: this should be 503, not 400 (see inline comment)

`APISystemNotReadyError` inherits `APIError`'s default `status = 400`, which aiohasupervisor maps to `SupervisorBadRequestError` (https://github.com/home-assistant-libs/python-supervisor-client/blob/main/aiohasupervisor/client.py). That collides with how Core interprets 400 from Supervisor. Concretely, https://github.com/home-assistant/core/pull/171129 (merged) calls `/supervisor/update` during onboarding and treats `SupervisorBadRequestError` as "no update available, proceed normally". This PR does not gate `/supervisor/update` today, so nothing breaks right now, but that route is the obvious next candidate for exactly this guard (the `supervisor_update` job has no `RUNNING` condition either), and the moment it is added Core will read "Supervisor not ready" as "no update needed" and load against the outdated Supervisor that PR exists to prevent. Silent, and exactly the class of edge case in https://github.com/home-assistant/supervisor/issues/6762.

With 503 the two designs compose: aiohasupervisor raises `SupervisorServiceUnavailableError`, which falls into Core's generic `except SupervisorError` branch and becomes `ConfigEntryNotReady`, so Core retries until Supervisor is `RUNNING` and then updates. There is precedent for a transient not-ready state right in this file (`APIDBMigrationInProgress`, `status = 503`), and semantically "try again later" is what 503 means.

### Core integrations will hit this on a normal boot

Supervisor starts Core before the `application` phase, so Core's integrations run `async_setup_entry` while Supervisor is still in `STARTUP`. Several of them start their managed add-on if it is not running yet: `AddonManager.async_schedule_start_addon()` from https://github.com/home-assistant/core/blob/dev/homeassistant/components/matter/__init__.py, `homeassistant_hardware/util.py`, and the Z-Wave JS flow. Any managed add-on still waiting for the `application` phase now yields an `AddonError` (logged as an error via `catch_error=True`) plus `ConfigEntryNotReady`. Before this PR that start went through and raced Supervisor's own boot, which is the bug you are fixing, so the new behavior is arguably the right one, but it is a visible change and worth a sentence in the description. It is also another reason for 503: it lets Core tell "retry quietly" apart from a real bad request.

### "fixes #7189" is a bit generous

This closes the API vector, which is the one that actually bit. Restore already required `RUNNING`. Two things remain from https://github.com/home-assistant/supervisor/issues/7189: the watchdog's `_restart_after_problem` path can still fire inside the reservation window (unlikely, but not excluded), and the structural point that only `Core.start()` in `supervisor/core.py` knows about the reservation while `HomeAssistantCore.start()`/`restart()` do not is untouched. I would merge this as the guard it is and keep the issue open for the reservation handoff rather than let it auto-close.

### Minor

- `check_conditions()` logs at CRITICAL when a checked condition is in `ignore_conditions`. Pre-existing behavior, but this puts it on a much hotter path: a user who ignores `running` now gets a CRITICAL line on every gated API call.
- New `error_key` `system_not_ready_error`: is there a frontend PR for the translation? Not linked in the description.

supervisor/exceptions.py
mdegat01: Sure yea that makes sense. I actually wanted it to be a 503 for sensible status code reasons but realized we used a 400 in similar spots elsewhere so left it for consistency. But fine with a 503.

As for the other two minor nits:
1. If someone is ignoring RUNNING I'm pretty ok with a critical log, that seems like a pretty bad one to ignore
2. There's no frontend pr, they aren't actually translated yet 😂

PR #7201, opened 2026-09-07
Start the Supervisor auto update on every version reload

balloob:
## Proposed change

A pending Supervisor update blocks Core, OS and app updates through the `SUPERVISOR_UPDATED` job condition. Only startup and the daily scheduled reload started the auto update. Every other reload, such as the one Core requests when the user opens Settings, made the new version known without installing it. The user then saw `supervisor needs to be updated first` until the daily task ran.

The updater now starts the auto update after every reload while the system is running. `Supervisor.auto_update_supervisor()` was moved out of the scheduled task so it can be reused, and it now returns the task performing the update instead of firing-and-forgetting it, so callers can check whether an update is genuinely in progress instead of guessing from `need_update`.

This surfaced a few related problems that are fixed here as well:

- **API update requests during an in-progress auto update**: `Supervisor.update()` is now a detached job (using the `detach` option added to the Job decorator in #7211). A concurrent call - whether a manual update request or the auto update - returns the same in-progress task instead of raising `SupervisorJobError`, so `auto_update_supervisor()` and manual update requests always share one real task regardless of who started it. The update API and the Home Assistant Core install retry loop now await that returned task directly, so they reflect its real outcome (the caller sees the Supervisor restart it triggers) instead of surfacing an update-failed error.
- **Restoring a backup that requires a newer Supervisor version**: this previously always returned the same generic error, even if an update was about to happen automatically. It now raises a specific, translatable `BackupSupervisorVersionError` (400) when auto update is disabled or unavailable, or `BackupSupervisorUpdateInProgressError` (503) telling the caller to retry once an update is under way. Before deciding, the version data is refreshed (if there's connectivity) in case a newer Supervisor was just released, so the check isn't based on stale data.
- **Version fetch throttling could mask a needed refresh**: `Updater.fetch_data()` is throttled and records its throttle window even when a fetch fails, so a caller checking for fresh data right after a concurrent/failed fetch could be silently skipped. `fetch_data()` is now a detached job as well: a call while one is already in-flight returns that same task instead of starting a redundant fetch, so `reload()` and the backup restore version check above can await it directly to get the real outcome instead of relying on the throttle window.

## Type of change

- [ ] Dependency upgrade

## Additional information

- Link to cli pull request:

## Checklist

https://claude.ai/code/session_01SAWqYzDiFEvYjZwfBWoxkC

---
_Generated by [Claude Code](https://claude.ai/code/session_01SAWqYzDiFEvYjZwfBWoxkC)_

### DIFF
--- a/supervisor/api/supervisor.py
+++ b/supervisor/api/supervisor.py
@@ -40,7 +40,7 @@ from ..const import (
     UpdateChannel,
 )
 from ..coresys import CoreSysAttributes
-from ..exceptions import APIError, SupervisorJobError
+from ..exceptions import APIError
 from ..store.validate import repositories
 from ..utils.blockbuster import BlockBusterManager
 from ..utils.sentry import close_sentry, init_sentry
@@ -262,11 +262,11 @@ class APISupervisor(CoreSysAttributes):
         if not self.sys_dev:
             version = self.sys_updater.version_supervisor
 
-        try:
-            await asyncio.shield(self.sys_supervisor.update(version))
-        except SupervisorJobError:
-            # The auto update already runs, the caller sees it through the restart
-            _LOGGER.info("Supervisor update is already in progress")
+        # update() is detached: this may return a task already started by a
+        # concurrent call (manual or auto-update) instead of a new one. Await
+        # it so this request reflects the real (possibly shared) result.
+        if task := await asyncio.shield(self.sys_supervisor.update(version)):
+            await task
 
     @api_process
     async def reload(self, request: web.Request) -> None:
--- a/supervisor/api/supervisor.py
+++ b/supervisor/api/supervisor.py
@@ -40,7 +40,7 @@ from ..const import (
     UpdateChannel,
 )
 from ..coresys import CoreSysAttributes
-from ..exceptions import APIError
+from ..exceptions import APIError, SupervisorJobError
 from ..store.validate import repositories
 from ..utils.blockbuster import BlockBusterManager
 from ..utils.sentry import close_sentry, init_sentry
@@ -262,7 +262,11 @@ class APISupervisor(CoreSysAttributes):
         if not self.sys_dev:
             version = self.sys_updater.version_supervisor
 
-        await asyncio.shield(self.sys_supervisor.update(version))
+        try:
+            await asyncio.shield(self.sys_supervisor.update(version))
+        except SupervisorJobError:
+            # The auto update already runs, the caller sees it through the restart
+            _LOGGER.info("Supervisor update is already in progress")
 
     @api_process
     async def reload(self, request: web.Request) -> None:
--- a/supervisor/backups/manager.py
+++ b/supervisor/backups/manager.py
@@ -840,16 +840,20 @@ class BackupManager(FileConfiguration, JobGroup):
             return
 
         if self.sys_updater.auto_update:
-            if self.sys_supervisor.need_update:
-                # Already known that an update is available, make sure it's kicked off
-                self.sys_create_task(self.sys_supervisor.auto_update_supervisor())
-            else:
-                # Refresh version info in case a newer Supervisor was just
-                # released. This also kicks off an auto update if one is now
-                # available.
-                await self.sys_updater.reload()
+            if not self.sys_supervisor.need_update:
+                # Ensure the latest version info is known before concluding
+                # there's no update to wait for. Await the fetch task
+                # directly (rather than calling reload()) so a fetch that's
+                # already in-flight - or one that just failed - isn't masked
+                # by fetch_data's throttle window.
+                await self.sys_updater.start_fetch_data()
 
-            if self.sys_supervisor.need_update:
+            # Start (or reuse) the auto update. We can't await it here - the
+            # update stops this very Supervisor instance once it completes,
+            # which would drop the connection before a response could be
+            # served. Use the task's existence/done status instead.
+            update_task = self.sys_supervisor.auto_update_supervisor()
+            if update_task and not update_task.done():
                 raise BackupSupervisorUpdateInProgressError(
                     _LOGGER.error,
                     backup_version=backup.supervisor_version,
--- a/supervisor/backups/manager.py
+++ b/supervisor/backups/manager.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import asyncio
 from collections.abc import Awaitable
+from contextlib import suppress
 import errno
 import logging
 from pathlib import Path
@@ -28,6 +29,7 @@ from ..exceptions import (
     BackupMountDownError,
     BackupSupervisorUpdateInProgressError,
     BackupSupervisorVersionError,
+    UpdaterError,
 )
 from ..jobs.const import JOB_GROUP_BACKUP_MANAGER, JobConcurrency, JobCondition
 from ..jobs.decorator import Job
@@ -842,17 +844,20 @@ class BackupManager(FileConfiguration, JobGroup):
         if self.sys_updater.auto_update:
             if not self.sys_supervisor.need_update:
                 # Ensure the latest version info is known before concluding
-                # there's no update to wait for. Await the fetch task
-                # directly (rather than calling reload()) so a fetch that's
-                # already in-flight - or one that just failed - isn't masked
-                # by fetch_data's throttle window.
-                await self.sys_updater.start_fetch_data()
+                # there's no update to wait for. fetch_data is a detached,
+                # REJECT-concurrency job, so this reuses an already in-flight
+                # fetch instead of masking it behind the throttle window.
+                with suppress(UpdaterError):
+                    if task := await self.sys_updater.fetch_data():
+                        await task
 
-            # Start (or reuse) the auto update. We can't await it here - the
-            # update stops this very Supervisor instance once it completes,
-            # which would drop the connection before a response could be
-            # served. Use the task's existence/done status instead.
-            update_task = self.sys_supervisor.auto_update_supervisor()
+            # Start (or reuse) the auto update. We can't await its completion
+            # here - the update stops this very Supervisor instance once it
+            # completes, which would drop the connection before a response
+            # could be served. Use the returned task's done status instead;
+            # awaiting auto_update_supervisor() itself only waits for the
+            # update to be started (or reused), not for it to finish.
+            update_task = await self.sys_supervisor.auto_update_supervisor()
             if update_task and not update_task.done():
                 raise BackupSupervisorUpdateInProgressError(
                     _LOGGER.error,
--- a/supervisor/backups/manager.py
+++ b/supervisor/backups/manager.py
@@ -840,12 +840,21 @@ class BackupManager(FileConfiguration, JobGroup):
             return
 
         if self.sys_updater.auto_update:
-            self.sys_create_task(self.sys_tasks.auto_update_supervisor())
-            raise BackupSupervisorUpdateInProgressError(
-                _LOGGER.error,
-                backup_version=backup.supervisor_version,
-                supervisor_version=self.sys_supervisor.version,
-            )
+            if self.sys_supervisor.need_update:
+                # Already known that an update is available, make sure it's kicked off
+                self.sys_create_task(self.sys_supervisor.auto_update_supervisor())
+            else:
+                # Refresh version info in case a newer Supervisor was just
+                # released. This also kicks off an auto update if one is now
+                # available.
+                await self.sys_updater.reload()
+
+            if self.sys_supervisor.need_update:
+                raise BackupSupervisorUpdateInProgressError(
+                    _LOGGER.error,
+                    backup_version=backup.supervisor_version,
+                    supervisor_version=self.sys_supervisor.version,
+                )
 
         raise BackupSupervisorVersionError(
             _LOGGER.error,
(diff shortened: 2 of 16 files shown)


supervisor/misc/tasks.py
```diff
@@ -398,7 +390,7 @@ async def _reload_updater(self) -> None:
         concurrency=JobConcurrency.REJECT,
         internal=True,
     )
-    async def _auto_update_supervisor(self):
+    async def auto_update_supervisor(self) -> None:
```
agners: With this change above this function seems rather out of place in the tasks module. Maybe better to place it in `sys_supervisor`?

agners (changes requested):

mdegat01:
@balloob @agners I updated it to essentially move the core of the logic from https://github.com/home-assistant/core/pull/181596 into Supervisor. I think it should live there. It's much easier for Supervisor to manage this as it has all the information already and doesn't need to sequence a long series of API calls and worry about what happened in between each.

The other PR will still have to put some of the logic in core for now to handle existing versions of Supervisor shipped on OOTB hardware. But long-term this should be a Supervisor problem not a core one with core just responding to the error with a wait loop or an exit depending on the error returned.

I'm also doing it this way because I don't see this as a temporary problem/fix. We always ship a version of Supervisor with OS. That version may always need an update. A user may proceed to the restore backup step in onboarding before that update has completed with a backup made on a newer Supervisor. Or they may just happen to try a restore during a Supervisor update or from one machine that got ahead of another. They should always be presented with a sensible error and experience.

supervisor/backups/manager.py
```diff
@@ -832,6 +834,34 @@ async def _validate_backup_location(
...
 
+    async def _check_supervisor_version(self, backup: Backup) -> None:
+        """Check backup is not from a newer Supervisor version, raise if so."""
+        if backup.supervisor_version <= self.sys_supervisor.version:
+            return
+
+        if self.sys_updater.auto_update:
+            if self.sys_supervisor.need_update:
+                # Already known that an update is available, make sure it's kicked off
+                self.sys_create_task(self.sys_supervisor.auto_update_supervisor())
```
agners: This reports "update in progress" before anything has actually started. `auto_update_supervisor` runs as a background task, and it can exit without doing anything: `OS_SUPPORTED` and `ARCHITECTURE_SUPPORTED` are not covered by the restore job's own conditions, and a failed pull is suppressed inside it. The caller then gets a 503 telling it to retry after completion, Core polls for five minutes and fails with a restart timeout.

Suggestion: have `auto_update_supervisor` return the running update task (or `None` when nothing was started), store it on the instance, and await the call here instead of spawning it. The `@Job` wrapper already returns `None` on a condition failure, so that lines up with "not started". Something like:

```python
async def auto_update_supervisor(self) -> asyncio.Task[None] | None:
    if self._update_task and not self._update_task.done():
        return self._update_task
    if not self.need_update:
        return None
    _LOGGER.info("Found new Supervisor version %s, updating", self.latest_version)
    self._update_task = self.sys_create_task(self._auto_update())
    return self._update_task
```

Returning the existing task also avoids a race: `reload()` spawns `auto_update_supervisor` itself when the state is `RUNNING`, so a second direct call here would contend for the `supervisor_auto_update` lock during the awaited condition checks and the loser would raise `JobException` out of the restore. With the stored task the `REJECT` on this wrapper is no longer needed; the one on `supervisor_update` stays as the backstop.

A failed pull stays asynchronous either way, which I think is fine: it already creates an `UPDATE_FAILED` issue and logs critically.

supervisor/supervisor.py
```diff
@@ -192,6 +192,8 @@ def write_profile() -> Path:
 
     @Job(
         name="supervisor_update",
+        on_condition=SupervisorJobError,
```
agners: Worth a comment here that this job has no `conditions`, so `SupervisorJobError` can currently only mean the concurrency reject. The API handler relies on that when it turns the error into a 200. If someone adds a condition later, condition failures would silently become success responses.

supervisor/backups/manager.py
```diff
@@ -832,6 +834,34 @@ async def _validate_backup_location(
...
+            else:
+                # Refresh version info in case a newer Supervisor was just
```
agners: Minor: `fetch_data` has a 30 second `THROTTLE`, so this reload is a silent no-op if any reload ran within that window, for example Core's own `reload_updates` call. Harmless in practice since that reload already spawned the auto update, but the comment overstates what this call guarantees.

agners (changes requested):
Thanks for moving the restore logic over, I agree this belongs in Supervisor. The shape looks good and I am positive this is a good change. The reload-triggers-auto-update part also closes the gap behind https://github.com/home-assistant/supervisor/issues/6762 for the case where Core is already installed, which is what happened on the Green in https://github.com/home-assistant/core/pull/181596: offline boot, connectivity arrives later, the updater refetches but nothing acts on it until the daily task.

Two more things:

`homeassistant/core.py` still catches only `SupervisorUpdateError` around its own `sys_supervisor.update()` in the install loop. With `REJECT` on `supervisor_update`, losing the race against the reload-spawned auto update raises `SupervisorJobError` out of `install()`. Rare, since Core install at `RUNNING` only happens on the recovery path, but the `except` should include `SupervisorJobError` too.

The PR description covers only the first four commits. Before merge it should describe the restore handling, the two new error keys and the 503, link https://github.com/home-assistant/core/pull/181596 and https://github.com/home-assistant/supervisor/issues/6762, and mention that `/reload_updates` and Core's Settings refresh can now restart Supervisor as a side effect.

supervisor/backups/manager.py
mdegat01: Not totally harmless actually. Because if our call is rejected due to the throttle window but it hasn't finished updating our version information then this proceeds and sees no Supervisor updates despite the reload and kicks out with an unresolvable version mismatch error. This actually can occur in the [#181596](https://github.com/home-assistant/core/pull/181596/) as well but I don't know if there's any way we can handle this edge case race condition there. Here we can but it does add complexity. We'd have to do something similar to your comment above where we return the existing task if there's already one in progress

mdegat01:
Reopening despite the failures. I'm not entirely sure what's going on but these checks all pass locally. There appears to be some kind of (hopefully) transient failure with the apt repository we're using in CI. Its failing the digest check on the public key for packages we're trying to pull, not something changed in here but not something with an easy workaround other then waiting from what I can tell.

supervisor/supervisor.py
```diff
@@ -244,6 +250,44 @@ async def update(self, version: AwesomeVersion | None = None) -> None:
...
+        """
+        if self._update_task and not self._update_task.done():
+            return self._update_task
+
+        if not self.need_update:
+            return None
+
+        _LOGGER.info("Found new Supervisor version %s, updating", self.latest_version)
+        # Eager start so a synchronous, immediate failure (e.g. a job condition
+        # that doesn't need to await anything) is reflected in the task's done
```
agners: Uh, the `await` really becomes relevant here:

The job wrapper checks conditions in a fixed order: HEALTHY, RUNNING, FREE_SPACE, INTERNET_SYSTEM, INTERNET_HOST, then OS_SUPPORTED, AUTO_UPDATE, ARCHITECTURE_SUPPORTED. FREE_SPACE runs get_disk_free_space in an executor, so the task suspends there on every call. Only HEALTHY and RUNNING fail synchronously, and both are already guaranteed by the restore job's own conditions.

In other words, this job check condition implementation detail dictates what the http API return here. This feels brittle.

Firstly, I think we can reorder the conditions and put the co-routine based checks to the end. It would help if we stick with this implementation (e.g. ARCHITECTURE_SUPPORTED would be raised immediately if that were the case when this API gets called). At the very least the reordering can be seen as a performance improvement, so I created this PR either way: #7210

But I would actually prefer this PR does not depend on this implementation detail of the Job system. Claude suggested to create a detached property for Jobs, which makes sure all the conditions gets checked first, and a task is returned to the caller. This would then clearly separate the Job condition part of the execution from the actual job. I've created this as a PR on its own: #7211 If this LGTY we can include it by a main backmerge or merge this one as is and do a follow up PR. Either way is fine by me.

supervisor/updater.py
```diff
@@ -58,20 +61,45 @@ async def load(self) -> None:
...
+    def start_fetch_data(self) -> asyncio.Task[None]:
+        """Start (or reuse) a version data fetch, returning its task.
+
+        Returns the task performing the fetch, which may already be in
+        progress from a previous call.
+
+        Callers that need up to date version info should await the
+        returned task directly rather than calling fetch_data() themselves.
+        fetch_data() is throttled, so a second direct call while one is
+        already in-flight (or one that just failed) would silently be
```
agners: The in-flight case works, but the just-failed case does not. The decorator stamps the throttle timestamp before the body runs (`set_last_call` in the wrapper), so a fetch that failed a few seconds ago still counts as a call. `start_fetch_data` then finds no running task, creates a new one, and that one is throttled inside fetch_data and returns without fetching. The restore check ends up with the stale data and reports the unresolvable version mismatch, which is the case you called not harmless in the other thread.

We can actually reuse #7211 here too: `fetch_data` can carry `detach=True` with `REJECT` concurrency and both `reload()` and the restore check become:

```
if task := await self.fetch_data():
    await task
```

That replaces `start_fetch_data`, `_fetch_task` and `_fetch_data_suppressed`, and the `REJECT`-returns-running-task behavior gives the in-flight reuse for free 🎉 .

agners (changes requested):

supervisor/supervisor.py
mdegat01: I'm a fan of this plan. Merged the referenced PR and incorporated detach in 5404ca7, it is a lot cleaner 🎉

supervisor/updater.py
mdegat01: Done, fully agree, way nicer 👍

supervisor/api/supervisor.py
```diff
@@ -262,7 +262,11 @@ async def update(self, request: web.Request) -> None:
         if not self.sys_dev:
             version = self.sys_updater.version_supervisor
 
-        await asyncio.shield(self.sys_supervisor.update(version))
+        # update() is detached: this may return a task already started by a
+        # concurrent call (manual or auto-update) instead of a new one. Await
+        # it so this request reflects the real (possibly shared) result.
+        if task := await asyncio.shield(self.sys_supervisor.update(version)):
+            await task
```
agners: The shield no longer protects the update. It wraps the wrapper call, which returns as soon as the task exists, and the `await task` below it is unshielded. Awaiting a task from a coroutine that gets cancelled cancels that task, so a client disconnect during the image pull now cancels the Supervisor update mid-flight. That is exactly what the original `asyncio.shield(self.sys_supervisor.update(version))` prevented.

The outer shield can go too: `update()` has no conditions or throttle, so the wrapper never suspends before the task exists, and if a suspension point is added later, cancelling there aborts before anything started, which is fine.

```suggestion
        if task := await self.sys_supervisor.update(version):
            await asyncio.shield(task)
```

supervisor/supervisor.py
```diff
@@ -244,6 +255,39 @@ async def update(self, version: AwesomeVersion | None = None) -> None:
...
+        restarts Supervisor once done, and awaiting that here could drop
+        the caller's connection before a response is sent. Callers that
+        need to wait for the result should await the returned task
+        themselves.
+        """
+        if not self.need_update:
+            return None
+
+        _LOGGER.info("Found new Supervisor version %s, updating", self.latest_version)
+        return await self.update()
```
agners, line 289: With `update()` detached, its errors surface on the returned task instead of here. Two callers drop that task: `Updater.reload()` discards the return value, and `_check_supervisor_version` only reads `done()`. A failed pull, an AppArmor fetch error or the "already installed" case then leaves an unretrieved exception, which asyncio logs with a full traceback at garbage collection. Before this PR `_auto_update_supervisor` suppressed `SupervisorUpdateError`, so a failed auto update showed up only through the job record and the critical log in `update()`.

Since this is the one caller that does not care about the outcome, I would restore that here by consuming the exception on the task:

```python
        if task := await self.update():
            task.add_done_callback(
                lambda t: None if t.cancelled() else t.exception()
            )
        return task
```

Callers that do care, the API and the Core install loop, still await the task and still see the error. A test that patches `DockerSupervisor.install` to raise, runs `reload()`, and asserts no "Task exception was never retrieved" in the asyncio log would pin it.

agners (changes requested):
The move of `detach` onto `update()` is the right call: manual and auto updates share one task, the restore check sees a running update regardless of who started it, and the 200-on-`SupervisorJobError` workaround is gone. Two things fell out of the relocation, see inline. Both are small.

supervisor/supervisor.py
agners: This actually is better solved with #7217. So instead of doing it here, simply leave it unresolved and once #7217 got merged.

supervisor/api/supervisor.py
agners: I take @sairon's :+1:  as an approval :smile: . The change is straight forward and make sense, so let's get this PR done.

agners (approved):

supervisor/supervisor.py
agners: With https://github.com/home-assistant/supervisor/pull/7217 merged this is resolved.

PR #7204, opened 2026-09-09
Reinstall the stored Core version when its image is lost

agners:
## Proposed change

When the Docker storage is wiped, for example by the containerd snapshotter migration on Home Assistant OS 17 (`ha docker migrate-storage-driver`) or a Docker storage reset, the image of the installed Home Assistant Core is gone. On the next start Supervisor treated this like a fresh installation: it installed the landingpage and then pulled the latest Core release. Users on an older Core unexpectedly ended up on the newest release, see the reports linked below.

This PR remembers the stored Core version in `HomeAssistantCore.load()`. If the image is missing and a real version is stored, Supervisor now pulls exactly that version directly, without going through the landingpage. Core is then started by the regular startup sequence once the image is present. Apps already behave this way through the repair path.

A registry rate limit retries the pull, like the landingpage install does. Any other pull failure, for example a version that is no longer published, logs a warning and falls back to the existing landingpage flow, which installs the latest release. New installations, where no version is stored, are unaffected.

The `install_image` property now falls back to the default image for this system when the updater has no image information yet, so all Core install paths select the image the same way. This makes the wait-for-updater loop in the landingpage install obsolete.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR is related to issue: https://github.com/home-assistant/operating-system/discussions/4867#discussioncomment-17953296 and https://github.com/home-assistant/operating-system/discussions/4487#discussioncomment-15566751
- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/homeassistant/core.py
+++ b/supervisor/homeassistant/core.py
@@ -26,6 +26,7 @@ from ..exceptions import (
     DockerContainerNotFoundError,
     DockerContainerNotRunningError,
     DockerError,
+    DockerRegistryRateLimitExceeded,
     DockerStatsTimeoutError,
     HomeAssistantCrashError,
     HomeAssistantError,
@@ -103,10 +104,11 @@ class HomeAssistantCore(JobGroup):
             BusEvent.SUPERVISOR_STATE_CHANGE, self._supervisor_state_changed
         )
 
+        stored_version = self.sys_homeassistant.version
         try:
             # Evaluate Version if we lost this information
-            if self.sys_homeassistant.version:
-                version = self.sys_homeassistant.version
+            if stored_version:
+                version = stored_version
             else:
                 self.sys_homeassistant.version = (
                     version
@@ -124,6 +126,15 @@ class HomeAssistantCore(JobGroup):
             _LOGGER.info(
                 "No Home Assistant Docker image %s found.", self.sys_homeassistant.image
             )
+            # The image of an installed Core got lost (e.g. Docker storage was
+            # wiped). Reinstall that version instead of installing the latest
+            # release through the landingpage.
+            if (
+                stored_version
+                and stored_version != LANDINGPAGE
+                and await self._reinstall(stored_version)
+            ):
+                return
             await self.install_landingpage()
         else:
             self.sys_homeassistant.version = self.instance.version or version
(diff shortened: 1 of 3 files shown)


supervisor/homeassistant/core.py
```diff
@@ -189,6 +200,63 @@ async def install_landingpage(self) -> None:
...
+    async def _reinstall(self, version: AwesomeVersion) -> bool:
+        """Reinstall the lost image of an installed Home Assistant version.
+
+        Return False if the image could not be pulled.
+        """
+        image = (
+            self.sys_homeassistant.image
+            if self.sys_homeassistant.override_image
+            else self.sys_homeassistant.default_image
+        )
```
mdegat01:
```suggestion
        image = self.sys_homeassistant.install_image
```
This is what `install` uses, is there a reason we do it differently here? I guess `install_image` relies on `updater.reload` having been called once and `default_image` does not. But it seems like in that case `install_image` should be updated to fallback on `default_image` instead of returning `None` right? Just feels like `reinstall` and `install` should behave the same here.

supervisor/homeassistant/core.py
```diff
@@ -189,6 +200,63 @@ async def install_landingpage(self) -> None:
...
+            except TimeoutError:
+                if (job := self.instance.active_job) and job.progress:
+                    _LOGGER.info(
+                        "Downloading Home Assistant Core image, %d%%",
+                        int(job.progress),
+                    )
+                else:
+                    _LOGGER.info("Home Assistant Core installation in progress")
+
+    async def _reinstall(self, version: AwesomeVersion) -> bool:
```
mdegat01, line 195: Instead of refactoring the progress log into a shared helper, can `install` just call this method to do the actual install of the HA image? I'm noticing there's a number of differences between the code in here and the code that does the image install bit in `install`, particularly around error handling. Like:
1. why does `install` use `self.instance.update` and this uses `self.instance.install`? `self.instance.update` basically just calls `self.instance.install` under the hood anyway.
2. Why does this code handle`DockerRegistryRateLimitExceeded` but `install` does not?
3. Why does this code use its own conditional for determining image but `install` uses `sys_homeassistant.install_image`

I know the logs in this method use the verb `reinstall` instead of `install` but I don't think it would be confusing to users if we just said `install` so that doesn't seem like enough of a reason. Maybe there's enough other differences its not worth it but it seems like it would be a nicer approach to have `install` simply call `reinstall` and share all this code.

tests/homeassistant/test_core.py
```diff
@@ -1009,3 +1010,126 @@ async def test_core_loads_wrong_image_for_architecture(
...
+    was installed before instead of jumping to the latest release, and it
+    must not install the landingpage in between.
+    """
+    coresys.homeassistant.version = AwesomeVersion("2026.2.3")
+
+    with (
+        patch.object(DockerHomeAssistant, "attach", side_effect=DockerError()),
+        patch.object(DockerHomeAssistant, "install") as install,
+        patch.object(HomeAssistantCore, "install_landingpage") as install_landingpage,
+        patch.object(HomeAssistantCore, "start") as start,
```
mdegat01: I know claude and all these AIs love patch mocking our code in tests. I still don't love it personally, it feels like it introduces unnecessary fragility and should be a tool used sparingly. But perhaps we should establish some limits in our AI guidance? These are high level methods within the same class.

I assume the goal here is to say with certainty that "we did not install landing page or start the home assistant container as part of this test". But if code was added that called `self.instance.start` for example, this test passes but the tested premise does not.

Maybe we should at least limit to avoiding mocking methods within the primary class under test? Since the test calls `HomeAssistantCore.load` as its main action step mocking methods of `HomeAssistantCore` feels like a bit of a code smell. Mocking the underlying docker methods feels ok.

I know this is not unique to this PR, we can take this up separately and not deal with it here. I do worry our tests are getting more fragile though.

mdegat01 (approved):
My comments are more along the line of nits about code organization. The `install_image` one is the main one potentially worth fixing but if you disagree I'm ok with it as is. Good catch though, this would be a frustrating issue.

supervisor/homeassistant/core.py
agners: Agreed, done in ac095c8d7. `install_image` now falls back to `default_image` when the updater has no image information, and `_reinstall()` uses it like `install()` does. That also made the wait-for-updater loop in `install_landingpage()` obsolete, so it is gone.

supervisor/homeassistant/core.py
agners: The remaining differences are intentional, though I agree the image selection was not. That one is fixed now.

`update()` vs `install()`: `install()` uses `instance.update()` since #1196 because it also stops the running landingpage container after the pull. Without that Core never came up. In the reinstall case there is no container at all, so plain `instance.install()` is correct. Sharing the call would need the stop to be added back conditionally.

Rate limit: `install()` retries every error forever, so it never needed to distinguish a rate limit. `_reinstall()` has to give up on a permanent failure, e.g. a version that is no longer published, so it can fall back to latest. That is why it is the first Core install path that needs to tell a transient rate limit apart, same as the app autofix does since #6816.

I also dropped the progress logging from the reinstall. It was added for the first installation where the landingpage shows the Supervisor logs, which does not apply here. So the two loops are separate and nothing is shared anymore.

tests/homeassistant/test_core.py
agners: Fair point. I reworked the new tests in ac095c8d7 to mock at the Docker layer only: container and image inspect return not found, `DockerAPI.pull_image` and `DockerAPI.run` are patched, and the tests assert which tags were pulled and that no container was created. Agree it is worth writing this down in the AI guidance separately.

mdegat01 (approved):
Oh you decided to drop the logging from reinstall then. Yea makes sense, progress reporting covers it anyway, no reason to add it to new places.

I didn't really think about the fact that adding `default_image` to `install_image` would mean we kill the retry loops to make sure we know the current image from the `version` repo. That might have unintended consequences but I think I prefer it too? Its pretty safe and we really don't want users setting up a new system to get stuck waiting in a boot sequence waiting for unknown reasons more then necessary. Better to get them through setup quickly and correct these things later if necessary.

bcutter:
> Good catch though, this would be a frustrating issue.

Nah, it WAS (/IS) a frustrating one 😁 

Good to see this gets solved - better late (storage driver migration was shipped when, beginning of 2026?) than never 👍🏼

PR #7207, opened 2026-09-10
Hide new ntp host feature from v1 /host/info features field

mdegat01:
## Proposed change

#7181 added `HostFeature.NTP` to the `features` list returned by `/host/info`. Unfortunately the `python-supervisor-client` library's `HostInfo` model types that field as the strict `list[HostFeature]` instead of `list[HostFeature | str]`:

https://github.com/home-assistant-libs/python-supervisor-client/blob/3459d25b7b223dfd2a49e5aa3f794ea822faab5c/aiohasupervisor/models/host.py#L49

As a result, the hassio integration currently fails to set up on nightly with:

```
Failed setup, will retry: Field "features" of type list[HostFeature] in HostInfo has invalid value ['reboot', 'shutdown', 'services', 'network', 'hostname', 'timedate', 'os_agent', 'haos', 'resolved', 'journal', 'disk', 'mount', 'ntp']
```

This PR works around the issue by hiding `ntp` from the v1 `/host/info` `features` field and exposing the full, unfiltered list under a new `all_features` field instead. The v2 API is unaffected (it has not shipped yet, and its `features` field is already typed as `list[HostFeature | str]` in the client library).

A test was added which enables every `HostFeature` value and fails if the v1 `features` field ever contains a value outside the currently known-safe set (excluding `ntp`), to catch this class of issue if a future feature is added without the same compatibility treatment.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR fixes or closes issue: fixes the nightly hassio integration setup failure caused by #7181
- This PR is related to issue: #7181
- Link to cli pull request:
- Link to client library pull request: https://github.com/home-assistant-libs/python-supervisor-client/pull/408

## Checklist


sairon (approved):
Ugh, nice catch! The typing was somewhat short-sighted for such field. I guess we need to carry the added field as a permanent workaround - or at least until we don't need to support Core that has the updated client library with the relaxed typing - in that case we should tell that and say it's for internal use until some deadline.

Nit: it should probably be added to `GET /host/info` docs and explained why there are two similar fields: https://developers.home-assistant.io/docs/api/supervisor/endpoints#host

agners (approved):
Nice catch!

In a way it shows that this "extending API is ok/not breaking" has its pitfalls 😅 .

PR #7210, opened 2026-09-10
Evaluate synchronous job conditions before awaiting ones

agners:
## Proposed change

Reorder the checks in `Job.check_conditions` so that all synchronous state reads run first, the checks that await run second, and `PLUGINS_UPDATED` stays last because it performs plugin updates rather than only checking.

The current order grew by accretion. When the conditions were introduced in https://github.com/home-assistant/supervisor/pull/2250 every check was a cheap synchronous read, so the order did not matter and each later condition was simply inserted next to a related one. Since then `FREE_SPACE` became an executor call and `INTERNET_SYSTEM`/`INTERNET_HOST` started awaiting a connectivity probe in https://github.com/home-assistant/supervisor/pull/6765, without being moved. A job that is going to be refused for an unsupported OS or architecture, disabled auto update, or a missing host feature therefore first pays for a disk stat and a network probe.

With this change a refusal is visible before the first suspension. That also lets callers eagerly start a job as a task and rely on the task being done immediately when the job was refused, which the Supervisor auto update in https://github.com/home-assistant/supervisor/pull/7201 depends on. `MOUNT_AVAILABLE` moves out from behind `PLUGINS_UPDATED` into the synchronous section. Two comments document the rule so new conditions land in the right section.

The only observable difference is which condition is named in the log when several fail at once, and that the `FREE_SPACE` resolution issue is not created from this path when an earlier synchronous check already refuses the job.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR is related to issue: https://github.com/home-assistant/supervisor/pull/7201
- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/jobs/decorator.py
+++ b/supervisor/jobs/decorator.py
@@ -382,6 +382,9 @@ class Job(CoreSysAttributes):
         coresys: CoreSysAttributes, conditions: set[JobCondition], method_name: str
     ):
         """Check conditions."""
+        # Evaluation order: synchronous state reads first, checks that await
+        # second, and PLUGINS_UPDATED last because it performs work. New
+        # conditions must be inserted in the matching section.
         used_conditions = set(conditions) - set(coresys.sys_jobs.ignore_conditions)
         ignored_conditions = set(conditions) & set(coresys.sys_jobs.ignore_conditions)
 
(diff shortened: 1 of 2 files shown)


mdegat01 (approved):
LGTM 👍

PR #7211, opened 2026-09-10
Add detach option to the Job decorator

agners:
## Proposed change

Add a `detach=True` option to the `Job` decorator. A detached job awaits its conditions, concurrency control and throttling exactly as before, but then runs the method in a separate task and returns that task to the caller. A refused job returns `None`, or raises `on_condition` when configured. With `REJECT` concurrency, a call while the job is running returns the running task instead of raising. The task releases the concurrency lock and cleans up the job record when it completes. Group concurrency is not supported in this mode and is rejected at decoration time.

This gives callers a definitive answer on whether a long running job was started, without awaiting its completion. The Supervisor auto update in https://github.com/home-assistant/supervisor/pull/7201 needs exactly that: a backup restore that requires a newer Supervisor has to know whether the update actually started before it tells the caller to retry later, and the update itself stops the Supervisor, so it cannot be awaited from within an API request. Today the only way to get that answer is to eagerly start the wrapped call as a task and inspect its `done()` state, which silently depends on none of the job conditions suspending before the refusal. With `detach=True` the auto update becomes `if task := await self.auto_update_supervisor(): raise UpdateInProgress` and the hand-rolled task tracking on `Supervisor` goes away.

The shared method body is factored into a local `execute()` coroutine used by both modes, so the non-detached path behaves as before. The `execute()` coroutine is only created after throttling has passed, so a refused call leaves no unawaited coroutine behind.

Tests cover: the task is returned and the job record is cleaned up after completion, refusal by a condition both with and without `on_condition`, throttling, `REJECT` returning the running task and starting a new one after it finishes, an error in the body surfacing when the task is awaited, and the group concurrency guard.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR is related to issue: https://github.com/home-assistant/supervisor/pull/7201
- Link to cli pull request:

## Checklist

### DIFF
--- a/supervisor/jobs/decorator.py
+++ b/supervisor/jobs/decorator.py
@@ -345,8 +345,13 @@ class Job(CoreSysAttributes):
                         # Record before the task is scheduled so a concurrent
                         # call is throttled instead of starting a second task
                         record_call()
+                        # Start eagerly so the runner is inside its try block
+                        # before the task is exposed. A task cancelled before
+                        # its first step never runs its finally, which would
+                        # leave the lock held and the job registered.
                         self._detached_task = self.sys_create_task(
-                            self._run_detached(job_group, job, cleanup, execute())
+                            self._run_detached(job_group, job, cleanup, execute()),
+                            eager_start=True,
                         )
                         detached = True
                         return self._detached_task
--- a/supervisor/jobs/decorator.py
+++ b/supervisor/jobs/decorator.py
@@ -23,6 +23,7 @@ from ..resolution.const import (
     IssueType,
     UnsupportedReason,
 )
+from ..utils.sentinel import DEFAULT
 from ..utils.sentry import async_capture_exception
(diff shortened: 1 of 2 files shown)


agners:
Wrt making this part of Concurrency, the flag can be useful in all concurrency options:

 - No concurrency + detach: every accepted call starts its own task. Fine for fire-and-forget work that tolerates overlap.
  - `QUEUE` + detach: the caller waits for the lock, then gets a task for its own run. Serialised background runs.
  - `REJECT` + detach: the caller gets the already running task. This is the case auto_update_supervisor and fetch_data want.

I've previously untangled such combination with #6013, so I rather not go down that road again :sweat_smile: .

mdegat01 (approved):
Yep, I like this, good idea! This is way better then what I was doing in the other PR and less error prone with how the conditions work. Lets get this in and then use it 👍

PR #7218, opened 2026-09-14
Fix supervisor config restore from encrypted backups

### LINKED ISSUE #7213: Mount and registry config is not restored from encrypted backups
cc @agners @mdegat01, you reviewed and merged #6411.

## What I observed

My dad's HAOS system had a network storage mount that was used as the backup destination. He made a backup in August (Supervisor 2026.08.0), reinstalled, and restored it. The restore during onboarding failed because the fresh install's Supervisor was older than the one in the backup (`Must update supervisor first`). He then onboarded normally and restored the same backup from the backup UI. Everything came back except the network storage.

## Where it was supposed to be fixed

#6411 (shipped in 2026.03.3) added `supervisor.tar` with `mounts.json` to every backup, and #6674 added `docker.json`. Both are restored by `Backup.restore_supervisor_config()`.

## Root cause

Core-created backups are encrypted by default. securetar opens an encrypted inner tar in streaming mode (`r|gz`) because it cannot seek in ciphertext. `restore_supervisor_config` calls `tar_file.getmember("mounts.json")` and then `tar_file.extractfile(member)`, which needs a backwards seek. That raises `tarfile.StreamError: seeking backwards is not allowed`. The `except (tarfile.TarError, json.JSONDecodeError)` clause catches it, logs a warning, and returns `(False, [])`. Core starts restores with `background=True` and the failure is not captured on the job, so the restore reports success.

Reproduced with securetar 2026.4.1, writing `supervisor.tar.gz` through `SecureTarArchive.create_tar()` like `store_supervisor_config` does, then reading it like `restore_supervisor_config` does:

| Written | Read with | Result |
|---|---|---|
| encrypted | correct password | `StreamError: seeking backwards is not allowed` |
| plain | no password | OK |

Home Assistant, apps and folders restore fine because they use `extractall`, which reads sequentially. The tests in `tests/backups/test_backup.py` never set a backup password, so they only cover the plain case.

Affected systems should have this in the Supervisor log: `Failed to read supervisor config from backup: seeking backwards is not allowed`

## Fix

Iterate the inner tar sequentially (`for member in tar_file:`) and read `mounts.json` and `docker.json` as they stream past. Add a test with a password set.

Related Core issue: encrypt/decrypt rewrites in Core leave `supervisor.tar` untouched, so the tar's encryption state can also mismatch the backup metadata. See home-assistant/core#182137.

agners:
## Proposed change

The network storage mount and Docker registry configuration stored in `supervisor.tar` was never restored from encrypted backups. securetar opens encrypted inner tars in streaming mode since it cannot seek in the ciphertext. `restore_supervisor_config()` used `getmember()` followed by `extractfile()`: to locate the member, tarfile reads all headers and thereby advances past the member data, so extracting it requires seeking backwards. This raises `tarfile.StreamError`, which was caught by the broad `TarError` handler and only logged as a warning. The restore reported success while the mounts were silently missing.

This affects every encrypted backup since the supervisor tar was introduced in 2026.03.3 (#6411), and Core-created backups are encrypted by default. The data in those backups is intact, restoring them again with this fix recovers the configuration.

The inner tar is now read sequentially, picking up `mounts.json` and `docker.json` as they stream past. A new test stores and restores mounts and registries with a backup password set, as the existing tests only covered plain tars.

Additionally, the `Backup` class docstring now describes the archive layout and documents that Core only encrypts/decrypts inner tars from a fixed list when rewriting a backup (`_get_expected_archives`), so any new inner tar needs to be added there as well. The missing `supervisor` entry is tracked in home-assistant/core#182137.

## Type of change

- [ ] Dependency upgrade

## Additional information

- This PR fixes or closes issue: fixes #7213
- This PR is related to issue: https://github.com/home-assistant/core/issues/182137
- Link to cli pull request:

## Checklist


sairon (approved):
Nice catch, LGTM!

